PR #53, opened 2014-12-17
[do not merge] initial change to remove cloud service managed cache scaffolding support

yugangw-msft:
Per discussion with azure cache team, we will start to remove the cache scaffolding on a cloud service project 
This pr is only for the purpose to evaluate the change scope overall.


ogail:
@yugangw-msft this PR is so old, closing for now

PR #55, opened 2014-12-18
Add new cmdlets for storage SAS token stored policy.

EmmaZhu:
Add new cmdlets for storage SAS token stored policy.


markcowl:
@EmmaZhu merge conflicts, please update

PR #61, opened 2014-12-19
Output warnings that Managed Cache scaffolding commands will be removed 

yugangw-msft:
For upcoming release (at beginning of Jan), we will output warning message. After it is released, I will go ahead to merge "https://github.com/Azure/azure-powershell/pull/53" so we finally remove the commands


yugangw-msft:
re-open, per the latest discussion that we do need to output warning, and we will remove the command at the next-next release

PR #108, opened 2015-01-24
Fixed null ref error in MockClientFactory during test recording

stankovski:


ogail:
Could we try record an existing test and make sure it passes? Otherwise LGTM

huangpf:
LGTM. Thanks~

stankovski:
Confirmed. I have successfully recorded a test and have verified that it worked. I just didn't include it in the PR.

PR #109, opened 2015-01-24
Make NewAzureDeployment param validation consistent with SetAzureDeploym...

ghost:
In `Set-AzureDeployment`, if the given package Uri is to a file, then it will first upload that file to Azure storage. The storage account which receives this package is defined by the `CurrentStorageAccountName` property of the current subscription. In the case where this is `null` and a package upload is required, an exception is thrown. However, if the package is already uploaded, then the value of `CurrentStorageAccountName` is irrelevant.

`New-AzureDeployment`, however, always checks whether `CurrentStorageAccountName` is `null` before executing any code, and thus, even if the package is already uploaded to storage, it will throw an exception if this field is `null`.

This change makes the behavior of `New-AzureDeployment` consistent with that of `Set-AzureDeployment` - that is, it will only throw an exception when `CurrentStorageAccountName` is `null` if the package needs to be uploaded.

redmond\rylewell

### DIFF
--- a/src/ServiceManagement/Compute/Commands.ServiceManagement/HostedServices/NewAzureDeployment.cs
+++ b/src/ServiceManagement/Compute/Commands.ServiceManagement/HostedServices/NewAzureDeployment.cs
@@ -123,6 +123,11 @@ namespace Microsoft.WindowsAzure.Commands.ServiceManagement.HostedServices
             }
             else
             {
+                if (string.IsNullOrEmpty(storageName))
+                {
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@rylewell-msft__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

stankovski:
@azuresdkci add to whitelist

PR #127, opened 2015-01-30
Batch first OM based cmdlet

jasper-schneider:
- Batch PowerShell data models and tool to generate them
- Infrastructure for cmdlets based on the Batch OM
- Get-AzureBatchWorkItem cmdlet + some unit tests


azurecla:
Hi __@jasper-schneider__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (jaschnei). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

stankovski:
@azuresdkci add to whitelist

stankovski:
@azuresdkci test this please

src/ResourceManager/Batch/Commands.Batch.Test/Commands.Batch.Test.csproj
```diff
@@ -186,6 +220,7 @@
       <ErrorText>This project references NuGet package(s) that are missing on this computer. Enable NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is {0}.</ErrorText>
     </PropertyGroup>
     <Error Condition="!Exists('$(SolutionDir)\.nuget\NuGet.targets')" Text="$([System.String]::Format('$(ErrorText)', '$(SolutionDir)\.nuget\NuGet.targets'))" />
+    <Error Condition="!Exists('..\..\..\packages\xunit.runner.visualstudio.2.0.0-rc1-build1030\build\net20\xunit.runner.visualstudio.props')" Text="$([System.String]::Format('$(ErrorText)', '..\..\..\packages\xunit.runner.visualstudio.2.0.0-rc1-build1030\build\net20\xunit.runner.visualstudio.props'))" />
```
matt-gibbs: Should follow up with Yugang on the changes to move all xunit use to the package instead of the extension.

src/ResourceManager/Batch/Commands.Batch.Test/Commands.Batch.Test.csproj
```diff
@@ -80,6 +92,17 @@
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\..\..\packages\Microsoft.IdentityModel.Clients.ActiveDirectory.2.11.10918.1222\lib\net45\Microsoft.IdentityModel.Clients.ActiveDirectory.WindowsForms.dll</HintPath>
     </Reference>
+    <Reference Include="Microsoft.WindowsAzure.Configuration">
+      <HintPath>..\..\..\packages\Microsoft.WindowsAzure.ConfigurationManager.1.8.0.0\lib\net35-full\Microsoft.WindowsAzure.Configuration.dll</HintPath>
+    </Reference>
+    <Reference Include="Microsoft.WindowsAzure.Storage, Version=2.1.0.4, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
```
matt-gibbs: Why such an old version of storage?

jasper-schneider: The Batch OM (in the Microsoft.Azure.Batch NuGet package) was built on the older version.  There were breaking changes in the newest version of the storage library, so the owners of the Batch OM opted to stay on the older version for the time being.

src/ResourceManager/Batch/Commands.Batch.Test/Commands.Batch.Test.csproj
jasper-schneider: I'll send him mail.

src/ResourceManager/Batch/Commands.Batch.Test/Commands.Batch.Test.csproj
matt-gibbs: ugh. There should be an issue filed to address the breaking changes there and move to something more current.

src/ResourceManager/Batch/Commands.Batch/WorkItems/GetBatchWorkItemCommand.cs
```diff
@@ -0,0 +1,85 @@
...
+
+using Microsoft.Azure.Batch;
+using Microsoft.Azure.Commands.Batch.Models;
+using Microsoft.Azure.Commands.Batch.Properties;
+using System;
+using System.Collections.Generic;
+using System.Management.Automation;
+using Constants = Microsoft.Azure.Commands.Batch.Utils.Constants;
+
+namespace Microsoft.Azure.Commands.Batch
```
ogail: All code inside cmdlets including checks and everything should be wrapped in the client. For example, this class should be (using `ExecuteCmdlet` as this is the expected method)
```csharp
{
    WorkItemsFilterOptions options = new WorkItemsFilterOptions { Name = Name, Filter = Filter };
    BatchClient.GetWorkItem(
}
```

src/ResourceManager/Batch/Commands.Batch/WorkItems/GetBatchWorkItemCommand.cs
```diff
@@ -0,0 +1,85 @@
...
+{
+    [Cmdlet(VerbsCommon.Get, "AzureBatchWorkItem", DefaultParameterSetName = Constants.NameParameterSet),
+        OutputType(typeof(PSCloudWorkItem), ParameterSetName = new string[] { Constants.NameParameterSet }),
+        OutputType(typeof(IEnumerableAsyncExtended<PSCloudWorkItem>), ParameterSetName = new string[] { Constants.ODataFilterParameterSet })]
+    public class GetBatchWorkItemCommand : BatchOMCmdletBase
+    {
+        [Alias("Name")]
+        [Parameter(Position = 0, ParameterSetName = Constants.NameParameterSet, Mandatory = false, ValueFromPipeline = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The name of the WorkItem to query.")]
+        [ValidateNotNullOrEmpty]
+        public string WorkItemName { get; set; }
```
ogail: Just have the parameter named `Name`

src/ResourceManager/Batch/Commands.Batch/BatchOMCmdletBase.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// </summary>
+        internal IEnumerable<BatchClientBehavior> AdditionalBehaviors { get; set; }
+
+        [Parameter(Mandatory = true, ValueFromPipeline = true, HelpMessage = "The Batch account information.")]
+        [ValidateNotNullOrEmpty]
+        public BatchAccountContext BatchContext { get; set; }
+
+        public override void ExecuteCmdlet()
+        {
+            WriteVerboseWithTimestamp(Resources.AccountAndKeyInUse, BatchContext.AccountName, BatchContext.KeyInUse);
```
ogail: you could override `BeingProcessing` and add this logic there and no need got `ServiceRequest` methods

src/ResourceManager/Batch/Commands.Batch/BatchOMCmdletBase.cs
```diff
@@ -0,0 +1,46 @@
...
+using Microsoft.Azure.Commands.Batch.Properties;
+using System.Collections.Generic;
+using System.Management.Automation;
+
+namespace Microsoft.Azure.Commands.Batch
+{
+    /// <summary>
+    /// Base class for cmdlets that use the Batch OM.
+    /// </summary>
+    public class BatchOMCmdletBase : BatchCmdletBase
```
ogail: What is `OM` stands for? it's better to have full names

ogail:
- Add a mocked scenario test
- Remove the code generation tool code

src/ResourceManager/Batch/Commands.Batch/WorkItems/GetBatchWorkItemCommand.cs
jasper-schneider: Updated to using ExecuteCmdlet()

Regarding putting everything into a client method, how does that work when the two parameter sets have different output types (single object vs. collection object)?  I'd rather not have the method have output type object, so it seems like I need to do some parameter processing in the cmdlet to choose the correct client method.  Looking at Storage's Get-AzureStorageBlob for example, it seems like they're doing something similar with choosing the method to use based on parameter set.  Is there a different example I should be following?

src/ResourceManager/Batch/Commands.Batch/WorkItems/GetBatchWorkItemCommand.cs
jasper-schneider: Updated

src/ResourceManager/Batch/Commands.Batch/BatchOMCmdletBase.cs
jasper-schneider: Updated

src/ResourceManager/Batch/Commands.Batch/BatchOMCmdletBase.cs
jasper-schneider: Renamed the file/class.  OM stands for "Object Model"

jasper-schneider:
Removed code gen tool.

Regarding the mocked scenario tests, the Batch OM isn't built on top of Hyak yet, so it doesn't hook into the Http interceptor.  We do have a plan to replace the Batch OM REST layer with Hyak in the future, in which case we should get this behavior.
We discussed this a bit at our last meeting, but we haven't used Hyak yet because of the way Batch auth works using signed headers.  We heard there's a supported way to add delegates to update the header, and we're in the process of investigating this, but we're not there yet.
We do have our own live scenario test system in place that we will be adding to as we move forward.
Feel free to email me and Charlie if you'd like to discuss this further.

src/ResourceManager/Batch/Commands.Batch/WorkItems/GetBatchWorkItemCommand.cs
ogail: In this case, your conv. method will return collection and you need to call write object with true flag. PowerShell will handle the single/multiple case:
```csharp
WriteObject(collection, true);
```

jasper-schneider: Hmm, I'd prefer to call WriteObject(collection, false) in the case that I'm actually returning a collection.  We have a custom enumerator that has some features that let it handle large scale data sets pretty well, such as only querying the service in chunks when the enumerator is actually being iterated through and supporting parallel iteration very well.  Walking through the entire collection up front removes some of these benefits.data into memory unnecessarily.  Would this level of logic in the cmdlet be ok?
if (name not null) { WriteObject(client.GetSingleItem()); }
else { WriteObject(client.GetCollection(), false); }

markcowl:
@azuresdkci test this please

src/ResourceManager/Batch/Commands.Batch.Test/Commands.Batch.Test.csproj
markcowl: I think it's worse than this - I don't believe we can accept this without the OM upgrading to staorage 4.0.  There will be confliucts with the other resources modules and too much possibility for regression

markcowl:
@jasper-schneider We cannot accept this PR that adds such an olde version of storage to the ARM clients.  Please upgrade to a later version of storage and submit for the next release (end of Feb).

ogail:
@jasper-schneider these tests are for guarding your cmdlets in case a change happened in PowerShell code so people can verify the change without need for pinging every single service team

PR #139, opened 2015-02-05
Remove dependency on AzureSdkInternal 

yugangw-msft:


ogail:
I hate to say this but ... the version for ADAL lib in Test.Framework packages is `2.12.111071459` while in AzurePowerShell and it's `2.11.10918.1222` so we need to align package versions somehow. I have two ideas:
- Bump PowerShell dependency to `2.12.111071459`
- Get rid of Azure PowerShell dependency on ADAL lib as Common.Authentication already have this dependency.

stankovski:
@ogail Agree - we need to bump up POSH version to 2.12. In fact we have already done it in profile branch so this PR can just target that branch...

yugangw-msft:
retest this please

yugangw-msft:
will post PR to the profile

PR #179, opened 2015-02-19
Modify HDInsight config cmdlets to use shared copy method to preserve parameters

ghost:
The following cmdlets currently operate on an input AzureHDInsightConfig object to modify or add to the configuration:
- Add-AzureHDInsightConfig
- Add-AzureHDInsightMetastore
- Add-AzureHDInsightScriptAction
- Add-AzureHDInsightStorage
- Set-AzureHDInsightDefaultStorage

The existing implementations have duplicated code in their Config parameter and some options have been missed as new configuration options are added.  Specifically, the Spark configuration is only maintained by Add-AzureHDInsightConfig but would be dropped by any of the others.

This Pull Request moves the duplicated code into a utility function (AzureHDInsightConfig.CopyFrom), and alters the implementation of each of these cmdlets to leverage the utility function.

I looked at moving the _Config_ parameter into a superclass and deriving each of these cmdlets from that class, but the fact that the Config parameter is operating on the _this.command_ object, which is a different class for each cmdlet, made that option significantly more difficult.

I also added a Unit Test _CanCallAllConfigCmdlets_PreserveAll_ that exercises the cmdlets by setting all of the possible options and verifying that they are maintained through a pipeline including all of the referenced cmdlets.  This unit test surfaced additional properties that were not being maintained properly in pipelined commands:
- ZookeeperNodeVMSize
- DataNodeVMSize
- OozieConfiguration.AdditionalActionExecutorLibraries
- OozieConfiguration.AdditionalSharedLibraries
- HBaseConfiguration.AdditionalLibraries
- HiveConfiguration.AdditionalLibraries

The CopyFrom method now copies all of these parameters.

### DIFF
--- a/src/ServiceManagement/HDInsight/Commands.HDInsight/Model/DataObjects/AzureHDInsightConfig.cs
+++ b/src/ServiceManagement/HDInsight/Commands.HDInsight/Model/DataObjects/AzureHDInsightConfig.cs
@@ -174,7 +174,7 @@ namespace Microsoft.WindowsAzure.Management.HDInsight.Cmdlet.DataObjects
             }
 
             this.ClusterSizeInNodes = value.ClusterSizeInNodes;
-            this.DefaultStorageAccount = value.DefaultStorageAccount;
+            this.DefaultStorageAccount = new AzureHDInsightDefaultStorageAccount(value.DefaultStorageAccount);
             this.AdditionalStorageAccounts.AddRange(value.AdditionalStorageAccounts);
             this.ConfigActions.AddRange(value.ConfigActions);
             this.HiveMetastore = value.HiveMetastore ?? this.HiveMetastore;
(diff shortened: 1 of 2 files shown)


azurecla:
Hi __@rickmsft__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

1mahesh:
Looks good. Can you verify the final cmdlet build to do a sanity validation?

rickmsft:
Built the debug MSI and installed, ran some tests and created an HBase in a VNET with custom metastore and mapred config settings.  Also ran some additional checks populating more of the config parameters offline without creating clusters & dumping the objects, and all look as expected.  Feel free to ping me offline if you want to see the scripts or results.

PR #182, opened 2015-02-20
Use MemoryDataStore in case of memory profile

ogail:


src/Common/Commands.Common/AzurePSCmdlet.cs
```diff
@@ -69,11 +69,20 @@ protected override void BeginProcessing()
         private void InitializeProfile()
         {
             // Load profile from disk 
-            var profileFromDisk = new AzureProfile(Path.Combine(AzureSession.ProfileDirectory, AzureSession.ProfileFile));
-            if (Profile == null ||
-                Profile.ProfilePath == profileFromDisk.ProfilePath)
+            var defaultProfile = new AzureProfile(Path.Combine(AzureSession.ProfileDirectory, AzureSession.ProfileFile));
```
markcowl: We are still readingf rom disk every time this way.  Why can't we do this only if the profile is null or matches the expected file path, and only create the defautl profile if it's actually needed?

PR #210, opened 2015-03-02
Hiding Profile property and CurrentProfile static property

markcowl:


src/Common/Commands.Common/AzurePSCmdlet.cs
```diff
@@ -27,18 +27,38 @@ namespace Microsoft.WindowsAzure.Commands.Utilities.Common
...
         private readonly RecordingTracingInterceptor _httpTracingInterceptor = new RecordingTracingInterceptor();
+        protected static AzureProfile _currentProfile = null;
 
         [Parameter(Mandatory = false, HelpMessage = "In-memory profile.")]
         public AzureProfile Profile { get; set; }
 
-        public static AzureProfile CurrentProfile { get; set; }
+        /// <summary>
+        /// Determines the current profile - the profile used when no Profile is explicitly passed in.  Should only be used by
+        /// Profile cmdlets and tests that needc to set up a particular profile
```
amarzavery: need - remove the letter c

PR #238, opened 2015-03-06
Adds new ARM cmdlets

chadiel:


azurecla:
Hi __@chadiel__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (chadiel). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

markcowl:
Shadi, it looks like what's happening is that a new API version in  the resources client is not matching the APIVersion in the recorded mocks for the providers list call used by Get-AzureLocation.  It is OK to do a search/replace on the mocks to change the API version, assuming the contracts haven't changed.

chadiel:
I actually figured out what the problem – the API version didn’t change, instead there seems to be a bit of a behavior change where the .NET SDK doesn’t append the extra ‘&’ to the request URI (seems like it was a bug fix somewhere.) Anyway, it seems like the MockHttpServer you guys use does a simple match against the URI. This has 2 problems: 1) The mock server does a simple string lookup and is not intelligent enough to handle cosmetic changes to the query such as this one and more importantly 2) https traffic has it’s URIs Base64 encoded which means that you look for the encoded version of the strings – so it’s not as simple as just doing a find and replace.

I’m going to only update the .NET SDK for the ARM library since I can (for the most part) re-record/patch those test cases to make them work but I cannot do it for all other recorded session. Will update the PR shortly.

PR #377, opened 2015-05-08
Move ARM storage account management cmdlets to a new project.

EmmaZhu:
Move ARM storage account management cmdlets to a new project.

### DIFF
--- a/src/ResourceManager/Storage/Commands.Management.Storage/Properties/AssemblyInfo.cs
+++ b/src/ResourceManager/Storage/Commands.Management.Storage/Properties/AssemblyInfo.cs
@@ -1,4 +1,18 @@
-﻿using System.Reflection;
+﻿// ----------------------------------------------------------------------------------
+//
+// Copyright Microsoft Corporation
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// ----------------------------------------------------------------------------------
+
+using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
--- a/src/ResourceManager/Storage/Commands.Management.Storage/StorageAccount/StorageManagementClient.cs
+++ b/src/ResourceManager/Storage/Commands.Management.Storage/StorageAccount/StorageManagementClient.cs
@@ -12,10 +12,10 @@
 // limitations under the License.
 // ----------------------------------------------------------------------------------
 
+using System;
 using Microsoft.Azure.Common.Authentication;
 using Microsoft.Azure.Common.Authentication.Models;
 using Microsoft.Azure.Management.Storage;
-using System;
 
 namespace Microsoft.Azure.Commands.Management.Storage
 {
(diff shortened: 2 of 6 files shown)


azurecla:
Hi __@EmmaZhu__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (emmazhu). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

src/ResourceManager/Storage/Commands.Management.Storage/StorageAccount/StorageManagementClient.cs
```diff
@@ -17,7 +17,7 @@
 using Microsoft.Azure.Management.Storage;
 using System;
 
-namespace Microsoft.Azure.Commands.Compute
+namespace Microsoft.Azure.Commands.Management.Storage
```
huangpf: Any other options?

EmmaZhu: You mean other options for namespace? We decided to use this namespace after discussed with Mark and according to his suggestion..

huangpf: Yes, saw that. But what do you think of this namespace? It seems to me it's a bit confusing to put ```Commands``` and ```Management``` together, whereas the latter one usually means management library.

It seems to me the outlier is the ```Storage Data``` cmdlets, which shall be in a specialized namespace, not the ```Storage Management``` ones.

@markcowl Thoughts?

EmmaZhu: @huangpf , can you reply in the mail thread I just forward to you? We can discuss and make a decision in the mail thread.  
BTW, maybe next time you can raise your concern earlier before I sent the PR...:-)

huangpf: As you may know most of the time it's clearer to directly look at the code. :-)

src/ResourceManager/Storage/Commands.Management.Storage.Test/TestController.cs
```diff
@@ -0,0 +1,171 @@
...
+using Microsoft.Azure.Management.Storage;
+using Microsoft.Azure.Subscriptions;
+using Microsoft.Azure.Test;
+using Microsoft.WindowsAzure.Commands.ScenarioTest;
+
+namespace Microsoft.Azure.Commands.Management.Storage.Test.ScenarioTests
+{
+    public class TestController
+    {
+        bool testViaCsm = true; // Currently set to true, we will get this from Environment varialbe.
```
huangpf: This should not be needed anymore, BTW it's always ```true```.

src/ResourceManager/Storage/Commands.Management.Storage.Test/TestController.cs
```diff
@@ -0,0 +1,171 @@
...
+        
+        public ResourceManagementClient ResourceManagementClient { get; private set; }
+
+        public SubscriptionClient SubscriptionClient { get; private set; }
+        
+        public AuthorizationManagementClient AuthorizationManagementClient { get; private set; }
+
+        public StorageManagementClient StorageClient { get; private set; }
+
+        public GalleryClient GalleryClient { get; private set; }
```
huangpf: I guess you don't really need this for SRP (and so does CRP). Just FYI.

huangpf:
LGTM, with some minor comments that are just for your information. Thanks.

src/ResourceManager/Storage/Commands.Management.Storage.Test/TestController.cs
EmmaZhu: I thought this too at the beginning. But the test framework will report an error without GalleryClient initialized. So I can just to put it here to resolve the issue.

src/ResourceManager/Storage/Commands.Management.Storage/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Reflection;
```
stankovski: Please add copyright headers to all cs files

markcowl:

PR #448, opened 2015-05-30
AzureRT - Add Dynamic Tests for Project and/or Test Runner

huangpf:


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

src/ResourceManager/Compute/Commands.Compute.Test/ScenarioTests/VMDynamicTests.cs
```diff
@@ -0,0 +1,30 @@
...
+using Microsoft.Azure.Test.HttpRecorder;
+using Microsoft.WindowsAzure.Commands.ScenarioTest;
+using Xunit;
+
+namespace Microsoft.Azure.Commands.Compute.Test.ScenarioTests
+{
+    public partial class VMDynamicTests
+    {
+        [Fact]
+        [Trait(Category.AcceptanceType, Category.CheckIn)]
```
hyonholee: Can we add another AcceptanceType for dynamic tests to distinguish runner tests from other tests?

huangpf: This test actually can be a check-in test in Playback mode.

huangpf: When the runner gets this test script and run it on the runner machine, it will be in Live mode.

huangpf:
retest this please

src/ResourceManager/Compute/Commands.Compute.Test/ScenarioTests/VMDynamicTests.ps1
```diff
@@ -0,0 +1,341 @@
...
+# limitations under the License.
+# ----------------------------------------------------------------------------------
+
+function get_all_vm_locations
+{
+    $locations = Get-AzureLocation | where { $_.Name -like 'Microsoft.Compute/virtualMachines' } | select -ExpandProperty Locations;
+
+    for ($i = 0; $i -lt $locations.Count; $i++)
+    {
+        $locations[$i] = ($locations[$i] -Replace ' ', '').ToLower();
```
hyonholee: Why do you remove space and change it to lower case?  Should the location name be used as it is?

huangpf: Yes, the canonical form (instructed by the ARM team) is to remove any white spaces and use lower cases.

src/ResourceManager/Compute/Commands.Compute.Test/ScenarioTests/VMDynamicTests.ps1
```diff
@@ -0,0 +1,341 @@
...
+    $osDiskCaching = 'ReadWrite';
+
+    $vmconfig = Set-AzureVMOSDisk -VM $vmconfig -Name $osDiskName -VhdUri $osDiskVhdUri -Caching $osDiskCaching -CreateOption FromImage;
+
+    # Image Reference;
+    $vmconfig.StorageProfile.SourceImage = $null;
+    $imgRef = Get-DefaultCRPImage;
+    $vmconfig = ($imgRef | Set-AzureVMSourceImage -VM $vmconfig);
+
+    # TODO: Remove Data Disks for now
```
hyonholee: Is it really 'TODO'?  Data disks is actually removed by the following line, so I think TODO is misleading.  Can you add more detailed comments why data disks are removed and what is to do (e.g. a bug number?).

huangpf: I can remove this.

src/ResourceManager/Compute/Commands.Compute.Test/ScenarioTests/VMDynamicTests.ps1
hyonholee: Is it a bug in ARM?  What happened if we don't remove space? I think creating a VM with 'West US' should work.

huangpf: I'm not sure whether it's a bug, and it's just safe to use this form. Creating VM is fine, but ```Get-AzureVMSize -Location "West US" would fail```.

hyonholee: I think you already fixed Get-AzureVMSize such that it works with "West US".   Is there any other cmdlet that does not change to a canonical form and fail with "West US" ?

huangpf: Good point. But there may be other cmdlets that are not compatible yet.

May I know what's your concern if only the canonical form is used?

hyonholee: If a cmdlet cannot handle a standard location name, I consider it a bug.  A customer should not be worried about canonical form and must be able to use a location name from Get-AzureLocation.  (Is there any document that a customer should convert a location name to a canonical form for a certain case?)  So, any test should not convert a standard location name into a canonical name.  It should be handled by cmdlet (or by client library or by server.)

huangpf: Well, I don't think all users must use the return from Get-AzureLocation, and for example, I may simply know that ```westus``` works, but I can understand your suggestion -- you would like to validate that the return values from ```Get-AzureLocation``` can be used here.

I would argue that the dynamic tests' focus is not yet about different forms of location strings (for which we have different test cases), as VM create/remove scenario is.

I would suggest we create a new item for follow-up.

hyonholee: Ok, I just wonder if this dynamic test fails if we use only standard location name instead of canonical name.  (If fails, where?)

AzureRT: I can try and let you know, :-)

If there is no error, I would update the PR, ;-)

huangpf: I can try and let you know, :-)

If there is no error, I would update the PR, ;-)

PR #543, opened 2015-06-25
[do not merge; review only] Add CloudException Handling to Compute Cmdlets

huangpf:


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

src/Common/Commands.Common/ComputeCloudException.cs
```diff
@@ -0,0 +1,51 @@
...
+        public ComputeCloudException(CloudException ex)
+            : base(GetErrorMessageWithRequestIdInfo(ex), ex)
+        {
+        }
+
+        protected static string GetErrorMessageWithRequestIdInfo(CloudException ex)
+        {
+            var sb = new StringBuilder(ex.Message);
+
+            if (ex.Response != null && ex.Response.Headers.ContainsKey(RequestIdHeaderInResponse))
```
hyonholee: What if ex.Response.Headers is null?

src/Common/Commands.Common/ComputeCloudException.cs
```diff
@@ -0,0 +1,51 @@
...
+        protected const string RequestIdHeaderInResponse = "x-ms-request-id";
+
+        public ComputeCloudException(CloudException ex)
+            : base(GetErrorMessageWithRequestIdInfo(ex), ex)
+        {
+        }
+
+        protected static string GetErrorMessageWithRequestIdInfo(CloudException ex)
+        {
+            var sb = new StringBuilder(ex.Message);
```
hyonholee: ex cannot be null?

src/ServiceManagement/Services/Commands.Utilities/Common/ServiceManagementBaseCmdlet.cs
```diff
@@ -215,11 +215,18 @@ protected OperationStatusResponse GetOperationNewSM(string operationId)
...
+                    {
+                        throw new ComputeCloudException(ce);
+                    }
                 }
-                catch (CloudException ex)
+                catch (Exception ex)
                 {
-                    WriteExceptionDetails(ex);
+                    WriteExceptionError(ex);
                 }
```
hyonholee: Can you avoid a nested try-catch?

src/Common/Commands.Common/ComputeCloudException.cs
huangpf: It's a ```get``` only field, so it can be empty, but I don't think it will be ```null``` : ```public IDictionary<string, IEnumerable<string>> Headers { get; }```

src/ServiceManagement/Services/Commands.Utilities/Common/ServiceManagementBaseCmdlet.cs
huangpf: This is expected - capturing the cloud exception and re-throwing the wrapped one. May I know why you think it should be avoided?

hovsepm:
Please reopen it towards dev branch.

PR #596, opened 2015-07-13
Reorder child resource output

DeepakRajendranMsft:
Reordering Child resource output. 

This is a just "good to have" in the current cycle, so doesnt need to be merged if it is going to delay the msi release.

### DIFF
--- a/src/ResourceManager/Network/Commands.Network/Models/PSBackendAddressPool.cs
+++ b/src/ResourceManager/Network/Commands.Network/Models/PSBackendAddressPool.cs
@@ -21,10 +21,13 @@ namespace Microsoft.Azure.Commands.Network.Models
 
     public class PSBackendAddressPool : PSChildResource
     {
+        [JsonProperty(Order = 1)]
         public List<PSResourceId> BackendIpConfigurations { get; set; }
 
+        [JsonProperty(Order = 1)]
         public List<PSResourceId> LoadBalancingRules { get; set; }
 
+        [JsonProperty(Order = 1)]
         public string ProvisioningState { get; set; }
 
         [JsonIgnore]
--- a/src/ResourceManager/Network/Commands.Network/Models/PSFrontendIpConfiguration.cs
+++ b/src/ResourceManager/Network/Commands.Network/Models/PSFrontendIpConfiguration.cs
@@ -21,18 +21,25 @@ namespace Microsoft.Azure.Commands.Network.Models
 
     public class PSFrontendIPConfiguration : PSChildResource
     {
+        [JsonProperty(Order = 1)]
         public string PrivateIpAddress { get; set; }
 
+        [JsonProperty(Order = 1)]
         public string PrivateIpAllocationMethod { get; set; }
 
+        [JsonProperty(Order = 1)]
         public PSResourceId Subnet { get; set; }
 
+        [JsonProperty(Order = 1)]
         public PSResourceId PublicIpAddress { get; set; }
 
+        [JsonProperty(Order = 1)]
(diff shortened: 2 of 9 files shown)


azurecla:
Hi __@DeepakRajendranMsft__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>You've already signed the contribution license agreement. Thanks!</span>
        <span>We will now validate the agreement and then real humans will evaluate your PR.</span>

TTYL, AZPRBOT;

src/ResourceManager/Network/Commands.Network/Models/PSBackendAddressPool.cs
```diff
@@ -21,10 +21,13 @@ namespace Microsoft.Azure.Commands.Network.Models
 
     public class PSBackendAddressPool : PSChildResource
     {
+        [JsonProperty(Order = 1)]
```
markcowl: How does making every property Order=1 change the behavior?  Do you have a test for the new serialization?

DeepakRajendranMsft:
@markcowl 
Since PSBackendaddresspool, inherits from PSChildResource, the properties from PSBackendAddressPool gets printed first followed by PSChildResource (contains name, id, etag), like 
BackendAddressPools      : [
                             {
                               "BackendIpConfigurations": [],
                               "LoadBalancingRules": [],
                               "ProvisioningState": "Succeeded",
                               "Name": "bepool1",
                               "Etag": "W/\"971e92ac-7882-4539-8950-18df75644e63\"",
                               "Id": "/subscriptions/9532a63e-f2eb-4649-bb23-5ed01077ce80/resourceGroups/ptest/provider
                           s/Microsoft.Network/loadBalancers/lbname1/backendAddressPools/bepool1"
                             }
                           ]

However, we want the output to be 

BackendAddressPools      : [
                             {
                          "Name": "bepool1",
                               "Etag": "W/\"971e92ac-7882-4539-8950-18df75644e63\"",
                               "Id": "/subscriptions/9532a63e-f2eb-4649-bb23-5ed01077ce80/resourceGroups/ptest/provider
                           s/Microsoft.Network/loadBalancers/lbname1/backendAddressPools/bepool1"
                             
                               "BackendIpConfigurations": [],
                               "LoadBalancingRules": [],
                               "ProvisioningState": "Succeeded",
                             }  
                           ]

By default, the order of a property is 0 and when you serialize an object in json, it serializes in the ascending order. So, we are now changing the Jsonproperties of the derived classes to be 1, so that the base class properties get printed first.

Regarding test, since this is only a formatting change and not a change to the output object, i am not sure what kind of test i can write. (i can write to a stream and read from it, but that seems to be overkill since we are not changing the property value but only the order)

DeepakRajendranMsft:
I just realized that the better way to fix this is in PSchildResource rather than the individual derived classes. Will update the PR shortly.

markcowl:
It seems like writing the object to and reading from a memory stream would be the best way to test this.  I won't hold up the PR on this, though, please just add this test the next time you are updatign tests oc cmdlets.

DeepakRajendranMsft:
@markcowl  Thanks, will add a test to verify the serialization later.

Also, this PR is good as it is. Changing on the base class doesnt work. So we are good to merge now.

Regarding reading and verifying from stream, we should have a bigger conversation with xrp folks since we dont have any tests which verify the formal.xml files either.

markcowl:
@DeepakRajendranMsft This is just protection from regression - you'll need to make sure to do this with each derived class from this point forward.

PR #602, opened 2015-07-14
Fix signing issues with api management library

markcowl:


markcowl:

PR #688, opened 2015-08-03
fix a bad merge on posh solution file which causes recovery commands not load

yugangw-msft:


yugangw-msft:
reviewed by jianghao

PR #698, opened 2015-08-04
Use the latest posh version in the test mock file

yugangw-msft:
This won't reset the release, and just want to get all in sync before integrate to the master branch


yugangw-msft:
also i update the release document, so to find and replace "v<version>", not just "<version>"

PR #806, opened 2015-08-26
Add Launch option to Get-RDP command

DeepakRajendranMsft:


azurecla:
Hi __@DeepakRajendranMsft__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>You've already signed the contribution license agreement. Thanks!</span>
        <span>We will now validate the agreement and then real humans will evaluate your PR.</span>

TTYL, AZPRBOT;

ogail:
Update release notes

DeepakRajendranMsft:
@ogail updated release notes

pcgeek86:
@DeepakRajendranMsft Thank you -- I checked out the PR and the parameter sets look good.

However, I'm curious about the implementation of the  mstsc.exe  invocation. Is there a reason that you're wrapping the invocation inside a batch file? Isn't that going to leave a bunch of  .bat  files on the user's filesystem?

Cheers,
Trevor Sullivan
Microsoft MVP: PowerShell

DeepakRajendranMsft:
@pcgeek86. thanks for the feedback,

TBH, I dont have a good explanation for this. this is the same implementation as in the existing non-arm cmdlet

pcgeek86:
@DeepakRajendranMsft Cool, thanks for validating. Seems like a good opportunity to fix it then. Right? :)

ogail:
On demand job: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-on-demand/149/

DeepakRajendranMsft:
The batch file takes care of deleting the rdp and the batch file after mstsc is invoked so we should be okay right. (?)

pcgeek86:
Gotcha, I guess I just don't see a need to create the batch file in the first place. It's adding an additional layer of overhead that isn't necessary. Just invoke `mstsc.exe` directly from the C# code. Does that create any challenges? I think it would be simpler.

Cheers,
Trevor Sullivan

ogail:
On demand #2: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-on-demand/150/

DeepakRajendranMsft:
@pcgeek86 yes i agree with you, we can invoke mstsc.exe directly. but the batch file is in there to delete the temporary rdp file(after the user is done with the RDP) that will be created when the user doesnt specify the "-localpath". 

If it is okay to leave the temp rdp file in the temp dir, if so we can remove the batch file (or is there a better way to deal with the temp rdp file)

pcgeek86:
@DeepakRajendranMsft If the user downloads the file using the `-LocalPath` parameter, then we don't need to worry about cleaning it up at all, agreed. I'm only talking about the `Launch` parameter set here. You can delete the RDP temp file using C#, right? Why do we need a batch file to delete it? Just use `System.IO.File.Delete()` to kill off the RDP file.

https://msdn.microsoft.com/en-us/library/system.io.file.delete(v=vs.110).aspx

Cheers,
Trevor Sullivan
Microsoft MVP: PowerShell

pcgeek86:
Oh wait, I think I see what you're saying. The batch file lifetime is outside the RDP process lifetime. When the `mstsc.exe` process exits, the batch file removes the RDP file and exits as well. Since the C# code is not running synchronously with the RDP session, it is unable to handle this. I understand now. Nevermind, leave it as-is. It was my misunderstanding.

Cheers,
Trevor Sullivan
Microsoft MVP: PowerShell

pcgeek86:
Actually wait, once the RDP file is invoked, does `mstsc.exe` still have a file lock on it? If not, I would just kick off `mstsc.exe` from C#, and then immediately delete the RDP file. There's no need to keep the RDP file around for the whole RDP session, is there? Of course, if `mstsc.exe` does have a lock on the file, then your existing solution is probably the simplest.

DeepakRajendranMsft:
hmmm let me try that (i dont there is a lock), will confirm

pcgeek86:
@DeepakRajendranMsft Thank you -- appreciate your discussion around this topic.

DeepakRajendranMsft:
@pcgeek86  mstsc doesnt seem to take a lock. but it looks like i need to put a sleep after the process has started and before i delete the file. (since the file sometimes gets deleted before we delete the process. I didnt find any api on process to wait on start). So i think the current implementation is probably the best and simple.

example of waiting:
bool found = 0;
while (!found)
{
    foreach (Process clsProcess in Process.GetProcesses())
        if (clsProcess.Name == Name)
            found = true;

    Thread.CurrentThread.Sleep(1000);
}

or use Process.waitForIdleTimeout()

Issue #387, opened 2015-05-14
Calling Set-AzureServiceDiagnosticsExtension with -Role failing


craig-wagner:
Based on the version number showing in "Programs and Features" I'm running 0.9.1 of the Azure Powershell. I installed it yesterday (2015/05/13) using the Web Platform Installer.

    $storage_name = "mystorageaccountname"
    $key = "mykey"
    $service_name = "cwagner-ipseity-worker"
    $public_config = "C:\tfs\SI\ipseity\trunk\BuildArtifacts\Deployment\Packages\AzureWorker\PaaSDiagnostics.Ipseity.Server.MessageProcessor.AzureWorkerRole.PubConfig.xml" 
    $storageContext = New-AzureStorageContext -StorageAccountName $storage_name -StorageAccountKey $key
    Set-AzureServiceDiagnosticsExtension -StorageContext $storageContext -DiagnosticsConfigurationPath $public_config -ServiceName $service_name -Slot Production -Role Ipseity.Server.MessageProcessor.AzureWorkerRole

I get the following output:

> VERBOSE: Setting PaaSDiagnostics configuration for Ipseity.Server.MessageProcessor.AzureWorkerRole.
> Set-AzureServiceDiagnosticsExtension : BadRequest: The extension ID
IpseityServerMessageProcessorAzureWorkerRole-PaaSDiagnostics-Production-Ext-0 is invalid.
> At line:1 char:1
> + Set-AzureServiceDiagnosticsExtension -StorageContext $storgeContext -Diagnostics ...
> + > ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>    + CategoryInfo          : NotSpecified: (:) [Set-AzureServiceDiagnosticsExtension], CloudException
>    + FullyQualifiedErrorId : Hyak.Common.CloudException,Microsoft.WindowsAzure.Commands.ServiceManagement.Extensions.
   SetAzureServiceDiagnosticsExtensionCommand

I realize this is similar to https://github.com/Azure/azure-powershell/issues/72 (dots in the role name) but it is my understanding that that issue has been fixed and the reported extension ID in error has had the dots stripped.

If we don't specify the `-Role` parameter it works.

devigned:
@matt-gibbs or @markcowl thoughts on this one?

bgambs:
Similar issue here; 

[...]

PR #905, opened 2015-09-14
AzureRT S54 PR #2 - Record Debug Messages & Fix Add-AzureVhd

huangpf:


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

huangpf:
Fix #387

src/Common/AzurePSCmdlet.cs
```diff
@@ -254,7 +257,7 @@ protected bool IsVerbose()
 
         protected new void WriteError(ErrorRecord errorRecord)
         {
-            FlushDebugMessages();
+            FlushDebugMessages(IsDataCollectionAllowed());
```
ogail: no need to pass `IsDataCollectionAllowed()` just use it in `FlushDebugMessages`

src/Common/AzurePSCmdlet.cs
```diff
@@ -362,8 +365,13 @@ protected void SafeWriteOutputPSObject(string typeName, params object[] args)
             WriteObject(customObject);
         }
 
-        private void FlushDebugMessages()
+        private void FlushDebugMessages(bool record = false)
         {
+            if (record)
```
ogail:
```csharp
if (IsDataCollectionAllowed())
{
    RecordDebugMessages();
}
```

ogail:
added a nit

src/Common/AzurePSCmdlet.cs
huangpf: You mean by default log everything?

src/Common/AzurePSCmdlet.cs
huangpf: ok

src/Common/AzurePSCmdlet.cs
huangpf: But we only want to log when there is an exception. Do you mean you want to log everything if data collection is allowed?

src/Common/AzurePSCmdlet.cs
huangpf: same above

PR #918, opened 2015-09-16
Added Get-AzureRMTenant [#102955260]

hovsepm:


src/Common/Commands.ResourceManager.Profile.Test/TenantCmdletTests.cs
```diff
@@ -0,0 +1,119 @@
+﻿// ----------------------------------------------------------------------------------
```
markcowl: I think we need to add some mocked tests for this - we can do it using the existing infrastructure, or using Moq infrastructure (I'm checking some in with the subscription cmdlet)

src/Common/Commands.ResourceManager.Profile/Tenant/GetAzureRMTenant.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// Cmdlet to get user tenant information. 
+    /// </summary>
+    [Cmdlet(VerbsCommon.Get, "AzureRMTenant")]
+    [Alias("Get-AzureRMDomain")]
+    [OutputType(typeof(List<AzureTenant>))]
+    public class GetAzureRMTenantCommand : AzureRMCmdlet
+    {
+        [Alias("Domain")]
+        [ValidateNotNullOrEmpty]
+        public string Tenant { get; set; }
```
markcowl: This needs a parameter attribute - it should be Mandatory=False, ValueFromPipelineByPropertyName = true

src/Common/Commands.ResourceManager.Profile/Tenant/GetAzureRMTenant.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Management.Automation;
+
+namespace Microsoft.Azure.Commands.Profile
+{    
+    /// <summary>
+    /// Cmdlet to get user tenant information. 
+    /// </summary>
+    [Cmdlet(VerbsCommon.Get, "AzureRMTenant")]
+    [Alias("Get-AzureRMDomain")]
+    [OutputType(typeof(List<AzureTenant>))]
```
markcowl: I believe this should just be AzureTenant

src/Common/Commands.ResourceManager.Profile/Tenant/GetAzureRMTenant.cs
hovsepm: addressed.

src/Common/Commands.ResourceManager.Profile/Tenant/GetAzureRMTenant.cs
hovsepm: According to the task details - "If no tenant is specified, list all tenants.". So it can be a list.

src/Common/Commands.ResourceManager.Profile.Test/TenantCmdletTests.cs
hovsepm: Done.

PR #922, opened 2015-09-17
Changed SelectAzureRMContext to SetAzureRMContext.

hovsepm:


stankovski:
Looks good. We can add piping of AzureSubscription scenario later.

src/Common/Commands.ResourceManager.Common/RMProfileClient.cs
```diff
@@ -78,55 +78,34 @@ public AzureRMProfile Login(AzureAccount account, AzureEnvironment environment,
...
-                   null,
-                   promptBehavior,
-                   TokenCache.DefaultShared);
-
-                account.Properties[AzureAccount.Property.Tenants] = accessToken.TenantId;
-                newTenant = new AzureTenant();
-                newTenant.Id = new Guid(accessToken.TenantId);
-                newTenant.Domain = accessToken.GetDomain();
-                _profile.Context = new AzureContext(account, envrionment, newTenant);
+                _profile.Context = new AzureContext(
```
stankovski: We should also update tenantId in context.Account and context.Subscription

hovsepm: Done.

PR #954, opened 2015-09-22
[#102955616, #103828190] Add output types for profile cmdlets and fix…

markcowl:
… piping and display scenarios


markcowl:

PR #977, opened 2015-09-25
Support to Build and test a specific cmdlet project through msbuild property of "scope"

yugangw-msft:


src/ApiManagement.sln
```diff
@@ -0,0 +1,100 @@
+Microsoft Visual Studio Solution File, Format Version 12.00
```
markcowl: I think we want the solutions in the individual project directories

yugangw-msft: yep, that is the plan, and will be in the next PR. For now, this PR was just to get the functionalities done and less chance to conflict with other people's PR

PR #986, opened 2015-09-26
Fix sign build break at "ps-gallery-sign"

yugangw-msft:
The failure of "... no strong name" was caused by the project "Commands.ServiceManagement.Common.csproj" was not in the "ResourceManager.sln",
and this caused the msbuild to pick default configure of "debug" which produced 
unsigned assembly.
I also fixed the build configuration mismatch error in commands.profile.csproj


yugangw-msft:
reviewed with Denis and Hovsep

PR #1086, opened 2015-10-08
Fixes for dsc and storage cmdlets

markcowl:


markcowl:
on demand:  http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-on-demand/237/

ogail:
Nit: would be good to add a small unit test to make sure this code works as expected and also does not throw upon failure.

PR #1120, opened 2015-10-15
Update AzureRM.psm1

rchaganti:
Modified -Force to be a switch parameter to bring it inline with the PowerShell way of using Force.

### DIFF
--- a/tools/AzureRM/AzureRM.psm1
+++ b/tools/AzureRM/AzureRM.psm1
@@ -38,8 +38,11 @@ function Test-AdminRights([string]$Scope)
   }
 }
 
-function CheckIncompatibleVersion([bool]$Force)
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@rchaganti__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

markcowl:
@rchaganti Thanks for your contribution. This already is a switch parameter in the actually exported function.  The code change you provided results in always throwing from this method.  Closing for now.  If you'd like to make a contribution that did not export this function, that would be acceptable, or one that propagated the -Force parameter from the calling function.

Issue #902, opened 2015-09-12
Disable-azuredatacollection prompts if you want to disable azure data collection in non-interactive session


TravisEz13:
Disable-azuredatacollection prompts if you want to disable azure data collection in non-interactive session

# Repro

```powershell
del "$env:appdata\Windows Azure Powershell\AzureDataCollectionProfile.json" -Force -ErrorAction Ignore
powershell -NonInteractive -command Disable-azuredatacollection
```

# Expected results
Data collection is disabled without prompting (since there is no parameter to suppress prompting)

# Actual results

> prompt> powershell -NonInteractive -command Disable-azuredatacollection
>WARNING: Microsoft Azure PowerShell collects data about how users use PowerShell cmdlets and some problems they encounter.  Microsoft uses this information to improve our PowerShell
>cmdlets.  Participation is voluntary and when you choose to participate your device automatically sends information to Microsoft about how you use Azure PowerShell.
>
>If you choose to participate, you can stop at any time by using Azure PowerShell as follows:
>1. Use the Disable-AzureDataCollection cmdlet to turn the feature Off. The cmdlet can be found in the AzureResourceManager module
>To disable data collection: PS > Disable-AzureDataCollection
>
>If you choose to not participate, you can enable at any time by using Azure PowerShell as follows:
>1. Use the Enable-AzureDataCollection cmdlet to turn the feature On. The cmdlet can be found in the AzureResourceManager module
>To enable data collection: PS > Enable-AzureDataCollection
>
>Select Y to enable data collection [Y/N]:
>WARNING: You choose not to participate in Microsoft Azure PowerShell data collection.
>WARNING: The setting profile has been saved to the following path 'C:\Users\tplunk\AppData\Roaming\Windows Azure Powershell\AzureDataCollectionProfile.json'.
>WARNING: The setting profile has been saved to the following path 'C:\Users\tplunk\AppData\Roaming\Windows Azure Powershell\AzureDataCollectionProfile.json'.

devigned:
We will address this issue. Thank you for reporting this issue.

thoutekier:
Is there already any feedback on this?  In encounter the same issue when using cmdlet New-AzureStorageContext in a powershell script executed by a Team Foundation Build 2015 Powershell task.  As a consequence, the build is blocked and times out (or should be cancelled).

devigned:
@thoutekier this should not cause the build to timeout. At the worst, the cmd should prompt and wait for only 60 seconds or so.

[...]

PR #1140, opened 2015-10-21
[#103424390] Disable prompts in non interactive mode

hovsepm:
Fix for issue #902


markcowl:
@hovsepm I wonder if this is the right fix.  SInce the Enable/Disable cmdlets are just setting a value in a  file, do they need to inherit from AzurePSCmdlet at all?

markcowl:
after discussion, lgtm

PR #1142, opened 2015-10-21
Role commands bugbash fixes 

shuagarw:
Role commands bugbash fixes


azurecla:
Hi __@shuagarw__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (shuagarw). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

ogail:
Is it guranteed that the name will match with one role def only?

shuagarw:
Name can match with multiple role definitions(edge case), but when deleting we ensure that if only 1 role is matched, then only delete, otherwise an error is thrown asking user to specify role by Id, as multiple roles with same name are found. 
So yes, when deleting a role, name will match to 1 role def only, otherwise it will be an error.

PR #1151, opened 2015-10-21
[#105517922] Update Select-AzureRmSubscription to pass in subscriptionId only or subscriptionName only

hovsepm:


src/ResourceManager/Profile/Commands.Profile/Context/SetAzureRMContext.cs
```diff
@@ -58,14 +56,40 @@ protected override void ProcessRecord()
...
             {
                 var profileClient = new RMProfileClient(AzureRmProfileProvider.Instance.Profile);
-                profileClient.SetCurrentContext(SubscriptionId, TenantId);
-                if (!string.IsNullOrWhiteSpace(SubscriptionName))
+                AzureSubscription subscription = null;
+                string tenantId = AzureRmProfileProvider.Instance.Profile.Context.Tenant.Id.ToString();
+
+                if (string.IsNullOrWhiteSpace(TenantId))
+                {
+                    WriteWarning(
```
markcowl: I think this should be WriteVerbose - the warning will cause some scripts to fail, depending on WarningPreference variable

src/ResourceManager/Profile/Commands.Profile/Context/SetAzureRMContext.cs
```diff
@@ -58,14 +56,40 @@ protected override void ProcessRecord()
...
+                    throw new ItemNotFoundException(
+                        string.Format(Resources.SubscriptionNameNotFoundError, SubscriptionName));
+                }
+
+                profileClient.SetCurrentContext(subscription.Id.ToString(), tenantId, verifySubscription: false);
+                AzureRmProfileProvider.Instance.Profile.Context.Subscription.Name = SubscriptionName;
+            }
+            else if (ParameterSetName == SubscriptionIdParameterSet)
+            {
+                var profileClient = new RMProfileClient(AzureRmProfileProvider.Instance.Profile);
```
markcowl: In this case, I think we need to do the same validation - make a call to get the subscription by Id and fill in the name - right now this will create a context with no subscription name.

markcowl:
One comment, also, wer need to add a test looking up subscription by name

Issue #1124, opened 2015-10-15
A parameter name for tenants should be consistent


alexandair:
Is it -Tenant or -TenantId? Some cmdlets use -Tenant, some others -TenantId:

```
PS C:\> gcm -module * -ParameterName tenant

CommandType     Name                                               Version    Source         
-----------     ----                                               -------    ------         
Alias           Login-AzureRmAccount                               0.10.0     AzureRM.profile
Cmdlet          Add-AzureAccount                                   0.9.11     Azure          
Cmdlet          Add-AzureRmAccount                                 0.10.0     AzureRM.profile
Cmdlet          Get-AzureRmTenant                                  0.10.0     AzureRM.profile
Cmdlet          New-AzureProfile                                   0.9.11     Azure          

PS C:\WINDOWS\system32> gcm -module * -ParameterName tenantid

CommandType     Name                                               Version    Source         
-----------     ----                                               -------    ------         
Alias           Select-AzureRmSubscription                         0.10.0     AzureRM.profile
Cmdlet          Get-AzureRmSubscription                            0.10.0     AzureRM.profile
Cmdlet          Set-AzureRmContext                                 0.10.0     AzureRM.profile

```

markcowl:
@alexandair  We can at least be consistent in the new cmdlets.

alexandair:
You can fix it for old cmdlets too. Just use parameter aliases.


Sent from a toaster.





On Tue, Oct 20, 2015 at 6:32 PM -0700, "Mark Cowlishaw" <notifications@github.com> wrote:
@alexandair  We can at least be consistent in the new cmdlets.

---

hovsepm:
Fixed in Dev branch. Will be available in the upcoming release.

PR #1155, opened 2015-10-22
[#106171538] Changed Tenant to TenantId in cmdletst

hovsepm:
Addressed issue #1124

### DIFF
--- a/src/ResourceManager/Profile/Commands.Profile/Account/AddAzureRmAccount.cs
+++ b/src/ResourceManager/Profile/Commands.Profile/Account/AddAzureRmAccount.cs
@@ -127,9 +127,9 @@ namespace Microsoft.Azure.Commands.Profile
                 password = Credential.Password;
             }
 
-            if (!string.IsNullOrEmpty(Tenant))
+            if (!string.IsNullOrEmpty(TenantId))
             {
-                azureAccount.SetProperty(AzureAccount.Property.Tenants, new[] { Tenant });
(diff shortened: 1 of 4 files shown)


src/ResourceManager/Profile/Commands.Profile/Account/AddAzureRmAccount.cs
```diff
@@ -48,8 +48,9 @@ public class AddAzureRMAccountCommand : AzureRMCmdlet , IModuleAssemblyInitializ
         [Parameter(ParameterSetName = "User", Mandatory = false, HelpMessage = "Optional tenant name or ID")]
         [Parameter(ParameterSetName = "ServicePrincipal", Mandatory = true, HelpMessage = "TenantId name or ID")]
         [Parameter(ParameterSetName = "AccessToken", Mandatory = false, HelpMessage = "TenantId name or ID")]
+        [Alias("Tenant")]
         [ValidateNotNullOrEmpty]
-        public string Tenant { get; set; }
+        public string TenantId { get; set; }
```
ogail: Why this is the only cmdlet that has parameter name as `TenantId` and alias as `Tenant` while others are vice versa?

ogail:
LGTM

src/ResourceManager/Profile/Commands.Profile/Account/AddAzureRmAccount.cs
hovsepm: We left Tenant for ASM with alias TenantId and did the opposite for ARM.

PR #1188, opened 2015-10-28
Disable sql tests to re-enable post-check-in run

markcowl:


markcowl:
onm demand run here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-on-demand/250/

PR #1199, opened 2015-10-29
[AzureRT] [Hot Fix] Fix the bug of New-AzureRmVM

hyonholee:
This update fixes the following bugs:
1. [Hot Fix - Very Important] New-AzureRmVM fails when OS disk uses a storage account created in a different resource group from that of the VM.
2. In Add-AzureRmVMDataDisk, VhdUri should be mandatory parameter.

### DIFF
--- a/src/ResourceManager/Compute/Commands.Compute/VirtualMachine/Operation/NewAzureVMCommand.cs
+++ b/src/ResourceManager/Compute/Commands.Compute/VirtualMachine/Operation/NewAzureVMCommand.cs
@@ -111,10 +111,15 @@ namespace Microsoft.Azure.Commands.Compute
                 }
                 catch (Exception e)
                 {
-                    if (!e.Message.Contains("ResourceNotFound"))
+                    if (e.Message.Contains("ResourceNotFound"))
                     {
                         WriteWarning(string.Format(
-                            Properties.Resources.ErrorDuringGettingStorageAccountForBootDiagnostics, storageAccountName));
+                            Properties.Resources.StorageAccountNotFoundForBootDiagnostics, storageAccountName));
+                    }
+                    else
+                    {
+                        WriteWarning(string.Format(
+                            Properties.Resources.ErrorDuringGettingStorageAccountForBootDiagnostics, storageAccountName, e.Message));
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


azurecla:
Hi __@hyonholee__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (hylee). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

src/ResourceManager/Compute/Commands.Compute/VirtualMachine/Operation/NewAzureVMCommand.cs
```diff
@@ -100,11 +100,22 @@ private Uri GetOrCreateStorageAccountForBootDiagnostics()
...
+                    var storageAccountResponse = storageClient.StorageAccounts.GetProperties(this.ResourceGroupName,
+                        storageAccountName);
+                    if (!storageAccountResponse.StorageAccount.AccountType.Equals(AccountType.PremiumLRS))
+                    {
+                        return storageAccountResponse.StorageAccount.PrimaryEndpoints.Blob;
+                    }
+                }
+                catch (Exception e)
+                {
+                    if (!e.Message.Contains("ResourceNotFound"))
```
huangpf: Exceptions are swallowed here. Need to throw it and/or output the error text.

huangpf:
LGTM

src/ResourceManager/Compute/Commands.Compute/VirtualMachine/Operation/NewAzureVMCommand.cs
hyonholee: The fix is to swallow every exception during getting storage account.  If the exception is "ResourceNotFound", then the given storage account name is not in the same resource group, so no warning message is shown.  If the exception is other than "ResourceNotFound", we show a warning message, but keeps going, because we can still use other storage account for boot diagnostics.

markcowl:
on demand run here:http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-on-demand/282/

LGTM once the on-demand run passes

markcowl:
still LGTM

PR #1203, opened 2015-10-29
Updating wix installer

markcowl:
- updating wix installer for new dlls


markcowl:
on demand run here: https://github.com/Azure/azure-powershell/pull/1203

PR #1211, opened 2015-10-30
Updating version numbers to 1.0

markcowl:


markcowl:
on demand run here: http://azuresdkci.cloudapp.net/job/powershell-on-demand/292/

PR #1237, opened 2015-11-04
Fix for broken remove-datasource cmdlet 

begoldsm:
This fixes the broken remove data source cmdlet, which had inverted
conditional logic that made it impossible to remove either type of data
source.
In addition to this fix, this change includes test cases specific to
testing adding and removing data sources, to ensure this doesn't happen
again.
Finally, all tests were run against production for both DataLake\* cmdlet
services so that we have production recordings.


markcowl:

markcowl:
on demand run here:  http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/324/

PR #1252, opened 2015-11-07
MSFT 4811945: [LIVESITE BLOCKING][HOTFIX] VMProvisioning: Collection creation often fails because service-account limit on domain-join too low, and we don't clean-up AD accounts

yadavbdev:


markcowl:
@yadavbdev Is Ssytem.DIrectoryServices always in the GAC, for all versions of .Net 4+ or does this put additional restrictions on the framework target?

vadimp-msft:
It's a standard lib, like System.dll, so yes, as long as .NET is installed, it is expected in GAC.

markcowl:
@yadavbdev Merging.  Please send a separate PR with Changelog.txt changes

PR #1273, opened 2015-11-11
Fixing getting vault case sensitivity in Get-AzureRmKeyVault and code cleanup

pomortaz:
1. Fixed the case sensitivity issue when a vault is retrieved using Get-AzureRmKeyVault and without providing the ResourceGroupName property.
2. Code cleanup.
3. Added test


azurecla:
Hi __@pomortaz__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (pomortaz). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

setup/azurecmdfiles.wxi
```diff
@@ -418,15 +418,6 @@
                         <Component Id="cmp4B5A611FF761BA928A53E0E56FB58D78" Guid="*">
```
markcowl: There is no reason for this file to be updated, please remove

markcowl:
LGTM, other than the wix changes

setup/azurecmdfiles.wxi
pomortaz: I have build failures on my clean devbox because it could not find the dependencies that I removed here. After removing them, the build got fixed. I will revert this, but it looks like these dependencies are not needed and should be removed.

markcowl:
@pomortaz Please add unit tests to ensure no regressions in this behavior

PR #1285, opened 2015-11-12
Fix a bug in filtering logic for job retrieval

begoldsm:
The formatting for the datetimeoffset string was invalid in some cases. Fixing it to explicitly format the datetimestring as a datetime offset.


markcowl:
@begoldsm  No test coverage?

PR #1292, opened 2015-11-13
retrieve latest dsc version in dsc tests

eshaparmar:


markcowl:
LGTM once the build passes.

PR #1294, opened 2015-11-13
[AzureRT] Sprint 56 update

hyonholee:
1. Enable BGInfo by default
2. Fix the issue when an OS disk is in a different resource group - Now New-AzureRmVM does not create a new storage account for boot diagnostics.
3. Add a new test when an empty list for additional unattend content is set for Update-AzureRmVM.
4. Re-record all compute tests because of New-AzureRmVM behavior change.

### DIFF
--- a/src/ResourceManager/Compute/Commands.Compute/Common/ComputeClientBaseCmdlet.cs
+++ b/src/ResourceManager/Compute/Commands.Compute/Common/ComputeClientBaseCmdlet.cs
@@ -63,7 +63,15 @@ namespace Microsoft.Azure.Commands.Compute
             }
             catch (CloudException ex)
             {
-                base.EndProcessing();
+                try
+                {
+                    base.EndProcessing();
+                }
+                catch
+                {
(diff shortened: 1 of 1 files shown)


markcowl:
on demand run here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/332/

src/ResourceManager/Compute/Commands.Compute/Common/ComputeClientBaseCmdlet.cs
```diff
@@ -63,6 +63,7 @@ protected void ExecuteClientAction(Action action)
             }
             catch (CloudException ex)
             {
+                base.EndProcessing();
```
huangpf: What if ```EndProcessing()``` shows exception? In that case, ```ComputeCloudException``` will be missed.

huangpf:
LGTM, other than 1 question about ```EndProcessing```.

PR #1333, opened 2015-11-20
Update test version Add DataLake tests

begoldsm:
Updated the test version to include generation of GUIDs, if necessary
Added tests for catalog, filesystem and permissions for DataLake
Updated catalog to support views procedures credentials types and get
secret
Updated file cmdlets to populate fields in a readable way for users.


markcowl:
on demand run here:  http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/341/

@begoldsm There is at least one very long looking recording file name - (AdlaTests).  Hopefully this will fit, but maybe shortening the namespace would help ensure it

PR #1395, opened 2015-12-02
Fix set resource

vivsriaus:
Do not call get when using patch semantics


vivsriaus:
On demand job here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/381/

PR #1469, opened 2015-12-10
Fixing broken PS command let:- Get-AzureRmVpnClientPackage

Nilambari:
Fixing the broken PS command let:- Get-AzureRmVpnClientPackage by adding code to retrieve actual contents from get locationResults response. We will fix this API properly in next release.

### DIFF
--- a/src/ResourceManager/Network/Commands.Network/Common/NetworkClient.cs
+++ b/src/ResourceManager/Network/Commands.Network/Common/NetworkClient.cs
@@ -54,7 +54,6 @@ namespace Microsoft.Azure.Commands.Network
         public NetworkClient()
         {
         }
-
         public string Generatevpnclientpackage(string resourceGroupName, string virtualNetworkGatewayName, VpnClientParameters parameters)
         {
             return Task.Factory.StartNew(() => GeneratevpnclientpackageAsync(resourceGroupName, virtualNetworkGatewayName, parameters)).Unwrap().GetAwaiter().GetResult();
@@ -139,7 +138,9 @@ namespace Microsoft.Azure.Commands.Network
             }
             // Send Request
             cancellationToken.ThrowIfCancellationRequested();
-            HttpClient httpClient = new HttpClient();
+
+            var client = this.NetworkManagementClient as NetworkManagementClient;
+            HttpClient httpClient = client.HttpClient;
             HttpResponseMessage httpResponse = await httpClient.SendAsync(httpRequest, cancellationToken).ConfigureAwait(false);
 
             HttpStatusCode statusCode = httpResponse.StatusCode;


Nilambari:
On demand job:- http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/416/

Nilambari:
@markcowl  As per our yesterday’s discussion, I have fixed the broken PS command let:- Get-AzureRmVpnClientPackage for now by adding code in Powershell repository itself to retrieve actual contents from get locationResults response. With this fix, now this API seem works fine. We will fix this API properly in next release. Can you Please help accept this PR to include in current release? Thank you very much!

src/ResourceManager/Network/Commands.Network/VirtualNetworkGateway/GenerateAzureVpnClientPackage.cs
```diff
@@ -65,10 +78,164 @@ public override void ExecuteCmdlet()
...
+
+            // Set Credentials
+            if (this.NetworkClient.NetworkManagementClient.Credentials != null)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                await this.NetworkClient.NetworkManagementClient.Credentials.ProcessHttpRequestAsync(httpRequest, cancellationToken).ConfigureAwait(false);
+            }
+            // Send Request
+            cancellationToken.ThrowIfCancellationRequested();
+            HttpClient httpClient = new HttpClient();
```
markcowl: why not use the http client from NetworkManagementClient here? (i.e. this.NetworkClient.NetworkManagementClient.HttpClient) This way you gate all the custom configuration that PowerShell does of that client

src/ResourceManager/Network/Commands.Network/VirtualNetworkGateway/GenerateAzureVpnClientPackage.cs
```diff
@@ -65,10 +78,164 @@ public override void ExecuteCmdlet()
...
+            {
+                throw new ValidationException(ValidationRules.CannotBeNull, "virtualNetworkGatewayName");
+            }
+            if (parameters == null)
+            {
+                throw new ValidationException(ValidationRules.CannotBeNull, "parameters");
+            }
+
+            // Construct URL
+            var url = new Uri(new Uri(baseUrl + (baseUrl.EndsWith("/") ? "" : "/")), "subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/virtualnetworkgateways/{virtualNetworkGatewayName}/generatevpnclientpackage").ToString();
```
markcowl: nit: Please break lines at around 80 columns

src/ResourceManager/Network/Commands.Network/VirtualNetworkGateway/GenerateAzureVpnClientPackage.cs
```diff
@@ -65,10 +78,164 @@ public override void ExecuteCmdlet()
...
+            {
+                throw new Exception(string.Format("Get-AzureRmVpnClientPackage Operation Failed as no valid status code received in response!"));
+            }
+
+            string operationId = result.RequestId;
+            #endregion
+
+            #region 2. Wait for Async operation to succeed 
+
+            Thread.Sleep(TimeSpan.FromSeconds(60));
```
markcowl: So it seems like here, you should use a while loop waiting for the response to come back as a 200.  Also, please use Microsoft.WindowsAzure.Commands.Utilities.CommonTestMockSupport.Delay  instead of calling Thread.Sleep directly, as this  will avoid delays when running mocked.

src/ResourceManager/Network/Commands.Network/VirtualNetworkGateway/GenerateAzureVpnClientPackage.cs
```diff
@@ -65,10 +78,164 @@ public override void ExecuteCmdlet()
...
+        /// Parameters supplied to the Begin Generating  Virtual Network Gateway Vpn
+        /// client package operation through Network resource provider.
+        /// </param>
+        /// <param name='customHeaders'>
+        /// Headers that will be added to request.
+        /// </param>
+        /// <param name='cancellationToken'>
+        /// The cancellation token.
+        /// </param>
+        public async Task<AzureOperationResponse<string>> GeneratevpnclientpackageWithHttpMessagesAsync(string resourceGroupName, string virtualNetworkGatewayName, VpnClientParameters parameters, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken))
```
markcowl: nit: consider placing this code in NetworkClient instead of this method

src/ResourceManager/Network/Commands.Network/VirtualNetworkGateway/GenerateAzureVpnClientPackage.cs
Nilambari: Done

src/ResourceManager/Network/Commands.Network/VirtualNetworkGateway/GenerateAzureVpnClientPackage.cs
Nilambari: Done.

src/ResourceManager/Network/Commands.Network/Common/NetworkClient.cs
```diff
@@ -41,5 +54,177 @@ public NetworkClient(INetworkManagementClient NetworkManagementClient)
...
+                    if (DateTime.UtcNow > giveUpAt)
+                    {
+                        if (!string.IsNullOrEmpty(newHttpResponse.Content.ReadAsStringAsync().Result))
+                        {
+                            result.Body = newHttpResponse.Content.ReadAsStringAsync().Result;
+                        }
+                        else
+                        {
+                            string newResponseContent = await newHttpResponse.Content.ReadAsStringAsync().ConfigureAwait(false);
+                            throw new Exception(string.Format("Get-AzureRmVpnClientPackage Operation returned an invalid status code '{0}' with Exception:{1} while retrieving " +
```
markcowl: I think you want to throw an exception whenever you are past the 'giveUpAt' time, so this exception throw belongs outside the else block, right?

src/ResourceManager/Network/Commands.Network/Common/NetworkClient.cs
```diff
@@ -41,5 +54,177 @@ public NetworkClient(INetworkManagementClient NetworkManagementClient)
...
+
+            // Set Credentials
+            if (NetworkManagementClient.Credentials != null)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                await NetworkManagementClient.Credentials.ProcessHttpRequestAsync(httpRequest, cancellationToken).ConfigureAwait(false);
+            }
+            // Send Request
+            cancellationToken.ThrowIfCancellationRequested();
+            HttpClient httpClient = new HttpClient();
```
markcowl: I think you should be using this.NetworkManagementClient.HttpClient

markcowl:
on demand run here:  http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/417/

PR #1509, opened 2015-12-16
Consolidate installer building script for osx & linuix

yugangw-msft:
1. per recommendation from @johanste, incorporate coreconsole from official released bits. Internal share are under \vwdbuild01\dev\yugangw\dotnet-cli
2. fix up the line endings of auzre.sh, otherwise it can't execute


johanste:
You should also be able to remove the reference to ConsoleHost from project.json in all Microsoft.CLU*/clurun projects as the contents of that nuget package is no longer used

yugangw-msft:
@johanste,sounds good and i made the change

PR #1513, opened 2015-12-17
Fixed CLU issue with ValidateElement when the object is not a string or enum.

hovsepm:


markcowl:
@johanste can you take a look?

johanste:
For value types other than enum, we really should pass in the correct (invariant) culture when calling ToString.

Best regards,
Johan

hovsepm:
@johanste addressed feedback. Also added version bump up to this PR (actually done by mistake, but I think it can go along with these changes).

hovsepm:
now have to bring other changes since the Nuget packages were bumped up.

PR #1520, opened 2015-12-18
Fixing resource string for log in and add logging and environment cha…

markcowl:
…nges for tests

### DIFF
--- a/src/CLU/Commands.Common.Authentication/Properties/Resources.resx
+++ b/src/CLU/Commands.Common.Authentication/Properties/Resources.resx
@@ -139,7 +139,7 @@
     <value>The environment name '{0}' is not found.</value>
   </data>
   <data name="ExpiredRefreshToken" xml:space="preserve">
-    <value>Your Microsoft Azure credential in the Windows PowerShell session has expired. Please log in again. </value>
+    <value>Your Microsoft Azure credential has expired. Please log in again.</value>
   </data>
   <data name="FilePathIsNotValid" xml:space="preserve">
     <value>File path is not valid</value>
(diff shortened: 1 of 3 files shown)


src/CLU/Commands.Common.Authentication/Properties/Resources.resx
```diff
@@ -139,7 +139,7 @@
     <value>The environment name '{0}' is not found.</value>
   </data>
   <data name="ExpiredRefreshToken" xml:space="preserve">
-    <value>Your Microsoft Azure credential in the Windows PowerShell session has expired. Please log in again. In PowerShell, execute Login-AzureRMAccount for Azure Resource Manager cmdlets or Add-AzureAccount for service management cmdlets.</value>
+    <value>Your Microsoft Azure credential in the Windows PowerShell session has expired. Please log in again. </value>
```
hovsepm: Windows? should it be just "Your Microsoft Azure credential in current session has expired"

src/CLU/Microsoft.Azure.Commands.Profile/Models/RMProfileClient.cs
```diff
@@ -403,7 +403,7 @@ public IEnumerable<AzureSubscription> ListSubscriptions()
                 catch (AadAuthenticationException)
                 {
                     WriteWarningMessage(string.Format("Could not authenticate user account {0} with tenant {1}.  " +
-                       "Subscriptions in this tenant will not be listed. Please login again using Login-AzureRmAccount " +
+                       "Subscriptions in this tenant will not be listed. Please ladd your account again " +
```
hovsepm: typo - should be "add" instead of "ladd"

PR #1521, opened 2015-12-18
upload win, osx, and linux installers to azure sdk team storage account

yugangw-msft:
I am taking the suggestions from @johanste and just upload all installers to sdk team's storage account


yugangw-msft:
@stankovski @amarzavery, please take a look and see whether this is enough for your testing scenario testing.
I am merging so that i can test the CI integration, but if you see any i should improve, please still comment here.

Issue #1411, opened 2015-12-04
New-AzureStorageContext with -Environment not setting endpoints to AzureChinaCloud


cordovinian:
I have noticed that after upgrading to v1.0.1 that the New-AzureStorageContext command is not returning the correct endpoint for AzureChinaCloud. This was found to work fine in previous versions. Specifically we upgraded from v0.9.7 to v1.0.1.
Surprisingly this is seems isolated to New-AzureStorageContext. Running the same test with Get-AzureEnvironment works fine.

**On Azure v1.0.1**
```powershell
S:\> (Get-Module azure).Version

Major  Minor  Build  Revision
-----  -----  -----  --------
1      0      1      -1

S:\> New-AzureStorageContext -Environment "AzureChinaCloud" -StorageAccountName "MyAccount" -StorageAccountKey "[key hidden]"

StorageAccountName : MyAccount
BlobEndPoint       : https://MyAccount.blob.core.windows.net/
TableEndPoint      : https://MyAccount.table.core.windows.net/
QueueEndPoint      : https://MyAccount.queue.core.windows.net/
Context            : Microsoft.WindowsAzure.Commands.Common.Storage.AzureStorageContext
Name               :
StorageAccount     : BlobEndpoint=https://MyAccount.blob.core.windows.net/;
					 QueueEndpoint=https://MyAccount.queue.core.windows.net/;
					 TableEndpoint=https://MyAccount.table.core.windows.net/;
					 FileEndpoint=https://MyAccount.file.core.windows.net/;
					 AccountName=MyAccount;
					 AccountKey=[key hidden]
EndPointSuffix     : core.windows.net/

D:\> Get-AzureEnvironment AzureChinaCloud

Name                                     : AzureChinaCloud
EnableAdfsAuthentication                 : False
ActiveDirectoryServiceEndpointResourceId : https://management.core.chinacloudapi.cn/
AdTenant                                 :
GalleryUrl                               : https://gallery.chinacloudapi.cn/
ManagementPortalUrl                      : http://go.microsoft.com/fwlink/?LinkId=301902
ServiceManagementUrl                     : https://management.core.chinacloudapi.cn/
PublishSettingsFileUrl                   : http://go.microsoft.com/fwlink/?LinkID=301776
ResourceManagerUrl                       : https://management.chinacloudapi.cn
SqlDatabaseDnsSuffix                     : .database.chinacloudapi.cn
StorageEndpointSuffix                    : core.chinacloudapi.cn
[...]

PR #1545, opened 2015-12-28
[#110771146]: Fix Endpoint Suffix when using New-AzureStorageContext specifying -Environment

hovsepm:
Fix of the issue #1411


hovsepm:

PR #1566, opened 2015-12-31
AzureRT PR - CLU Test Updates

huangpf:


examples/compute-management/01-VirtualMachineSizes.sh
```diff
@@ -0,0 +1,34 @@
...
+result=`echo "$vmSizeResult" | grep -q "$queryString"`
+if [ "$vmSizeResult" = "" ] ; then
+    printf "\nFailure: VM Size Not Found: '%s'.\n" "$queryString"
+    exit 1
+else
+    printf "\nSuccess: VM Size Found in Results: '%s'.\n" "$queryString"
+fi
+
+queryString=Standard_G1
+result=`echo "$vmSizeResult" | grep -q "$queryString"`
```
huangpf: @stankovski @hovsepm @markcowl 
What's the correct way to do string operations and test the not-containing case?

stankovski: This looks like one of the options: http://stackoverflow.com/questions/229551/string-contains-in-bash

huangpf: I think I had try the ```[[``` method, but the test runner throws error. @stankovski Could you please help try it on your end? Thanks.

stankovski: This code worked for me:
```sh
string='My long string';

if [[ $string != *"short"* ]]
then
  echo "It's not there!";
fi
```

stankovski: Don't forget the spaces around `[[` - they are important

huangpf: It would throw the following error ``` [[: command not found```, using the same code above. Any configuration I need to make? Thanks.

stankovski: Here is a working test:
```sh
#!/bin/bash
set -e
printf "\n=== Managing Virtual Machine Sizes in Azure Compute ===\n"

printf "\n1. Showing VM size results in location: %s.\n" "$location"
azure vmsize get --location "$location"

printf "\n2. Checking VM size results in location: %s.\n" "$location"
vmSizeResult=`azure vmsize get --location "$location"`

if [[ $vmSizeResult == "" ]]; then
    echo "Failure: No VM sizes!" 1>&2
    exit 1
else
    echo "Success: Non-empty Results."
fi

filterResult=`azure vmsize get --location "$location" | cat | jq 'select(.name | contains("Standard_A0"))' --raw-output`
if [[ "$filterResult" == "" ]]; then
    echo "Failure: Standard_A0 vm size not found." 1>&2
    exit 1
else
    echo "Success: Standard_A0 vm size found."
fi

filterResult=`azure vmsize get --location "$location" | cat | jq 'select(.name | contains("Standard_G1"))' --raw-output`
if [[ "$filterResult" == "" ]]; then
    echo "Failure: Standard_G1 vm size not found." 1>&2
    exit 1
else
    echo "Success: Standard_G1 vm size found."
fi
```

huangpf: Thanks @stankovski . ```bash 4.0``` works. I have updated the PR. Could you please help review again? Thanks.

PR #1576, opened 2016-01-04
AzureRT - VM Creation Update & Tests

huangpf:


examples/compute-management/02-VirtualMachineCreation.sh
```diff
@@ -0,0 +1,38 @@
...
+contextResult=`azure context get`
+
+subId=`echo $contextResult | jq '.Subscription.SubscriptionId' --raw-output`
+
+subnetId="/subscriptions/$subId/resourceGroups/$groupName/providers/Microsoft.Network/virtualNetworks/test/subnets/test"
+
+echo -n "$subnetId" > "$BASEDIR/$groupName.subnetIdFile"
+
+printf "\n4. Create network interface with:\r\nsubId='%s' \r\n& \r\nsubnetId='$subnetId'.\n" "$subId"
+azure network interface create --name test --resourcegroupname "$groupName" --location "$location" --subnetid @"$BASEDIR/$groupName.subnetIdFile"
```
huangpf: @markcowl @stankovski 
```--subnetid /subscription/xxx``` would result in ```c:\Program Files\Git\subscriptions...``` and thus fail the request in the tests.

```--subnetid @file``` would pass.

Why?

Comment (stankovski) [inline examples/compute-management/02-VirtualMachineCreation.sh]: You need to disable POSIX path conversion in Git Bash by setting MSYS_NO_PATHCONV env variable to 1 before your command (and disabling it after)
```bash
export MSYS_NO_PATHCONV=1
azure network interface create --name test ...
export MSYS_NO_PATHCONV=
```

PR #1579, opened 2016-01-05
Update Compute library to Swagger-generated one

hyonholee:


src/ResourceManager/Compute/Commands.Compute/VirtualMachineSizes/GetAzureVMSizeCommand.cs
```diff
@@ -78,30 +79,30 @@ public override void ExecuteCmdlet()
...
                 }
                 else
                 {
-                    result = this.VirtualMachineSizeClient.List(this.Location.Canonicalize());
+                    result = this.VirtualMachineSizeClient.ListWithHttpMessagesAsync(this.Location.Canonicalize()).GetAwaiter().GetResult();
                 }
 
                 List<PSVirtualMachineSize> psResultList = new List<PSVirtualMachineSize>();
-                foreach (var item in result.VirtualMachineSizes)
+                foreach (var item in result.Body)
```
huangpf: The converted field's name ```Body``` is weird... Is it possible to define ```VirtualMachineSizes```?

src/ResourceManager/Compute/Commands.Compute/VirtualMachine/Operation/GetAzureVMCommand.cs
```diff
@@ -78,51 +79,64 @@ public override void ExecuteCmdlet()
                 {
                     if (Status)
                     {
-                        var result = this.VirtualMachineClient.GetWithInstanceView(this.ResourceGroupName, this.Name);
+                        var result = this.VirtualMachineClient.Get(this.ResourceGroupName, this.Name, InstanceViewExpand);
```
huangpf: Can the 3rd parameter be optional and overloaded?

huangpf: @stankovski

src/ResourceManager/Compute/Commands.Compute/VirtualMachineSizes/GetAzureVMSizeCommand.cs
huangpf: @stankovski

src/ResourceManager/Compute/Commands.Compute.Test/ScenarioTests/VirtualMachineTests.ps1
```diff
@@ -455,11 +455,6 @@ function Test-VirtualMachineImageList
                             {
                                 $versions = $s4 | select -ExpandProperty Version;
 
-                                $s5 = Get-AzureRmVMImage -Location $locStr -PublisherName $pub -Offer $offer -Sku $sku -FilterExpression ('name -eq *');
```
huangpf: It seems this is the original ```FilterExpression``` test case. Why removing it?

huangpf:
LGTM, in case the ```FilterExpression``` test case concern is addressed. Thanks @hyonholee!!

src/ResourceManager/Compute/Commands.Compute/VirtualMachine/Operation/GetAzureVMCommand.cs
hyonholee: The 3rd parameter (expand) is optional.  Default is null.

src/ResourceManager/Compute/Commands.Compute.Test/ScenarioTests/VirtualMachineTests.ps1
hyonholee: FilterExpression is not supported for List VM Image API.   I think we have to remove the parameter as well.

src/ResourceManager/Compute/Commands.Compute/VirtualMachineSizes/GetAzureVMSizeCommand.cs
stankovski: You're getting Body because you're using ListWithHttpMessagesAsync(). If you were to use just List() you would get a collection right away.

src/ResourceManager/Compute/Commands.Compute.Test/ScenarioTests/VirtualMachineTests.ps1
huangpf: OK. I think we shall add a filter expression test case for VM extension then, if feasible.

hyonholee: That part is already there: line 496 in the same test file.

src/ResourceManager/Compute/Commands.Compute/VirtualMachineSizes/GetAzureVMSizeCommand.cs
huangpf: OK, but ```List``` doesn't contain the operation info that we want.

Will it be feasible to return a named collection in the async method? [feature request]

PR #1593, opened 2016-01-07
WebApp bash tests and modifications

amarzavery:


examples/lib/setup.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
stankovski: This whole file should be removed

src/CLU/Microsoft.Azure.Commands.Websites/Cmdlets/WebApps/SetAzureWebApp.cs
```diff
@@ -68,7 +68,7 @@ public class SetAzureWebAppCmdlet : WebAppBaseCmdlet
         [Parameter(ParameterSetName = ParameterSet1Name, Position = 10, Mandatory = false, HelpMessage = "Web app connection strings. Example: -ConnectionStrings @{ ConnectionString1 = @{ Type = \"MySql\"; Value = \"MySql Connection string\"}; ConnectionString2 = @{ Type = \"SQLAzure\"; Value = \"SqlAzure Connection string 2\"} }")]
         [ValidateNotNullOrEmpty]
         [ValidateConnectionStrings]
-        public Hashtable ConnectionStrings { get; set; }
```
stankovski: Add 
```cs
// TODO: CLU
//public Hashtable ConnectionStrings { get; set; }
```

PR #1594, opened 2016-01-07
[#110120350] Subscriptions and tenants can be listed correctly without specifying -tenantID 

hovsepm:
- [#110120350] Subscriptions and tenants can be listed correctly without specifying -tenantID 
- [#110225916] Do not select subscriptions that are disabled when logging in

### DIFF
--- a/src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
+++ b/src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
@@ -485,8 +485,8 @@ namespace Microsoft.Azure.Commands.ResourceManager.Common
                     {
                         var subscriptions = (subscriptionClient.Subscriptions.List().Subscriptions ?? 
                                                 new List<Microsoft.Azure.Subscriptions.Models.Subscription>())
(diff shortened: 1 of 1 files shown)


src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
```diff
@@ -481,19 +483,24 @@ private AzureEnvironment MergeEnvironmentProperties(AzureEnvironment environment
                     }
                     else
                     {
-                        var subscriptions = subscriptionClient.Subscriptions.List().Subscriptions;
-                        if (subscriptions != null && subscriptions.Any())
+                        var subscriptions = (subscriptionClient.Subscriptions.List().Subscriptions ?? 
+                                                new List<Microsoft.Azure.Subscriptions.Models.Subscription>())
+                                            .Where(s => s.State.Equals("enabled", StringComparison.OrdinalIgnoreCase) ||
```
stankovski: May be safer to compare constant string to property in case server ever returns a null:
`"enabled".Equals(s.State, StringComparison.OrdinalIgnoreCase)`

PR #1614, opened 2016-01-08
Improve diagnostics extension command

yantang-msft:
This code change has gone through internal code review. Reviewer: @bfung, @rapandit

And the changes including:
1. Allow user specify storage account name, key, endpoint instead of a mandatory StorageContext parameter
2. Allow user specify an array of configuration for SetAzureServiceDiagnosticsExtension commands, so we can apply different configurations within one call.
3. Bug fix
4. Add scenario tests

### DIFF
--- a/src/ServiceManagement/Compute/Commands.ServiceManagement/Common/DiagnosticsHelper.cs
+++ b/src/ServiceManagement/Compute/Commands.ServiceManagement/Common/DiagnosticsHelper.cs
@@ -71,18 +71,18 @@ namespace Microsoft.WindowsAzure.Commands.ServiceManagement.Common
             int wadCfgBeginIndex = config.IndexOf("<WadCfg>");
             if (wadCfgBeginIndex == -1)
             {
-                throw new ArgumentException("Cannot find the WadCfg element in the config.");
+                throw new ArgumentException(Properties.Resources.DiagnosticsExtensionXmlConfigNoWadCfgStartTag);
             }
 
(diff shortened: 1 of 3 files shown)


bfung:
I don't believe I'm the _bfung_ you're looking for, there are some simple fixes I could suggest, but perhaps they've been caught by the build? ;)

yantang-msft:
@bfung, didn't noticed the alias points to you by mistake.
What kind of simple fix would you suggest? The build failed for scenario test, while it succeeded on my local machine. I'm looking into it.

src/ServiceManagement/Compute/Commands.ServiceManagement/Common/DiagnosticsHelper.cs
```diff
@@ -0,0 +1,317 @@
...
+            }
+        }
+
+        public static Hashtable GetPublicDiagnosticsConfiguration(string config, string storageAccountName)
+        {
+            // find the <WadCfg> element and extract it
+            int wadCfgBeginIndex = config.IndexOf("<WadCfg>");
+            if (wadCfgBeginIndex == -1)
+            {
+                throw new ArgumentException("Cannot find the WadCfg element in the config.");
```
hovsepm: Please add all strings for the exceptions to the resources and reference it from there.

yantang-msft: > Please add all text strings for the exceptions to the resources and reference it from there.

Thanks! Done.

PR #1651, opened 2016-01-14
Format web output [#110362888], Fix generic resource display [#111658848]

markcowl:
- renamed login, vm and vnet new -> create, removed extra resourcemanager from provider cmdlets
- added logout cmdlet


BurtBiel:
Changes look fine to me and the cmdlet naming changes make sense

PR #1670, opened 2016-01-16
Update cmdlet names for resource cmdlets

markcowl:


markcowl:

examples/storage-management/01-Storage-Accounts.sh
```diff
@@ -25,28 +25,28 @@ rm -f $BASEDIR/$accountName.json
...
 key1=$(cat $BASEDIR/$accountName.json | jq '.key1' --raw-output)
 key2=$(cat $BASEDIR/$accountName.json | jq '.key2' --raw-output)
 [ $key1 != $key2 ]
 rm -f $BASEDIR/$accountName.json
 
 printf "\n5. Renew account key1 of storage account %s in group %s\n" $accountName $groupName
-azure storage account key new -g $groupName -n $accountName -k "key1"
-azure storage account key get -g $groupName -n $accountName > $BASEDIR/$accountName.json
+az storage account key create -g $groupName -n $accountName -k "key1"
+az storage account key get -g $groupName -n $accountName > $BASEDIR/$accountName.json
```
BurtBiel: should this be az storage account key ls (instead of get)?

markcowl: Once these cmdlets are aliased, yes.  Right now they are using the default with rename.cfg.  I plan to update this in an upcoming PR (one PR per service).

PR #1708, opened 2016-01-22
Support Json format config file when enable diagnostics extension

yantang-msft:
This  change has gone through internal code review.
Reviewer alias: bfung, rapandit


hyonholee:
LGTM

PR #1728, opened 2016-01-27
[#111280330] Enhance Get-AzureRmSubscription to support RBAC changes (support Subscription pagination)

hovsepm:


hovsepm:
on-demand: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/453/

src/ResourceManager/Profile/Commands.Profile/Subscription/GetAzureRMSubscription.cs
```diff
@@ -92,7 +95,41 @@ public override void ExecuteCmdlet()
...
+
+                    foreach (var tenantId in tenantsList)
+                    {
+                        try
+                        {
+                            string listNextLink = null;
+                            do
+                            {
+                                var subscriptions = _client.ListSubscriptions(tenantId, ref listNextLink);
+                                WriteObject(subscriptions.Select((s) => (PSAzureSubscription)s), enumerateCollection: true);
```
markcowl: @hovsepm I'm not sure what actually happens in this case, where we are writing mulsiple collections to the output stream.  To be safe, I would perform a WriteObject for each subscription.

markcowl:
@hovsepm one change, go ahead and merge after this

PR #1743, opened 2016-01-29
Revert ARM to package using older api version

vivsriaus:


vivsriaus:
On demand job here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/462/

PR #1763, opened 2016-02-03
Use explicit exports in module manifest

lzybkr:
When wildcards are used in the module manifest, PowerShell is slower when analyzing modules to see what commands are exported.

This is typically a one time cost (first time you start up), but in a container, it is a cost paid pretty much every time you run PowerShell in the container.

This analysis PowerShell does costs around 250ms for just the Azure module, so explicitly declaring the exports is a noticable win.

This change should not cause any change in behavior, it should only impact performance.

Note that if this change is accepted, you'll need to remember to change the module manifest whenever you add a new cmdlet, whereas before this change, that wasn't necessary.


azurecla:
Hi __@lzybkr__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
Hi __@lzybkr__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

markcowl:
@lzybkr Thanks for the contribution! Can you share the source of the information for these performance gains?  If we can validate the performance gains, we would add in automation to generate the exported cmdlets at build time rather than keep a static list per module.

lzybkr:
This is how I capture the data.  It requires a recent public Win10 builds:

* Delete the file $env:USERPROFILE\AppData\Local\Microsoft\Windows\PowerShell\ModuleAnalysisCache (very safe to delete)
* Start an ETW trace, e.g. with perfview.  Turn on the provider *Microsoft-PowerShell-CommandDiscovery
* Run: powershell -noprofile -command "CommandNotFound"
(The actual command here doesn't matter, the goal is to search all modules.)
* Then in the captured etw data, look for the events GetModuleExportedCommandsStart and GetModuleExportedCommandsStop.  The payload has the module path.  Subtract the times b/w Stop and Start to get the elapsed time analyzing the module.

My 250ms measurement was warm - it should be much better for cold because the Azure module has a number of dlls that are processed w/o this change.

markcowl:
@lzybkr Had a discussion with the PowerShell core team and they confirmed that listing modules will impact the performance of module autoload (at least for the first load of the module).  We will add in a build process step to populate the ExportedCommands for each module.  Thanks again for researching this!

markcowl:
@lzybkr closing for now, sincew we will go with build-time generation of these fields. We are tracking this item in pivotal here: https://www.pivotaltracker.com/story/show/113641745

kilasuit:
@markcowl I personally highly disagree with closing issues in Github before the issue is actually resolved for the following reasons

-  This is the location where the general public can track where issues are upto without having to sign up for another service (even though it is open too)

- The Issues in here should really be the single source of truth of the state of the project 

- Additional tooling should improve the management of issue tracking and the allocation of work for bug fixing but should not replace the existing tools available by integrating with the existing tools 

I would suggest that you make use of the Github API https://developer.github.com/v3/issues/ to close this and future issues when you've closed the corresponding item in your project tracker

Also as an aside your tracker doesn't have any integration with the issues here - i.e its a completely separate process and therefore has the possibility for the Azure PowerShell Team to miss issues being raised therefore making the outward view being one of disorganisation. I would suggest using https://waffle.io/ for Issue Management as it integrates well with GitHub and is already being used by the PSScriptAnalyzer team - though this also brings the issue of the use of very much different tool sets across the Microsoft OSS efforts.

trondhindenes:
I completely agree with @kilasuit on this. if you want community feedback you need to treat github issues as an integral part of issue tracking for the codebase, not just a dumping ground for stuff that doesnt work. 

I have yet still not seen any official wording from the team on how they actually want this github repo to be used. The messiness and lack of proper labelling of issues has come up before (and pretty much why I stopped submitting issues, since it was responded to so randomly by the team).

If you're gonna use a tool, use it right. And if you're gonna actually use github issues, then use them right. In any case, create a contributing.md file in the repo root explaining how the public can interact. This manually moving issues from one system to another and closing the source issue is just crazy. You're Microsoft, FGS. Do it right!

markcowl:
@kilasuit @lzybkr @trondhindenes This is a pull request and not just an issue.  I have personally talked to Jason about how we are going to address this and it is not by adding a static list to the module, so this code change as it stands is not going to be accepted and he is fine with that.

I'm not sure I understand the value of keeping open a PR that the owner doesn't think we should accept as is.  Can you please carify what I am missing.

PR #1792, opened 2016-02-09
AzureCP PR - Fix Extension ID Issue #387

huangpf:
AzureCP PR - Fix Extension ID Issue #387


huangpf:
@hyonholee @hovsepm Please help review.

src/ServiceManagement/Compute/Commands.ServiceManagement/Extensions/Common/ExtensionRole.cs
```diff
@@ -22,12 +23,28 @@ public class ExtensionRole
...
+            var match = disallowedCharactersRegex.Match(roleName);
+
+            while (match.Success)
+            {
+                roleName = roleName.Remove(match.Index, match.Length);
+                match = disallowedCharactersRegex.Match(roleName);
+            }
+
+            return roleName;
+        }
```
hovsepm: this method seems too complex for the thing it is doing. 
Just put following on line 37 and you should be good:
```
return disallowedCharactersRegex.Replace(roleName,"");
```

src/ServiceManagement/Compute/Commands.ServiceManagement/Extensions/Common/ExtensionRole.cs
```diff
@@ -63,12 +78,16 @@ public override string ToString()
...
-            normalizedExtName = normalizedExtName.Replace("_", string.Empty);
+            var normalizedExtName = RemoveDisallowedCharacters(extensionName);
 
             var suffix = new StringBuilder();
             suffix.AppendFormat(ExtensionIdSuffixTemplate, normalizedExtName, slot, index);
+            if (suffix.Length > MaxSuffixLength)
+            {
+                int lenDiff = suffix.Length - MaxSuffixLength;
+                int startIndex = 1; // Suffix starts with '-'
+                suffix.Remove(startIndex + normalizedExtName.Length - lenDiff, lenDiff);
```
hovsepm: what do you want to achieve here? string.Remove returns a new string which you do not use anywhere.

hovsepm: BTW, please cover this case in the tests as well ;)

src/ServiceManagement/Compute/Commands.ServiceManagement/Extensions/Common/ExtensionRole.cs
```diff
@@ -63,12 +78,16 @@ public override string ToString()
...
+            }
```
hovsepm: is this for limiting the string length? If so, this should be enough 
```
suffix = siffix.Substring(0, MaxSuffixLength);
```
right?

huangpf: Actually no... So the format of the extension ID is ```{role-name-part} + {suffix}```, where ```{suffix} = -{extension-name-part}-{slot}-Ext-{num}``` (all from legacy design request).

I want to limit the ```suffix``` length, but I can't simply cut it off by ```MaxSuffixLength```, and I have to truncate the ```{extension-name-part}```.

src/ServiceManagement/Compute/Commands.ServiceManagement/Extensions/Common/ExtensionRole.cs
huangpf: Yes, it's actually covered, :-)... -- The test case is to validate that a very long extension name will be truncated in the suffix, while the other parts, i.e. slot & number are preserved.

huangpf:
Ping... @hovsepm

markcowl:
@huangpf please fix merge conflicts, pull the latest, and re-generate the wix file.

huangpf:
@markcowl Done~

PR #1806, opened 2016-02-11
Reduce Batch PR test time

jasper-schneider:
Removed the Checkin/Acceptance attribute on many of the recorded tests as an initial step to reduce test time.  Will revisit restructure/reorganization later.


jasper-schneider:
On demand build here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/478/

PR #1830, opened 2016-02-18
Disable Set-AzureSqlServerCommunicationLink

adumitr:
update semantics are not supported on server communication links


adumitr:
I would really like this to be in the next release for which we're a couple a days from the cutoff date. Can we please pull this in quicker?

hovsepm:
@adumitr this approach does not let users understand why the cmdlet was removed. It would be better to throw an exception with some meaningful explanation rather than just commenting out the cmdlet which is there for several releases.

adumitr:
I see your point. We considered it, in this case we'd like for the cmdlet to not show up at all. It was never documented and it never worked before, update semantics are not allowed on the service side so the exceptions would come from there anyway. Let's merge as is

markcowl:
@adumitr @hovsepm Commenting out the cmdlet is not acceptable - you must remove the file and remove the reference from the .csproj file.

adumitr:
Okay, done, can we merge it now?

PR #1832, opened 2016-02-18
Fix for set and remove arm cmdlet

eshaparmar:


eshaparmar:
we already have existing automation in our CI and azure sdk code base

PR #1845, opened 2016-02-23
HDInsight CMDLets for script actions

dulems:
HDInsight CMDLets for script actions

### DIFF
--- a/src/ResourceManager/HDInsight/Commands.HDInsight.Test/UnitTests/ConfigurationTests.cs
+++ b/src/ResourceManager/HDInsight/Commands.HDInsight.Test/UnitTests/ConfigurationTests.cs
@@ -42,7 +42,7 @@ namespace Microsoft.Azure.Commands.HDInsight.Test
             {
                 CommandRuntime = commandRuntimeMock.Object,
                 HDInsightManagementClient = hdinsightManagementMock.Object,
-                ClusterType = HDInsightClusterType.Hadoop
+                ClusterType = ClusterType
             };
 
             newconfigcmdlet.ExecuteCmdlet();
@@ -51,7 +51,7 @@ namespace Microsoft.Azure.Commands.HDInsight.Test
                     f.WriteObject(
                         It.Is<AzureHDInsightConfig>(
                             c =>
-                                c.ClusterType == HDInsightClusterType.Hadoop &&
+                                c.ClusterType == ClusterType &&
                                 c.AdditionalStorageAccounts.Count == 0 &&
                                 c.Configurations.Count == 0 &&
                                 string.IsNullOrEmpty(c.WorkerNodeSize) &&
--- a/src/ResourceManager/HDInsight/Commands.HDInsight.Test/UnitTests/DataLakeStoreTests.cs
+++ b/src/ResourceManager/HDInsight/Commands.HDInsight.Test/UnitTests/DataLakeStoreTests.cs
@@ -56,7 +56,7 @@ namespace Commands.HDInsight.Test.UnitTests
             };
         }
 
-        [Fact]
+        [Fact(Skip="Test currently failing. To be fixed in next release.")]
         [Trait(Category.AcceptanceType, Category.CheckIn)]
         public void CanCreateNewHDInsightDataLakeStoreCluster()
         {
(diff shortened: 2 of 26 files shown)


azurecla:
Hi __@dulems__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (dule). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

hovsepm:
@azuresdkci add to whitelist

src/ResourceManager/HDInsight/Commands.HDInsight.Test/UnitTests/DataLakeStoreTests.cs
```diff
@@ -56,7 +56,7 @@ public DataLakeStoreTests()
             };
         }
 
-        [Fact]
+        //[Fact]
```
hovsepm: why this was removed? you don't need this test anymore?

src/ResourceManager/HDInsight/Commands.HDInsight.Test/UnitTests/ConfigurationTests.cs
```diff
@@ -51,7 +51,7 @@ public void CanCreateNewConfig()
                     f.WriteObject(
                         It.Is<AzureHDInsightConfig>(
                             c =>
-                                c.ClusterType == HDInsightClusterType.Hadoop &&
+                                c.ClusterType == "Hadoop" &&
```
hovsepm: Please use constant string instead of putting "Hadoop" string everywhere. you should also consider making a case insensitive comparison especially if this is something that customer may provide as parameter.

src/ResourceManager/HDInsight/Commands.HDInsight.Test/UnitTests/DataLakeStoreTests.cs
dulems: This test was temporarily removed since it has been failing for 3 months. The owner of this test agreed to fix it in the next release.

hovsepm: then do not remove [fact] but add [Fact(Skipped="Reason why it was skipped")]

src/ResourceManager/HDInsight/Commands.HDInsight.Test/UnitTests/ConfigurationTests.cs
dulems: Previously we have an Enum, but have changed the contracts to accept any string, which will be processed by the back end. All but one instance of "Hadoop" is in the test code and don't need to be "Hadoop." Is using constant in this case?

dulems: Fixed by adding constants. Insensitive comparison is not needed as the comparison only happens in test code.

src/ResourceManager/HDInsight/Commands.HDInsight.Test/UnitTests/DataLakeStoreTests.cs
dulems: Fixed. Thanks!

src/ResourceManager/HDInsight/Commands.HDInsight.Test/UnitTests/ConfigurationTests.cs
shefaliv: There really isn't a need for a constant here. These are tests and are subject to change as we feel given the coverage we want for the code. Definitely doesn't need to be case insensitive.

markcowl:
on demand run here: http://azuresdkci.cloudapp.net/job/powershell-demand/

hovsepm:
Please pull from upstream dev. After that add link to the new on-demand job to this PR

dulems:
On demand run http://azuresdkci.cloudapp.net/job/powershell-demand/534/

dulems:
It looks like something went wrong with my pull from upstream dev, but the on demand run outcome is still valid. Please only merge the last commit (2d62a46).

hovsepm:
please create a PR to dev branch.

PR #1873, opened 2016-02-26
Fix issue with Find-AzureRmResource when listing resources in type

vivsriaus:


vivsriaus:
On demand here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/555/

PR #1876, opened 2016-02-26
CDN powershell

kuangweizhang:
Adding powershell cmdlets for Azure CDN


azurecla:
Hi __@kuangweizhang__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (kzhan). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

README.md
```diff
@@ -119,6 +119,9 @@ This repository contains a set of PowerShell cmdlets for developers and administ
 * Azure Automation
   * Manage automation accounts
   * Manage automation jobs, runbooks and schedules
+* CDN
```
markcowl: Please remove.  The readme should not contain a reference to each service

src/ResourceManager/Cdn/Commands.Cdn/CustomDomain/GetAzureCdnCustomDomain.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Management.Automation;
+using Microsoft.Azure.Commands.Cdn.Common;
+using Microsoft.Azure.Commands.Cdn.Helpers;
+using Microsoft.Azure.Commands.Cdn.Models.CustomDomain;
+using Microsoft.Azure.Commands.Cdn.Properties;
+using Microsoft.Azure.Management.Cdn;
+
+namespace Microsoft.Azure.Commands.Cdn.CustomDomain
+{
+    [Cmdlet(VerbsCommon.Get, "AzureCdnCustomDomain", ConfirmImpact = ConfirmImpact.None), OutputType(typeof(PSCustomDomain))]
```
markcowl: ARM cmdlets must use AzureRm as their prefix, as in Get-AzureRmCdnCustomDomain

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/CustomDomainInvalidReason.cs
```diff
@@ -0,0 +1,25 @@
+﻿////////////////////////////////////////////////////////////////////////////////
```
markcowl: License header

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/DeepCreatedOrigin.cs
```diff
@@ -0,0 +1,62 @@
+﻿////////////////////////////////////////////////////////////////////////////////
```
markcowl: license header

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/DeepCreatedOriginProperties.cs
```diff
@@ -0,0 +1,60 @@
+﻿////////////////////////////////////////////////////////////////////////////////
```
markcowl: License header

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/EndpointResourceState.cs
```diff
@@ -0,0 +1,18 @@
+﻿////////////////////////////////////////////////////////////////////////////////
```
markcowl: License header

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/LoadInput.cs
```diff
@@ -0,0 +1,26 @@
+﻿////////////////////////////////////////////////////////////////////////////////
```
markcowl: License header

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/PurgeAzureCdnEndpointContent.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Management.Automation;
+using Microsoft.Azure.Commands.Cdn.Common;
+using Microsoft.Azure.Commands.Cdn.Models.Endpoint;
+using Microsoft.Azure.Commands.Cdn.Properties;
+using Microsoft.Azure.Management.Cdn;
+
+namespace Microsoft.Azure.Commands.Cdn.Endpoint
+{
+    [Cmdlet("Purge", "AzureCdnEndpointContent", ConfirmImpact = ConfirmImpact.Medium), OutputType(typeof(bool))]
+    public class PurgeAzureCdnEndpointContent : AzureCdnCmdletBase
```
markcowl: Please use PowerShell specific names.  Remove-AzureRmCdnENdpointContent would likely be the appropriate verb

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/PurgeInput.cs
```diff
@@ -0,0 +1,26 @@
+﻿////////////////////////////////////////////////////////////////////////////////
```
markcowl: header

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/QueryStringCachingBehavior.cs
```diff
@@ -0,0 +1,36 @@
+﻿////////////////////////////////////////////////////////////////////////////////
```
markcowl: Header

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/RemoveAzureCdnEndnpoint.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Net;
+using Microsoft.Azure.Commands.Cdn.Common;
+using Microsoft.Azure.Commands.Cdn.Models.Endpoint;
+using Microsoft.Azure.Commands.Cdn.Properties;
+using Microsoft.Azure.Management.Cdn;
+using Microsoft.Azure.Management.Cdn.Models;
+
+namespace Microsoft.Azure.Commands.Cdn.Endpoint
+{
+    [Cmdlet(VerbsCommon.Remove, "AzureCdnEndpoint", ConfirmImpact = ConfirmImpact.High), OutputType(typeof(bool))]
```
markcowl: SupportsShouldProcess=true

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/RemoveAzureCdnEndnpoint.cs
```diff
@@ -0,0 +1,89 @@
...
+                {
+                    throw new PSArgumentException(string.Format(
+                        Resources.Error_DeleteNonExistingEndpoint,
+                        EndpointName,
+                        ProfileName,
+                        ResourceGroupName));
+                }
+            }
+
+            ConfirmAction(
```
markcowl: ConfirmAction usage will be altered in the next release - you should use ShouldProcess to determine whether to continue, users will specify ```-Confirm:$false``` to bypass prompts.

-Force should be reserved for especially brutal actions that would normally fail, like deleting a folder that contained files.

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/RemoveAzureCdnEndnpoint.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        [Parameter(Mandatory = true, ParameterSetName = FieldsParameterSet, HelpMessage = "The resource group of the Azure Cdn Profile")]
+        [ValidateNotNullOrEmpty]
+        public string ResourceGroupName { get; set; }
+
+        [Parameter(Mandatory = true, ParameterSetName = ObjectParameterSet, ValueFromPipeline = true, HelpMessage = "The endpoint.")]
+        [ValidateNotNull]
+        public PSEndpoint CdnEndpoint { get; set; }
+
+        [Parameter(Mandatory = false, HelpMessage = "Do not ask for confirmation.")]
```
markcowl: Probably not necessary.  This usage of force to get rid of prompts is being deprecated.

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/RemoveAzureCdnEndnpoint.cs
```diff
@@ -0,0 +1,89 @@
...
+            }
+
+            ConfirmAction(
+                        Force.IsPresent,
+                        string.Format(Resources.Confirm_RemoveEndpoint, EndpointName, ProfileName, ResourceGroupName),
+                        () => { CdnManagementClient.Endpoints.DeleteIfExists(EndpointName, ProfileName, ResourceGroupName); });
+
+            WriteVerbose(Resources.Success);
+            WriteVerbose(string.Format(Resources.Success_RemoveEndpoint, EndpointName, ProfileName, ResourceGroupName));
+            WriteObject(true);
```
markcowl: In most cases, you will want to write nothing on success.  Users should be able to pass -PassThrough to get the true for success printed.

markcowl:
@kuangweizhang Thanks for the PR.  It is past code complete, so we will be unlikely to take this PR for the current release.  There are a few issues that I noted, but since this is the first release, would like to go over the cmdlets in detail.

We will also need to know who on your team will field GitHub issues. And where these should be field internally.

On naming, your module should eb anmed AzureRM.Cdn and the build directory should have the same name.

hovsepm:
@azuresdkci add to whitelist

hovsepm:
@azuresdkci test this please

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/RemoveAzureCdnEndnpoint.cs
kuangweizhang: ok, I see lots of other command has both -Force and SupportsShouldProcess in the code. Should we just directly remove the -Force and only use ShouldProcess?
For example:
\src\ResourceManager\Sql\Commands.Sql\Database\Cmdlet\RemoveAzureSqlDatabase.cs
This command has both Force and ShouldProcess

src/ResourceManager/Cdn/Commands.Cdn.Test/Commands.Cdn.Test.csproj
```diff
@@ -0,0 +1,191 @@
...
+    <None Include="ScenarioTests\EndpointTests.ps1">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Include="ScenarioTests\OriginTests.ps1">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Include="ScenarioTests\ProfileTests.ps1">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Include="SessionRecords\Microsoft.Azure.Commands.Cdn.Test.ScenarioTests.ScenarioTest.CustomDomainTests\TestCustomDomainGetRemoveWithRunningEndpoint.json" />
```
markcowl: These will need to be set to 'copy if newer'

markcowl:
@kuangweizhang  You will need to add your project into Azure.powershell.test.targets for inclusion in the check-in tests

src/ResourceManager/Cdn/Commands.Cdn/Common/AzureCdnCmdletBase.cs
```diff
@@ -0,0 +1,73 @@
...
+                return _defaultRequestHeaders ??
+                       (_defaultRequestHeaders =
+                           new Dictionary<string, List<string>> { { "UserAgent", new List<string> { "PowerShell" } } });
+            }
+            set { _defaultRequestHeaders = value; }
+        }
+
+        public void ConfirmAction(bool force, string actionMessage, Action action)
+        {
+            if (force || ShouldContinue(actionMessage, ""))
```
markcowl: ConfirmAction is being rewritten to look like this:
```C#
if (ShouldProcess()) {
     if (!shouldContinueRequired || Force || ShouldContinue()) {
                  <do cmdlet actions>
   }
}
```

If you cann AzurePSCmdlet.SHouldContinue(), you will get this behavior by default

src/ResourceManager/Cdn/Commands.Cdn/CustomDomain/GetAzureRmCdnCustomDomain.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Management.Automation;
+using Microsoft.Azure.Commands.Cdn.Common;
+using Microsoft.Azure.Commands.Cdn.Helpers;
+using Microsoft.Azure.Commands.Cdn.Models.CustomDomain;
+using Microsoft.Azure.Commands.Cdn.Properties;
+using Microsoft.Azure.Management.Cdn;
+
+namespace Microsoft.Azure.Commands.Cdn.CustomDomain
+{
+    [Cmdlet(VerbsCommon.Get, "AzureRmCdnCustomDomain", ConfirmImpact = ConfirmImpact.None), OutputType(typeof(PSCustomDomain))]
```
markcowl: SupportsShouldProcess=true

markcowl: You don't need ConfirmImpact if you don't have ShouldProcess support

src/ResourceManager/Cdn/Commands.Cdn/CustomDomain/GetAzureRmCdnCustomDomain.cs
```diff
@@ -0,0 +1,51 @@
...
+        [ValidateNotNullOrEmpty]
+        public string ProfileName { get; set; }
+
+        [Parameter(Mandatory = true, HelpMessage = "The resource group of the Azure Cdn Profile")]
+        [ValidateNotNullOrEmpty]
+        public string ResourceGroupName { get; set; }
+
+        public override void ExecuteCmdlet()
+        {
+            var customDomain = CdnManagementClient.CustomDomains.Get(CustomDomainName, EndpointName, ProfileName, ResourceGroupName);
```
markcowl: lien breaks

src/ResourceManager/Cdn/Commands.Cdn/CustomDomain/NewAzureRmCdnCustomDomain.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        public override void ExecuteCmdlet()
+        {
+            try
+            {
+                CdnManagementClient.CustomDomains.GetWithHttpMessagesAsync(
+                    CustomDomainName, 
+                    EndpointName, 
+                    ProfileName,
+                    ResourceGroupName).Wait();
```
markcowl: use ```.ConfigureAwait(false)```

src/ResourceManager/Cdn/Commands.Cdn/CustomDomain/NewAzureRmCdnCustomDomain.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        [Parameter(Mandatory = true, HelpMessage = "The resource group of the Azure Cdn Profile")]
+        [ValidateNotNullOrEmpty]
+        public string ResourceGroupName { get; set; }
+
+        public override void ExecuteCmdlet()
+        {
+            try
+            {
+                CdnManagementClient.CustomDomains.GetWithHttpMessagesAsync(
```
markcowl: Thius doesn't seem like the best way to handle this.  Can't you use the if-none-match header in your call to create to ensure that the domain doesn't already exist?

src/ResourceManager/Cdn/Commands.Cdn/CustomDomain/RemoveAzureCdnCustomDomain.cs
```diff
@@ -0,0 +1,104 @@
...
+using System.Net;
+using Microsoft.Azure.Commands.Cdn.Common;
+using Microsoft.Azure.Commands.Cdn.Models.CustomDomain;
+using Microsoft.Azure.Commands.Cdn.Properties;
+using Microsoft.Azure.Management.Cdn;
+using Microsoft.Azure.Management.Cdn.Models;
+
+namespace Microsoft.Azure.Commands.Cdn.CustomDomain
+{
+    [Cmdlet(VerbsCommon.Remove, "AzureRmCdnCustomDomain", ConfirmImpact = ConfirmImpact.High, SupportsShouldProcess = true), OutputType(typeof(bool))]
```
markcowl: line breaks

src/ResourceManager/Cdn/Commands.Cdn/CustomDomain/RemoveAzureCdnCustomDomain.cs
```diff
@@ -0,0 +1,104 @@
...
+                return;
+            }
+
+            CdnManagementClient.CustomDomains.DeleteIfExists(
+                            CustomDomainName,
+                            EndpointName,
+                            ProfileName,
+                            ResourceGroupName);
+
+            WriteObject(true);
```
markcowl: Consider using a PassThrough parameter, so the cmdlet returns nothing if successful, or true if PassThrough is specified

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/LoadAzureRmCdnEndpointContent.cs
```diff
@@ -0,0 +1,60 @@
...
+
+using System.Management.Automation;
+using Microsoft.Azure.Commands.Cdn.Common;
+using Microsoft.Azure.Commands.Cdn.Models.Endpoint;
+using Microsoft.Azure.Commands.Cdn.Properties;
+using Microsoft.Azure.Management.Cdn;
+
+namespace Microsoft.Azure.Commands.Cdn.Endpoint
+{
+    [Cmdlet("Load", "AzureRmCdnEndpointContent", ConfirmImpact = ConfirmImpact.Low), OutputType(typeof(bool))]
```
markcowl: In general, don't use ConfirmImpact unless you support ShouldProcess

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/LoadAzureRmCdnEndpointContent.cs
```diff
@@ -0,0 +1,60 @@
...
+            if (ParameterSetName == ObjectParameterSet)
+            {
+                ResourceGroupName = CdnEndpoint.ResourceGroupName;
+                ProfileName = CdnEndpoint.ProfileName;
+                EndpointName = CdnEndpoint.Name;
+            }
+
+            CdnManagementClient.Endpoints.LoadContent(EndpointName, ProfileName, ResourceGroupName, LoadContent);
+            WriteVerbose(Resources.Success);
+            WriteObject(true);
```
markcowl: Consider implementing PassThrough pattern

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/LoadAzureRmCdnEndpointContent.cs
```diff
@@ -0,0 +1,60 @@
...
+using Microsoft.Azure.Commands.Cdn.Models.Endpoint;
+using Microsoft.Azure.Commands.Cdn.Properties;
+using Microsoft.Azure.Management.Cdn;
+
+namespace Microsoft.Azure.Commands.Cdn.Endpoint
+{
+    [Cmdlet("Load", "AzureRmCdnEndpointContent", ConfirmImpact = ConfirmImpact.Low), OutputType(typeof(bool))]
+    public class LoadAzureRmCdnEndpointContent : AzureCdnCmdletBase
+    {
+        [Parameter(Mandatory = true, ParameterSetName = FieldsParameterSet, HelpMessage = "Azure Cdn endpoint name.")]
```
markcowl: If this represents a resource, you should have an overload to load the endpoint by Id

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/GetAzureRmCdnEndpointNameAvailability.cs
```diff
@@ -0,0 +1,42 @@
...
+using Microsoft.Azure.Commands.Cdn.Common;
+using Microsoft.Azure.Commands.Cdn.Helpers;
+using Microsoft.Azure.Commands.Cdn.Models.Endpoint;
+using Microsoft.Azure.Commands.Cdn.Properties;
+using Microsoft.Azure.Management.Cdn;
+using Microsoft.Azure.Management.Cdn.Models;
+
+namespace Microsoft.Azure.Commands.Cdn.Endpoint
+{
+    [Cmdlet(VerbsCommon.Get, "AzureRmCdnEndpointNameAvailability", ConfirmImpact = ConfirmImpact.None), OutputType(typeof(PSCheckNameAvailabilityOutput))]
```
markcowl: This should use the verb Check

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/GetAzureRmCdnEndpoint.cs
```diff
@@ -0,0 +1,46 @@
...
+using Microsoft.Azure.Commands.Cdn.Models.Endpoint;
+using Microsoft.Azure.Commands.Cdn.Properties;
+using Microsoft.Azure.Management.Cdn;
+
+namespace Microsoft.Azure.Commands.Cdn.Endpoint
+{
+    [Cmdlet(VerbsCommon.Get, "AzureRmCdnEndpoint", ConfirmImpact = ConfirmImpact.None), OutputType(typeof(PSEndpoint))]
+    public class GetAzureRmCdnEndpoint : AzureCdnCmdletBase
+    {
+        [Parameter(Mandatory = true, HelpMessage = "Azure Cdn endpoint name.")]
```
markcowl: Should have a parameter set for retrieving by id

src/ResourceManager/Cdn/Commands.Cdn/CustomDomain/RemoveAzureCdnCustomDomain.cs
```diff
@@ -0,0 +1,104 @@
...
+using Microsoft.Azure.Commands.Cdn.Models.CustomDomain;
+using Microsoft.Azure.Commands.Cdn.Properties;
+using Microsoft.Azure.Management.Cdn;
+using Microsoft.Azure.Management.Cdn.Models;
+
+namespace Microsoft.Azure.Commands.Cdn.CustomDomain
+{
+    [Cmdlet(VerbsCommon.Remove, "AzureRmCdnCustomDomain", ConfirmImpact = ConfirmImpact.High, SupportsShouldProcess = true), OutputType(typeof(bool))]
+    public class RemoveAzureCdnCustomDomain : AzureCdnCmdletBase
+    {
```
markcowl: Should have a parameter set for retrieving by id

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/NewAzureRmCdnEndpoint.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        [Parameter(Mandatory = false,
+            HelpMessage = "The tags to associate with the Azure Cdn Endpoint")]
+        public Hashtable Tags { get; set; }
+
+        public override void ExecuteCmdlet()
+        {
+            try
+            {
+                CdnManagementClient.Endpoints.GetWithHttpMessagesAsync(EndpointName, ProfileName, ResourceGroupName).Wait();
```
markcowl: Again, this seems like not the best pattern - you could use if-none-match to guarantee you are creating an endpoint that does not exist

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/PurgeAzureRmCdnEndpointContent.cs
```diff
@@ -0,0 +1,60 @@
...
+
+using System.Management.Automation;
+using Microsoft.Azure.Commands.Cdn.Common;
+using Microsoft.Azure.Commands.Cdn.Models.Endpoint;
+using Microsoft.Azure.Commands.Cdn.Properties;
+using Microsoft.Azure.Management.Cdn;
+
+namespace Microsoft.Azure.Commands.Cdn.Endpoint
+{
+    [Cmdlet("Purge", "AzureRmCdnEndpointContent", ConfirmImpact = ConfirmImpact.Medium), OutputType(typeof(bool))]
```
markcowl: No way this will pass policheck.  You should use Remove

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/PurgeAzureRmCdnEndpointContent.cs
```diff
@@ -0,0 +1,60 @@
...
+    public class PurgeAzureRmCdnEndpointContent : AzureCdnCmdletBase
+    {
+        [Parameter(Mandatory = true, ParameterSetName = FieldsParameterSet, HelpMessage = "Azure Cdn endpoint name.")]
```
markcowl: Again, consider implementing the PassThrough pattern

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/PurgeAzureRmCdnEndpointContent.cs
```diff
@@ -0,0 +1,60 @@
...
+        [ValidateNotNullOrEmpty]
+        public string ResourceGroupName { get; set; }
+
+        [Parameter(Mandatory = true, ValueFromPipeline = true, HelpMessage = "The endpoint.", ParameterSetName = ObjectParameterSet)]
+        [ValidateNotNull]
+        public PSEndpoint CdnEndpoint { get; set; }
+
+        [Parameter(Mandatory = true, HelpMessage = "The resource group of the Azure Cdn Profile")]
+        [ValidateCount(Constants.PurgeLoadMinimumCollectionCount, Constants.PurgeLoadMaximumCollectionCount)]
+        public string[] PurgeContent { get; set; }
```
markcowl: rename

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/RemoveAzureRmCdnEndnpoint.cs
```diff
@@ -0,0 +1,86 @@
...
+using Microsoft.Azure.Commands.Cdn.Properties;
+using Microsoft.Azure.Management.Cdn;
+using Microsoft.Azure.Management.Cdn.Models;
+
+namespace Microsoft.Azure.Commands.Cdn.Endpoint
+{
+    [Cmdlet(VerbsCommon.Remove, "AzureRmCdnEndpoint", ConfirmImpact = ConfirmImpact.High, SupportsShouldProcess = true), OutputType(typeof(bool))]
+    public class RemoveAzureCdnEndpoint : AzureCdnCmdletBase
+    {
+        [Parameter(Mandatory = true, ParameterSetName = FieldsParameterSet, HelpMessage = "Azure Cdn endpoint name.")]
```
markcowl: should have a by id parameter set

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/RemoveAzureRmCdnEndnpoint.cs
```diff
@@ -0,0 +1,86 @@
...
+            {
+                var errorResponseException = exception.InnerException as ErrorResponseException;
+                if (errorResponseException == null)
+                {
+                    throw;
+                }
+
+                if (errorResponseException.Response.StatusCode.Equals(HttpStatusCode.NotFound))
+                {
+                    throw new PSArgumentException(string.Format(
```
markcowl: shouldn't delete of a non-existent endpoint indicate success.  I would consider using a warning here

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/SetAzureRmCdnEndpoint.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.Azure.Commands.Cdn.Helpers;
+using Microsoft.Azure.Commands.Cdn.Models.Endpoint;
+using Microsoft.Azure.Commands.Cdn.Properties;
+using Microsoft.Azure.Management.Cdn;
+using Microsoft.Azure.Management.Cdn.Models;
+using SdkQueryStringCachingBehavior = Microsoft.Azure.Management.Cdn.Models.QueryStringCachingBehavior;
+
+namespace Microsoft.Azure.Commands.Cdn.Endpoint
+{
+    [Cmdlet(VerbsCommon.Set, "AzureRmCdnEndpoint", ConfirmImpact = ConfirmImpact.Medium), OutputType(typeof(PSEndpoint))]
```
markcowl: Con sider supporting shouldprocess

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/StartAzureRmCdnEndpoint.cs
```diff
@@ -0,0 +1,57 @@
...
+
+using System.Management.Automation;
+using Microsoft.Azure.Commands.Cdn.Common;
+using Microsoft.Azure.Commands.Cdn.Models.Endpoint;
+using Microsoft.Azure.Commands.Cdn.Properties;
+using Microsoft.Azure.Management.Cdn;
+
+namespace Microsoft.Azure.Commands.Cdn.Endpoint
+{
+    [Cmdlet(VerbsLifecycle.Start, "AzureRmCdnEndpoint", ConfirmImpact = ConfirmImpact.Low), OutputType(typeof(bool))]
```
markcowl: Consider implementing PassThrough

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/StopAzureRmCdnEndpoint.cs
```diff
@@ -0,0 +1,66 @@
...
+using Microsoft.Azure.Commands.Cdn.Common;
+using Microsoft.Azure.Commands.Cdn.Models.Endpoint;
+using Microsoft.Azure.Commands.Cdn.Properties;
+using Microsoft.Azure.Management.Cdn;
+
+namespace Microsoft.Azure.Commands.Cdn.Endpoint
+{
+    [Cmdlet(VerbsLifecycle.Stop, "AzureRmCdnEndpoint", ConfirmImpact = ConfirmImpact.High, SupportsShouldProcess = true), OutputType(typeof(bool))]
+    public class StopAzureRmCdnEndpoint : AzureCdnCmdletBase
+    {
```
markcowl: Consider implementing PassThrough

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/ValidateAzureRmCdnCustomDomain.cs
```diff
@@ -0,0 +1,67 @@
...
+using Microsoft.Azure.Commands.Cdn.Models.Endpoint;
+using Microsoft.Azure.Commands.Cdn.Properties;
+using Microsoft.Azure.Management.Cdn;
+
+namespace Microsoft.Azure.Commands.Cdn.Endpoint
+{
+    [Cmdlet("Validate", "AzureRmCdnCustomDomain", ConfirmImpact = ConfirmImpact.None), OutputType(typeof(PSValidateCustomDomainOutput))]
+    public class ValidateAzureRmCdnCustomDomain : AzureCdnCmdletBase
+    {
+        [Parameter(Mandatory = true, ParameterSetName = FieldsParameterSet, HelpMessage = "Azure Cdn endpoint name.")]
```
markcowl: Consider having a ById parameter set

src/ResourceManager/Cdn/Commands.Cdn/Microsoft.Azure.Commands.Cdn.dll-help.psd1
```diff
@@ -0,0 +1,86 @@
...
+DotNetFrameworkVersion = '4.0'
+
+# Minimum version of the common language runtime (CLR) required by this module
+CLRVersion='4.0'
+
+# Processor architecture (None, X86, Amd64, IA64) required by this module
+ProcessorArchitecture = 'None'
+
+# Modules that must be imported into the global environment prior to importing this module
+RequiredModules = @()
```
markcowl: Need a dependency on AzureRM.Profile

src/ResourceManager/Cdn/Commands.Cdn/Models/Endpoint/PSDeepCreatedOrigin.cs
```diff
@@ -0,0 +1,27 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// ----------------------------------------------------------------------------------
+
+namespace Microsoft.Azure.Commands.Cdn.Models.Endpoint
+{
+    public class PSDeepCreatedOrigin
```
markcowl: Name?

src/ResourceManager/Cdn/Commands.Cdn/Models/Profile/PSProfileResourceState.cs
```diff
@@ -0,0 +1,26 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// ----------------------------------------------------------------------------------
+
+namespace Microsoft.Azure.Commands.Cdn.Models.Profile
+{
+    public enum PSProfileResourceState
```
markcowl: Do you really want this as an enum?  Enums on deserialization are essentially just a timebomb waiting to throw when the enum value changes on the service.  Think carefully about whether your enums are actually guaranteed not to add new values, for those enums used on deserialization.

Enums on serialization are entirely appropriate

src/ResourceManager/Cdn/Commands.Cdn/Profile/NewAzureRmCdnProfile.cs
```diff
@@ -0,0 +1,107 @@
...
+using SdkSku = Microsoft.Azure.Management.Cdn.Models.Sku;
+using SdkSkuName = Microsoft.Azure.Management.Cdn.Models.SkuName;
+
+namespace Microsoft.Azure.Commands.Cdn.Profile
+{
+    /// <summary>
+    /// Defines the New-AzureRmCdnProfile cmdlet.
+    /// </summary>
+    [Cmdlet(VerbsCommon.New, "AzureRmCdnProfile", ConfirmImpact = ConfirmImpact.Low), OutputType(typeof(PSProfile))]
+    public class NewAzureRmCdnProfile : AzureCdnCmdletBase
```
markcowl: Is Profile a well=known CDN concept?  Profile is overloaded int he powershell context, and if this is not a well-known concept, a different name would be preferrable

src/ResourceManager/Cdn/Commands.Cdn/Profile/RemoveAzureRmCdnProfile.cs
```diff
@@ -0,0 +1,80 @@
...
+using Microsoft.Azure.Commands.Cdn.Properties;
+using Microsoft.Azure.Management.Cdn;
+using Microsoft.Azure.Management.Cdn.Models;
+
+namespace Microsoft.Azure.Commands.Cdn.Profile
+{
+    [Cmdlet(VerbsCommon.Remove, "AzureRmCdnProfile", ConfirmImpact = ConfirmImpact.High, SupportsShouldProcess = true), OutputType(typeof(bool))]
+    public class RemoveAzureRmCdnProfile : AzureCdnCmdletBase
+    {
+        [Parameter(Mandatory = true, ParameterSetName = FieldsParameterSet, HelpMessage = "The name of the profile.")]
```
markcowl: by id parameter set

hovsepm:
@kuangweizhang As Mark mentioned - "your module should eb anmed AzureRM.Cdn and the build directory should have the same name". Make sure that you will rename AzureCdn.psd1 to AzureRM.CDN.psd1

Also please change dependency in the AzureRm.CDN.psd1 form 
```
ModuleName = 'AzureRM.Profile'; ModuleVersion = '1.0.4' 
```
to version 1.0.5

hovsepm:
@kuangweizhang  why do you have 2 modules?

1. src/ResourceManager/Cdn/AzureCdn.psd1
2. src/ResourceManager/Cdn/Commands.Cdn/AzureCdn.psd1

hovsepm:
Please add your test project to AzurePowershell.Test.targets file in the root folder. Otherwise your tests will not run on CI server during Checkin and post-checkin jobs.

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/GetAzureRmCdnEndpointNameAvailability.cs
kilasuit: Check is not an approved verb - so No Check **shouldnt** be used as the verb - options are as below as pulled from Windows10 14279 build

Add
Approve
Assert
Backup
Block
Checkpoint
Clear
Close
Compare
Complete
Compress
Confirm
Connect
Convert
ConvertFrom
ConvertTo
Copy
Debug
Deny
Disable
Disconnect
Dismount
Edit
Enable
Enter
Exit
Expand
Export
Find
Format
Get
Grant
Group
Hide
Import
Initialize
Install
Invoke
Join
Limit
Lock
Measure
Merge
Mount
Move
New
Open
Optimize
Out
Ping
Pop
Protect
Publish
Push
Read
Receive
Redo
Register
Remove
Rename
Repair
Request
Reset
Resize
Resolve
Restart
Restore
Resume
Revoke
Save
Search
Select
Send
Set
Show
Skip
Split
Start
Step
Stop
Submit
Suspend
Switch
Sync
Test
Trace
Unblock
Undo
Uninstall
Unlock
Unprotect
Unpublish
Unregister
Update
Use
Wait
Watch
Write


I would suggest using Test and forcing this cmdlet to return a boolean value instead

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/LoadAzureRmCdnEndpointContent.cs
kilasuit: Load is not an approved verb - Please rethink this and only use an approved verb in your cmdlets

src/ResourceManager/Cdn/Commands.Cdn.Test/Commands.Cdn.Test.csproj
kuangweizhang: "PreserveNewest" is "copy if newer"

src/ResourceManager/Cdn/Commands.Cdn/CustomDomain/NewAzureRmCdnCustomDomain.cs
kuangweizhang: Our REST doesn't support this yet.
I think "if-none-match" is a improvement we should do...
I believe we have same problem with traffic manager https://github.com/Azure/azure-powershell/blob/dev/src/ResourceManager/TrafficManager/Commands.TrafficManager2/Profile/NewAzureTrafficManagerProfile.cs#L75

src/ResourceManager/Cdn/Commands.Cdn/CustomDomain/RemoveAzureCdnCustomDomain.cs
kuangweizhang: Make sense

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/NewAzureRmCdnEndpoint.cs
kuangweizhang: Copy comment from above

Our REST doesn't support this yet.
I think "if-none-match" is an improvement we should do...
I believe we have same problem with traffic manager https://github.com/Azure/azure-powershell/blob/dev/src/ResourceManager/TrafficManager/Commands.TrafficManager2/Profile/NewAzureTrafficManagerProfile.cs#L75

src/ResourceManager/Cdn/Commands.Cdn/Endpoint/RemoveAzureRmCdnEndnpoint.cs
kuangweizhang: Sorry, I don't understand. Are you saying that if user is trying to delete a non-existent endpoint, we should write a warning instead of throwing an exception?

src/ResourceManager/Cdn/Commands.Cdn/Models/Endpoint/PSDeepCreatedOrigin.cs
kuangweizhang: Sorry, what do you mean?

src/ResourceManager/Cdn/Commands.Cdn/Models/Profile/PSProfileResourceState.cs
kuangweizhang: I think enum is more accurate.
One version of powershell is corresponding to one version of sdk, one version of SDK is corresponding to one version of API. Since SDK have this value as enum, powershell should be consistent.

src/ResourceManager/Cdn/Commands.Cdn/Profile/NewAzureRmCdnProfile.cs
kuangweizhang: Profile is the name of the CDN resource and we should not change this.
https://msdn.microsoft.com/en-us/library/azure/mt634454.aspx

kuangweizhang:
@hovsepm @markcowl @kilasuit 
Thanks a lot for your review. I've fixed most of the comments and I have some questions.
1. For the verb of check name availability. I can't use check as Ryan said. I don't want to use Test because I don't want the output of the cmdlet to be only boolean. The reason for that is there are reasons why this name is not available in the response if the name is not available. We want to expose this to users. I've looked at existing powershell and just to be consistent, I used get
https://github.com/Azure/azure-powershell/blob/dev/src/ResourceManager/Storage/Commands.Management.Storage/StorageAccount/GetAzureStorageAccountNameAvailability.cs#L21

2. For purge and load command. Both Purge and Load is very well known CDN terms. Is it possible to leave those two verbs? Remove is not a very good replacement to purge because remove makes people think that we are removing the files from the origin. But what we actually do is to invalid the cache from CDN PoPs. What's your suggestions?

3. I saw the comments about retrieving resources by resource Id. For all the Set,Remove,Start,Stop commands, the powershell has the piping option allow user to pass in an object from Gets. I would think this is equivalent or even easier than asking for a resource Id here. For Get commands, I don't see a reason to use Id because the Id come from a property of the result of a Get. What do you think? What is the standard for powershell? I've looked at existing commands and didn't see the "Get by Id" https://github.com/Azure/azure-powershell/blob/dev/src/ResourceManager/TrafficManager/Commands.TrafficManager2/Profile/GetAzureTrafficManagerProfile.cs

kuangweizhang:
I'm fixing the failed tests

kilasuit:
Regards 2 - I would suggest Revoke/Remove/Unpublish/Unregister over adding a non standard verb of Purge - This needs to be consistent with the rest of the core PowerShell Language and non-standard verbs shouldn't be used when at all possible.
Also instead of Load - perhaps Publish/Register would be better Verb Choices?

kuangweizhang:
oops, I think I messed up, let me fix this

kilasuit:
I'm confused at why you've closed the PR?

kuangweizhang:
I think I didn't do the rebase from upstream correctly result in a lot more changes in my pull request. Let me send a new one

kuangweizhang:
New pull request submitted: https://github.com/Azure/azure-powershell/pull/1926

PR #1922, opened 2016-03-09
[DO NOT MERGE]: Implement Force/Confirm according to powershell guidelines

markcowl:


src/Common/Storage/Commands.Storage/Blob/Cmdlet/SetAzureStorageBlobContent.cs
```diff
@@ -80,6 +80,13 @@ public string File
             set { FileName = value; }
         }
 
+        [Parameter(HelpMessage = "Force to overwrite the existing blob or file")]
```
EmmaZhu: Here is only for blob, you can remove the "or file" from the HelpMessage.

src/Common/Storage/Commands.Storage/Blob/Cmdlet/SetAzureStorageContainerStoredAccessPolicy.cs
```diff
@@ -25,7 +25,7 @@ namespace Microsoft.WindowsAzure.Commands.Storage.Blob.Cmdlet
     /// <summary>
     /// create a new azure container
     /// </summary>
-    [Cmdlet(VerbsCommon.Set, StorageNouns.ContainerStoredAccessPolicy), OutputType(typeof(String))]
+    [Cmdlet(VerbsCommon.Set, StorageNouns.ContainerStoredAccessPolicy, SupportsShouldProcess=true, ConfirmImpact = ConfirmImpact.Medium), OutputType(typeof(String))]
     public class SetAzureStorageContainerStoredAccessPolicyCommand : StorageCloudBlobCmdletBase
```
EmmaZhu: Set StoredAccessPolicy should be fine, it's OK not to have SupportShouldProcess.

src/Common/Storage/Commands.Storage/File/Cmdlet/GetAzureStorageFileContent.cs
```diff
@@ -25,7 +25,7 @@ namespace Microsoft.WindowsAzure.Commands.Storage.File.Cmdlet
     using LocalDirectory = System.IO.Directory;
     using LocalPath = System.IO.Path;
 
-    [Cmdlet(VerbsCommon.Get, LocalConstants.FileContentCmdletName, SupportsShouldProcess = true, ConfirmImpact = ConfirmImpact.High, DefaultParameterSetName = LocalConstants.ShareNameParameterSetName)]
```
EmmaZhu: GetAzureStorageFileContent may need to overwrite local file, should keep SupportsShouldProcess.

src/ResourceManager/Compute/Commands.Compute/Extension/DSC/SetAzureVMDscExtensionCommand.cs
```diff
@@ -460,20 +460,6 @@ private ConfigurationUris UploadConfigurationDataToBlob()
                     configurationDataBlobReference.Uri.AbsoluteUri,
                     () =>
                     {
-                        if (!Force && configurationDataBlobReference.Exists())
```
eshaparmar: removing force here would mean that we by default always update the storage blob - no warning to the user

markcowl: @eshaparmar This is just refactoring the code to have the same behavior using the new ConfirmAction overload.  the new code, after the ShouldProcess check, will show the ShouldContinue prompt unless the user has specified the Force parameter, when the configurationDataBlobReference exists.

src/Common/Storage/Commands.Storage/Blob/Cmdlet/SetAzureStorageBlobContent.cs
markcowl: @EmmaZhu Good catch.

src/Common/Storage/Commands.Storage/Blob/Cmdlet/SetAzureStorageContainerStoredAccessPolicy.cs
markcowl: @EmmaZhu Every cmdlet that makes permanent chenges should implement ShouldProcess

src/Common/Storage/Commands.Storage/File/Cmdlet/GetAzureStorageFileContent.cs
markcowl: @EmmaZhu Yes, this is one of the cmdlets on the list that will require a small amount of refactoring - the cmdlet should implement ShouldProcess, and should prompt with a ShouldContinue prompt when attempting to overwrite a file (unless the user specifies -Force, so the cmdlet also needs a Force parameter)

PR #1926, opened 2016-03-11
Adding Azure CDN powershell

kuangweizhang:


azurecla:
Hi __@kuangweizhang__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (kzhan). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

kuangweizhang:
Copied from closed pull request
@hovsepm @markcowl @kilasuit 
Thanks a lot for your review. I've fixed most of the comments and I have some questions.

For the verb of check name availability. I can't use check as Ryan said. I don't want to use Test because I don't want the output of the cmdlet to be only boolean. The reason for that is there are reasons why this name is not available in the response if the name is not available. We want to expose this to users. I've looked at existing powershell and just to be consistent, I used get
https://github.com/Azure/azure-powershell/blob/dev/src/ResourceManager/Storage/Commands.Management.Storage/StorageAccount/GetAzureStorageAccountNameAvailability.cs#L21

For purge and load command. Both Purge and Load is very well known CDN terms. Is it possible to leave those two verbs? Remove is not a very good replacement to purge because remove makes people think that we are removing the files from the origin. But what we actually do is to invalid the cache from CDN PoPs. What's your suggestions?

I saw the comments about retrieving resources by resource Id. For all the Set,Remove,Start,Stop commands, the powershell has the piping option allow user to pass in an object from Gets. I would think this is equivalent or even easier than asking for a resource Id here. For Get commands, I don't see a reason to use Id because the Id come from a property of the result of a Get. What do you think? What is the standard for powershell? I've looked at existing commands and didn't see the "Get by Id" https://github.com/Azure/azure-powershell/blob/dev/src/ResourceManager/TrafficManager/Commands.TrafficManager2/Profile/GetAzureTrafficManagerProfile.cs

kuangweizhang:
I've change verb of purge and load to publish and unpublish

kuangweizhang:
I looked at the error of tests and very confused.

Errors:
Unable to find recorded mock directory 'SessionRecords\Microsoft.Azure.Commands.Cdn.Test.ScenarioTests.ScenarioTest.OriginTests'

Could not find a part of the path 'D:\workspace\powershell-4\src\ResourceManager\KeyVault\Commands.KeyVault.Test\bin\Debug\SessionRecords\Microsoft.Azure.Commands.Cdn.Test.ScenarioTests.ScenarioTest.OriginTests\TestOriginGetSetWithStoppedEndpoint.json'.

Unable to find recorded mock directory 'SessionRecords\Microsoft.Azure.Commands.Cdn.Test.ScenarioTests.ScenarioTest.ProfileTests'

For first and third error, the folder definitely exist, otherwise, other Origin tests and Profile tests can't pass.
For second error, it's trying to look for CDN tests in Key Vault folder which I don't see why. This test is same with all other tests.
Can anyone please help?

kuangweizhang:
Ping... Can any one please take a look at this PR?

hovsepm:
To fix the test add following lines to the RunPsTestWorkflow method: 

```
HttpMockServer.RecordsDirectory = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "SessionRecords");
```

hovsepm:
@kuangweizhang please pull from upstream dev branch, and update CDN reference to ClienRuntime and ClientRuntime.Azure to latest.

kuangweizhang:
@hovsepm I've updated all the reference as build error suggested. I don't understand the current error message in the test.
The running command stopped because the preference variable "ErrorActionPreference" or common parameter is set to Stop: Unable to find a matching HTTP request for URL 'HEAD /subscriptions/27cafca8-b9a4-4264-b399-45d0c9cca1ab/resourcegroups/onesdk4223?'. Calling method Item().

I checked the records and find the request. Anything else is wrong?

kuangweizhang:
Newer version of the pull request has been created https://github.com/Azure/azure-powershell/pull/1972

PR #1949, opened 2016-03-15
[#114730201][#114888193] Automation for selecting subscription and tenant in authentication scenarios

haocs:


src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
```diff
@@ -121,8 +121,16 @@ public RMProfileClient(AzureRMProfile profile)
                         newTenant == null &&                         
                         TryGetTenantSubscription(token, account, environment, tenant, subscriptionId, subscriptionName, out tempSubscription, out tempTenant))
                     {
-                        newTenant = tempTenant;
-                        newSubscription = tempSubscription;
+                        if (tempSubscription == null && i + 1 < tenants.Count())
```
hovsepm: No need for the empty bock. You can invert the if to work for one case. Doesn't this work?
```
If( tempSubscription != null || i == (tenants.Count() -1) ) 
```

hovsepm: BTW, this will change the logic of selecting the tenants. Now if tenant does not have any subscription - Last tenant will be selected. Somewhere in the code we say that "First subscription will be selected"

haocs: @hovsepm 
- yes.
- To keep the tenant selecting order, how about iterating tenants in a reverse fashion? Personally, I don't think "first tenant/subscription" makes much sense here if tenants are not sorted(or sorted alphabetically).

markcowl:
@azuresdkci add to whitelist

PR #1950, opened 2016-03-16
Refactoring commands.common.storage to remove storage management depe…

markcowl:
…ndency

### DIFF
--- a/src/Common/Commands.Common.Storage/Adapters/ARM.Storage.3/ARMStorageService.cs
+++ b/src/Common/Commands.Common.Storage/Adapters/ARM.Storage.3/ARMStorageService.cs
@@ -65,11 +65,20 @@ namespace Microsoft.Azure.Commands.Management.Storage.Models
             get { return _authenticationKeys; }
         }
 
-        public static string ParseResourceGroupFromId(string idFromServer)
+        /// <summary>
+        /// Get the resource group name from a storage account resource Id
+        /// </summary>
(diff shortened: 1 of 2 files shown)


src/Common/Commands.Common.Storage/Adapters/ARM.Storage.3/ARMStorageService.cs
```diff
@@ -0,0 +1,81 @@
...
+            get { return _authenticationKeys; }
+        }
+
+        public static string ParseResourceGroupFromId(string idFromServer)
+        {
+            if (!string.IsNullOrEmpty(idFromServer))
+            {
+                string[] tokens = idFromServer.Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
+
+                return tokens[3];
```
hovsepm: Since this is public static method - it may fail with index outside the bounds of the array. Could you check that it has 3 elements before accessing it?

src/Common/Commands.Common.Storage/IStorageService.cs
```diff
@@ -0,0 +1,52 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.WindowsAzure.Commands.Common.Storage
+{
+    public interface IStorageService
+    {
+        /// <summary>
+        /// The eblob service endpoint
```
hovsepm: eblob? just blob?

PR #1962, opened 2016-03-17
Auto fill metrics resource id

yantang-msft:
Automatically construct the resource id for the user, so they don't need to provide it by themselves. Moreover, they don't need to have separate config files for different resources because of the resource id.


yantang-msft:
I'm rerunning the scenario test to fix the build failure.

PR #2029, opened 2016-03-30
Updating the Azure US Gov Cloud STS and Traffic Manager

zakramer:
Updating the STS URL for Azure Gov to the US Only STS that was stood up
for Azure Gov. - The AAD team has stood up an STS endpoint (login-us.microsoftonline.com) that will limit traffic to US Only servers.  This is the STS that the Azure US Government portal (https://manage.windowsazure.us) is now using.  To ensure that all Azure US Government customers are using the US Only servers (to ensure compliance) we want to update the Environment for Azure US Government to use this STS.

Adding the Traffic Manager Suffix - Noticed that this was missing so I added it.  I can remove and make a separate PR is that is better.

Updating the Add-AzureAccount and Add-AzureRMAccount documentation to include AzureUSGovernment in the text and to change Environment to EnvironmentName to be easier to use.  Noticed that the documentation was very confusing during testing so I updated some of the docs to include AzureUSGovernment as an option.  

Also, I noticed that everyone uses EnvironmentName for Add-AzureRMAccount but that was not in the docs.  Just updated Environment to EnvironmentName.  I can add both if that makes more sense but it seemed that having less is better since EnvironmentName is the one that is easy to use.


azurecla:
Hi __@zakramer__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (zakramer). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

hovsepm:
@azuresdkci add to whitelist

hovsepm:
re-targeted to release branch in #2072

PR #2052, opened 2016-04-04
Add TenantId parameter for new Key Vault

mattjohnsonpint:
The [Key Vault REST API](https://msdn.microsoft.com/en-us/library/azure/mt620025.aspx) allows you to specify a `tenantId`.  This is important, as you might want to create a Key Vault in a specific AAD tenant that is not necessarily the same one as your sign in.

Currently, both the Powershell and X-Plat CLI don't allow you to supply the Tenant ID.   When you go through the [Getting started with Azure Key Vault](https://azure.microsoft.com/en-us/documentation/articles/key-vault-get-started/) documentation, there's a warning half-way through that says:

> Important:
> To complete the tutorial, your account, the vault, and the application that you will register in this step must all be in the same Azure directory.

Unfortunately, if the user's account belongs to a different AD, then it is impossible to continue.  This is especially problematic for co-admins with an MSA  (hotmail.com, etc.)

Allowing the `TenantId` to be an optional parameter will solve the problem.

(Apologies in advance, but I couldn't figure out where to add unit tests for this change.)

### DIFF
--- a/src/ResourceManager/KeyVault/Commands.KeyVault/Commands/NewAzureKeyVault.cs
+++ b/src/ResourceManager/KeyVault/Commands.KeyVault/Commands/NewAzureKeyVault.cs
@@ -90,6 +90,11 @@ namespace Microsoft.Azure.Commands.KeyVault
             HelpMessage = "A hash table which represents resource tags.")]
         public Hashtable[] Tag { get; set; }
 
+        [Parameter(Mandatory = false,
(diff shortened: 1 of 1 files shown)


mj1856:
X-Plat version is here: Azure/azure-xplat-cli#2712

hovsepm:
@azuresdkci test this please

mj1856:
Closing per comments in the xplat version, and discussion with KeyVault team.

PR #2069, opened 2016-04-08
Add support for "add" and "create" permission when genreating SAS token against blob/container/file/share

blueww:
1. Change the permission parse function written by us to  Client library function PermissionsFromString(), so PowerShell will naturally support new added permission in client library. (Help don't need change as we don't mention which permission is supported in help)
2. remove the Unit test for the permission parse function written by us, since we don't use these function now.
3. Add the change to the change log . (please note, the release date and version are up to change by one SDK team when get it.)


markcowl:
@blueww Are there unit tests for the AccessPolicyHelper function?

blueww:
No, there are no unit test for AccessPolicyHelper. But we have full regression test for it and has passed. 
Actually, there are no unit test for it before also. I only remove the unit test for permission parse function of each service B/T/Q/F, since we has replace that with XSCL functions. We don't lose test coverage by remove these unit test.

markcowl:
@blueww yes, but this removes all end-to-end tests, so the lack of unit tests is problematic.  Can you add in unit tests or end-to-end tests covering the SAS token scenario within the next week or so?  I am worried about the lack of coverage in this repo for this important scenario.

markcowl:
LGTM.  Will merge once the release branch is merged.

blueww:
As we add E@R and Cool feature, close this PR and open a new one to make it more clear.

PR #2118, opened 2016-04-19
Fix a typo

yantang-msft:
A simple fix for a typo.

### DIFF
--- a/src/ServiceManagement/Compute/Commands.ServiceManagement/IaaS/Extensions/Diagnostics/SetAzureVMDiagnosticsExtension.cs
+++ b/src/ServiceManagement/Compute/Commands.ServiceManagement/IaaS/Extensions/Diagnostics/SetAzureVMDiagnosticsExtension.cs
@@ -241,7 +241,7 @@ namespace Microsoft.WindowsAzure.Commands.ServiceManagement.IaaS.Extensions
         private void GetResourceId()
         {
             var vmRoleContext = VM as PersistentVMRoleContext;
-            if (VM != null)
+            if (vmRoleContext != null)
             {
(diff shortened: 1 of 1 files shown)


markcowl:
@yantang-msft Is there some functionality this fixes?  If so, can you please add or modify a test to prevent future regressions?

yantang-msft:
> Is there some functionality this fixes?  If so, can you please add or modify a test to prevent future regressions?

The functionality is the same. Just make the logic right for the type casting.

markcowl:
@yantang-msft The functionality is not the same.  There are now cases where this code will not execute, and I want to make sure this is tested

yantang-msft:
> The functionality is not the same.  There are now cases where this code will not execute, and I want to make sure this is tested

The functionality for what we announced is the same. The code path that should execute still executes. The path that should not execute is now really prevented by the fix.
To create a test to verify such simple fix probably doesn't deserve the effort.

hovsepm:
@yantang-msft 
no need to add a new test - you can modify the existing one. Adding test coverage for bugfixes makes sure that your new feature works as designed. Plus it will be a gatekeeper  for future code changes to exclude regression bugs. 
Otherwise please explicitly sign-off  on this fix and specify that tests here are not required.

markcowl:
@yantang-msft Looking at this again, the 'VM' property is not used anywhere in the code block, and any violation would throw at the second statement.  I would still like to see coverage of this, as adequate testing should have prevented this issue in the first place, but won't gate this fix on it.

markcowl:
on-demand run here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/695/

PR #2119, opened 2016-04-20
Add E@R and Cool Support when Create Account, Add "add" and "create" permission when genreating SAS token

blueww:
Add E@R and Cool Support when Create Account:
1. Upgrade SRP SDK t0 5.0 preview
2. Add Parameter to New/Set AzureRMstoageAccount cmdlets to support E@R and Cool 
3. All changes has passed Regression test.

Add "add" and "create" permission:
 This part has already passed review in PR: #2069

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,3 +1,20 @@
+##2016.05.03 version 1.4.0
+* Azure Storage
+  * Upgrade to Microsoft.Azure.Management.Storage nuget package v5.0 preview
+  * Add Encryption and Hot/Cool features support to resource mode storage account cmdlets
+    - New-AzureRmStorageAccount
+    - Set-AzureRmStorageAccount
+  * Add "Add" and "Create" permission to Blob SAS cmdlets
+    - New-AzureStorageBlobSASToken
+    - New-AzureStorageContainerSASToken
+    - New-AzureStorageContainerStoredAccessPolicy
+    - Set-AzureStorageContainerStoredAccessPolicy
+  * Add "Create" permission to File SAS cmdlets
+    - New-AzureStorageFileSASToken
+    - New-AzureStorageShareSASToken
+    - New-AzureStorageShareStoredAccessPolicy
+    - Set-AzureStorageShareStoredAccessPolicy
+
 ##2016.04.19 version 1.3.2
 * Add support for specifying NIC/VMSS as application gateway backend address
 * Fix HDI ADL cluster creation and live test


azurecla:
Hi __@blueww__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (weiwei). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

blueww:
The on-demand test passed in http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/689/ 
(I have a change on change log after it, but that should not impact the Unit test result.)

src/ResourceManager/Storage/Commands.Management.Storage/Commands.Management.Storage.csproj
```diff
@@ -135,12 +136,6 @@
     <Reference Include="System.Management.Automation, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL" />
   </ItemGroup>
   <ItemGroup>
-    <Compile Include="..\..\..\Common\Commands.Common.Storage\Adapters\ARM.Storage.3\ARMStorageProvider.cs">
```
markcowl: @blueww I think the right thing here is to write the needed adapters for the new storage management client.  It looks like you copied over the methods instead.

src/ResourceManager/Storage/Commands.Management.Storage/StorageAccount/StorageAccountBaseCmdlet.cs
```diff
@@ -76,37 +93,97 @@ public string SubscriptionId
...
             {
-                return AccountType.PremiumLRS;
+                accountEncryption.Services.Blob = new EncryptionService();
+                accountEncryption.Services.Blob.Enabled = false;
             }
-            throw new ArgumentOutOfRangeException("accountType");
+            return accountEncryption;
+        }
+
+        public static StorageCredentials GenerateStorageCredentials(IStorageManagementClient storageClient, string resourceGroupName, string accountName)
```
markcowl: @blueww I *think* these are largely copied from commands.common.storage.  If possible, it would be better to write the version 5 adapters ans use the common functions instead.  This way other can reuse your adapter if they decide to update their storage management versions.

src/ResourceManager/Storage/Commands.Management.Storage/Commands.Management.Storage.csproj
blueww: As there are no one else need to use SRP SDK 5.0, and we are on a tight schedule for the PR, implement the adapter interface for the new storage version might NOT be our priority now.
We can write an adapt in the future when someone else need that.

src/ResourceManager/Storage/Commands.Management.Storage/StorageAccount/StorageAccountBaseCmdlet.cs
blueww: As there are no one else need to use SRP SDK 5.0, and we are on a tight schedule for the PR, implement the adapter interface for the new storage version might NOT be our priority now.
We can write an adapt in the future when someone else need that.

PR #2129, opened 2016-04-22
[DO NOT MERGE] Fix tags usage in ARM cmdlets

vivsriaus:


vivsriaus:
@TianoMS is added to the review. #Closed

src/ResourceManager/Tags/Commands.Tags/Model/TagsConversionHelper.cs
```diff
@@ -24,21 +25,23 @@ public class TagsConversionHelper
     {
         public static PSTagValuePair Create(Hashtable hashtable)
```
markcowl: We need to move the ConversionHelper into Copmmands.ResourceManager.Common, so RPs can depend on it without having to depend on the Tags cmdlets.

PR #2154, opened 2016-04-26
Made some parameter name changes for GetRP and restore

mkherani:


mkherani:
Created PR against wrong repo. closing it

PR #2157, opened 2016-04-27
Fix list intelligence packs so it returns a full list

chfunk:
Was able to get an updated session record.  Minor updates to list function and updates version of sdk.


hovsepm:
on-demand: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/712/

PR #2266, opened 2016-05-14
Application package cmdlets for Batch

brnleehng:


hovsepm:
add licence header to all newly added files

PR #2281, opened 2016-05-18
RDBug 6410988:[PSH] Change the default value of -KeyName in New-Azure…

blueww:
Set the default value of -KeyName in New-AzureRMStorageAccountKey to lower case "key1" "key2", and only lower case is valid value for SRP SDK interface.


blueww:
On Demand build is in http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/689/

Issue #2280, opened 2016-05-18
Definition of New-AzureVirtualNetworkGatewayConnection shows EnableBgp parameter is optional, but command fails if it is not specified


cg6:
This code has worked in a previous version
```
New-AzureVirtualNetworkGatewayConnection  -gatewayConnectionName "$source-to-$dest" `
                                              -gatewayConnectionType IPsec `
                                              -sharedKey $lgwconnection.SharedKey `
                                              -virtualNetworkGatewayId $lgwconnection.VirtualNetworkGatewayId `
                                              -connectedEntityId $lgwconnection.ConnectedEntityId
```

Now it returns: 
> New-AzureVirtualNetworkGatewayConnection : Value cannot be null.
> Parameter name: value

This worked
```
    New-AzureVirtualNetworkGatewayConnection  -gatewayConnectionName "$source-to-$dest" `
                                              -gatewayConnectionType IPsec `
                                              -sharedKey $lgwconnection.SharedKey `
                                              -virtualNetworkGatewayId $lgwconnection.VirtualNetworkGatewayId `
                                              -connectedEntityId $lgwconnection.ConnectedEntityId `
                                              -EnableBgp $false
```

The definition shows EnableBgp is optional:
```
get-command New-AzureVirtualNetworkGatewayConnection |% definition

New-AzureVirtualNetworkGatewayConnection [-ConnectedEntityId] <string> [-GatewayConnectionName] <string> [-GatewayConnectionType] <string> [[-RoutingWeight] <int>] [[-SharedKey] <string>] [-VirtualNetworkGatewayId] <string> [[-EnableBgp] <string>] [-Profile <AzureSMProfile>] [<CommonParameters>]
```

cg6:
This bug was introduced here:
https://github.com/Azure/azure-powershell/commit/659f2cf0243decaa898afa632498bb4d5eac175c

The EnableBgp parameter was changed from a boolean to a string, then later bool.Parse(EnableBgp) is called. bool.Parse returns the "Value cannot be null" error if EnableBgp is null.

bool.Parse could also return an unhandled exception if EnableBgp is not equivalent to TrueString or FalseString. If EnableBgp is going to be a string parameter there should be a validator against it. However I don't see why it shouldn't just remain a boolean.

lamchester:
Thanks for catching this! 

[...]

PR #2287, opened 2016-05-19
Prevent unhandled exception if EnableBgp is not specified (issue #2280)

lamchester:
Only call bool.Parse if a string that's not null or empty is passed to the -EnableBgp parameter


azurecla:
Hi __@lamchester__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (chelam). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

markcowl:
@lamchester can we add a test to prevent regressions in this behavior?

PR #2288, opened 2016-05-19
Add new PowerShell Cmdlet for ResourceManager\MachineLearning

NonStatic2014:
New PowerShell cmdlet for the MachineLearning WebServices.

### DIFF
--- a/src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ImportWebServiceDefinition.cs
+++ b/src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ImportWebServiceDefinition.cs
@@ -50,7 +50,9 @@ namespace Microsoft.Azure.Commands.MachineLearning.Cmdlets
                         ImportWebServiceDefinition.ImportFromFileParamSet, 
                         StringComparison.OrdinalIgnoreCase))
             {
-                jsonDefinition = CmdletHelpers.GetWebServiceDefinitionFromFile(this.InputFile);
+                jsonDefinition = CmdletHelpers.GetWebServiceDefinitionFromFile(
+                                        this.SessionState.Path.CurrentFileSystemLocation.Path, 
+                                        this.InputFile);
             }
 
             WebService serviceDefinition = ModelsSerializationUtil.GetAzureMLWebServiceFromJsonDefinition(jsonDefinition);
--- a/src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/NewAzureMLWebService.cs
+++ b/src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/NewAzureMLWebService.cs
@@ -83,8 +83,12 @@ namespace Microsoft.Azure.Commands.MachineLearning.Cmdlets
                                 NewAzureMLWebService.CreateFromFileParameterSet,
                                 StringComparison.OrdinalIgnoreCase))
                     {
-                        string jsonDefinition = CmdletHelpers.GetWebServiceDefinitionFromFile(this.DefinitionFile);
-                        this.NewWebServiceDefinition = ModelsSerializationUtil.GetAzureMLWebServiceFromJsonDefinition(jsonDefinition);
+                        string jsonDefinition = 
+                                CmdletHelpers.GetWebServiceDefinitionFromFile(
+                                                this.SessionState.Path.CurrentFileSystemLocation.Path,
+                                                this.DefinitionFile);
+                        this.NewWebServiceDefinition = 
+                                ModelsSerializationUtil.GetAzureMLWebServiceFromJsonDefinition(jsonDefinition);
                     }
 
                     WebService newWebService =
--- a/src/ResourceManager/MachineLearning/Commands.MachineLearning/Utilities/CmdletHelpers.cs
+++ b/src/ResourceManager/MachineLearning/Commands.MachineLearning/Utilities/CmdletHelpers.cs
@@ -49,10 +49,19 @@ namespace Microsoft.Azure.Commands.MachineLearning.Utilities
             return false;
         }
 
-        internal static string GetWebServiceDefinitionFromFile(string definitionFilePath)
+        internal static string GetWebServiceDefinitionFromFile(string currentPath, string definitionFilePath)
         {
-            var definitionFileFullPath = Path.GetFullPath(definitionFilePath);
-            if (!File.Exists(definitionFileFullPath))
+            string definitionFileFullPath;
+            if (Path.IsPathRooted(definitionFilePath))
+            {
+                definitionFileFullPath = definitionFilePath;
+            }
+            else
+            {
+                definitionFileFullPath = Path.Combine(currentPath, definitionFilePath);
+            }
+
+            if (!File.Exists(definitionFilePath))
             {
                 throw new FileNotFoundException(Resources.MissingDefinitionFile, definitionFilePath);
             }


azurecla:
Hi __@NonStatic2014__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (bohu). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

markcowl:
on demand run here:http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/797/

.gitignore
```diff
@@ -201,3 +201,8 @@ FakesAssemblies/
 /tools/*.dll
 *.GhostDoc.xml
 pingme.txt
+/tools/localFeed/Microsoft.Rest.ClientRuntime.Azure.3.1.0.nupkg
```
markcowl: you shouldn't need to push these to the main fork

src/ResourceManager/MachineLearning/Commands.MachineLearning.Test/ScenarioTests/WebServicesTestController.cs
```diff
@@ -0,0 +1,216 @@
...
+                        cleanup();
+                    }
+                }
+            }
+        }
+
+        private void SetupManagementClients(MockContext context)
+        {
+            this.ResourceManagementClient = LegacyTest.TestBase.GetServiceClient<ResourceManagementClient>(this.csmTestFactory);
+            this.WebServicesManagementClient = context.GetServiceClient<AzureMLWebServicesManagementClient>(TestEnvironmentFactory.GetTestEnvironment());
```
markcowl: line breaks

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ExportWebServiceDefinition.cs
```diff
@@ -0,0 +1,55 @@
...
+        public override void ExecuteCmdlet()
+        {
+            string serializedDefinition = ModelsSerializationUtil.GetAzureMLWebServiceDefinitionJsonFromObject(this.WebService);
+            if (!string.IsNullOrWhiteSpace(this.ToFile))
+            {
+               File.WriteAllText(this.ToFile, serializedDefinition);
+            }
+            else
+            {
+                Console.WriteLine(serializedDefinition);
```
markcowl: Use WriteOutput

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ExportWebServiceDefinition.cs
```diff
@@ -0,0 +1,55 @@
...
+        private const string ExportToFileParamSet = "Export to file.";
+        private const string ExportToStringParamSet = "Export to JSON string.";
+
+        [Parameter(Mandatory = true, HelpMessage = "The web service definition object to export.")]
+        [ValidateNotNullOrEmpty]
+        public WebService WebService { get; set; }
+
+        [Parameter(ParameterSetName = ExportWebServiceDefinition.ExportToFileParamSet, Mandatory = true, HelpMessage = "Path to a file on disk where to export the web service definition in JSON format.")]
+        [ValidateNotNullOrEmpty]
+        public string ToFile { get; set; }
```
markcowl: This parameter name should probably be -OutputFile

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ExportWebServiceDefinition.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.IO;
+using System.Management.Automation;
+using Microsoft.Azure.Commands.ResourceManager.Common;
+using Microsoft.Azure.Management.MachineLearning.WebServices.Models;
+using Microsoft.Azure.Management.MachineLearning.WebServices.Util;
+
+namespace Microsoft.Azure.Commands.MachineLearning.Cmdlets
+{
+    [Cmdlet(VerbsData.Export, WebServicesCmdletBase.CommandletSuffix)]
+    public class ExportWebServiceDefinition : AzureRMCmdlet
```
markcowl: This cmdlet needs to implement ShouldProcess and  needs to check if the output file would be overwritten and implement Force/ShouldContinue to prompt for overwrite (or force overwriting)

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/GetAzureMLWebService.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.Commands.ResourceManager.Cmdlets.Components;
+using Microsoft.Azure.Management.MachineLearning.WebServices.Models;
+
+namespace Microsoft.Azure.Commands.MachineLearning
+{
+    [Cmdlet(VerbsCommon.Get, WebServicesCmdletBase.CommandletSuffix)]
+    public class GetAzureMLWebService : WebServicesCmdletBase
+    {
+        private const string GetMlWebServiceByNameGroupParameterSet = "Get an Azure ML web service resource details by name and group.";
```
markcowl: Please fix the line braks in this file (80 columns, they should fit into this window)

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ExportWebServiceDefinition.cs
```diff
@@ -0,0 +1,55 @@
...
+using System;
+using System.IO;
+using System.Management.Automation;
+using Microsoft.Azure.Commands.ResourceManager.Common;
+using Microsoft.Azure.Management.MachineLearning.WebServices.Models;
+using Microsoft.Azure.Management.MachineLearning.WebServices.Util;
+
+namespace Microsoft.Azure.Commands.MachineLearning.Cmdlets
+{
+    [Cmdlet(VerbsData.Export, WebServicesCmdletBase.CommandletSuffix)]
```
markcowl: Neeed to specify the output type of the cmdlet, using an OutputType attribute.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/GetAzureMLWebService.cs
```diff
@@ -0,0 +1,77 @@
...
+        [ValidateNotNullOrEmpty]
+        public string ResourceGroupName { get; set; }
+
+        [Parameter(ParameterSetName = GetAzureMLWebService.GetMlWebServiceByNameGroupParameterSet, Mandatory = true, HelpMessage = "The name of the web service.")]
+        [ValidateNotNullOrEmpty]
+        public string Name { get; set; }
+
+        [Parameter(ParameterSetName = GetAzureMLWebService.GetMlWebServicesByGroupParameterSet, Mandatory = true, HelpMessage = "When specified, ensures that the query is run against a paginated list of services instead of a resource.")]
+        [Parameter(ParameterSetName = GetAzureMLWebService.GetMlWebServicesInSubscriptionParameterSet, Mandatory = true, HelpMessage = "When specified, ensures that the query is run against a paginated list of services instead of a resource.")]
+        public SwitchParameter IsCollection { get; set; }
```
markcowl: You should not use a switch parameter to differentiate between listing all wevservices in the current scope versus getting details for a specific one.  This should be specified based on the other parameters (no Name means list)

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/GetAzureMLWebService.cs
```diff
@@ -0,0 +1,77 @@
...
+using System;
+using System.Management.Automation;
+using System.Threading.Tasks;
+using Microsoft.Azure.Commands.ResourceManager.Cmdlets.Components;
+using Microsoft.Azure.Management.MachineLearning.WebServices.Models;
+
+namespace Microsoft.Azure.Commands.MachineLearning
+{
+    [Cmdlet(VerbsCommon.Get, WebServicesCmdletBase.CommandletSuffix)]
+    public class GetAzureMLWebService : WebServicesCmdletBase
```
markcowl: OutputType attribute

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/GetAzureMLWebServiceKeys.cs
```diff
@@ -0,0 +1,58 @@
...
+
+using System;
+using System.Management.Automation;
+using Microsoft.Azure.Commands.MachineLearning.Utilities;
+using Microsoft.Azure.Management.MachineLearning.WebServices.Models;
+
+namespace Microsoft.Azure.Commands.MachineLearning.Cmdlets
+{
+    [Cmdlet(VerbsCommon.Get, WebServicesCmdletBase.CommandletSuffix + "Keys")]
+    public class GetAzureMLWebServiceKeys : WebServicesCmdletBase
```
markcowl: OutputType attribute

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/GetAzureMLWebService.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        [Parameter(ParameterSetName = GetAzureMLWebService.GetMlWebServiceByNameGroupParameterSet, Mandatory = true, HelpMessage = "The name of the web service.")]
+        [ValidateNotNullOrEmpty]
+        public string Name { get; set; }
+
+        [Parameter(ParameterSetName = GetAzureMLWebService.GetMlWebServicesByGroupParameterSet, Mandatory = true, HelpMessage = "When specified, ensures that the query is run against a paginated list of services instead of a resource.")]
+        [Parameter(ParameterSetName = GetAzureMLWebService.GetMlWebServicesInSubscriptionParameterSet, Mandatory = true, HelpMessage = "When specified, ensures that the query is run against a paginated list of services instead of a resource.")]
+        public SwitchParameter IsCollection { get; set; }
+
+        protected override void RunCmdlet()
```
markcowl: Why not use ExecuteCmdlet

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ImportWebServiceDefinition.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Management.Automation;
+using Microsoft.Azure.Commands.MachineLearning.Utilities;
+using Microsoft.Azure.Commands.ResourceManager.Common;
+using Microsoft.Azure.Management.MachineLearning.WebServices.Models;
+using Microsoft.Azure.Management.MachineLearning.WebServices.Util;
+
+namespace Microsoft.Azure.Commands.MachineLearning.Cmdlets
+{
+    [Cmdlet(VerbsData.Import, WebServicesCmdletBase.CommandletSuffix)]
+    public class ImportWebServiceDefinition : AzureRMCmdlet
```
markcowl: same general comments

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
```diff
@@ -0,0 +1,137 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Management.Automation;
+using Microsoft.Azure.Management.MachineLearning.WebServices.Models;
+using Microsoft.WindowsAzure.Commands.Common;
+
+namespace Microsoft.Azure.Commands.MachineLearning
+{
+    [Cmdlet(VerbsData.Update, WebServicesCmdletBase.CommandletSuffix)]
```
markcowl: attribute for supportsshouldprocess

markcowl:
Put in some specific examples of general comments for all cmdlets, these should be applied across the board:

(1) Do not put types from the amangement clients as parameter types or output types for cmdlets
(2) Specify OutputType attribute for each cmdlet
(3) PowerShell cmdlets should decide between whether a collection is returned based on the scope of the request, not based on switch parameters
(4) Please break lines in source code to allow to fit in GitHub windows (~ 80 columns)

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ExportWebServiceDefinition.cs
NonStatic2014: @markcowl Would you mind explain a little bit why we need to avoid outputting types?

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/GetAzureMLWebService.cs
NonStatic2014: Because Get*, New*, Remove* and Update* want to share the same exception handler in ExecuteCmdlet. So we create this RunCmdlet() in base class and implement it in each of sub-class.

src/ResourceManager/MachineLearning/Commands.MachineLearning.Test/ScenarioTests/WebServicesTestController.cs
NonStatic2014: @markcowl I will try to make the line break at between 80 - 100. But is there any tools or extensions to do it automatically?

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/GetAzureMLWebService.cs
NonStatic2014: Actually IsCollection is not really in use. Removed.

markcowl:
new on demand run here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/817/

src/ResourceManager/MachineLearning/AzureRM.MachineLearning.psd1
```diff
@@ -0,0 +1,110 @@
...
+DotNetFrameworkVersion = '4.0'  
+  
+# Minimum version of the common language runtime (CLR) required by this module  
+CLRVersion='4.0'  
+  
+# Processor architecture (None, X86, Amd64, IA64) required by this module  
+ProcessorArchitecture = 'None'  
+  
+# Modules that must be imported into the global environment prior to importing this module  
+RequiredModules = @( @{ ModuleName = 'AzureRM.Profile'; ModuleVersion = '1.0.4'})
```
markcowl: 1.0.8

src/ResourceManager/MachineLearning/Commands.MachineLearning.Test/ScenarioTests/Common.ps1
```diff
@@ -0,0 +1,179 @@
...
+    return "southcentralus"
+}
+
+<#
+.SYNOPSIS
+Gets the latest API Version for the resource type
+#>
+function Get-ProviderAPIVersion($providerNamespace, $resourceType)
+{ 
+    if ([Microsoft.Azure.Test.HttpRecorder.HttpMockServer]::Mode -ne [Microsoft.Azure.Test.HttpRecorder.HttpRecorderMode]::Playback)
```
markcowl: nit: line breaks. you can use the backtick to break lines in powershell scripts

src/ResourceManager/MachineLearning/Commands.MachineLearning.Test/ScenarioTests/WebServiceTests.ps1
```diff
@@ -0,0 +1,330 @@
...
+<#
+.SYNOPSIS
+Tests creating and updating an Azure ML web service resource
+#>
+function Test-UpdateWebService
+{
+    $actualTest = {
+        param([string] $rgName, [string] $location, [string] $webServiceName, [string] $commitmentPlanId, [object] $storageAccount)        
+        try 
+        {
```
markcowl: nit: line breaks

src/ResourceManager/MachineLearning/Commands.MachineLearning.Test/ScenarioTests/WebServiceTests.ps1
```diff
@@ -0,0 +1,330 @@
...
+    $svcDefinition.Properties.StorageAccount.Key = $storageAccount.Key
+
+    return $svcDefinition
+}
+
+function ValidateWebServiceResult([string] $rgName, [string] $webServiceName, [string] $location, [Microsoft.Azure.Management.MachineLearning.WebServices.Models.WebService] $svc)
+{
+    $subscriptionId = ((Get-AzureRmContext).Subscription).SubscriptionId        
+    $expectedServiceResourceId = "/subscriptions/$subscriptionId/resourceGroups/$rgName/providers/Microsoft.MachineLearning/webservices/$webServiceName"
+    LogOutput "Checking that the created webservice's resource id $($svc.Id) matches the expected value $expectedServiceResourceId"
```
markcowl: Please don't do this, this subverts the current logging of powershell scripts.  Use the standard powershell streams in this case the Verbose stream and you'll get to see this as part of test logs.

src/ResourceManager/MachineLearning/Commands.MachineLearning.Test/ScenarioTests/Common.ps1
```diff
@@ -0,0 +1,179 @@
...
+}
+
+<#
+.SYNOPSIS
+Writes a timestamped message to stdout
+#>
+function LogOutput($message)
+{
+    $timestamp = Get-Date -UFormat "%Y-%m-%d %H:%M:%S %Z"
+    Write-Output "[$timestamp]: $message"
```
markcowl: Generally speaking, you don't want to write messages to the output stream - use the Verbose stream or the Debug stream for these kinds of messages.

src/ResourceManager/MachineLearning/Commands.MachineLearning.Test/TestData/GraphWebServiceDefinition_Dogfood.json
```diff
@@ -0,0 +1,190 @@
+﻿{
+  "properties": {
+    "title": "ServiceTitle",
+    "description": "This is some service description",
+    "readOnly": "false",
+    "keys": {
+      "primary": "S8ayf24PAsbQqy8Nm67uGf3MbOeorYKPY8QfXen2+JsBZV5Q0NFfk9tgC7mjJ3aCsUPY8yKCFVcI/sCcOlmkKw==",
```
markcowl: Are these real credentials?  Do we need to define these explicitly in this fashion?

src/ResourceManager/MachineLearning/Commands.MachineLearning.Test/TestData/GraphWebServiceDefinition_Prod.json
```diff
@@ -0,0 +1,190 @@
+﻿{
+  "properties": {
+    "title": "ServiceTitle",
+    "description": "This is some service description",
+    "readOnly": "false",
+    "keys": {
+      "primary": "S8ayf24PAsbQqy8Nm67uGf3MbOeorYKPY8QfXen2+JsBZV5Q0NFfk9tgC7mjJ3aCsUPY8yKCFVcI/sCcOlmkKw==",
```
markcowl: real credentials?

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ExportWebServiceDefinition.cs
```diff
@@ -0,0 +1,80 @@
...
+        SupportsShouldProcess = true)]
+    [OutputType(typeof(string))]
+    public class ExportWebServiceDefinition : AzureRMCmdlet
+    {
+        private const string ExportToFileParamSet = "Export to file.";
+        private const string ExportToStringParamSet = "Export to JSON string.";
+
+        [Parameter(Mandatory = true, HelpMessage = "The web service definition object to export.")]
+        [ValidateNotNullOrEmpty]
+        public WebService WebService { get; set; }
```
markcowl: How does a user create or get a web service to pass to this command?  Is this something you would expect to come from the pipeline?

src/ResourceManager/MachineLearning/Commands.MachineLearning.Test/TestData/GraphWebServiceDefinition_Dogfood.json
NonStatic2014: It is the key to access the web service you created with the cmdlet. It is part of the data structure here.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ExportWebServiceDefinition.cs
```diff
@@ -0,0 +1,80 @@
...
+        [Parameter(
+            Mandatory = false, 
+            HelpMessage = "Do not ask for confirmation.")]
+        public SwitchParameter Force { get; set; }
+
+        public override void ExecuteCmdlet()
+        {
+            string serializedDefinition = 
+                ModelsSerializationUtil.GetAzureMLWebServiceDefinitionJsonFromObject(this.WebService);
+            if (!string.IsNullOrWhiteSpace(this.OutputFile))
```
markcowl: In this case, the pattern should be:
```
if ShouldProcess(target, action) {
   if (!File.Exists(outputFile) || Force || ShouldContinue(...) { ... actions to write file text...}
}
```

Unfortunately, there isn't any good common code that implements this pattern right now.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/GetAzureMLWebService.cs
```diff
@@ -0,0 +1,92 @@
...
+using System;
+using System.Management.Automation;
+using System.Threading.Tasks;
+using Microsoft.Azure.Commands.ResourceManager.Cmdlets.Components;
+using Microsoft.Azure.Management.MachineLearning.WebServices.Models;
+
+namespace Microsoft.Azure.Commands.MachineLearning
+{
+    [Cmdlet(VerbsCommon.Get, WebServicesCmdletBase.CommandletSuffix)]
+    [OutputType(typeof(WebService), typeof(Array))]
```
markcowl: You can make the second type WebService[]

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/GetAzureMLWebService.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            // If this is a simple get web service by name operation, resolve it as such
+            if (!string.IsNullOrWhiteSpace(this.ResourceGroupName) && 
+                !string.IsNullOrWhiteSpace(this.Name))
+            {
+                WebService service =
+                    this.WebServicesClient.GetAzureMlWebService(this.SubscriptionId, this.ResourceGroupName, this.Name);
+                this.WriteObject(service);
+            }
+            else
```
markcowl: Can't the user specify just a resource group or just a name?

src/ResourceManager/MachineLearning/Commands.MachineLearning.Test/TestData/GraphWebServiceDefinition_Prod.json
NonStatic2014: You could consider it the access key to the web service and only the web services.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ImportWebServiceDefinition.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        private const string ImportFromFileParamSet = "Import from JSON file.";
+        private const string ImportFromStringParamSet = "Import from JSON string.";
+
+        [Parameter(
+            ParameterSetName = ImportWebServiceDefinition.ImportFromFileParamSet, 
+            Mandatory = true, 
+            HelpMessage = "Path to a file on disk containing the web service definition in JSON format.")]
+        [ValidateNotNullOrEmpty]
+        public string FromFile { get; set; }
```
markcowl: I would suggest -InputFile or -File

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/GetAzureMLWebService.cs
NonStatic2014: They could, and the "else" part will take care them.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ImportWebServiceDefinition.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        [Parameter(
+            ParameterSetName = ImportWebServiceDefinition.ImportFromStringParamSet, 
+            Mandatory = true, 
+            HelpMessage = "The actual web service definition as a JSON string.")]
+        [ValidateNotNullOrEmpty]
+        public string FromJsonString { get; set; }
```
markcowl: Again, probably no From needed

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ImportWebServiceDefinition.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public override void ExecuteCmdlet()
+        {
+            string jsonDefinition = this.FromJsonString;
```
markcowl:
```
if (ShouldProcess(target, action).) {... execution code...}

File: src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ImportWebServiceDefinition.cs [hunk 28]
```diff
@@ -0,0 +1,60 @@
...
+using System;
+using System.Management.Automation;
+using Microsoft.Azure.Commands.MachineLearning.Utilities;
+using Microsoft.Azure.Commands.ResourceManager.Common;
+using Microsoft.Azure.Management.MachineLearning.WebServices.Models;
+using Microsoft.Azure.Management.MachineLearning.WebServices.Util;
+
+namespace Microsoft.Azure.Commands.MachineLearning.Cmdlets
+{
+    [Cmdlet(VerbsData.Import, WebServicesCmdletBase.CommandletSuffix)]
```
markcowl: SupportsShouldProcess=true

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/NewAzureMLWebService.cs
```diff
@@ -0,0 +1,81 @@
...
+using System;
+using System.Management.Automation;
+using Microsoft.Azure.Commands.MachineLearning.Utilities;
+using Microsoft.Azure.Management.MachineLearning.WebServices.Models;
+using Microsoft.Azure.Management.MachineLearning.WebServices.Util;
+
+namespace Microsoft.Azure.Commands.MachineLearning.Cmdlets
+{
+    [Cmdlet(VerbsCommon.New, WebServicesCmdletBase.CommandletSuffix), OutputType(typeof(WebService))]
+    public class NewAzureMLWebService : WebServicesCmdletBase
```
markcowl: SuppoprtsShouldProcess=true

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ExportWebServiceDefinition.cs
NonStatic2014: Yes, this command expect user call Get/New/Update and get an instance of WebService, then call this command to serialize to a file or string.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/NewAzureMLWebService.cs
```diff
@@ -0,0 +1,81 @@
...
+            ParameterSetName = NewAzureMLWebService.CreateFromObjectParameterSet, 
+            Mandatory = true, 
+            HelpMessage = "The definition of the new web service.", 
+            ValueFromPipeline = true)]
+        [ValidateNotNullOrEmpty]
+        public WebService NewWebServiceDefinition { get; set; }
+
+        protected override void RunCmdlet()
+        {
+            if (string.Equals(
```
markcowl:
```
if (ShouldProcess(target, action)) {... execution actions...}
```

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/RemoveAzureMLWebService.cs
```diff
@@ -0,0 +1,92 @@
...
+using System;
+using System.Management.Automation;
+using Microsoft.Azure.Commands.MachineLearning.Utilities;
+using Microsoft.Azure.Management.MachineLearning.WebServices.Models;
+using Microsoft.WindowsAzure.Commands.Common;
+
+namespace Microsoft.Azure.Commands.MachineLearning
+{
+    [Cmdlet(VerbsCommon.Remove, WebServicesCmdletBase.CommandletSuffix)]
+    [OutputType(typeof(void))]
```
markcowl: SupportsShouldProcess=true

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ImportWebServiceDefinition.cs
NonStatic2014: I think this command does not change anything in local system or cloud, so do we still need the "ShouldProcess"?

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ImportWebServiceDefinition.cs
NonStatic2014: Same comments as above one.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/RemoveAzureMLWebService.cs
```diff
@@ -0,0 +1,92 @@
...
+                                    out webServiceName))
+                {
+                    throw new ValidationMetadataException(Resources.InvalidWebServiceIdOnObject);
+                }
+
+                this.ResourceGroupName = resourceGroup;
+                this.Name = webServiceName;
+            }
+
+            this.ConfirmAction(
```
markcowl: This is likely the wrong thing to do (although this pattern is used in many places, it is actually incorrect and is in process of beign changed across the board).

In this case, you probably don't want a -Force parameter, unless there are some special circumstances in which you want an extra prompt for the user (for example, if you can detect that the web service is in use, you might want an extra prompt).

For the no special deletion, just use:
```
if (ShouldProcess(target, action) {...normal execution...}
```
and do not have a force parameter.  If there are special conditions under which you'd like to do an extra prompt:
```
if (ShouldProcess(target, action)) {
... actions
  if (!SpecialConditionPredicate() || Force || ShouldContinue(...)) { ... perform deletion ...}
}
```

LMK if this isn't clear

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
```diff
@@ -0,0 +1,195 @@
...
+namespace Microsoft.Azure.Commands.MachineLearning
+{
+    [Cmdlet(
+        VerbsData.Update, 
+        WebServicesCmdletBase.CommandletSuffix, 
+        SupportsShouldProcess = true)]
+    [OutputType(typeof(WebService))]
+    public class UpdateAzureMLWebService : WebServicesCmdletBase
+    {
+        protected const string UpdateFromArgumentsParameterSet =
```
markcowl: Are there any piping scenarios for this command?  It is a frequent scenario to use 
```
Get-AzureRmMLWebService <params> | { code block to do some transformations} | Update-AzureRmMLWebService
```
In this case, you might want to take a pipeline parameter to identify the web service and additional parameters for changeable properties.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
```diff
@@ -0,0 +1,195 @@
...
+            ParameterSetName = UpdateAzureMLWebService.UpdateFromArgumentsParameterSet, 
+            Mandatory = false, 
+            HelpMessage = "New realtime endpoint runtime settings for the web service.")]
+        public RealtimeConfiguration RealtimeConfiguration { get; set; }
+
+        [Parameter(
+            ParameterSetName = UpdateAzureMLWebService.UpdateFromArgumentsParameterSet, 
+            Mandatory = false, 
+            HelpMessage = "Updated assets for the web service.")]
+        public IDictionary<string, AssetItem> Assets { get; set; }
```
markcowl: PowerShell doesn't deal well with these - Hashtable is a better bet

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
```diff
@@ -0,0 +1,195 @@
...
+            ParameterSetName = UpdateAzureMLWebService.UpdateFromArgumentsParameterSet, 
+            Mandatory = false, 
+            HelpMessage = "New diagnostics settings for the web service.")]
+        public DiagnosticsConfiguration Diagnostics { get; set; }
+
+        [Parameter(
+            ParameterSetName = UpdateAzureMLWebService.UpdateFromArgumentsParameterSet, 
+            Mandatory = false, 
+            HelpMessage = "New realtime endpoint runtime settings for the web service.")]
+        public RealtimeConfiguration RealtimeConfiguration { get; set; }
```
markcowl: If this is an enum, this shoudl be fine.  If not, how does the user create this?

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
```diff
@@ -0,0 +1,195 @@
...
+            ParameterSetName = UpdateAzureMLWebService.UpdateFromArgumentsParameterSet, 
+            Mandatory = false, 
+            HelpMessage = "New access key for the storage account associated with the web service. This allows for key rotation if needed.")]
+        public string StorageAccountKey { get; set; }
+
+        [Parameter(
+            ParameterSetName = UpdateAzureMLWebService.UpdateFromArgumentsParameterSet, 
+            Mandatory = false, 
+            HelpMessage = "New diagnostics settings for the web service.")]
+        public DiagnosticsConfiguration Diagnostics { get; set; }
```
markcowl: How does the user create this?

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
```diff
@@ -0,0 +1,195 @@
...
+            ParameterSetName = UpdateAzureMLWebService.UpdateFromArgumentsParameterSet, 
+            Mandatory = false, 
+            HelpMessage = "Mark the service as readonly.")]
+        public SwitchParameter IsReadOnly { get; set; }
+
+        [Parameter(
+            ParameterSetName = UpdateAzureMLWebService.UpdateFromArgumentsParameterSet, 
+            Mandatory = false, 
+            HelpMessage = "New access keys for the web service.")]
+        public WebServiceKeys Keys { get; set; }
```
markcowl: How to create?

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
```diff
@@ -0,0 +1,195 @@
...
+            ParameterSetName = UpdateAzureMLWebService.UpdateFromArgumentsParameterSet, 
+            Mandatory = false, 
+            HelpMessage = "Updated assets for the web service.")]
+        public IDictionary<string, AssetItem> Assets { get; set; }
+
+        [Parameter(
+            ParameterSetName = UpdateAzureMLWebService.UpdateFromArgumentsParameterSet, 
+            Mandatory = false, 
+            HelpMessage = "Updated input schema for the web service.")]
+        public ServiceInputOutputSpecification Input { get; set; }
```
markcowl: How to create?

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
```diff
@@ -0,0 +1,195 @@
...
+
+        [Parameter(
+            ParameterSetName = UpdateAzureMLWebService.UpdateFromArgumentsParameterSet, 
+            Mandatory = false, 
+            HelpMessage = "Updated output schema for the web service.")]
+        public ServiceInputOutputSpecification Output { get; set; }
```
markcowl: How to create?

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
NonStatic2014: We do have a test case shows how to do it. So I think our users should be able to do the same thing. For the definition of each structure, the user could find it out from the web service definition.

            # Update realtime endpoint settings and the service key by using in line parameters
            $newPrimaryKey = 'highly secure key'
            LogOutput "Updating in line properties on service $($svc.Id)"
            $updatedSvc2 = Update-AzureRmMlWebService -ResourceGroupName $rgName -Name $webServiceName -RealtimeConfiguration @{ MaxConcurrentCalls = 30 } -Keys @{ Primary = $newPrimaryKey }
            Assert-NotNull $updatedSvc2
            LogOutput "Update has completed."
            $update2ModifiedOn = [datetime]::Parse($updatedSvc2.Properties.ModifiedOn)
            LogOutput "Web service's last modified time stamp: $update2ModifiedOn"

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
```diff
@@ -0,0 +1,195 @@
...
+
+        [Parameter(
+            ParameterSetName = UpdateAzureMLWebService.UpdateFromArgumentsParameterSet, 
+            Mandatory = false, 
+            HelpMessage = "Updated global parameter values definition for the web service.")]
+        public IDictionary<string, string> Parameters { get; set; }
```
markcowl: Hashtable

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
NonStatic2014: Similar to above.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
NonStatic2014: Similar to above.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
```diff
@@ -0,0 +1,195 @@
...
+
+        [Parameter(
+            ParameterSetName = UpdateAzureMLWebService.UpdateFromArgumentsParameterSet, 
+            Mandatory = false, 
+            HelpMessage = "Updated graph package for the web service.")]
+        public GraphPackage Package { get; set; }
```
markcowl: How to create?

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
NonStatic2014: Similar to above.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
NonStatic2014: Similar to above.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
```diff
@@ -0,0 +1,195 @@
...
+
+        [Parameter(
+            ParameterSetName = UpdateAzureMLWebService.UpdateFromObjectParameterSet, 
+            Mandatory = true, 
+            HelpMessage = "An updated definition object to update the referenced web service with.", 
+            ValueFromPipeline = true)]
+        public WebService ServiceUpdates { get; set; }
```
markcowl: So, does it make sense to take the value from the pipeline and update with additional parameters?

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
```diff
@@ -0,0 +1,195 @@
...
+                        StringComparison.OrdinalIgnoreCase))
+            {
+                isUpdateToReadonly = this.ServiceUpdates.Properties != null &&
+                                     this.ServiceUpdates.Properties.ReadOnlyProperty.HasValue &&
+                                     this.ServiceUpdates.Properties.ReadOnlyProperty.Value;
+            }
+
+            if (isUpdateToReadonly)
+            {
+                this.ConfirmAction(
```
markcowl: See previous comment.  You do not need a Force in this case unless you want to prevent overwriting certain properties in some circumstances.  Most likely you justw ant no Force parameter and:

```
if (ShouldProcess(target, action)) { ... actions ...}
```

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/WebServicesCmdletBase.cs
```diff
@@ -0,0 +1,261 @@
...
+using Microsoft.Azure.Commands.ResourceManager.Cmdlets.Extensions;
+using Microsoft.Azure.Commands.ResourceManager.Common;
+using Microsoft.Rest.Azure;
+using PSResourceManagerModels = Microsoft.Azure.Commands.Resources.Models;
+
+namespace Microsoft.Azure.Commands.MachineLearning
+{
+    public abstract class WebServicesCmdletBase : AzureRMCmdlet
+    {
+        public const string CommandletSuffix = "AzureRmMlWebService";
```
markcowl: Wow do I hate that suffix, but it is, unfortunately, correct :-)

markcowl: Note that you can cosider adding shorter cmdlet aliases for this release or in a subsequent release.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/WebServicesCmdletBase.cs
```diff
@@ -0,0 +1,261 @@
...
+        protected override void BeginProcessing()
+        {
+            try
+            {
+                if (this.cancellationSource == null)
+                {
+                    this.cancellationSource = new CancellationTokenSource();
+                }
+
+                this.SubscriptionId = this.DefaultProfile.Context.Subscription.Id.ToString();
```
markcowl: Why is this needed?

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/WebServicesCmdletBase.cs
```diff
@@ -0,0 +1,261 @@
...
+                base.BeginProcessing();
+            }
+            catch (Exception ex)
+            {
+                if (ex.IsFatal())
+                {
+                    throw;
```
markcowl: Instead, consider calling ```Cmdlet.ThrowTerminatingError(...)```

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/WebServicesCmdletBase.cs
```diff
@@ -0,0 +1,261 @@
...
+        {
+            try
+            {
+                base.EndProcessing();
+            }
+            catch (Exception ex)
+            {
+                if (ex.IsFatal())
+                {
+                    throw;
```
markcowl: consider calling ```Cmdlet.ThrowTerminatingError(...)```

src/ResourceManager/MachineLearning/Commands.MachineLearning/Commands.MachineLearning.csproj
```diff
@@ -0,0 +1,160 @@
...
+    </ProjectReference>
+    <ProjectReference Include="..\..\Common\Commands.ResourceManager.Common\Commands.ResourceManager.Common.csproj">
+      <Project>{3819d8a7-c62c-4c47-8ddd-0332d9ce1252}</Project>
+      <Name>Commands.ResourceManager.Common</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\..\Resources\Commands.ResourceManager\Cmdlets\Commands.Resources.Rest.csproj">
+      <Project>{8058d403-06e3-4bed-8924-d166ce303961}</Project>
+      <Name>Commands.Resources.Rest</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\..\Resources\Commands.Resources\Commands.Resources.csproj">
```
markcowl: Dop not reference other cmdlets in your cmdlet project.  This makes it difficult to version these modules independently.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Commands.MachineLearning.csproj
```diff
@@ -0,0 +1,160 @@
...
+    </ProjectReference>
+    <ProjectReference Include="..\..\..\Common\Commands.Common\Commands.Common.csproj">
+      <Project>{5ee72c53-1720-4309-b54b-5fb79703195f}</Project>
+      <Name>Commands.Common</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\..\Common\Commands.ResourceManager.Common\Commands.ResourceManager.Common.csproj">
+      <Project>{3819d8a7-c62c-4c47-8ddd-0332d9ce1252}</Project>
+      <Name>Commands.ResourceManager.Common</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\..\Resources\Commands.ResourceManager\Cmdlets\Commands.Resources.Rest.csproj">
```
markcowl: Do not reference other cmdlets in your cmdlet project.  This makes it difficult to version these modules independently.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Extensions/ErrorResponseMessageExceptionExtensions.cs
```diff
@@ -0,0 +1,112 @@
...
+                errorReport.AppendLine("Error Code: {0}".FormatInvariant(exception.ErrorResponseMessage.Error.Code));
+                errorReport.AppendLine("Error Message: {0}".FormatInvariant(exception.ErrorResponseMessage.Error.Message));
+                errorReport.AppendLine("Error Target: {0}".FormatInvariant(exception.ErrorResponseMessage.Error.Target));
+                
+                if (exception.ErrorResponseMessage.Error.Details.Any())
+                {
+                    errorReport.AppendLine("Error Details:");
+                    foreach (var errorDetail in exception.ErrorResponseMessage.Error.Details)
+                    {
+                        errorReport.AppendLine("\t[Code={0}, Message={1}]".FormatInvariant(errorDetail.Code, errorDetail.Message));
```
markcowl: line breaks

src/ResourceManager/MachineLearning/Commands.MachineLearning/Microsoft.Azure.Commands.MachineLearning.dll-help.xml
```diff
@@ -0,0 +1,1477 @@
...
+			<maml:name>Parameters</maml:name>
+			<maml:description>
+				<maml:para>The set of global parameters values defined for the web service, given as a global parameter name -&gt; default value collection. If no default value is specified, the parameter is considered to be required.</maml:para>
+			</maml:description>
+			<command:parameterValue required="true" variableLength="true">IDictionary`2[String]</command:parameterValue>
+			<dev:type>
+				<maml:name>IDictionary`2[String]</maml:name>
+				<maml:uri/>
+			</dev:type>
+
```
markcowl: Thank you for adding examples!

src/ResourceManager/MachineLearning/Commands.MachineLearning/Utilities/CmdletHelpers.cs
```diff
@@ -0,0 +1,61 @@
...
+using Microsoft.Azure.Management.MachineLearning.WebServices.Models;
+using Microsoft.Azure.Management.MachineLearning.WebServices.Util;
+
+namespace Microsoft.Azure.Commands.MachineLearning.Utilities
+{
+    internal static class CmdletHelpers
+    {
+        private static readonly Regex MlWebServiceResourceIdRegex = 
+            new Regex(
+                    @"^\/subscriptions\/(?<subscriptionId>[^\/]+)\/resourceGroups\/(?<resourceGroupName>[^\/]+)\/providers\/Microsoft.MachineLearning\/webservices\/(?<webServiceName>[^\/]+)$",
```
markcowl: line breaks

src/ResourceManager/MachineLearning/Commands.MachineLearning/Utilities/WebServicesClient.cs
```diff
@@ -0,0 +1,142 @@
...
+{
+    public class WebServicesClient
+    {
+        private const int AsyncOperationPollingIntervalSeconds = 5;
+
+        private readonly APIClient apiClient;
+
+        public WebServicesClient(AzureContext context)
+        {
+            this.apiClient = AzureSession.ClientFactory.CreateArmClient<APIClient>(context, AzureEnvironment.Endpoint.ResourceManager);
```
markcowl: line breaks

src/ResourceManager/MachineLearning/Commands.MachineLearning/Utilities/WebServicesClient.cs
```diff
@@ -0,0 +1,142 @@
...
+            this.apiClient.LongRunningOperationRetryTimeout = WebServicesClient.AsyncOperationPollingIntervalSeconds;
```
markcowl: You should not reset the timeout, unless you set it to zero when runniung in playback mode - otherwise your playback tests will wait for no reason.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Properties/AssemblyInfo.cs
```diff
@@ -29,6 +29,8 @@
 [assembly: AssemblyFileVersion("1.1.0")]
```
markcowl: Do not include this assembly, or make changes to it in this PR.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ExportWebServiceDefinition.cs
```diff
@@ -0,0 +1,80 @@
...
+        WebServicesCmdletBase.CommandletSuffix, 
+        SupportsShouldProcess = true)]
+    [OutputType(typeof(string))]
+    public class ExportWebServiceDefinition : AzureRMCmdlet
+    {
+        private const string ExportToFileParamSet = "Export to file.";
+        private const string ExportToStringParamSet = "Export to JSON string.";
+
+        [Parameter(Mandatory = true, HelpMessage = "The web service definition object to export.")]
+        [ValidateNotNullOrEmpty]
```
markcowl: Repeat of original PR comment.  The recommendation is not to expose types from the management client in the signature of your cmdlets.  This is for two reasons:
1. It shields users from breaking changes in the management client types
2. It allows you to use types that are more natural to PowerShell.

This is not a requirement, but it is a recommendation.

markcowl:
@NonStatic2014 Left several comments.  The laregst ones are abotu types (recommendation) and the design of the Update cmdlet.  It's important that you have scenarios in mind for how the customer will use this cmdlet in scripts.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
NonStatic2014: Similar to the "RealtimeConfiguration"

NonStatic2014:
@markcowl Thank you Mark! I will address your comments and update the PR soon. 

Basically, it is not easy to not expose WebService in output type. Actually we expect user update the WebService instance in memory and then call update command to update the web service. The backward compatibility is the thing we will keep in mind and don't break user scenarios after the future update.

For the update command, I will keep what we have now to let user could update only the part he/she wants to update. Otherwise, we have to expose the whole data structure and the user have to create a whole one.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ExportWebServiceDefinition.cs
NonStatic2014: Thanks for the suggestion! I have answered in below comments.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ExportWebServiceDefinition.cs
NonStatic2014: I think the AzurePSCmdlet.ConfirmAction() in the next couple lines has taken care the ShouldProcess(). I don't think I need to do the same thing twice in my part of code.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/WebServicesCmdletBase.cs
NonStatic2014: Here we try to get the subscription Id and store it: 1) don't need to do it again anywhere we use it; 2) try to handle exceptions with more friendly error message.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Commands.MachineLearning.csproj
NonStatic2014: Seems other projects are doing the same thing here. Could we do refactoring in the later check-ins?

src/ResourceManager/MachineLearning/Commands.MachineLearning/Commands.MachineLearning.csproj
NonStatic2014: Same feedback as above.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Utilities/WebServicesClient.cs
NonStatic2014: May I know how could I set the interval only for playback testing? Since it is a retry timeout, I think it is fine for playback test because there will no need to retry in the local playback.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
NonStatic2014: I would expect the "ConfirmAction()" takes care the "Force" logic. I think it is ok if the user want to force an update in a script.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/WebServicesCmdletBase.cs
NonStatic2014: Will consider this in an update of the cmdlet.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/NewAzureMLWebService.cs
NonStatic2014: Will use ConfirmAction()

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/RemoveAzureMLWebService.cs
NonStatic2014: I think we probably still need a -Force to make sure the user does not delete a web service by accident no matter it is in-use or not. So I think the consumption of ConfirmAction() is the right thing. In case the behavior needs to be changed, someone just need to update the ConfirmAction() and it is fixed.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
NonStatic2014: We currently only expect whole service update could use pipeline:

        [Parameter(
            ParameterSetName = UpdateAzureMLWebService.UpdateFromObjectParameterSet, 
            Mandatory = true, 
            HelpMessage = "An updated definition object to update the referenced web service with.", 
            **ValueFromPipeline = true**)]
        public WebService ServiceUpdates { get; set; }

src/ResourceManager/MachineLearning/Commands.MachineLearning.Test/TestData/GraphWebServiceDefinition_Prod.json
markcowl: It is not allowed to specify real credentials used for a real service, even if this service is only created for a test and changed afterward.  Please find an alternate way to specify or generate this credential

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ExportWebServiceDefinition.cs
markcowl: So then it needs to be attributed to allow pipeline input.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
NonStatic2014: ServiceUpdates and other parameters belong to different ParameterSet. So it is not possible to have them together.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/GetAzureMLWebService.cs
markcowl: Got it, missed the if after the function definitions.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ImportWebServiceDefinition.cs
markcowl: Yes.  The main point is to support standard mechanisms for learning about cmdlets (-WhatIf and -Confirm).  Since this makes changes in Azure, it is appropriate

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/NewAzureMLWebService.cs
markcowl: Do not use ConfirmAction, please

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/RemoveAzureMLWebService.cs
markcowl: It is not. ConfirmAction implements this incorrectly and is being changed. Please implement as I have specified.  There is no need for -Force if it is *always* applied.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
markcowl: This is not a good experience for updating a WebService.  You force the user to create several .Net types without providing a good way to get them from the pipeline or construct them through cmdlets.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
markcowl: Same comment as above.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
markcowl: That is the point.  Is it a better experience to be able to pipe an existing object and change some parameters in the cmdlet through changoign the parameter sets?  Right now the user doesn't even know which properties (other than name and resource group) are required.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/ExportWebServiceDefinition.cs
markcowl: It does not.  Please follow the explicit advice in the review comments.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/WebServicesCmdletBase.cs
markcowl: Where is it used?  The subscription id is aready injected in the client when you create it, and is always accessbile from the context, if needed (cmdlets should not need it).

src/ResourceManager/MachineLearning/Commands.MachineLearning/Commands.MachineLearning.csproj
markcowl: No, we are no longer accepting this in new projects.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Commands.MachineLearning.csproj
markcowl: Same response.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Utilities/WebServicesClient.cs
markcowl: It is the polling interval.  You can use TestSupport.RunningMocked to determine if the cmdlet is running in the context of a mocked test.

markcowl:
@NonStatic2014  I think the Set cmdlet ncould use simplification - as it is, it has so many settings that it is difficult to use. This is soemthign to clean up in a future revision with additional parameter sets.  In the future, having an early design review on new cmdlets will help to make sure cmdlets meet standards and are usable in powershell scripts.

You need to remove credentials from the tests, update the ShouldProcess implementation as specified, and make the other noted minor corrections.

markcowl:
@NonStatic2014 Today is code complete.  Please update the PR as specificed above.  Otherwise, this will have to be targeted for the next release.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/UpdateAzureMLWebService.cs
NonStatic2014: We may need to revisit the Update- command in the future release to make it more friendly.

NonStatic2014:
@markcowl I think I have addressed all comments you mentioned above. Please take a look the latest one. Thanks!

NonStatic2014:
Seems there is build issue in the lab. I'm on it.

NonStatic2014:
on demand run passed: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/864/

PR #2291, opened 2016-05-19
Fix for application tests

brnleehng:
Added absolute location
Avoiding the storage call since SAS token expires for the blob upload


markcowl:
on demand run here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/796/

PR #2293, opened 2016-05-20
Commiting DevTestLabs powershell cmdlets

iouri-s:


azurecla:
Hi __@iouri-s__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (iouris). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

markcowl:

src/ResourceManager/DevTestLabs/Commands.DevTestLabs.Test/ScenarioTests/DevTestLabsTestFixture.ps1
```diff
@@ -0,0 +1 @@
+﻿
```
markcowl: Please remove

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands.DevTestLabs.csproj
```diff
@@ -0,0 +1,194 @@
...
+      <Project>{5ee72c53-1720-4309-b54b-5fb79703195f}</Project>
+      <Name>Commands.Common</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\..\Common\Commands.ResourceManager.Common\Commands.ResourceManager.Common.csproj">
+      <Project>{3819d8a7-c62c-4c47-8ddd-0332d9ce1252}</Project>
+      <Name>Commands.ResourceManager.Common</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\..\Resources\Commands.Resources\Commands.Resources.csproj">
+      <Project>{e1f5201d-6067-430e-b303-4e367652991b}</Project>
+      <Name>Commands.Resources</Name>
```
markcowl: Why do you need a reference to the resource manager cmdlets?  This kind of cross-dependency, unfortunately makes it impossible to ship the cmdlets independently,  If there is some helper function in this assembly that you are using, we should work on getting it into Commands.ResourceManager.Common

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands.DevTestLabs.csproj
```diff
@@ -0,0 +1,194 @@
...
+    </ProjectReference>
+    <ProjectReference Include="..\..\Tags\Commands.Tags\Commands.Tags.csproj">
+      <Project>{2493a8f7-1949-4f29-8d53-9d459046c3b8}</Project>
+      <Name>Commands.Tags</Name>
```
markcowl: Same comment

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/DevTestLabsCmdletBase.cs
```diff
@@ -0,0 +1,71 @@
...
+        #endregion Input Parameter Definitions
+
+        internal IDevTestLabsClient DataServiceClient
+        {
+            get
+            {
+                if (_dataServiceClient == null)
+                {
+                    this._dataServiceClient = AzureSession.ClientFactory.CreateArmClient<DevTestLabsClient>(DefaultContext, AzureEnvironment.Endpoint.ResourceManager);
+                    this._dataServiceClient.SubscriptionId = DefaultContext.Subscription.Id.ToString();
```
markcowl: This should not be necessary, this is already done in the factory if a subscription is present in the context

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/DtlPolicyCmdletBase.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        /// <summary>
+        /// Enable.
+        /// </summary>
+        [Parameter(Mandatory = false,
+            Position = 2,
+            ParameterSetName = ParameterSetEnable,
+            HelpMessage = "Whether to enable the policy.")]
+        [ValidateNotNullOrEmpty]
+        public SwitchParameter Enable
```
markcowl: Is it necessary to have explicit parameters for enable and disable?  Can enable be implied from other parameter sets?  I wonder if this creates confusing parameter set issues, since disable usaually only needs parameters for identification, while Enable requires additional parameters.  As it is, this definition of the parameter set in a base cmdlet prevents combining these parameters with any parameter set other than the default.

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Get/GetAzureRmDtlAutoStartPolicy.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.Azure.Management.DevTestLabs.Models;
+using System;
+using System.Linq;
+using System.Management.Automation;
+
+namespace Microsoft.Azure.Commands.DevTestLabs
+{
+    [Cmdlet(VerbsCommon.Get, "AzureRmDtlAutoStartPolicy", HelpUri = Constants.DevTestLabsHelpUri)]
+    [OutputType(typeof(Schedule))]
+    public class GetAzureRmDtlAutoStartPolicy : DevTestLabsCmdletBase
```
markcowl: Not having thought about this deeply, I wonder if this is the best use of the cmdlets?  Would it be better to have a single Get-AzureRMDtlPolicy cmdloet and allow the user to specify the required policies as a list?

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Get/GetAzureRmDtlAutoStartPolicy.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.Azure.Management.DevTestLabs;
+using Microsoft.Azure.Management.DevTestLabs.Models;
+using System;
+using System.Linq;
+using System.Management.Automation;
+
+namespace Microsoft.Azure.Commands.DevTestLabs
+{
+    [Cmdlet(VerbsCommon.Get, "AzureRmDtlAutoStartPolicy", HelpUri = Constants.DevTestLabsHelpUri)]
+    [OutputType(typeof(Schedule))]
```
markcowl: Make sure that output types do not expose types from the management library - this allows cmdlets to use types that are conducive to PowerShell scripting and shield users from changes in the underlying management types.

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Get/GetAzureRmDtlVMsPerLabPolicy.cs
```diff
@@ -0,0 +1,35 @@
...
+// ----------------------------------------------------------------------------------
+
+using Microsoft.Azure.Management.DevTestLabs;
+using System;
+using System.Management.Automation;
+
+namespace Microsoft.Azure.Commands.DevTestLabs
+{
+    [Cmdlet(VerbsCommon.Get, "AzureRmDtlVMsPerLabPolicy", HelpUri = Constants.DevTestLabsHelpUri)]
+    public class GetAzureRmDtlVMsPerLabPolicy : DevTestLabsCmdletBase
```
markcowl: OutputType

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Get/GetAzureRmDtlVMsPerUserPolicy.cs
```diff
@@ -0,0 +1,35 @@
...
+// ----------------------------------------------------------------------------------
+
+using Microsoft.Azure.Management.DevTestLabs;
+using System;
+using System.Management.Automation;
+
+namespace Microsoft.Azure.Commands.DevTestLabs
+{
+    [Cmdlet(VerbsCommon.Get, "AzureRmDtlVMsPerUserPolicy", HelpUri = Constants.DevTestLabsHelpUri)]
+    public class GetAzureRmDtlVMsPerUserPolicy : DevTestLabsCmdletBase
```
markcowl: Output Type

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Set/SetAzureRmDtlVMsPerUserPolicy.cs
```diff
@@ -0,0 +1,57 @@
...
+
+using Microsoft.Azure.Management.DevTestLabs;
+using Microsoft.Azure.Management.DevTestLabs.Models;
+using System;
+using System.Management.Automation;
+
+namespace Microsoft.Azure.Commands.DevTestLabs
+{
+    [Cmdlet(VerbsCommon.Set, "AzureRmDtlVMsPerUserPolicy", HelpUri = Constants.DevTestLabsHelpUri, DefaultParameterSetName = ParameterSetEnable)]
+    public class SetAzureRmDtlVMsPerUserPolicy : DtlPolicyCmdletBase
```
markcowl: Output Type

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Set/SetAzureRmDtlVMsPerLabPolicy.cs
```diff
@@ -0,0 +1,57 @@
...
+
+using Microsoft.Azure.Management.DevTestLabs;
+using Microsoft.Azure.Management.DevTestLabs.Models;
+using System;
+using System.Management.Automation;
+
+namespace Microsoft.Azure.Commands.DevTestLabs
+{
+    [Cmdlet(VerbsCommon.Set, "AzureRmDtlVMsPerLabPolicy", HelpUri = Constants.DevTestLabsHelpUri, DefaultParameterSetName = ParameterSetEnable)]
+    public class SetAzureRmDtlVMsPerLabPolicy : DtlPolicyCmdletBase
```
markcowl: OutputType

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Set/SetAzureRmDtlVMsPerLabPolicy.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        #region Input Parameter Definitions
+
+        /// <summary>
+        /// Maximum number of virtual machines.
+        /// </summary>
+        [Parameter(Mandatory = true,
```
markcowl: Does it make sense to have a single set cmdlets and set the scope of maxVMs to peruser or perlab?

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Get/GetAzureRmDtlAutoShutdownPolicy.cs
```diff
@@ -0,0 +1,37 @@
...
+    public class GetAzureRmDtlAutoShutdownPolicy : DevTestLabsCmdletBase
+    {
+        public override void ExecuteCmdlet()
+        {
+            var schedule = DataServiceClient.Schedule.GetResource(
+                ResourceGroupName,
+                LabName,
+                WellKnownPolicyNames.LabVmsShutdown);
+
+            WriteObject(schedule.DuckType<ScheduleDisplay>());
```
markcowl: This seems like an inefficient (in terms of processor cycles) means of doing type conversion.  Is the idea to convert into a display type?  Why not write a first class type conversion for this?

Also, your OutputType should reflect the actual return type

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Get/GetAzureRmDtlAutoStartPolicy.cs
```diff
@@ -0,0 +1,38 @@
...
+    public class GetAzureRmDtlAutoStartPolicy : DevTestLabsCmdletBase
+    {
+        public override void ExecuteCmdlet()
+        {
+            var schedule = DataServiceClient.Schedule.GetResource(
+                ResourceGroupName,
+                LabName,
+                WellKnownPolicyNames.LabVmsAutoStart);
+
+            WriteObject(schedule.DuckType<ScheduleDisplay>());
```
markcowl: See previous comment on using a serialization / deserialization roundtrip here, and on appropriate attributing for output type

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Models/ScheduleDisplay.cs
```diff
@@ -0,0 +1,107 @@
+﻿using Microsoft.Azure.Management.DevTestLabs.Models;
+using System;
+using System.Linq;
+
+namespace Microsoft.Azure.Commands.DevTestLabs.Models
+{
+    internal class ScheduleDisplay : Schedule
```
markcowl: An output type should not be internal

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Models/ScheduleDisplay.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            if (hours > 23 || minutes > 59)
+            {
+                return militaryTime;
+            }
+
+            return new DateTime(2000, 1, 1, hours, minutes, 0, 0).ToShortTimeString();
+        }
+
+        public new WeekDetailsDisplay WeeklyRecurrence
```
markcowl: Output types should use proeprties with easily usable types in PowerShell, e.g., string, int, array, DateTime

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using System;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("Microsoft Azure Powershell - Key Vault")]
```
markcowl: real assembly title

markcowl:
@Iouri-s please respond to review comments

src/ResourceManager/DevTestLabs/Commands.DevTestLabs.Test/ScenarioTests/DevTestLabsTestFixture.ps1
iouri-s: This file is required because of DevTestLabsTestFixture.cs. I preferred put the two scripts in common.ps1 that's also loaded. That's why the file is empty.

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands.DevTestLabs.csproj
iouri-s: Removed.

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands.DevTestLabs.csproj
iouri-s: Is this about Commands.Tags - removed. 

We only had these references because I started with a copy of the KeyVault project. They have all these references.

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/DevTestLabsCmdletBase.cs
iouri-s: removed

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/DtlPolicyCmdletBase.cs
iouri-s: Good point. The intent was that both Enable and Disable can be supplied with identification alone (labName  + resource group) or with other parameters, but never both at the same time. I made some changes that I will publish shortly.

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Get/GetAzureRmDtlAutoStartPolicy.cs
iouri-s: This is not the right place and time to discuss requirements for the user story. If you feel strongly about it, we can schedule a meeting with the PO (product owner) to change requirements.

markcowl:
on demand run here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/823/

src/ResourceManager/DevTestLabs/Commands.DevTestLabs.Test/ScenarioTests/Common.ps1
```diff
@@ -0,0 +1,92 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ----------------------------------------------------------------------------------
+
+<#
+.SYNOPSIS
+Gets the location for the Lab. Default to West US if none found.
+#>
+function Get-Location
+{
```
markcowl: nit: line breaks.  You can continue lines in powershell using the backtick

src/ResourceManager/DevTestLabs/Commands.DevTestLabs.Test/ScenarioTests/DevTestLabsTestFixture.cs
```diff
@@ -0,0 +1,88 @@
...
+
+            _synchronizationContext = SynchronizationContext.Current;
+
+            _controller = DevTestLabsController.NewInstance;
+
+            var callingClassType = TestUtilities.GetCallingClass(1);
+            var mockName = TestUtilities.GetCurrentMethodName(1);
+
+            _controller.RunPsTestWorkflow(
+                () => new[] { "Setup-Test-ResourceGroup " + _resourceGroupName + " " + _labName },
```
markcowl: Why not use the common infrastructure for creating resource group names?

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/DevTestLabsCmdletBase.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        #endregion Input Parameter Definitions
+
+        internal IDevTestLabsClient DataServiceClient
+        {
+            get
+            {
+                if (_dataServiceClient == null)
+                {
+                    this._dataServiceClient = AzureSession.ClientFactory.CreateArmClient<DevTestLabsClient>(DefaultContext, AzureEnvironment.Endpoint.ResourceManager);
```
markcowl: nit: line breaks

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Set/SetAzureRmDtlAllowedVMSizesPolicy.cs
```diff
@@ -0,0 +1,122 @@
...
+                    FactName = PolicyFactName.LabVmSize,
+                    Threshold = JsonConvert.SerializeObject(VmSizes),
+                    EvaluatorType = PolicyEvaluatorType.AllowedValuesPolicy,
+                    Status = Disable ? PolicyStatus.Disabled : PolicyStatus.Enabled
+                };
+            }
+            else
+            {
+                // Do nothing if user cancelled the operation
+                var actionDescription = string.Format(CultureInfo.CurrentCulture, Resources.SavePolicyDescription, PolicyName, LabName);
```
markcowl: nit: line breaks

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Set/SetAzureRmDtlAutoShutdownPolicy.cs
```diff
@@ -0,0 +1,124 @@
...
+using Microsoft.Azure.Management.DevTestLabs;
+using Microsoft.Azure.Management.DevTestLabs.Models;
+using Microsoft.Rest.Azure;
+using System;
+using System.Globalization;
+using System.Management.Automation;
+
+namespace Microsoft.Azure.Commands.DevTestLabs
+{
+    [Cmdlet(VerbsCommon.Set, "AzureRmDtlAutoShutdownPolicy", HelpUri = Constants.DevTestLabsHelpUri, DefaultParameterSetName = ParameterSetEnable, SupportsShouldProcess = true)]
```
markcowl: nit: line breaks

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/DevTestLabsCmdletBase.cs
```diff
@@ -0,0 +1,70 @@
...
+    public class DevTestLabsCmdletBase : AzureRMCmdlet
+    {
+        internal IDevTestLabsClient _dataServiceClient;
+
+        #region Input Parameter Definitions
+
+        /// <summary>
+        /// Lab name
+        /// </summary>
+        [Parameter(Mandatory = true,
```
markcowl: This makes these parameters mandatory across all parameter sets.  This will prevent, for example, listing across resource groups

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/DtlPolicyCmdletBase.cs
markcowl: @iouri-s are the changes to enable/disable in the PR?

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Get/GetAzureRmDtlAutoStartPolicy.cs
markcowl: @iour-s This is more about UI than requirements, but will defer to you that this meets your users needs in powershell. It is very different from other cmdlets, however.

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands.DevTestLabs.csproj
```diff
@@ -0,0 +1,189 @@
...
+      <HintPath>..\..\..\packages\Microsoft.Net.Http.2.2.28\lib\net45\System.Net.Http.Primitives.dll</HintPath>
+    </Reference>
+    <Reference Include="System.Net.Http.WebRequest" />
+    <Reference Include="System.Runtime.Serialization" />
+    <Reference Include="System.Security" />
+    <Reference Include="System.XML" />
+    <Reference Include="System.Xml.Linq" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\Common\Commands.Common.Authentication\Commands.Common.Authentication.csproj">
```
markcowl: @iour-s thanks for removing the extra cmdlet project references.  This is a task we have planned for existing projects that violated the rule, but it's a big help to not increase the scope of the problem.

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/DevTestLabsCmdletBase.cs
iouri-s: Listing across resource groups is not a supported operation for nested resources.

src/ResourceManager/DevTestLabs/Commands.DevTestLabs.Test/ScenarioTests/DevTestLabsTestFixture.cs
iouri-s: I am not sure what you mean. We are using Microsoft.Azure.Test.HttpRecorder.HttpMockServer.GetAssetName("DevTestLabsTests", "onesdk"); 

Isn't that common infrastructure? We wanted to make sue that the costly setup only happens once (hence the use of a feature) and that the fixture gets cleaned up. This is the best I could come up with.

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Set/SetAzureRmDtlAllowedVMSizesPolicy.cs
```diff
@@ -0,0 +1,122 @@
...
+                var actionWarning = string.Format(CultureInfo.CurrentCulture, Resources.SavePolicyWarning, PolicyName);
+                if (!ShouldProcess(
```
markcowl: This is in the wrong place, it needs to protect the actual CreateOrUpdate call, regardless of whether input policy is null

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Set/SetAzureRmDtlAutoShutdownPolicy.cs
```diff
@@ -0,0 +1,124 @@
...
+                        Time = Time.Value.ToString("HHmm")
+                    },
+                    Status = Disable ? PolicyStatus.Disabled : PolicyStatus.Enabled
+                };
+            }
+            else
+            {
+                // Do nothing if user cancelled the operation
+                var actionDescription = string.Format(CultureInfo.CurrentCulture, Resources.SavePolicyDescription, PolicyName, LabName);
+                var actionWarning = string.Format(CultureInfo.CurrentCulture, Resources.SavePolicyWarning, PolicyName);
```
markcowl: Same comment

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Set/SetAzureRmDtlAutoStartPolicy.cs
```diff
@@ -0,0 +1,137 @@
...
+                    },
+                    Status = Disable ? PolicyStatus.Disabled : PolicyStatus.Enabled
+                };
+            }
+            else
+            {
+                // Do nothing if user cancelled the operation
+                var actionDescription = string.Format(CultureInfo.CurrentCulture, Resources.SavePolicyDescription, PolicyName, LabName);
+                var actionWarning = string.Format(CultureInfo.CurrentCulture, Resources.SavePolicyWarning, PolicyName);
+                if (!ShouldProcess(
```
markcowl: same comment

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Set/SetAzureRmDtlVMsPerLabPolicy.cs
```diff
@@ -0,0 +1,121 @@
...
+                    Threshold = MaxVMs.Value.ToString(),
+                    Status = Disable ? PolicyStatus.Disabled : PolicyStatus.Enabled
+                };
+            }
+            else
+            {
+                // Do nothing if user cancelled the operation
+                var actionDescription = string.Format(CultureInfo.CurrentCulture, Resources.SavePolicyDescription, PolicyName, LabName);
+                var actionWarning = string.Format(CultureInfo.CurrentCulture, Resources.SavePolicyWarning, PolicyName);
+                if (!ShouldProcess(
```
markcowl: same comment

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Commands/Set/SetAzureRmDtlVMsPerUserPolicy.cs
```diff
@@ -0,0 +1,121 @@
...
+                    EvaluatorType = PolicyEvaluatorType.MaxValuePolicy,
+                    Status = Disable ? PolicyStatus.Disabled : PolicyStatus.Enabled
+                };
+            }
+            else
+            {
+                // Do nothing if user cancelled the operation
+                var actionDescription = string.Format(CultureInfo.CurrentCulture, Resources.SavePolicyDescription, PolicyName, LabName);
+                var actionWarning = string.Format(CultureInfo.CurrentCulture, Resources.SavePolicyWarning, PolicyName);
+                if (!ShouldProcess(
```
markcowl: same comment

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Extensions.cs
```diff
@@ -0,0 +1,19 @@
...
+    internal static class Extensions
+    {
+        internal static T DuckType<T>(this object target)
+        {
+            if (target == null)
+            {
+                return default(T);
+            }
+
+            return JsonConvert.DeserializeObject<T>(JsonConvert.SerializeObject(target));
```
markcowl: As per previous comment, this seems computationally inefficient.  Please use a proper type conversion.

src/ResourceManager/DevTestLabs/Commands.DevTestLabs/Models/PSPolicy.cs
```diff
@@ -0,0 +1,84 @@
...
+        // Summary:
+        //     The status of the policy. Possible values include: 'Enabled', 'Disabled'
+        [JsonProperty(PropertyName = "properties.status")]
+        public string Status { get; set; }
+
+        //
+        // Summary:
+        //     The tags of the resource.
+        [JsonProperty(PropertyName = "tags")]
+        public IDictionary<string, string> Tags { get; set; }
```
markcowl: Hashtable

tools/Test/Logon.ps1
```diff
@@ -0,0 +1,17 @@
+[CmdletBinding()]
```
markcowl: Please remove.

markcowl:
@iouri-s three main remaining issues:
1. Implementing SHouldProcess to protect the change action in all cases (marked inline)
2. Using type conversion instead of serialization / deserialization - it is more code, but faster and safer.  There are a small number of types here, so this should be relatively small.  Utilities like AutoMapper are available (and used in other modules) for similar tasks.
3. Looks like you were going to make a change to Enable/Disable.  I think what you want is to have two parameter sets, one that includes the identity parameters and Disable, one that includes all the other parameters.  I don't think you actually need enable, as this is implied by not providing -Disable.  This may mean moving these parameters out of the common class, which is unfortunate.  You may *also* be able to reuse the parameter set name for disable in both the common class and the inheriting classes, but I have not tried this.

markcowl:

markcowl:
discussed offline.  Will fix (1) and defer on (2), (3) is implemented correctly for DevTestLab scenarios

tools/Test/Logon.ps1
markcowl: @iouri-s don't see this change, please update.

markcowl:
on demand run here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/844/

tools/Test/Logon.ps1
iouri-s: Deleted the file.

markcowl:
@azuresdkci retset this please

PR #2305, opened 2016-05-23
Added import/export powershell documentation xml

brnleehng:
The GUID is the same as the Azure Batch module. Can both of these modules have the same GUID?


brnleehng:
On-Demand Job: http://azuresdkci.cloudapp.net/job/powershell-demand/820/

PR #2315, opened 2016-05-24
Fix wxi file generation. Some projects indirectly use Microsoft.Web.A…

iouri-s:
…dministration that should not be copied to OUTDIR because it is in the GAC. However, in some cases (e.g. Azure SDK is installed), it will be found outside of the GAC and copied. Explicitly stating that it is <Private>False</Private>

### DIFF
--- a/src/ResourceManager/Compute/Commands.Compute/Commands.Compute.csproj
+++ b/src/ResourceManager/Compute/Commands.Compute/Commands.Compute.csproj
@@ -48,6 +48,9 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Microsoft.Web.Administration, Version=7.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35">
+      <Private>False</Private>
+    </Reference>
     <Reference Include="AutoMapper, Version=3.1.1.0, Culture=neutral, PublicKeyToken=be96cd2c38ef1005, processorArchitecture=MSIL">
       <SpecificVersion>False</SpecificVersion>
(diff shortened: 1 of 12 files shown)


azurecla:
Hi __@iouri-s__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (iouris). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

src/ResourceManager/Compute/Commands.Compute/Commands.Compute.csproj
```diff
@@ -48,6 +48,9 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Microsoft.Web.Administration, Version=7.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35">
```
markcowl: This isn't safe.  Microsoft.Web.Administration is platform-specific.  The actual methods in this library are unused in the deployment code.  Closing this PR for now (thanks for looking at this, though)

iouri-s: Microsoft.Web.Administration, Version=7.9.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=**MSIL**.

Doesn't look platform-specific to me... 
This change is harmless - the only thing it'll do is prevent the assembly from being copied. If the assembly cannot be found, it will have no effect.

Also note the corrected reference to SMA in src/ServiceManagement/ExpressRoute/Commands.ExpressRoute/Commands.ExpressRoute.csproj.

PR #2339, opened 2016-05-26
Revert changes to GetAzureResourceGroupCmdlet

vivsriaus:
Revert changes to GetAzureResourceGroupCmdlet to ignore exception when trying to import ResourceManagerStartup.ps1 during tests


hovsepm:
this issue is addressed in #2341

PR #2362, opened 2016-05-29
Updating build process to use tags embedded in modules

markcowl:
- Added tags for Azure module


markcowl:
@azuresdkci add to whitelist

PR #2368, opened 2016-05-31
Fix regression caused by commit 8b26aa3d

krkhan:
This reverts commit 8b26aa3df3ce559f6ed21bf6e4e1fe738ef2a140.


markcowl:

src/ResourceManager/Compute/Commands.Compute.Test/ScenarioTests/VirtualMachineExtensionTests.cs
```diff
@@ -66,8 +66,7 @@ public void TestVirtualMachineAccessExtension()
             ComputeTestController.NewInstance.RunPsTest("Test-VirtualMachineAccessExtension");
         }
 
-        [Fact(Skip = "TODO: only works for live mode")]
```
markcowl: You need to make this [Fact]

Issue #1961, opened 2016-03-17
IOTHub - Proper error message not shown when sku is missing while creating a new IotHub using New-AzureRmResource


Pradebban:
Hi,

Am trying to create IOTHub with the following command.
Please look into the same.

New-AzureRmResource -ResourceName "IotHubSample1" -ResourceType Microsoft.Devices/Iothubs -ResourceGroupName Poc -Location "East US" -Properties @{
Location="East US"}

Confirm
Are you sure you want to create the following resource:
/subscriptions/xxxxxxxxxxxx/resourceGroups/Poc/providers/Microsoft.Devices/Iothubs/IotHubSample1
[Y] Yes  [N] No  [S] Suspend  [?] Help (default is "Y"): y
**New-AzureRmResource : The pipeline has been stopped.**
At line:1 char:1
+ New-AzureRmResource -ResourceName "IotHubSample1" -ResourceType Microsoft.Device ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : CloseError: (:) [New-AzureRmResource], PipelineStoppedException
    + FullyQualifiedErrorId : Microsoft.Azure.Commands.ResourceManager.Cmdlets.Implementation.NewAzureResourceCmdlet

**New-AzureRmResource : A task was canceled.**
At line:1 char:1
+ New-AzureRmResource -ResourceName "IotHubSample1" -ResourceType Microsoft.Device ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : CloseError: (:) [New-AzureRmResource], TaskCanceledException
    + FullyQualifiedErrorId : A task was canceled.,Microsoft.Azure.Commands.ResourceManager.Cmdlets.Implementation.NewAzureResourceCmdlet

Thanks,
Pradebban Raja

ravbhatnagar:
@Pradebban - If I specify a Sku for the IOT hub, the hub gets created successfully.
[...]

PR #2412, opened 2016-06-06
Dispose response messages when we retry sending requests. 

TianoMS:
Fix issue: https://github.com/Azure/azure-powershell/issues/1961


TianoMS:
On demand job: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/886/

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Handlers/RetryHandler.cs
```diff
@@ -56,15 +56,15 @@ public class RetryHandler : DelegatingHandler
         protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
         {
             HttpResponseMessage response = null;
+
             for (int attempt = 0; attempt < RetryHandler.MaxAttempts; ++attempt)
```
vivsriaus: This is not very readable, IMHO. We should just do this:
for(int attempt = 1; attempt <= RetryHandler.MaxAttempts; ++attempt)
{
..
}

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Handlers/RetryHandler.cs
```diff
@@ -56,15 +56,15 @@ public class RetryHandler : DelegatingHandler
...
                 try
                 {
-                    response = await base
-                        .SendAsync(request: request, cancellationToken: cancellationToken)
+                    response = await base.SendAsync(request: request, cancellationToken: cancellationToken)
                         .ConfigureAwait(continueOnCapturedContext: false);
-
-                    if (attempt == RetryHandler.MaxAttempts ||
+                    
+                    if (attempt == RetryHandler.MaxAttempts - 1 ||
```
vivsriaus: Why do we need the attempt == RetryHandler.MaxAttempts -1 condition here? The loop condition will ensure attempt won't go above MaxAttempts (3).

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Handlers/RetryHandler.cs
```diff
@@ -74,17 +74,22 @@ public class RetryHandler : DelegatingHandler
                 }
                 catch (Exception ex)
                 {
-                    if (ex.IsFatal() || attempt == RetryHandler.MaxAttempts)
+                    if (ex.IsFatal() || attempt == RetryHandler.MaxAttempts - 1)
```
vivsriaus: Please see my comment above, regarding changing the for loop condition. If we keep it as for(int attempt = 1; attempt <= RetryHandler.MaxAttempts; ++attempt), this should just read attempt == RetryHandler.MaxAttempts

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Handlers/RetryHandler.cs
```diff
@@ -74,17 +74,22 @@ public class RetryHandler : DelegatingHandler
...
+                {
+                    response.Dispose();
+                }
+
                 await Task.Delay(delay: RetryHandler.GetDelay(attempt), cancellationToken: cancellationToken)
                     .ConfigureAwait(continueOnCapturedContext: false);
             }
 
-            return response;
+            return null;
```
vivsriaus: Why? Why not response?

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/RestClients/ResourceManagerRestClientBase.cs
```diff
@@ -219,27 +219,39 @@ public ResourceManagerRestClientBase(Components.HttpClientHelper httpClientHelpe
...
+                            errorMessage: message);
+                    }
+
+                    return response;
                 }
+                catch (Exception exception)
+                {
+                    if (exception is OperationCanceledException && !cancellationToken.IsCancellationRequested)
+                    {
+                        throw new TimeoutException("A request times out.");
```
vivsriaus: Can we throw a more detailed exception, instead of just "A request timed out" ? What was the request? What is the content of exception.Message?

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Handlers/RetryHandler.cs
TianoMS: Previously the condition was "attempt == RetryHandler.MaxAttempts", which would never be true. I think the point here is directly return the response after MaxAttempts of tries, and don't do the rest of the stuff.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Handlers/RetryHandler.cs
TianoMS: We should actually never reach here. The response would already have been returned from the loop after MaxAttempts tries.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/RestClients/ResourceManagerRestClientBase.cs
TianoMS: When we get here, the inner exception is "A task was canceled". Here we are trying to catch on "await base.SendAsync(...).ConfigureAwait(...)".

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Handlers/RetryHandler.cs
vivsriaus: The return is needed for this to successfully compile. I'd keep this as response (which is set to null either way), instead of returning null explicitly

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Handlers/RetryHandler.cs
vivsriaus: Per your logic, we will return response in the 3rd attempt, which is not what we want.

TianoMS: Hmm... I thought we want to attempt three times in total..

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Handlers/RetryHandler.cs
TianoMS: Fixed.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Handlers/RetryHandler.cs
TianoMS: This is fixed.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Handlers/RetryHandler.cs
TianoMS: Added a comment in code.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Handlers/RetryHandler.cs
gandhiniraj: How can attempt > RetryHandler.MaxAttempts in for (int attempt = 1; attempt <= RetryHandler.MaxAttempts; ++attempt) ? :)  

---
In reply to: [66539492](https://github.com/Azure/azure-powershell/pull/2412#discussion_r66539492) [](ancestors = 66539492)

gandhiniraj:
>                         return response;

For better readability and not explicitly return null value, how about this ?
- do break; here
- instead of return null just do return response;


---
Refers to: src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Handlers/RetryHandler.cs:72 in 73325f4. [](commit_id = 73325f4f64d08c6ab32b7fc3689052d05b7fc782, deletion_comment = False)

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Handlers/RetryHandler.cs
gandhiniraj: And combining with my response above if we want to throw if we have reached max attempt, initialize attempt outside of the for loop and if  attempt > RetryHandler.MaxAttempts then throw outside of the for loop

---
In reply to: [66560166](https://github.com/Azure/azure-powershell/pull/2412#discussion_r66560166) [](ancestors = 66560166,66539492)

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/RestClients/ResourceManagerRestClientBase.cs
```diff
@@ -219,27 +220,39 @@ public ResourceManagerRestClientBase(Components.HttpClientHelper httpClientHelpe
...
-                    .SendAsync(request: request, cancellationToken: cancellationToken)
-                    .ConfigureAwait(continueOnCapturedContext: false);
-
-                if (!response.StatusCode.IsSuccessfulRequest())
+                try
                 {
-                    var errorResponse = await ResourceManagerRestClientBase
-                        .TryReadErrorResponseMessage(response, rewindContentStream: true)
+                    var response = await httpClient
+                        .SendAsync(request: request, cancellationToken: cancellationToken)
```
gandhiniraj: nit:For better readability use: await httpClient.SendAsync(request: request, cancellationToken: cancellationToken) and next line .Configure...

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/RestClients/ResourceManagerRestClientBase.cs
```diff
@@ -219,27 +220,39 @@ public ResourceManagerRestClientBase(Components.HttpClientHelper httpClientHelpe
...
                         .ConfigureAwait(continueOnCapturedContext: false);
 
-                    var message = await ResourceManagerRestClientBase
-                        .GetErrorMessage(request: request, response: response, errorResponse: errorResponse)
-                        .ConfigureAwait(continueOnCapturedContext: false);
+                    if (!response.StatusCode.IsSuccessfulRequest())
+                    {
+                        var errorResponse = await ResourceManagerRestClientBase
+                            .TryReadErrorResponseMessage(response, rewindContentStream: true)
+                            .ConfigureAwait(continueOnCapturedContext: false);
```
gandhiniraj: nit: same as above

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Properties/Resources.resx
```diff
@@ -396,4 +396,7 @@
   <data name="UpdatedResourceGroup" xml:space="preserve">
     <value>Updated resource group '{0}' in location '{1}'</value>
   </data>
+  <data name="OperationFailedWithTimeOut" xml:space="preserve">
+    <value>Operation failed because a request time out.</value>
```
gandhiniraj: nit: the request timed out.

TianoMS: Checked wiki. It's a verb. "timed out" is correct.

gandhiniraj: I am not saying it's not a word - I am just saying grammatically - "Operation failed because the request timed out"

---
In reply to: [66566451](https://github.com/Azure/azure-powershell/pull/2412#discussion_r66566451) [](ancestors = 66566451)

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Handlers/RetryHandler.cs
TianoMS: attempt shouldn't be > MaxAttempts. my latest change uses attempt == MaxAttempts as exit criteria.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Properties/Resources.resx
TianoMS: Sorry for the confusion... I was actually agreeing with you :)  I'm changing it to "timed out".

gandhiniraj: There isn't any confusion. all good

---
In reply to: [66567017](https://github.com/Azure/azure-powershell/pull/2412#discussion_r66567017) [](ancestors = 66567017)

TianoMS:
>                         return response;

Will update it. 

---
In reply to: [225104925](https://github.com/Azure/azure-powershell/pull/2412#issuecomment-225104925) [](ancestors = 225104925,225090348)

---
Refers to: src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Handlers/RetryHandler.cs:72 in 73325f4. [](commit_id = 73325f4f64d08c6ab32b7fc3689052d05b7fc782, deletion_comment = False)

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/RestClients/ResourceManagerRestClientBase.cs
TianoMS: Checked server code. Our practice seems to be putting them at separate lines. Will follow the practice here.

TianoMS:
On demand job: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/893/

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Properties/Resources.Designer.cs
```diff
@@ -493,6 +493,15 @@ internal class Resources {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to Operation failed because a request timed out..
```
gandhiniraj: nit: operation

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/RestClients/ResourceManagerRestClientBase.cs
gandhiniraj: sounds good

---
In reply to: [66568621](https://github.com/Azure/azure-powershell/pull/2412#discussion_r66568621) [](ancestors = 66568621)

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Properties/Resources.Designer.cs
TianoMS: This is generated code. Looks it uses the original resource string.

gandhiniraj:
>         protected async Task<HttpResponseMessage> SendRequestAsync(HttpRequestMessage request,

we should add a test for this

---
Refers to: src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/RestClients/ResourceManagerRestClientBase.cs:218 in c591214. [](commit_id = c59121483281fa3d350ea08f380c9671fadd66aa, deletion_comment = False)

TianoMS:
>         protected async Task<HttpResponseMessage> SendRequestAsync(HttpRequestMessage request,

This is a corner case. It happens when an async task is stuck and eventually times out. When this happens we catch the exception and throw a more readable exception. I haven't figured out a good way to test this, but I'll try this later as separate efforts. 

---
In reply to: [226050060](https://github.com/Azure/azure-powershell/pull/2412#issuecomment-226050060) [](ancestors = 226050060)

---
Refers to: src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/RestClients/ResourceManagerRestClientBase.cs:218 in c591214. [](commit_id = c59121483281fa3d350ea08f380c9671fadd66aa, deletion_comment = False)

TianoMS:
@markcowl All comments were addressed except one on test coverage. I didn't figure out a good way to test the specific piece of fix yet. Regression is covered, though. I think I'll try to figure that out as separate tasks.

PR #2429, opened 2016-06-08
Batch test cleanup

jasper-schneider:
Autogenerate test certs per request from Kirthi and Mark.
Cleaned up some other tests related to certs and MPI as well.


jasper-schneider:
Chatted over email, will close this PR and create a new one later after reorganizing some of the changes.

PR #2435, opened 2016-06-10
ADL Updates: Defaults and Convenience

begoldsm:
This PR does the following:
1. Fixes default thread counts for file and folder upload
2. Updates Get-AzureRMDataLakeAnalyticsDataSource "list" operation to return all data sources as a new common data source object. This also indicates which data source is the default.
3. Update documentation and comments to reflect the above two changes.
4. Update tests


markcowl:
on demand run here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/898/

LGTM once the build passes, though I would really like to know where the need for the new client comes from.

begoldsm:
@markcowl I agree, it is quite confusing to me as well. I suspect it will go away once the "tags" stuff is moved into a more common location (maybe)?

PR #2442, opened 2016-06-13
[Do not Merge] Ported reference of DNX libraries to netcore (Clean version)

hovsepm:


hovsepm:
@azuresdkci add to whitelist

Issue #2416, opened 2016-06-07
New-AzureRmResourceGroupDeployment fail when the first interval check


LiliDeng:
Run below command, it will fail in a high rate when the first 5 seconds interval check, do we have any workaround to increase the interval or any other parameter can be used to avoid this issue?
Command:
New-AzureRmResourceGroupDeployment -Name $ResourceGroupDeploymentName -ResourceGroupName $RGName -TemplateFile $TemplateFile -TemplateParameterFile $TemplateParameterFile -Verbose
Failure:
VERBOSE: 7:16:38 PM - Checking deployment status in 5 seconds
New-AzureRmResourceGroupDeployment : Deployment 'ICA-RG-A1-liliarmde-6-6-19-16-22-deployment' could not be found.

markcowl:
@LiliDeng Can you try executing with ```$DebugPreference="Continue"``` to get debugging information for this issue.

@ravbhatnagar can you take a look?

ZACommando:
My team and I are experiencing the same issue.  We are running the latest version of the Azure PowerShell (1.5 released June 1st) and the three of us keep getting the failure at the same point: Checking deployment status in 5 seconds.

The deployments are completing successfully, but the remainder of our deployment script is dependent upon the New-AzureRmResourceGroupDeployment cmdlet succeeding.

Here is the cmdlet call:
```
New-AzureRmResourceGroupDeployment 
  -Name azuredeploy-0610-2206 
  -ResourceGroupName [ResourceGroupName] 
  -TemplateFile C:\azuredeploy.json
  -TemplateParameterFile C:\azuredeploy.parameters.json
  -Force
  -Verbose
  -Mode Incremental
```

Here is the output from the cmdlet with `$DebugPreference="Continue"` (shortened for brevity and security):
```
VERBOSE: 5:07:07 PM - Template is valid.
DEBUG: Microsoft.IdentityModel.Clients.ActiveDirectory Information: 2 : 
DEBUG: 6/10/2016 10:07:07 PM: bbacb506-eba4-4988-ad4a-e959f3723eed - AcquireTokenHandlerBase: === Token Acquisition started:
	Authority: https://login.microsoftonline.com/f2e3e4c3-e58f-4b8a-bde8-550f8708312a/
	Resource: https://management.core.windows.net/
	ClientId: 1950a258-227b-4e31-a9cf-717495945fc2
[...]

PR #2446, opened 2016-06-13
- Fix for ExecuteDeployment where deployment creation takes longer th…

ZACommando:
Issue #2416 Fix
- Changed CreateOrUpdateAsync to CreateOrUpdate in ExecuteDeployment to fix and issue where deployment creation takes longer than 5 seconds and ProvisionDeploymentStatus fails as a result
- Test to validate changes to ExecuteDeployment

### DIFF
--- a/src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/SdkClient/ResourceManagerSdkClient.cs
+++ b/src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/SdkClient/ResourceManagerSdkClient.cs
@@ -707,7 +707,7 @@ namespace Microsoft.Azure.Commands.ResourceManager.Cmdlets.SdkClient
                 WriteVerbose(ProjectResources.TemplateValid);
             }
 
-            ResourceManagementClient.Deployments.CreateOrUpdateAsync(parameters.ResourceGroupName, parameters.DeploymentName, deployment);
+            ResourceManagementClient.Deployments.CreateOrUpdate(parameters.ResourceGroupName, parameters.DeploymentName, deployment);
             WriteVerbose(string.Format(ProjectResources.CreatedDeployment, parameters.DeploymentName));
             DeploymentExtended result = ProvisionDeploymentStatus(parameters.ResourceGroupName, parameters.DeploymentName, deployment);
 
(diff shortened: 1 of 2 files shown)


azurecla:
Hi __@ZACommando__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
__@ZACommando__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

markcowl:
@azuresdkci add to whitelist

markcowl:
@vivsriaus can you take a look?

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/SdkClient/ResourceManagerSdkClient.cs
```diff
@@ -707,7 +707,7 @@ public virtual PSResourceGroupDeployment ExecuteDeployment(PSCreateResourceGroup
                 WriteVerbose(ProjectResources.TemplateValid);
             }
 
-            ResourceManagementClient.Deployments.CreateOrUpdateAsync(parameters.ResourceGroupName, parameters.DeploymentName, deployment);
+            ResourceManagementClient.Deployments.CreateOrUpdate(parameters.ResourceGroupName, parameters.DeploymentName, deployment);
```
vivsriaus: We have a pending PR to change this to BeginCreateOrUpdate instead, here: https://github.com/Azure/azure-powershell/pull/2451/files#diff-0e1fbafcda48a1d6c2fefd0fec655bfcR719

CreateOrUpdate is a synchronous call, which blocks the thread execution. This will mean you won't get the progress messages in the console when the deployment is running.

markcowl:
@vivsriaus Does the scenario for the contributed test make sense as an add?  Clearly the implementation would have to change since you're no longer using that call.

vivsriaus:
@markcowl Yes, I think there is some value to the included test in this PR. @TianoMS can we get this added to your [PR](https://github.com/Azure/azure-powershell/pull/2451)?

markcowl:
@ZACommando Closing as this is being handled separately by the service team - thanks for the contributed soluttion and the test, which they will likely use.

PR #2457, opened 2016-06-15
Added range for ASN number

aicaffre:
Currently we allow people to add an ASN that is out of the defined range for ASNs. 
If they then try and use the Gateway later to add 'Connections' with BGP enabled - they will create but never reach  the Connected state - with no visible errors to the user. 
Adding in this validation will prevent this scenario occurring and save effort on complicated support cases etc.

### DIFF
--- a/src/ResourceManager/Network/Commands.Network/VirtualNetworkGateway/NewAzureVirtualNetworkGatewayCommand.cs
+++ b/src/ResourceManager/Network/Commands.Network/VirtualNetworkGateway/NewAzureVirtualNetworkGatewayCommand.cs
@@ -122,7 +122,8 @@ namespace Microsoft.Azure.Commands.Network
         [Parameter(
             Mandatory = false,
             ValueFromPipelineByPropertyName = true,
-            HelpMessage = "The virtual network gateway's ASN for BGP over VPN")]
+            HelpMessage = "The virtual network gateway's ASN for BGP over VPN. Must be between 0 and 65535.")]
+            [ValidateRange(0,65535)]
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@aicaffre__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

markcowl:
@azuresdkci add to whitelist

markcowl:
@aicaffre Note that this is a good change in the code, but it will not impact the online help.  To change the online help, you would need to make this change here:  https://github.com/Azure/azure-powershell/blob/dev/src/ResourceManager/Network/Commands.Network/Microsoft.Azure.Commands.Network.dll-Help.xml

markcowl:
@deepakrajendranmsft Can you have someone take a look at this:  (1) Is the range restriction correct? (2) Can we add the text of the helpcomment to the parameter help in online helpf or this cmdlet?

markcowl:
@deepakrajendranmsft ping?

DeepakRajendranMsft:
@lamchester please take a look

lamchester:
What was the ASN that you tried to set? 32-bit ASNs should be supported - that's why a uint was used instead of a 16 bit data type. We should figure out why the connected state wasn't reached.

markcowl:
@aicaffre Closing this PR as it looks like this range is not required.  Please file an issue in https://github.com/azure/azure-powershell/issues to track the solution to your connection issues with the cmdlet.

PR #2469, opened 2016-06-17
Azure Site Recovery Bug Fixes and adding deprecation warning for RDFE cmdlets.

avneeshrai:


src/ResourceManager/SiteRecovery/Commands.SiteRecovery/ProtectionEntity/StartAzureSiteRecoveryPlannedFailoverJob.cs
```diff
@@ -83,10 +83,23 @@ public class StartAzureSiteRecoveryPlannedFailoverJob : SiteRecoveryCmdletBase
...
-        [ValidateSet(Constants.ForDowntime, Constants.ForSynchronization)]
+        [ValidateSet(Constants.ForDownTime, Constants.ForSynchronization)]
         public string Optimize { get; set; }
 
         /// <summary>
+        /// Gets or sets the recovery vm creation value.
+        /// </summary>
+        [Parameter]
+        [ValidateSet(Constants.Yes, Constants.No)]
+        public string CreateVmIfNotFound { get; set; }
```
markcowl: Shouldn't this be a switch parameter?

src/ResourceManager/SiteRecovery/Commands.SiteRecovery/ProtectionEntity/StartAzureSiteRecoveryPlannedFailoverJob.cs
```diff
@@ -83,10 +83,23 @@ public class StartAzureSiteRecoveryPlannedFailoverJob : SiteRecoveryCmdletBase
...
         /// </summary>
         [Parameter]
-        [ValidateSet(Constants.ForDowntime, Constants.ForSynchronization)]
+        [ValidateSet(Constants.ForDownTime, Constants.ForSynchronization)]
         public string Optimize { get; set; }
 
         /// <summary>
+        /// Gets or sets the recovery vm creation value.
+        /// </summary>
+        [Parameter]
```
markcowl: Need more detail on these parameters.  Do they take pipeline values?  Are they required or optional?

src/ResourceManager/SiteRecovery/Commands.SiteRecovery/Vault/SetAzureSiteRecoveryVaultSettings.cs
```diff
@@ -29,6 +29,13 @@ public class SetAzureSiteRecoveryVaultSettings : SiteRecoveryCmdletBase
         #region Parameters
 
         /// <summary>
+        /// Gets or sets vault Object.
+        /// </summary>
+        [Parameter(ParameterSetName = ASRParameterSets.Default, Mandatory = true)]
+        [ValidateNotNullOrEmpty]
+        public System.Object Vault { get; set; }
```
markcowl: Object?  How is this created by the user - how are they going to get this value to pass to the cmdlet?

markcowl:
@avneeshrai A few comments, please address.

src/ResourceManager/SiteRecovery/Commands.SiteRecovery/ProtectionEntity/StartAzureSiteRecoveryPlannedFailoverJob.cs
avneeshrai: This is a PM call. They want it as an explicit Yes or No from user.

src/ResourceManager/SiteRecovery/Commands.SiteRecovery/Vault/SetAzureSiteRecoveryVaultSettings.cs
avneeshrai: We have two different type of Vaults (ASRVault and ARSVault) out of which ASRVault will get deprecated some time in future.
Released cmdlet has 2 different parameter sets (one for each Vault type).
We want to get rid of different vault types and parameter names for this cmdlet and so added warnings for the existing 2 types and introduced a new common type with Object as parameter type for now - so that the same parameter name / set can be used for both the types till we deprecate the older one.
And once we deprecate the older Vault type, we will move the parameter type from Object to newly supported Vault type. In this case, customers interface doesn't change from cmdlet perspective. Moreover, our PMs will educate customers via documentation on the usage of new parameter set.
As for Object type, they can pass any value - but, in our code we are handling this to be either one of our known Vault types.Customers can pass objects returned by either Get-AzureRmSiteRecoveryVault or Get-AzureRmRecoveryServicesVault.

src/ResourceManager/SiteRecovery/Commands.SiteRecovery/ProtectionEntity/StartAzureSiteRecoveryPlannedFailoverJob.cs
avneeshrai: These are non-mandatory and they don't take value from pipeline. We have not added anything as by default these values are false.

src/ResourceManager/SiteRecovery/Commands.SiteRecovery/Vault/SetAzureSiteRecoveryVaultSettings.cs
markcowl: The problem is, how will the user know how to create this type - there is no indication of what properties it should have and its value cannot come from the pipeline, because you have attributes it to exclude this.  Please change this parameter type to be a strongly typed parameter.  In the worst case, you can have two parametsr sets, one that accepts ASRVault and the other that acepts ARSVault.  As it is, this cmdlet will be very difficult to use.

bsiva: @markcowl  The problem is as follows. We've got two different resource types - "Site Recovery vault" and "Recovery services vault" . Look at Site Recovery vault as the ARM representation of our Classic vault and Recovery Services vault as our newer and better vault. The same set of PowerShell cmdlets work on both theses resources as essentially all exposed operations within them are the same. Our customer's don't really differentiate between the two resource types when it comes to the actions they perform in them, and to them both or either of them are just vaults. The particular cmdlet in question is used to set the session context of the recovery services vault or the site recovery vault within which all further operations are to be performed. Once this is done the actions are pretty much the same. 

We want to abstract out this distinction for our customers when they set the vault context.So we chose the vault parameter as a way to do this. With the objects for the two resource types being different we need a higher level abstratction or class from which we inherit to serve as the object type for this parameter - hence the choice of System Object.

As for how a customer creates this object, the object definition isnt something that they necessarily concern themselves with since this is never something they would have to create or manipulate by accessing the object. The vault is created using one of these cmdlets:-
- New-AzureRmSiteRecoveryVault or -New-AzureRmRecoveryServicesVault and can be retrieved later using the corresponding Get cmdlet. The returned object will be passed unmodified as an input to the cmdlet in question against the -Vault parameter and will get cast as a System Object

The ASR PowerShell user is typically an advanced customer who is familiar with the our UX and uses PowerShell to automate repeated tasks and is therefore unlikely to be confused at this step. Further, this will be more than adequately covered in our documentation

markcowl:
@bsiva This does not explain why you can't have two parameter sets, one for each vault type, which would be more discoverable, and which would allow you to accept input from the pipeline.  PowerShell users use PowerShell reflection tools to discover cmdlet usage, and this design prevents them from doing this. 

Covering over issues with the cmdlet signature with documentation is not an acceptable alternative - you should either have a common type that both of these types can be converted from, or you should have two parameter sets.

bsiva:
@markcowl : We already have separate parameter sets in this cmdlet to support both vault types. Additionally there a separate get cmdlets for each of the resource type, and the output objects for these get methods will return only either of the ASRvault type or the ARSvault type and these returned objects can be used for reflection. We aren't changing the output type of these get cmdlets.

The nuanced user or PowerShell expert can use the cmdlet in question with the appropriate parameter set depending on the resource type he is working with. We aren't stopping this or doing away with this ability.

But, the typical ASR user might not be a PowerShell expert, and isn't typically one who would do all their operations from PowerShell. More often than not, they will have created the vault and performed their initial configurations from the portal. The reason they want to use PowerShell is to automate certain repeat taks. So they wont be creating vaults from PowerShell. What they are attempting to do is really set the context for their vault and then perform one or two operations within the vault. Might be a simple get cmdlet  to check replication health in a scheduled PS script that they use for monitoring. These users, tend to get confused at the step where they need to set context.

We received feedback from a couple of users, who ended up being blocked at this step and were wondering why they were unable to proceed - because they ended up using the wrong parameter set. This enhancement is to make life simple for such customers as they needn't concern themselves with some of the underlying nuances.

So while we have two parameter sets (one for each resource type), we are adding this third parameter set to make life simple for some of the less advanced (in PowerShell terms) ASR user. 

This cmdlet will only ever take a single parameter and we shouldn't be requiring any more parameter sets. So with this enhancement, the Power users can continue to use the appropriate parameter set of the cmdlet based on the resource type they are working with, while we also have a third parameter set that will work with both resource(object) types.

src/ResourceManager/SiteRecovery/Commands.SiteRecovery/Vault/SetAzureSiteRecoveryVaultSettings.cs
```diff
@@ -31,7 +31,7 @@ public class SetAzureSiteRecoveryVaultSettings : SiteRecoveryCmdletBase
         /// <summary>
         /// Gets or sets vault Object.
         /// </summary>
-        [Parameter(ParameterSetName = ASRParameterSets.Default, Mandatory = true)]
+        [Parameter(ParameterSetName = ASRParameterSets.Default, Mandatory = true, ValueFromPipeline = true)]
```
markcowl: This is very, very bad practice.  Everything on the pipeline is convertible to object

markcowl:
@bsiva   The problem is that 'Object'  will bind to any value they pass to this parameter, whehter it is a Vault, a string, an int, or whatever.  This produces far more potential for customer issues than you are trying to solve.  You may be able to get the functionality you want by using ValueFromPipeline in your existing parameter sets.  Let me query the powershell team for some advice on this.

bsiva:
@markcowl : Yes, makes sense. I was thinking the same when I saw your initial comment. I think we can have the other two parameter sets accept value from pipeline and not accept pipeline values for the new parameter set we are adding and should be able to achieve the same effect. We'll try this out and push the changes if this works. Thanks.

bsiva:
@avneeshrai : Avneesh, can you address Mark's comment on accepting pipeline values for the original parameter sets and drop pipelining for the new parameter set (-Vault) we're adding.

Let's validate that pipelining works for both the site recovery vault and recovery services vault with these new changes and then push them back up

markcowl:
@avneeshrai It looks like there are failures in running credscan, this probably means you need to pull the latest.  The exceptions file is here:  https://github.com/Azure/azure-powershell/blob/dev/tools/GlobalFilters.xml

markcowl:
@avneeshrai doisregard - the credscan errors are not yours.  Will rerun once the exceptions are updated

src/ResourceManager/SiteRecovery/Commands.SiteRecovery/Vault/SetAzureSiteRecoveryVaultSettings.cs
```diff
@@ -31,21 +31,21 @@ public class SetAzureSiteRecoveryVaultSettings : SiteRecoveryCmdletBase
         /// <summary>
         /// Gets or sets vault Object.
         /// </summary>
-        [Parameter(ParameterSetName = ASRParameterSets.Default, Mandatory = true, ValueFromPipeline = true)]
+        [Parameter(ParameterSetName = ASRParameterSets.Default, Mandatory = true)]
```
markcowl: You would remove this parameter altogether in this case.

markcowl:

src/ResourceManager/SiteRecovery/Commands.SiteRecovery/Vault/SetAzureSiteRecoveryVaultSettings.cs
bsiva: We still want to retain this for customers who are not passing objects through the pipeline

kirthik:
@avneeshrai - It looks like below file has some credentials that you need to either remove (if they are valid ones) or add exceptions

`src\ResourceManager\RecoveryServices\Commands.RecoveryServices.Test\SessionRecords\Microsoft.Azure.Commands.RecoveryServices.Test.ScenarioTests.RecoveryServicesTests\VaultCRUDTests.json`

avneeshrai:
@markcowl @kirthik I have added the file to the exceptions list and build is triggered again.

avneeshrai:
@markcowl Build is green.

src/ResourceManager/SiteRecovery/Commands.SiteRecovery/Vault/SetAzureSiteRecoveryVaultSettings.cs
markcowl: Verified with the PowerShell core team.  Please remove the parameter set with an Object parameter

avneeshrai:
@markcowl please let us know if any action is needed on our end for this PR to be merged?

src/ResourceManager/SiteRecovery/Commands.SiteRecovery/Vault/SetAzureSiteRecoveryVaultSettings.cs
markcowl: This parameter with type object is still there.  Remove this parameter, please.  This has been verified with the PowerShell core team.  You will need to remove this parameter to get this checked in.

markcowl:
@avneeshrai Yes, please implement the final code comment

avneeshrai:
@markcowl Taken your comment.

markcowl:
on demand run here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/963/

markcowl:
@avneeshrai Please pull the latest - the failing test is fixed in the tip of the dev branch

markcowl:
http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/978/

PR #2480, opened 2016-06-21
Change Tags parameter name to Tag, and modify some help text

blueww:
1. Change the Name of Tags parameter to Tag, and add Tags as alais
2. Modify the help text as feedback and Tags name change

The change has pass Storage team regression test.


blueww:
On Demond test Pass: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/906/

Issue #2482, opened 2016-06-21
Add E2E live tests for HDInsight


shefaliv:

cormacpayne:
@z01c can you take a look at this? Does it still apply to the HDInsight tests?

PR #2483, opened 2016-06-21
HDInsight - update references

shefaliv:


shefaliv:
https://github.com/Azure/azure-powershell/issues/2482

shefaliv:
on demand build: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/907/

MSI: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-msi/679/

markcowl:
@shefaliv, looks like there is an assembly version conflict in your own directory with Newtonsoft.Json.  caused by the downgrade of the library. This is causing the build failure. If you can verify that this scenario works with the conflict (hard to imagine if storage is actually being used), then you can add this as an exception here: https://github.com/Azure/azure-powershell/blob/dev/tools/StaticAnalysis/Exceptions/AssemblyVersionConflict.csv 

Regardless, it looks like this isn't going to make the release train for the hotfix, so please retarget the fix at the dev branch.  Closing for now.

PR #2486, opened 2016-06-21
Modify objectIds from Guid to string in RBAC cmdlets

shuagarw:


markcowl:
on demand run here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/912/

jkapil:
Is this a breaking change? What happens to the user who pass a GUID to these cmdlets for interacting with Azure

jkapil:
what about https://github.com/Azure/azure-powershell/blob/dev/src/ResourceManager/KeyVault/Commands.KeyVault/Models/KeyVaultManagementCmdletBase.cs
protected Guid GetObjectId(Guid objectId, string upn, string spn) 
protected Guid GetCurrentUsersObjectId()

PR #2490, opened 2016-06-22
Feature/app pack hotfix

matthew-rollitt:
When activating application packages you don't need a filePath.


markcowl:
@azuresdkci add to whitelist

PR #2493, opened 2016-06-22
fix VM size information

MSSedusch:
add new DS5_v2 and DS15_v2 sizes and fix values for other sizes


markcowl:
@MSSedusch No test updates?

MSSedusch:
no, the tests still work

PR #2505, opened 2016-06-23
[DO NOT MERGE]: Add ShoudContinue/ShouldProcess Warning Message

blueww:
Add ShoudContinue/ShouldProcess  Warning Message per our mail thread 

> RE: [BREAKING CHANGE NOTICE]:  Azure PowerShell cmdlet changes for ShouldProcess, ShouldContinue, and Force.

Please reply the original mail thread and comment if you have any comments on the warning message. 

As this is only text change, no necessary for storage regression test. And I have manually verified all added warning message will be show when execute the related cmdlets.


blueww:
On Demand test: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/918/

src/ResourceManager/Storage/Commands.Management.Storage/StorageAccount/RemoveAzureStorageAccount.cs
```diff
@@ -44,6 +44,8 @@ public override void ExecuteCmdlet()
         {
             base.ExecuteCmdlet();
 
+            WriteWarning("Will add confirmation of executing the cmdlet and Force parameter in a future release. ");
```
markcowl: Rather than writing a warnign like this, we should attribute the obsoleted parameter.  Please hold off until the ShouldProcess PR comes in.

blueww:
Thanks Mark!
I will hold it until the ShouldProcess PR is in.

src/ResourceManager/Storage/Commands.Management.Storage/StorageAccount/RemoveAzureStorageAccount.cs
EmmaZhu: Here in this cmdlet, we'll add a Force parameter instead of removing one as we'll add ShouldProcess&ShouldContinue confirmation logic. What do you think about the warning message?

PR #1846, opened 2016-02-23
KeyVault: Fix bug for tests in powershell

karlaugsten:
powershell 1.2.x does not import the Microsoft.WindowsAzure.Testing.TestUtilities module.

PR #2509, opened 2016-06-23
Fix issues with running RunKeyVaultTests.ps1.

adarce:
## Background

When attempting to run the Azure Key Vault PowerShell tests via the `RunKeyVaultTests.ps1` script, the following 5 tests would fail:

```
Test_ImportByokWithDefaultAttributes
Test_ImportByokWith1024BitKey
Test_ImportByokWithCustomAttributes
Test_GetAllKeys
Test_GetKeyVersions
```

The first 3 BYOK tests failed because they test against dummy `.byok` files that were generated with a specific subscription ID that may not match the subscription ID of whomever is running the tests. The last 2 tests failed because they attempt to reference `[Microsoft.WindowsAzure.Testing.TestUtilities]`, which cannot be found.

Moreover, this PowerShell script was difficult to use if we didn't want to provide our own vault to test against for both the data plane and the control plane, but we instead wanted the script to generate its own temporary vault. Also, we sometimes would like to programmatically skip Active Directory related tests for certain environments (instead of having it hardcoded to always skip for Fairfax environments). Finally, the PowerShell script could use some cleanup with respect to formatting whitespace, fleshing out docstrings, and general organization.
## The Change

This change addresses the above issues by:
- Guarding the 3 BYOK tests with an if condition that tests the subscription ID.
- Replacing `Wait($timeout * 1000)` with `Start-Sleep -Seconds $timeout`.
- Saving/restoring temporary vault state for both the control plane and data plane tests.
- Exposing the `$NoADCmdLetMode` parameter.
- Generally improving the coding style.


azurecla:
Hi __@adarce__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (adarce). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

src/ResourceManager/KeyVault/Commands.KeyVault.Test/Scripts/PSHCommon/Common.ps1
```diff
@@ -286,68 +357,78 @@ function Wait-Function
...
 {
-    param([int] $timeout)
-    
-    [Microsoft.WindowsAzure.Testing.TestUtilities]::Wait($timeout * 1000)
-}
+    param(
+        [int] $timeout
+    )
 
+    Start-Sleep -Seconds $timeout
```
hovsepm: revert this back to [Microsoft.WindowsAzure.Testing.TestUtilities]::Wait. We do not want to do sleep waits in playback mode.

hovsepm:
@adarce please squash your commits.

hovsepm:
@pomortaz do you need to sign-off? :)

pomortaz:
@hovsepm thanks for the review. I've already signed off. :)

src/ResourceManager/KeyVault/Commands.KeyVault.Test/Scripts/PSHCommon/Common.ps1
adarce: @hovsepm  It seems that this issue has come up before (see: https://github.com/Azure/azure-powershell/pull/1846#discussion_r53832755). I have updated my pull request to use the workaround recommended by @markcowl here: (https://github.com/Azure/azure-powershell/pull/1846#issuecomment-188571285).

adarce:
@hovsepm I have squashed my commits, as requested. Please let me know if there's anything else I can improve!

src/ResourceManager/KeyVault/Commands.KeyVault.Test/Scripts/ControlPlane/KeyVaultManagementTests.ps1
```diff
@@ -37,6 +37,8 @@ Param($rgName, $location, $tagName, $tagValue)
     Assert-AreEqual "Standard" $actual.Sku	
     Assert-AreEqual $false $actual.EnabledForDeployment
 
+    if ($global:noADCmdLetMode) {return;}
```
markcowl: I really don't like this.  This just skips the test without letting us know that's what happened.  A better solution would be to store the needed subscriptionId in the mocks, which is how most tests solve similar issues.

src/ResourceManager/KeyVault/Commands.KeyVault.Test/Scripts/PSHCommon/Common.ps1
```diff
@@ -13,51 +13,78 @@
 # ----------------------------------------------------------------------------------
```
markcowl: I'm not sure why you have captured a version of the Common.ps1 from the scenario tests common.  Seems like it woule be better to simply use the common ones.

src/ResourceManager/KeyVault/Commands.KeyVault.Test/Scripts/PSHCommon/Common.ps1
```diff
@@ -77,23 +104,21 @@ function Run-Test
...
+        $success = $true
     }
-    finally 
+    finally
     {
         Test-Cleanup
-        $oldPref = $ErrorActionPreference	 
+        $oldPref = $ErrorActionPreference
         $ErrorActionPreference = "SilentlyContinue"
-        #Stop-Transcript
```
markcowl: No need for setting error action preference if no transcript cmdlets are used.

src/ResourceManager/KeyVault/Commands.KeyVault.Test/Scripts/PSHCommon/Common.ps1
```diff
@@ -77,23 +104,21 @@ function Run-Test
     {
          Write-Log "[run-test]: Running test without file comparison"
     }
-        
-    $oldPref = $ErrorActionPreference	 
+
+    $oldPref = $ErrorActionPreference
     $ErrorActionPreference = "SilentlyContinue"
-    #Start-Transcript -Path $transFile
```
markcowl: Same comment - the ErrorActionPreference setting is not necessary if tthe transcript command is removed.

src/ResourceManager/KeyVault/Commands.KeyVault.Test/Scripts/PSHCommon/Common.ps1
```diff
@@ -286,68 +357,98 @@ function Wait-Function
...
-    
-    [Microsoft.WindowsAzure.Testing.TestUtilities]::Wait($timeout * 1000)
-}
+    param(
+        [int] $timeout
+    )
 
+    try
+    {
+        if ([Microsoft.Azure.Test.HttpRecorder.HttpMockServer]::Mode -eq [Microsoft.Azure.Test.HttpRecorder.HttpRecorderMode]::Playback)
```
markcowl: You should use TestMockSupport.RunningMocked, instead

markcowl:
@adarce a few comments

adarce:
@markcowl I've pushed a commit that hopefully addresses each of your concerns. Let me know if there's anything else I can improve!

markcowl:
on demand run here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/933/

markcowl:
LGTM once the on-demand run passes

PR #2522, opened 2016-06-28
PATCH /tenant/access/git is No Longer Required

solankisamir:
After the Git Feature went GA, the resource PATCH /tenant/access/git is deprecated. Removing the cmdlet and associated tests.


markcowl:
on demand run here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/928/

PR #2544, opened 2016-07-01
Adding code to remove dependency with AzureKeyVaultServiceEndpointResourceId

jkapil:
Adding code to remove dependency with static value specified in AzureKeyVaultServiceEndpointResourceId of Get-AzureRmEnvironment. The keyvault resource Id is now obtained from server using 401 challenge. The Azure PowerShell common infrastructure currently do not support this in an intuitive fashion, so tweaking its usage by overriding the environment endpoint value before acquiring the token.

### DIFF
--- a/src/ResourceManager/KeyVault/Commands.KeyVault/Models/DataServiceCredential.cs
+++ b/src/ResourceManager/KeyVault/Commands.KeyVault/Models/DataServiceCredential.cs
@@ -24,26 +24,23 @@ namespace Microsoft.Azure.Commands.KeyVault.Models
 {
     internal class DataServiceCredential
     {
+        private readonly IAuthenticationFactory _authenticationFactory;
+        private readonly AzureContext _context;
+        private readonly AzureEnvironment.Endpoint _endpointName;
+
         public DataServiceCredential(IAuthenticationFactory authFactory, AzureContext context, AzureEnvironment.Endpoint resourceIdEndpoint)
         {
             if (authFactory == null)
                 throw new ArgumentNullException("authFactory");
             if (context == null)
                 throw new ArgumentNullException("context");
-
-            var bundle = GetToken(authFactory, context, resourceIdEndpoint);
-            this.token = bundle.Item1;
+            _authenticationFactory = authFactory;
+            _context = context;
+            _endpointName = resourceIdEndpoint;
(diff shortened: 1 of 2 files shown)


azurecla:
Hi __@jkapil__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

src/ResourceManager/KeyVault/Commands.KeyVault/Models/DataServiceCredential.cs
```diff
@@ -74,14 +84,18 @@ public Task<string> OnAuthentication(string authority, string resource, string s
                 throw new ArgumentException(string.Format(KeyVaultProperties.Resources.UnsupportedAccountType, context.Account.Type));
 
             if (context.Subscription != null && context.Account != null)
-                TenantId = context.Subscription.GetPropertyAsArray(AzureSubscription.Property.Tenants)
+                tenantId = context.Subscription.GetPropertyAsArray(AzureSubscription.Property.Tenants)
```
markcowl: Why not use context.tenant?

markcowl:
@jkapil one minor comment.  I would really like to see a mocked test for the scenario - can you mock this by mocking out the authentication factory?

src/ResourceManager/KeyVault/Commands.KeyVault/Models/DataServiceCredential.cs
jkapil: We might, didn't wanted to change logic for tenant ID

markcowl:
@jkapil is this also intended for AzureStack TP2?  If so, you'll need to open a PR against the release-1.2.4 branch

src/ResourceManager/KeyVault/Commands.KeyVault/Models/DataServiceCredential.cs
jkapil: Yes this will be integrated into release 1.2.4

hovsepm:
on-demand: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/973/

PR #2595, opened 2016-07-12
Fix the parameter set for New And Set Resource Lock cmdlet

gandhiniraj:


azurecla:
Hi __@gandhiniraj__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

hovsepm:
@vivsriaus please sign-off.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Lock/ResourceLockManagementCmdletBase.cs
```diff
@@ -91,6 +91,7 @@ public abstract class ResourceLockManagementCmdletBase : ResourceManagerCmdletBa
         /// </summary>
         [Parameter(ParameterSetName = ResourceLockManagementCmdletBase.ResourceGroupResourceLevelLock, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The resource group name.")]
         [Parameter(ParameterSetName = ResourceLockManagementCmdletBase.ResourceGroupLevelLock, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The resource group name.")]
+        [Parameter(ParameterSetName = ResourceLockManagementCmdletBase.SubscriptionResourceLevelLock, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The resource group name.")]
         [ValidateNotNullOrEmpty]
```
TianoMS: What scenario does this enable? Trying to figure out a use case.

gandhiniraj: Currently if you don't provide inline parameter for Add and Set Resource Lock cmdlets, it will prompt you for the value of the parameter, currently this scenario is broken - it doesn't ask for the ResourceGroupName and will fail since it is null

vivsriaus: @gandhiniraj  ResourceGroupName should not be a part of the SubscriptionResourceLevelLock parameter set.  The SubscriptionResourceLevelLock is for locks at the subscription level, like /subscriptions/{subId}/RPNamespace/RPType/lockname.

Is there an issue opened for the New-AzureRmResourceLock cmdlet? I just tried creating a lock at using the -Scope and it works. If there is an issue with the default parameter set, we should instead investigate that. Currently, I see the subscription level lock being set as the default for the new cmdlet. We should instead use the ScopeLevel parameter set as the default one. Agree @ravbhatnagar ?

gandhiniraj: Talked to Vivek offline, there isn't any open issue by end user, found it while debugging an issue. per chat we should change that back to scope level since we have to fix the root level issue.

gandhiniraj:
Per chat with Vivek we have to fix the underlying issue differently, I will send out another PR

PR #2636, opened 2016-07-19
Remove exist check from New-AzureRmBatchAccount

jasper-schneider:
Removing some outdated client side validation.


markcowl:
@jasper-schneider isn't there a test recording file that should be deleted for this test as well?

Otherwise, LGTM

jasper-schneider:
@markcowl Doh, good call on the recording file.  I'll update the PR shortly.

PR #2667, opened 2016-07-26
Updating Tag usage in SQL cmdlets

akromm-zz:
Updating Tag usage in SQL cmdlets

Converting the Tags parameter in SQL cmdlets to use the Azure built Tags helper classes and follow the standard.

### DIFF
--- a/src/ResourceManager/Sql/Commands.Sql/Replication/Cmdlet/NewAzureSqlDatabaseCopy.cs
+++ b/src/ResourceManager/Sql/Commands.Sql/Replication/Cmdlet/NewAzureSqlDatabaseCopy.cs
@@ -13,8 +13,10 @@
 // ----------------------------------------------------------------------------------
 
 using Hyak.Common;
+using Microsoft.Azure.Commands.ResourceManager.Common.Tags;
 using Microsoft.Azure.Commands.Sql.Properties;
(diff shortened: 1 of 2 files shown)


cormacpayne:
@akromm are there any test changes required for the parameter change? Was this part ever covered with tests?

akromm:
@cormacpayne I ran all tests and non required any updates.

src/ResourceManager/Sql/Commands.Sql/Server/Cmdlet/SetAzureSqlServer.cs
```diff
@@ -50,8 +52,8 @@ public class SetAzureSqlServer : AzureSqlServerCmdletBase
         /// </summary>
         [Parameter(Mandatory = false,
             HelpMessage = "The tags to associate with the server.")]
-        [ValidateNotNull]
-        public Dictionary<string, string> Tags { get; set; }
+        [Alias("Tag")]
```
hovsepm: What is the value of having Tag alias over property with name "Tag"?

akromm: the property name is "Tags" not "Tag".

PR #2682, opened 2016-07-28
[AzureRT] Sprint 62

hyonholee:
Fix Service name issue in Set-AzureServiceDiagnosticsExtension cmdlets
Update cloud exception error output


src/ServiceManagement/Services/Commands.Utilities/Common/ServiceSettings.cs
```diff
@@ -232,7 +231,6 @@ private static string GetDefaultStorageName(string localStorageName, string glob
             return name;
         }
 
-        /// <summary>
```
hovsepm: please revert this deletion

hovsepm:
@hyonholee please address my comment.

PR #2705, opened 2016-08-02
Update GB18030 test to use the appropriate path

markcowl:
The local directory may change during execution.


markcowl:
on demand run here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1055/

PR #2709, opened 2016-08-02
ADL Update for PowerShell

begoldsm:
Update the Cmdlets to use the latest SDK
Bug fix for resume support in export/import cmdlets
Test updates for tags.


markcowl:
@begoldsm build failures, missing reference?

begoldsm:
@markcowl I am reverting the update for ADLA. There is something bizarre going on with our latest ADLA package that seems to think I need version 3.0 of Microsoft.Rest.Clientruntime (not .azure), which doesn't exist. I will investigate tomorrow, but for this release I am reverting that update.

begoldsm:
@markcowl I need to re-run my tests but I am blocked from doing so due to an issue in our production environment. I am going to close out this PR for now and, if they are able to get it fixed soon, I will retarget for the release branch. These changes aren't critical (in terms of fixing customer reported issues), but I would like to get them in, if it doesn't delay the release.

PR #2736, opened 2016-08-04
Updating the changelog with Azure Sql blob auditing addition

yoavrubin:


cormacpayne:
These updates to the change log were handled in PR https://github.com/Azure/azure-powershell/pull/2743

PR #2774, opened 2016-08-13
Fix documentation script for Tags duplication

markcowl:
Tags manifest is duplicated.  This removes the duplication


shahabhijeet:
Should we document, why we are excluding AzureRM.Tags.psd1
from
"..\..\..\src\Package\Release\ResourceManager\AzureRM.Tags.psd1"
but using below instead
"..\..\..\src\Package\Release\ResourceManager\AzureResourceManager\AzureRM.Tags\AzureRM.Tags.psd1"
Has this module manifest been moved to the new location?

markcowl:
@shahabhijeet No, it appears in both locations, and, unfortunately websites sorts lower in the list, so that is the one that they attempt to grab documentation from (and fail)

PR #2804, opened 2016-08-18
Add script to create registry entry for PS release

cormacpayne:
This script will create a new folder for the release registry entry, copy the necessary files and folders from the previous release registry entry and move them over, and move the signed .msi and packages from the sign job over.


hovsepm:
:shipit:

PR #2810, opened 2016-08-19
Write current module version to Verbose channel.

NonStatic2014:
Sometimes, our customer may use an old version of AzureML PS cmdlet and cause unexpected result. We have to let them know how to use other PowerShell cmdlet to get the module number. With this change, they just need to add `-vb` in the same command line and then they can easily get it and help the investigation.

### DIFF
--- a/src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/WebServicesCmdletBase.cs
+++ b/src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/WebServicesCmdletBase.cs
@@ -82,7 +82,7 @@ namespace Microsoft.Azure.Commands.MachineLearning
             }
             catch (Exception ex)
             {
-                this.WriteVersionInfoToVerboseChannel();
+                this.WriteVersionInfoToDebugChannel();
                 if (this.IsFatalException(ex))
                 {
                     ThrowTerminatingError(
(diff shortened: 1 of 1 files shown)


markcowl:
@NonStatic2014 This is not recommended practice for finding the module version.  If you would prefer to make this part of exception handling, or include the module version int he debug stream, this would be preferable.

NonStatic2014:
@markcowl I tried to write it into Debug channel, but it will require user do the rest of work step by step, or setup an environment variable before run the command. I'd like to use the Verbose channel to make the user experience easier for them other more commands or action.

markcowl:
@NonStatic2014 The assembly version is not useful information in this case, you want the module version.  Please change this to be the module version, and please put this in exception handling code and in the debug stream.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/WebServicesCmdletBase.cs
```diff
@@ -138,6 +140,7 @@ protected override void StopProcessing()
             }
             catch (Exception ex)
             {
+                this.WriteVersionInfoToDebugChannel();
```
markcowl: I was thinking more of adding this information to the error itself, but you could do this in a subsequent PR

markcowl:
ON demand run here:  http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1124/

LGTM once the on-demand run passes.

src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/WebServicesCmdletBase.cs
NonStatic2014: Yep, that will be a bigger change: derive from ErrorRecord, put the module version in it and return that object. Let me do that when we have more information needs to be displayed during exception handling.

PR #2817, opened 2016-08-23
Add ComputeImageAttributes parameters to OSImageDetailsContext object…

hyonholee:

## Overall Changes
- [ ] [**MANDATORY** - General changes](#general)
- [ ] [**MANDATORY** - Add/remove/edit test(s)](#tests)
### General
### Tests
- [ ] Tests must use xunit, and should either use Moq to mock management client calls, or use the scenario test framework

… in Get-AzurePlatformVMImage cmdlet


markcowl:
@hyonholee Please fill out the checklist as appropriate for these changes.

markcowl:
@hyonholee Tests for the new properties?

hyonholee:
@markcowl  this is a publisher's cmdlet and included in PIR module.  (this is not automatically loaded in Azure Powershell.  Publishers must manually import this module.)  So, the test of this cmdlets is done manually since a normal user does not have a permission to use it. (they can import the module, but will get an error during execution.)   This update only exposes the properties that were hidden before, so I did not add an automated test for this.   Thanks.

PR #2824, opened 2016-08-24
AzureSQL - Adding ConfirmAction usage, threat detection policy for servers, blob auditing policy properties

yoavrubin:
ConfirmAction is now handled at the base class for all AzureSQL cmdlets.
Added AzureSQL threat detection policy management at the server level. This includes added tests and rerecording relevant tests.
Extended the storage related properties of blob auditing policy and rerecorded the auditing tests.

### DIFF
--- a/src/ResourceManager/Sql/Commands.Sql/ThreatDetection/Services/ThreatDetectionEndpointsCommunicator.cs
+++ b/src/ResourceManager/Sql/Commands.Sql/ThreatDetection/Services/ThreatDetectionEndpointsCommunicator.cs
@@ -19,6 +19,7 @@ using Microsoft.Azure.Management.Sql;
 using Microsoft.Azure.Management.Sql.Models;
 using System;
 using System.Threading.Tasks;
+using Microsoft.WindowsAzure.Commands.Utilities.Common;
 
 namespace Microsoft.Azure.Commands.Sql.ThreatDetection.Services
 {
@@ -80,7 +81,7 @@ namespace Microsoft.Azure.Commands.Sql.ThreatDetection.Services
                 {
                     break;
                 }
-                Task.Delay(2000); // wait 2 seconds between each poll
+                TestMockSupport.Delay(2000); // wait 2 seconds between each poll
             }
         }
 


markcowl:
@yoavrubin - either rebase, or cherry-pick the commits into a new branch so that the commit history looks a bit better.

markcowl:
Please fill out the appropriate sections of the PR template:

This checklist is used to make sure that common issues in a pull request are covered by the creator. You can find a more complete discussion of PowerShell cmdlet best practices [here](https://msdn.microsoft.com/en-us/library/dd878270(v=vs.85).aspx).

## Overall Changes 
- [ ] [**MANDATORY** - General changes](#general)
- [ ] [**MANDATORY** - Add/remove/edit test(s)](#tests)
- [ ] [Add/remove/edit cmdlet(s)](#cmdlet-signature)
- [ ] [Add/remove/edit parameter(s)](#parameters)

### General
- [ ] Title of the PR is clear and informative
- [ ] All files have the Microsoft copyright header
- [ ] Cmdlets refer to management libraries through nuget references - no dlls are checked in
- [ ] The PR does not introduce breaking changes (unless a major version change occurs in the assembly and module)

### Tests
- [ ] Tests must use xunit, and should either use Moq to mock management client calls, or use the scenario test framework

### Cmdlet Signature
- [ ] Cmdlet name uses an approved PowerShell verb - use the enums for `VerbsCommon`, `VerbsCommunication`, `VerbsLifecycle`, `VerbsOther` whenever possible
- [ ] Cmdlet noun name uses the AzureRm prefix for management cmdlets, and the Azure prefix for data plane cmdlets
- [ ] Cmdlet specifies the `OutputType` attribute if any output is produced; if the cmdlet produces no output, it should implement a `PassThrough` parameter
- [ ] If the cmdlet makes changes or has side effects, it should implement `ShouldProcess` and have `SupportShouldProcess = true` specified in the cmdlet attribute. See a discussion about correct `ShouldProcess` implementation [here](https://gist.github.com/markcowl/338e16fe5c8bbf195aff9f8af0db585d#what-is-the-change)
- [ ] Cmdlets should derive from AzureRmCmdlet for management cmdlets, and AzureDataCmdlet for data cmdlets
- [ ] If multiple parameter sets are implemented, the cmdlet should specify a `DefaultParameterSetName` in its cmdlet attribute

### Parameters
- [ ] Cmdlets should have no more than four positional parameters
- [ ] Cmdlet parameter sets should be mutually exclusive - each parameter set must have at least one mandatory parameter not in other parameter sets
- [ ] Parameter types should not expose types from the management library - complex parameter types should be defined in the module
- [ ] Complex parameter types are discouraged - a parameter type should be simple types as often as possible. If complex types are used, they should be shallow and easily creatable from a constructor or another cmdlet
- [ ] Parameters should be explicitly marked as Mandatory or not, and should contain a HelpMessage
- [ ] No parameter is of type `object`.
        - Tag, type `HashTable`

src/ResourceManager/Sql/Commands.Sql/ThreatDetection/Services/ThreatDetectionEndpointsCommunicator.cs
```diff
@@ -52,6 +53,47 @@ public ThreatDetectionEndpointsCommunicator(AzureContext context)
...
+                return;
+            }
+            for (var iterationCount = 0; iterationCount < 1800; iterationCount++) // wait for at most an hour
+            {
+                var status = GetServerCreateOrUpdateOperationStatus(statusLink, clientRequestId);
+                if (status == OperationStatus.Succeeded)
+                {
+                    break;
+                }
+                Task.Delay(2000); // wait 2 seconds between each poll
```
hovsepm: @yoavrubin Please use TestMockSupport.Delay() instead

PR #2832, opened 2016-08-25
Include environment name in the key of Api-Version cache. 

TianoMS:


markcowl:
@TianoMS Thanks, Tiano, how are we testing this?  How difficult would it be to add a regression test?
Also, can you add this to the release-1.2.4 branch as well?

TianoMS:
@markcowl To test this we need an account which has subscriptions in a different environment. And we need to get a resource type which supports different versions in different environments, which is a very rare case. I tried to work this out but didn't succeed... 

Right now I'm trying to ask the issue opener to get a consistent repro from their side. They didn't get back yet. Once I get that I'll see if I can build a test case from their scenario.

TianoMS:
@markcowl I manually verified the fix with Bala's help. The fix works. Right now we don't have a good way to test it in an automation. We may have to skip it this time.

markcowl:
@TianoMS Please file an issue for this - it would be fine to have some separate tests of the caching mechanism rather than end-to-end test automation.  Also, please fix in the release-1.2.4 branch as well.

PR #2835, opened 2016-08-25
Azure RM Dns Bug Fixes and Update to Help File

corinagum:
1.  [No work item#] Update to Help File
2.  [No work item#] Fix -Confirm parameter to accept $false & remove -Force
3.  Bug 6412699 PowerShell Error: can’t convert …DnsRecordSet to type …DnsRecordSet
4.  Bug 3172031 PowerShell should warn if record name suffix matches zone name
5.  Bug 5600833 PowerShell New/Remove-AzureRmDnsRecordSet supports SOA
6.  Bug 6059441 PowerShell should expose NumRecordSets and MaxNumRecordSets

## Overall Changes
- [ x ] [**MANDATORY** - General changes](#general)
- [ x ] [**MANDATORY** - Add/remove/edit test(s)](#tests)
- [ x ] [Add/remove/edit cmdlet(s)](#cmdlet-signature)
- [ x ] [Add/remove/edit parameter(s)](#parameters)
### General
- [x] If it applies, references the bug/issue that the PR fixes
### Tests
### Cmdlet Signature
- [x] Cmdlet name uses an approved PowerShell verb - use the enums for `VerbsCommon`, `VerbsCommunication`, `VerbsLifecycle`, `VerbsOther` whenever possible
- [x] Cmdlet noun name uses the AzureRm prefix for management cmdlets, and the Azure prefix for data plane cmdlets
- [x] Cmdlet specifies the `OutputType` attribute if any output is produced; if the cmdlet produces no output, it should implement a `PassThrough` parameter
- [x] If the cmdlet makes changes or has side effects, it should implement `ShouldProcess` and have `SupportShouldProcess = true` specified in the cmdlet attribute. See a discussion about correct `ShouldProcess` implementation [here](https://gist.github.com/markcowl/338e16fe5c8bbf195aff9f8af0db585d#what-is-the-change)
- [x] Cmdlets should derive from AzureRmCmdlet for management cmdlets, and AzureDataCmdlet for data cmdlets
- [x] If multiple parameter sets are implemented, the cmdlet should specify a `DefaultParameterSetName` in its cmdlet attribute
### Parameters
- [x] Cmdlets should have no more than four positional parameters
- [x] Cmdlet parameter sets should be mutually exclusive - each parameter set must have at least one mandatory parameter not in other parameter sets
- [x] Parameter types should not expose types from the management library - complex parameter types should be defined in the module
- [x] Complex parameter types are discouraged - a parameter type should be simple types as often as possible. If complex types are used, they should be shallow and easily creatable from a constructor or another cmdlet
- [x] Parameters should be explicitly marked as Mandatory or not, and should contain a HelpMessage
- [x] No parameter is of type `object`.
    - Tag, type `HashTable`

### DIFF
--- a/src/ResourceManager/Dns/Commands.Dns/Records/NewAzureDnsRescordSet.cs
+++ b/src/ResourceManager/Dns/Commands.Dns/Records/NewAzureDnsRescordSet.cs
@@ -14,6 +14,7 @@
 
 using Microsoft.Azure.Commands.Dns.Models;
 using Microsoft.Azure.Management.Dns.Models;
+using System;
 using System.Collections;
 using System.Management.Automation;
 using ProjectResources = Microsoft.Azure.Commands.Dns.Properties.Resources;
@@ -60,6 +61,10 @@ namespace Microsoft.Azure.Commands.Dns
 
         [Parameter(Mandatory = false, HelpMessage = "Do not fail if the record set already exists.")]
         public SwitchParameter Overwrite { get; set; }
+
+        [Parameter(Mandatory = false, HelpMessage = "Do not ask for confirmation.")]
+        [Obsolete("This parameter is obsolete; use Confirm instead")]
+        public SwitchParameter Force { get; set; }
 
         public override void ExecuteCmdlet()
         {
--- a/src/ResourceManager/Dns/Commands.Dns/Records/RemoveAzureDnsRecordSet.cs
+++ b/src/ResourceManager/Dns/Commands.Dns/Records/RemoveAzureDnsRecordSet.cs
@@ -14,6 +14,7 @@
 
 using Microsoft.Azure.Commands.Dns.Models;
 using Microsoft.Azure.Management.Dns.Models;
+using System;
 using System.Management.Automation;
 using ProjectResources = Microsoft.Azure.Commands.Dns.Properties.Resources;
 
@@ -54,6 +55,10 @@ namespace Microsoft.Azure.Commands.Dns
 
         [Parameter(Mandatory = false, HelpMessage = "Do not use the ETag field of the RecordSet parameter for optimistic concurrency checks.", ParameterSetName = "Object")]
         public SwitchParameter Overwrite { get; set; }
+
+        [Parameter(Mandatory = false, HelpMessage = "Do not ask for confirmation.")]
+        [Obsolete("This parameter is obsolete; use Confirm instead")]
+        public SwitchParameter Force { get; set; }
 
         [Parameter(Mandatory = false)]
         public SwitchParameter PassThru { get; set; }
--- a/src/ResourceManager/Dns/Commands.Dns/Zones/RemoveAzureDnsZone.cs
+++ b/src/ResourceManager/Dns/Commands.Dns/Zones/RemoveAzureDnsZone.cs
@@ -43,6 +43,11 @@ namespace Microsoft.Azure.Commands.Dns
         [Parameter(Mandatory = false, HelpMessage = "Do not use the ETag field of the Zone parameter for optimistic concurrency checks.", ParameterSetName = "Object")]
         public SwitchParameter Overwrite { get; set; }
 
+        [Parameter(Mandatory = false, HelpMessage = "Do not ask for confirmation.")]
+        [Obsolete("This parameter is obsolete; use Confirm instead")]
+        public SwitchParameter Force { get; set; }
+
+
         [Parameter(Mandatory = false)]
         public SwitchParameter PassThru { get; set; }
 


azurecla:
Hi __@corinajacobson__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (v-cojaco). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

azurecla:
__@corinajacobson__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

src/ResourceManager/Dns/Commands.Dns.Test/Commands.Dns.Test.csproj
```diff
@@ -215,6 +215,7 @@
     <None Include="..\..\Common\Commands.ScenarioTests.ResourceManager.Common\AzureRM.Resources.ps1">
       <Link>ScenarioTests\AzureRM.Resources.ps1</Link>
     </None>
+    <None Include="app.config" />
```
markcowl: remove, this will cause failures, ass it is on .gitignore

src/ResourceManager/Dns/Commands.Dns/Zones/RemoveAzureDnsZone.cs
```diff
@@ -79,7 +76,7 @@ public override void ExecuteCmdlet()
                 }
 
                 ConfirmAction(
-                    Force.IsPresent,
+                    true,
```
markcowl: In this case, call the overload that does not contain a force parameter

src/ResourceManager/Dns/Commands.Dns/Zones/SetAzureDnsZone.cs
```diff
@@ -76,12 +76,21 @@ public override void ExecuteCmdlet()
                 zoneToUpdate.Name = zoneToUpdate.Name.TrimEnd('.');
                 this.WriteWarning(string.Format("Modifying zone name to remove terminating '.'.  Zone name used is \"{0}\".", zoneToUpdate.Name));
             }
+            ConfirmAction(
+                true,
```
markcowl: Same comment.  Call the overload with no Force parameter if no force parameter is present on the cmdlet

src/ResourceManager/Dns/Commands.Dns.Test/Commands.Dns.Test.csproj
```diff
@@ -357,4 +358,4 @@
     </EmbeddedResource>
   </ItemGroup>
```
markcowl: It doesn't look like the new test recordings are added here.  You'll need to add these explicitly, with 'Copy If Newer' or 'Copy Always' property

markcowl:
@azuresdkci add to whitelist

corinajacobson:
All comments have been addressed and I will push to pull request after running tests one more time. Thank you!

src/ResourceManager/Dns/Commands.Dns/Records/NewAzureDnsRescordSet.cs
```diff
@@ -61,15 +61,17 @@ public class NewAzureDnsRecordSet : DnsBaseCmdlet
         [Parameter(Mandatory = false, HelpMessage = "Do not fail if the record set already exists.")]
         public SwitchParameter Overwrite { get; set; }
 
-        [Parameter(Mandatory = false, HelpMessage = "Do not ask for confirmation.")]
```
markcowl: This is a breaking change.  The advice for this breaking change is to leave the parameeter add the ```[Obsolete(...)]``` attribute to the property, and not use it in cmdlet execution.  This gives users time to adapt to the change.

src/ResourceManager/Dns/Commands.Dns/Records/RemoveAzureDnsRecordSet.cs
```diff
@@ -55,9 +55,6 @@ public class RemoveAzureDnsRecordSet : DnsBaseCmdlet
         [Parameter(Mandatory = false, HelpMessage = "Do not use the ETag field of the RecordSet parameter for optimistic concurrency checks.", ParameterSetName = "Object")]
         public SwitchParameter Overwrite { get; set; }
 
-        [Parameter(Mandatory = false, HelpMessage = "Do not ask for confirmation.")]
```
markcowl: Same comment

markcowl:
@corinajacobson Sorry, I had forgotten about the breaking change policy, so missed this last time - weh should leave the Force parameters, not use them, and makr them as obsolete - this gives users a chance to update their scripts.

jtuliani:
@markcowl Does the breaking change policy apply to services in Preview?

markcowl:
@jtuliani Right now there is no preview for powershell.  WHen there is, it will not, but for now, there is only one way for customers to get azure powershell

markcowl:
on demand run here:http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1135/

LGTM once the build passes

jtuliani:
@markcowl The Azure DNS service is in Preview.  As with other Azure Previews, we clearly state that it is provided without SLA, and without Azure Support, should not be used for production workloads, and is subject to a 50% billing discount.

To me, 'Preview' includes every part of the service, including the PowerShell cmdlets and CLI.  I therefore don't think the breaking change policy should apply (indeed, we have made other breaking changes during Preview, although we prefer not to of course).

corinajacobson:
Looks like tests have passed! Please let me know if I can do anything else to help out.

markcowl:
@jtuliani  Unfortunately, we don't have any current way of differentiating the preview from the stable modules in azure powershell.  Expect some discussion of new policy in this area over the next couple of weeks.

PR #2848, opened 2016-08-28
Update ResourceManager SDK to 1.2.0-preview version. 

TianoMS:

## Overall Changes
### General
### Tests

### DIFF
--- a/src/ResourceManager/Resources/Commands.Resources.Test/Commands.Resources.Test.csproj
+++ b/src/ResourceManager/Resources/Commands.Resources.Test/Commands.Resources.Test.csproj
@@ -71,7 +71,7 @@
       <HintPath>..\..\..\packages\Microsoft.Azure.Management.Authorization.2.0.0\lib\net40\Microsoft.Azure.Management.Authorization.dll</HintPath>
     </Reference>
     <Reference Include="Microsoft.Azure.Management.ResourceManager">
-      <HintPath>..\..\..\packages\Microsoft.Azure.Management.ResourceManager.1.1.5-preview\lib\net45\Microsoft.Azure.Management.ResourceManager.dll</HintPath>
+      <HintPath>..\..\..\packages\Microsoft.Azure.Management.ResourceManager.1.2.0-preview\lib\net45\Microsoft.Azure.Management.ResourceManager.dll</HintPath>
     </Reference>
     <Reference Include="Microsoft.Azure.ResourceManager, Version=2.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
       <SpecificVersion>False</SpecificVersion>
--- a/src/ResourceManager/Common/Commands.ScenarioTests.ResourceManager.Common/Commands.ScenarioTests.ResourceManager.Common.csproj
+++ b/src/ResourceManager/Common/Commands.ScenarioTests.ResourceManager.Common/Commands.ScenarioTests.ResourceManager.Common.csproj
@@ -50,7 +50,7 @@
       <HintPath>..\..\..\packages\Microsoft.Azure.Common.2.1.0\lib\net45\Microsoft.Azure.Common.NetFramework.dll</HintPath>
     </Reference>
     <Reference Include="Microsoft.Azure.Management.ResourceManager">
-      <HintPath>..\..\..\packages\Microsoft.Azure.Management.ResourceManager.1.1.5-preview\lib\net45\Microsoft.Azure.Management.ResourceManager.dll</HintPath>
+      <HintPath>..\..\..\packages\Microsoft.Azure.Management.ResourceManager.1.2.0-preview\lib\net45\Microsoft.Azure.Management.ResourceManager.dll</HintPath>
     </Reference>
     <Reference Include="Microsoft.Azure.ResourceManager, Version=2.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
       <HintPath>..\..\..\packages\Microsoft.Azure.Management.Resources.2.20.0-preview\lib\net40\Microsoft.Azure.ResourceManager.dll</HintPath>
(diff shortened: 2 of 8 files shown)


TianoMS:
@markcowl My change upgrades to Microsoft.Azure.Management.ResourceManager.1.1.5-preview. I did it only for the Resources solution. Most of the other projects are still using 1.1.1-preview version. Let me know if you see a problem in it.

TianoMS:
On demand job: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1112/

vivsriaus:
@markcowl @hovsepm Adding on @TianoMS comment above regarding different versions of the same assembly. I recall there was an issue with the way powershell loaded assemblies that differed in just the maintenance version (1.1.1 vs 1.1.5). Since some projects reference 1.1.1 and ours (with this PR) reference 1.1.5, we want to ensure that this doesn't pose any problems for our cmdlets. i.e. - we don't have to force the powershell to load the 1.1.5 assembly via Add-Type to use it in our cmdlet.

src/ResourceManager/Resources/Commands.Resources.Test/Commands.Resources.Test.csproj
```diff
@@ -116,7 +116,7 @@
     </Reference>
     <Reference Include="Microsoft.Rest.ClientRuntime.Azure.TestFramework, Version=1.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
       <SpecificVersion>False</SpecificVersion>
-      <HintPath>..\..\..\packages\Microsoft.Rest.ClientRuntime.Azure.TestFramework.1.3.4-preview\lib\net45\Microsoft.Rest.ClientRuntime.Azure.TestFramework.dll</HintPath>
+      <HintPath>..\..\..\packages\Microsoft.Rest.ClientRuntime.Azure.TestFramework.1.3.6-preview\lib\net45\Microsoft.Rest.ClientRuntime.Azure.TestFramework.dll</HintPath>
```
vivsriaus: Was this intentional?

vivsriaus:
@TianoMS Since this is an api-version bump, why didn't we update the assembly's minor version instead (1.2.0 instead of 1.1.5)?

src/ResourceManager/Resources/Commands.Resources.Test/Commands.Resources.Test.csproj
TianoMS: This is about another issue. I think it's already fixed. Should not be a problem.

TianoMS:
@vivsriaus I have another change in SDK (the parenthesis in resource group names) recently which updated the version to 1.1.6-preview. It's not published yet. Maybe we can update the version to 1.2.0 there.

vivsriaus:
Yes, let's please do that @TianoMS. api-version bumps definitely warrants at least a minor version update. We should be fine keeping the powershell changes as-is, since the implementation is internal.

cormacpayne:
@TianoMS Hey Tian, your on-demand build failed because of a bug in a test that we merged a fix for yesterday evening, so if you pull from dev and rebase, that should fix the problem.

markcowl:
LGTM once the on-demand build passes.

on demand: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1122/

TianoMS:
@markcowl Please hold on with the merge. We have a newer version of SDK coming out. I'll push one more commit to upgrade to that version (1.2.0) directly.

TianoMS:
This is now ready to merge. On demand job:
http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1126/

src/ResourceManager/Resources/Commands.Resources.Test/Providers/GetAzureProviderCmdletTests.cs
```diff
@@ -153,7 +153,7 @@ public void GetsResourceProviderTests()
             };
             var pagableLocations = new Page<Location>();
             pagableLocations.SetItemValue<Location>(locationList);
-            var locationsResult = new AzureOperationResponse<IPage<Location>>()
+            var locationsResult = new AzureOperationResponse<IEnumerable<Location>>()
```
markcowl: This is a breaking change - why is this operation no longer pageable?

markcowl:
@TianoMS One question about the removal of a pageable operation

src/ResourceManager/Resources/Commands.Resources.Test/Providers/GetAzureProviderCmdletTests.cs
TianoMS: The change really happens here: https://github.com/Azure/azure-rest-api-specs/commit/5129caaf3816b18fa0e32f3e8e5ece7b9e70f6e6. 

At the bottom, it sets "nextLinkName" to null. This makes the generated SDK not pageable. ISubscriptionsOperations.ListLocationsWithHttpMessagesAsync now returns IEnumerable. 

It's not a breaking change here. It's a breaking change in .NET SDK. I can sync up with Vivek on that change in swagger spec. 

This should be a

TianoMS:
@markcowl Addressed.

src/ResourceManager/Resources/Commands.Resources.Test/Providers/GetAzureProviderCmdletTests.cs
vivsriaus: @TianoMS @markcowl That is correct. The list subscription locations was never pageable to begin with, it was incorrectly added. The above change to swagger spec @TianoMS referenced fixed the issue.

markcowl: As long as it's not a breaking change here, it's OK, but IMO, you want to fix this in the sdk, you don't want to go from pageange to not pageable back to pageable

Issue #2929, opened 2016-09-13
Automate creation of build drop


markcowl:
### Description

Create automation that, given a powershell-sign build and msi version data, create a build drop including
- MSI and packages
- scripts abd webpi installer
- xml webpi files
- a new branch/commit in webpi codeplex

### Cost: 2

markcowl:
Fixed

PR #2956, opened 2016-09-16
Add script to create a build drop from a sign job

cormacpayne:
## Comments

Fix for issue https://github.com/Azure/azure-powershell/issues/2929

This script will be used for Azure PowerShell releases.

It takes the build artifacts from a sign job and creates a new entry in the shared folder that will contain the msi, packages, scripts and WebPI installers from the PSReleaseDrop folder, and other xml files used for WebPI testing.

This script will also create and push a new branch to the given CodePlex WebPI fork containing the changes to the two xml files.

---

Below in **Overall Changes**, check off the boxes that apply to your PR. For the categories that you did not check off, you can remove them from this body. Within each of the categories that you did select, make sure that you can check off **all** of the boxes. 

For information on cleaning up the commits in your pull request, click [here](../documentation/cleaning-up-commits.md).
## Overall Changes
### General
- [x] If it applies, references the bug/issue that the PR fixes
### Tests

### DIFF
--- a/tools/BuildDrop.ps1
+++ b/tools/BuildDrop.ps1
@@ -116,12 +116,18 @@ for ($idx = 0; $idx -lt $content.Length; $idx++)
 
         # Change the buffer size to include the three lines just added
         $buffer = 3
+
+        # Flag that we are no longer in the VSFeed entry
+        $VSFeedSeen = $False
     }
 
     # Check if we are looking at the entry for DH_WindowsAzurePowerShellGet
     if ($PSGetSeen -and $content[$idx] -like "*msiProductCode*")
     {
         $content[$idx] = "   <msiProductCode>$ProductCode</msiProductCode>"
+
+        # Flag that we are no longer in the PSGet entry
+        $PSGetSeen = $False
     }
 
     $newContent[$idx + $buffer] = $content[$idx]
@@ -171,6 +177,11 @@ for ($idx = 0; $idx -lt $content.Length; $idx++)
         if ($content[$idx] -like "*<trackingURL>*")
         {
             $content[$idx] = "        <trackingURL>http://www.microsoft.com/web/handlers/webpi.ashx?command=incrementproddownloadcount&amp;prodid=WindowsAzurePowershell&amp;version=$PSVersion&amp;prodlang=en</trackingURL>"
+        }
+
+        if ($content[$idx] -like "*</entry>*")
+        {
+            $PSGetSeen = $False
         }
     }
 


tools/BuildDrop.ps1
```diff
@@ -0,0 +1,240 @@
...
+    # Check if we are looking at the DiscoveryHints for DH_WindowsAzurePowerShellVSFeed
+    # and if we have reached the end of the entry so we can add the new msi Product Code
+    if ($VSFeedSeen -and $content[$idx] -like "*</or>*")
+    {
+        $newContent[$idx] =     "      <discoveryHint>"
+        $newContent[$idx + 1] = "        <msiProductCode>$ProductCode</msiProductCode>"
+        $newContent[$idx + 2] = "      </discoveryHint>"        
+
+        # Change the buffer size to include the three lines just added
+        $buffer = 3
```
markcowl: Should we change $VSFeedSeen to false at this point?

tools/BuildDrop.ps1
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    # Check if we are looking at the entry for DH_WindowsAzurePowerShellGet
+    if ($PSGetSeen -and $content[$idx] -like "*msiProductCode*")
+    {
+        $content[$idx] = "   <msiProductCode>$ProductCode</msiProductCode>"
+    }
```
markcowl: Shoudl we change $PSGetSeen to false at this point?

tools/BuildDrop.ps1
```diff
@@ -0,0 +1,240 @@
...
+        {
+            $content[$idx] = "  <updated>$($ReleaseDate)T12:00:00Z</updated>"
+        }
+
+        if ($content[$idx] -like "*<trackingURL>*")
+        {
+            $content[$idx] = "        <trackingURL>http://www.microsoft.com/web/handlers/webpi.ashx?command=incrementproddownloadcount&amp;prodid=WindowsAzurePowershell&amp;version=$PSVersion&amp;prodlang=en</trackingURL>"
+        }
+    }
+
```
markcowl: Should have a condition here which ends the $PSGetSeen, to make this more robust over changed orderings in the file

tools/BuildDrop.ps1
cormacpayne: I thought about it, but didn't account for the case where the ordering of the file could change. That is a good idea. I will mark as false here.

tools/BuildDrop.ps1
cormacpayne: Same as comment for $VSFeedSeen.

tools/BuildDrop.ps1
cormacpayne: Same as comment for $VSFeedSeen.

PR #2957, opened 2016-09-17
Support paginated response in GET-AzureRmMlWebService

NonStatic2014:
Paginated response will happen when there are more than 1000 items returned back from the server. This is a corner case for current Azure Machine Learning users. So this change is for future-proof.

### DIFF
--- a/src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/GetAzureMLWebService.cs
+++ b/src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/GetAzureMLWebService.cs
@@ -17,6 +17,7 @@ using System.Collections.Generic;
 using System.Management.Automation;
 using System.Threading.Tasks;
 using Microsoft.Azure.Management.MachineLearning.WebServices.Models;
+using Microsoft.Azure.Commands.ResourceManager.Common;
 
 namespace Microsoft.Azure.Commands.MachineLearning
 {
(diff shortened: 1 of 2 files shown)


src/ResourceManager/MachineLearning/Commands.MachineLearning/Cmdlets/GetAzureMLWebService.cs
```diff
@@ -48,25 +49,27 @@ protected override void RunCmdlet()
...
-                                                        this.CancellationToken).Result;
-                }
-                else
-                {
-                    services = this.WebServicesClient.GetAzureMlWebServicesBySubscriptionAsync(
-                                                        null,
-                                                        this.CancellationToken).Result;
+                    // This is a call for resource retrieval within a resource group
+                    getFirstServicesPageCall = () => this.WebServicesClient.GetAzureMlWebServicesBySubscriptionAndGroupAsync(this.ResourceGroupName, null, this.CancellationToken);
+                    getNextPageCall = nextLink => this.WebServicesClient.GetAzureMlWebServicesBySubscriptionAndGroupAsync(this.ResourceGroupName, nextLink, this.CancellationToken);
```
cormacpayne: @NonStatic2014 it looks like if `this.ResourceGroupName` isn't null, then these two methods will be called twice (once before the if-statement, and once inside. Is there a reason for doing this instead of doing the following:

```cs
if (this.ResourceGroupName != null)
{
     Func<Task<ResponseWithContinuation<WebService[]>>> getFirstServicesPageCall = () => this.WebServicesClient.GetAzureMlWebServicesBySubscriptionAsync(this.ResourceGroupName, null, this.CancellationToken);
     Func<string, Task<ResponseWithContinuation<WebService[]>>> getNextPageCall = nextLink => this.WebServicesClient.GetAzureMlWebServicesBySubscriptionAsync(this.ResourceGroupName, nextLink, this.CancellationToken);
}
else
{
     Func<Task<ResponseWithContinuation<WebService[]>>> getFirstServicesPageCall = () => this.WebServicesClient.GetAzureMlWebServicesBySubscriptionAsync(null, this.CancellationToken);
     Func<string, Task<ResponseWithContinuation<WebService[]>>> getNextPageCall = nextLink => this.WebServicesClient.GetAzureMlWebServicesBySubscriptionAsync(nextLink, this.CancellationToken);
}
```

NonStatic2014: I think it is the same thing, just an overwrite if this.ResourceGroupName is not null. But your comment reminds me, I will update the if-statement to !string.IsNullOrWhitespace(this.ResourceGroupName)

NonStatic2014: I think it is the same thing, just an overwrite if this.ResourceGroupName is not null.

cormacpayne:
LGTM

PR #2991, opened 2016-09-22
Integrating new SDK .Net generated from Swagger spec

gucalder:
## Comments

---

Below in **Overall Changes**, check off the boxes that apply to your PR. For the categories that you did not check off, you can remove them from this body. Within each of the categories that you did select, make sure that you can check off **all** of the boxes. 

For information on cleaning up the commits in your pull request, click [here](../documentation/cleaning-up-commits.md).
## Overall Changes
- [x] [Add/remove/edit cmdlet(s)](#cmdlet-signature)
- [x] [Add/remove/edit parameter(s)](#parameters)
### General
- [ ] Title of the PR is clear and informative
- [ ] The PR does not introduce breaking changes (unless a major version change occurs in the assembly and module)
### Tests
### Cmdlet Signature
- [x] Cmdlet name uses an approved PowerShell verb - use the enums for `VerbsCommon`, `VerbsCommunication`, `VerbsLifecycle`, `VerbsOther` whenever possible
- [x] Cmdlet noun name uses the AzureRm prefix for management cmdlets, and the Azure prefix for data plane cmdlets
- [ ] Cmdlet specifies the `OutputType` attribute if any output is produced; if the cmdlet produces no output, it should implement a `PassThrough` parameter
- [ ] If the cmdlet makes changes or has side effects, it should implement `ShouldProcess` and have `SupportShouldProcess = true` specified in the cmdlet attribute. See a discussion about correct `ShouldProcess` implementation [here](https://gist.github.com/markcowl/338e16fe5c8bbf195aff9f8af0db585d#what-is-the-change)
- [x] Cmdlets should derive from AzureRmCmdlet for management cmdlets, and AzureDataCmdlet for data cmdlets
- [ ] If multiple parameter sets are implemented, the cmdlet should specify a `DefaultParameterSetName` in its cmdlet attribute
### Parameters
- [ ] Cmdlets should have no more than four positional parameters
- [x] Cmdlet parameter sets should be mutually exclusive - each parameter set must have at least one mandatory parameter not in other parameter sets
- [x] Parameter types should not expose types from the management library - complex parameter types should be defined in the module
- [x] Complex parameter types are discouraged - a parameter type should be simple types as often as possible. If complex types are used, they should be shallow and easily creatable from a constructor or another cmdlet
- [x] Parameters should be explicitly marked as Mandatory or not, and should contain a HelpMessage
- [x] No parameter is of type `object`.
    - Tag, type `HashTable`

### DIFF
--- a/src/ResourceManager/Insights/Commands.Insights/Alerts/RemoveAzureRmAlertRuleCommand.cs
+++ b/src/ResourceManager/Insights/Commands.Insights/Alerts/RemoveAzureRmAlertRuleCommand.cs
@@ -50,7 +50,7 @@ namespace Microsoft.Azure.Commands.Insights.Alerts
         /// </summary>
         protected override void ProcessRecordInternal()
         {
-            WriteWarning("The output of this cmdlet has changed and will change again in the future. The current request Id is empty. In the future the cmdlet will not return anything.");
+            WriteWarning("The output of this cmdlet will change. The cmdlet will not return anything in future releases.");
             var result = this.InsightsManagementClient.AlertRules.DeleteWithHttpMessagesAsync(resourceGroupName: this.ResourceGroup, ruleName: this.Name).Result;
 
             // Keep this response for backwards compatibility.
--- a/src/ResourceManager/Insights/Commands.Insights/Autoscale/RemoveAzureRmAutoscaleSettingCommand.cs
+++ b/src/ResourceManager/Insights/Commands.Insights/Autoscale/RemoveAzureRmAutoscaleSettingCommand.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Azure.Commands.Insights.Autoscale
         /// </summary>
         protected override void ProcessRecordInternal()
         {
-            WriteWarning("The output of this cmdlet has changed and will change again in the future. The current request Id is empty. In the future the cmdlet will not return anything.");
+            WriteWarning("The output of this cmdlet will change. The cmdlet will not return anything in future releases.");
             var result = this.InsightsManagementClient.AutoscaleSettings.DeleteWithHttpMessagesAsync(resourceGroupName: this.ResourceGroup, autoscaleSettingName: this.Name).Result;
 
             // Keep this response for backwards compatibility.
--- a/src/ResourceManager/Insights/Commands.Insights/Alerts/AddAzureRmAlertRuleCommandBase.cs
+++ b/src/ResourceManager/Insights/Commands.Insights/Alerts/AddAzureRmAlertRuleCommandBase.cs
@@ -77,7 +77,7 @@ namespace Microsoft.Azure.Commands.Insights.Alerts
         /// </summary>
         protected override void ProcessRecordInternal()
         {
-            WriteWarning("This output of this cmdlet will change in the next release to return the newly created object");
+            WriteWarning("This output of this cmdlet will change in the next release to return the updated or newly created object");
             AlertRuleResource parameters = this.CreateSdkCallParameters();
 
             // Part of the result of this operation is operation (result.Body ==> a AutoscaleSettingResource) is being discarded for backwards compatibility
--- a/src/ResourceManager/Insights/Commands.Insights/LogProfiles/RemoveAzureRmLogProfileCommand.cs
+++ b/src/ResourceManager/Insights/Commands.Insights/LogProfiles/RemoveAzureRmLogProfileCommand.cs
@@ -39,7 +39,7 @@ namespace Microsoft.Azure.Commands.Insights.LogProfiles
 
         protected override void ProcessRecordInternal()
         {
-            WriteWarning("The output of this cmdlet will change to be empty in the next release.");
+            WriteWarning("The output of this cmdlet will change. The cmdlet will not return anything in future releases.");
             this.InsightsManagementClient.LogProfiles.DeleteAsync(logProfileName: this.Name, cancellationToken: CancellationToken.None).Wait();
             WriteObject(true);
         }


cormacpayne:
On-demand: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1223/

src/ResourceManager/Insights/Commands.Insights/Alerts/RemoveAzureRmAlertRuleCommand.cs
```diff
@@ -15,13 +15,14 @@
...
 using System.Management.Automation;
+using Microsoft.Azure.Management.Insights.Models;
 
 namespace Microsoft.Azure.Commands.Insights.Alerts
 {
     /// <summary>
     /// Remove an Alert rule
     /// </summary>
-    [Cmdlet(VerbsCommon.Remove, "AzureRmAlertRule"), OutputType(typeof(List<AzureOperationResponse>))]
+    [Cmdlet(VerbsCommon.Remove, "AzureRmAlertRule"), OutputType(typeof(PSObject))]
```
cormacpayne: @gucalder this is a breaking change; if I had a script that did the following
```powershell
$foo = Remove-AzureRmAlertRule -ResourceGroup $ResourceGroup -Name $Name
```
and I tried to do anything with `$foo`, it will no longer work

src/ResourceManager/Insights/Commands.Insights/Alerts/RemoveAzureRmAlertRuleCommand.cs
```diff
@@ -49,8 +50,8 @@ public class RemoveAzureRmAlertRuleCommand : ManagementCmdletBase
         /// </summary>
         protected override void ProcessRecordInternal()
         {
-            AzureOperationResponse result = this.InsightsManagementClient.AlertOperations.DeleteRuleAsync(resourceGroupName: this.ResourceGroup, ruleName: this.Name).Result;
-            WriteObject(result);
+            this.InsightsManagementClient.AlertRules.DeleteAsync(resourceGroupName: this.ResourceGroup, ruleName: this.Name).Wait();
+            WriteObject("");
```
cormacpayne: @gucalder why was the output type of the cmdlet changed so that it could return nothing?

src/ResourceManager/Insights/Commands.Insights/Autoscale/AddAzureRmAutoscaleSettingCommand.cs
```diff
@@ -27,7 +27,7 @@ namespace Microsoft.Azure.Commands.Insights.Autoscale
     /// <summary>
     /// Create or update an Autoscale setting
     /// </summary>
-    [Cmdlet(VerbsCommon.Add, "AzureRmAutoscaleSetting"), OutputType(typeof(AzureOperationResponse))]
+    [Cmdlet(VerbsCommon.Add, "AzureRmAutoscaleSetting"), OutputType(typeof(PSAutoscaleSetting))]
```
cormacpayne: @gucalder same comment about output type breaking change

markcowl:
@gucalder this is too big a change to accept after code complete.  This will be targeted at the next release.

src/ResourceManager/Insights/Commands.Insights/Autoscale/NewAzureRmAutoscaleRuleCommand.cs
```diff
@@ -94,12 +94,6 @@ public class NewAzureRmAutoscaleRuleCommand : AzureRMCmdlet
         public ScaleDirection ScaleActionDirection { get; set; }
 
         /// <summary>
-        /// Gets or sets the scale action type
-        /// </summary>
-        [Parameter(Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The scale action type for the setting")]
-        public ScaleType ScaleActionScaleType { get; set; }
```
cormacpayne: @gucalder this is a breaking change; you are removing a **mandatory** parameter from this cmdlet, so any script that uses this cmdlet will break

src/ResourceManager/Insights/Commands.Insights/Autoscale/RemoveAzureRmAutoscaleSettingCommand.cs
```diff
@@ -14,13 +14,14 @@
...
 using System.Management.Automation;
+using Microsoft.Azure.Management.Insights.Models;
 
 namespace Microsoft.Azure.Commands.Insights.Autoscale
 {
     /// <summary>
     /// Remove an autoscale setting.
     /// </summary>
-    [Cmdlet(VerbsCommon.Remove, "AzureRmAutoscaleSetting"), OutputType(typeof(AzureOperationResponse))]
+    [Cmdlet(VerbsCommon.Remove, "AzureRmAutoscaleSetting"), OutputType(typeof(PSObject))]
```
cormacpayne: @gucalder same comment about output type breaking change

src/ResourceManager/Insights/Commands.Insights/Diagnostics/SetAzureRmDiagnosticSettingCommand.cs
```diff
@@ -21,13 +23,14 @@
...
 using System.Xml;
+using Microsoft.Rest.Azure;
 
 namespace Microsoft.Azure.Commands.Insights.Diagnostics
 {
     /// <summary>
     /// Get the list of events for at a subscription level.
     /// </summary>
-    [Cmdlet(VerbsCommon.Set, "AzureRmDiagnosticSetting"), OutputType(typeof(PSServiceDiagnosticSettings))]
+    [Cmdlet(VerbsCommon.Set, "AzureRmDiagnosticSetting"), OutputType(typeof(PSObject))]
```
cormacpayne: @gucalder same comment about output type breaking change

src/ResourceManager/Insights/Commands.Insights/LogProfiles/AddAzureRmLogProfileCommand.cs
```diff
@@ -24,7 +26,7 @@ namespace Microsoft.Azure.Commands.Insights.LogProfiles
     /// <summary>
     /// Get the log profiles.
     /// </summary>
-    [Cmdlet(VerbsCommon.Add, "AzureRmLogProfile"), OutputType(typeof(PSLogProfile))]
+    [Cmdlet(VerbsCommon.Add, "AzureRmLogProfile"), OutputType(typeof(PSObject))]
```
cormacpayne: @gucalder same comment about output type breaking change

src/ResourceManager/Insights/Commands.Insights/Metrics/GetAzureRmMetricCommand.cs
```diff
@@ -37,33 +44,35 @@ public class GetAzureRmMetricCommand : InsightsClientCmdletBase
...
         public string ResourceId { get; set; }
 
         /// <summary>
-        /// Gets or sets the timegrain parameter of the cmdlet
+        /// Gets or sets the aggregation type parameter of the cmdlet
         /// </summary>
-        [Parameter(Position = 1, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The time grain of the query")]
+        [Parameter(ParameterSetName = GetAzureRmAMetricFullParamGroup, Position = 1, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The aggregation type of the query")]
         [ValidateNotNullOrEmpty]
-        public TimeSpan TimeGrain { get; set; }
```
cormacpayne: @gucalder this is a breaking change; removing a **mandatory** parameter will break any script that uses this cmdlet

src/ResourceManager/Insights/Commands.Insights/Autoscale/RemoveAzureRmAutoscaleSettingCommand.cs
gucalder: The new spec from Swagger is not returning anything for Delete operations. The Hydra spec (old one) used to return an object with request Id and status code.

src/ResourceManager/Insights/Commands.Insights/Alerts/RemoveAzureRmAlertRuleCommand.cs
gucalder: I'll return List<Microsoft.Azure.AzureOperationResponse> with a singe element.
The element contains: status = OK, and an empty request Id.
This is since the new rules for the specs are that remove operations return nothing. This is still sort of a breaking change I know.
Returning a list is a bug btw, but I'll keep it like that for backwards compatibility.

src/ResourceManager/Insights/Commands.Insights/Alerts/RemoveAzureRmAlertRuleCommand.cs
gucalder: When I was writing the Swagger spec, I was told that "the rule" for delete operations is to return nothing.
Previously, the Hydra spec was returning the request Id and HttpStatus.

src/ResourceManager/Insights/Commands.Insights/Autoscale/AddAzureRmAutoscaleSettingCommand.cs
gucalder: Changed back to the previous type for backwards compatibility.
In the future we could just return the new resource and favor pipe lining since the new SDk .Net actually returns the new resource.

src/ResourceManager/Insights/Commands.Insights/Autoscale/NewAzureRmAutoscaleRuleCommand.cs
gucalder: That is another consequence of the new spec. We currently support only one type of auto scale action (ChangeCount). The spec team suggested to remove the parameter and just send the constant value through the cable only.
I just added the parameter back, made it optional and made default to the only value supported ChangeCount. I am not using the parameter at all, but I'll keep it for backwards compatibility.

src/ResourceManager/Insights/Commands.Insights/Autoscale/RemoveAzureRmAutoscaleSettingCommand.cs
gucalder: All Remove operations will return the same type of object they returned before.

src/ResourceManager/Insights/Commands.Insights/Diagnostics/SetAzureRmDiagnosticSettingCommand.cs
gucalder: My bad, returning type back to its original.

src/ResourceManager/Insights/Commands.Insights/LogProfiles/AddAzureRmLogProfileCommand.cs
gucalder: My bad, return type back to its previous value.

src/ResourceManager/Insights/Commands.Insights/Metrics/GetAzureRmMetricCommand.cs
gucalder: Metric is a special case.  The whole API is new. I cannot "not break it" in this case.
What would be the procedure to introduce breaking changes?

gucalder:
The last commit fails for reasons that are not in my control:

Microsoft.Azure.Commands.Sql.Test.ScenarioTests.ElasticPoolCrudTests.TestElasticPoolMetricGet [FAIL] [D:\workspace\powershell\build.proj]

All our tests have been enabled and are running and passing. None is being skipped.

16:40:24      Microsoft.Azure.Commands.Insights.Test                          Total:   56, Errors: 0, Failed: 0, Skipped:  0, Time:   279.402s

cormacpayne:
@gucalder that test is failing because it contains the following line:

```powershell
$metrics = $ep1 | Get-AzureRmMetric -TimeGrain "0:5:0" -StartTime "2015-04-22T16:00:00Z" -EndTime "2015-04-22T17:00:00Z"
```

and because your changes removed the `TimeGrain` parameter, this command fails

gucalder:
Ok, I am going to fix that issue with the time grain from the Swagger spec.
Apparently, the API changed from what I used to create the Swagger spec and time grain is there as an optional parameter. So the TimeGrain parameter will be back in the cmd let soon.

gucalder:
BTW: In the test "Test-GetElasticPoolMetric" the call to Get-AzureRmMetric was probably taking its MetricNames parameter from the pipe before. I do not see any explicit parameter in the call to Get-AzureRmMetric, but the recorded session contains a list of metric names in the requestUri. 
If the test was passing, the result of the Get-AzureRmSqlElasticPool cmdlet must have contained a property called MetricNames or the matching to the recorded session was very loose. Otherwise the test would have failed to find a matching record.

I am going to add the explicit list of metric names in the call and try again.

gucalder:
I talked to the Sql team member in charge of the tests and agreed with him that we can temporarily comment the section of the test that was failing.
So I did that and the all the tests are successful now.
I also added warning messages to the users about changes in the future like the output types.

gucalder:
Any updates on this?

src/ResourceManager/Insights/Commands.Insights/Alerts/RemoveAzureRmAlertRuleCommand.cs
markcowl: You should not use ```WriteObject ("")``` - this pushes a string onto the output stream.  The policy is that you must signal breaking changes to users a release or more in advance using a deprecation warning.

markcowl:
@cormacpayne @gucalder since this is a large sql change, and many of your tests are in the on demand run, we need the on demand run to pass for this as well:

http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1241/

gucalder:
That on demand looks green.

src/ResourceManager/Insights/Commands.Insights/Autoscale/NewAzureRmAutoscaleRuleCommand.cs
markcowl: If you add a ValidateSet attribute for the parameter, then this will prevent setting it to other values as well.

src/ResourceManager/Insights/Commands.Insights/Metrics/GetAzureRmMetricCommand.cs
markcowl: The procedure is to add a deprecation warning for the parameter / cmdlet you are going to deprecate a few releases in advance of the breaking change, then make the breaking change in a subsequent release.

src/ResourceManager/Insights/Commands.Insights/Autoscale/NewAzureRmAutoscaleProfileCommand.cs
```diff
@@ -105,14 +105,14 @@ public class NewAzureRmAutoscaleProfileCommand : AzureRMCmdlet
         /// </summary>
         [Parameter(ParameterSetName = AddAutoscaleProfileRecurrenceParamGroup, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The schedule hours for the recurrence")]
         [ValidateNotNull]
-        public List<int> ScheduleHours { get; set; }
+        public List<int?> ScheduleHours { get; set; }
```
shahabhijeet: >?> [](start = 23, length = 2)

Any reason this is nullable, when you have NotNull Validator?

src/ResourceManager/Insights/Commands.Insights/Autoscale/NewAzureRmAutoscaleProfileCommand.cs
```diff
@@ -105,14 +105,14 @@ public class NewAzureRmAutoscaleProfileCommand : AzureRMCmdlet
...
 
         /// <summary>
         /// Gets or sets the ScheduleMinutes
         /// </summary>
         [Parameter(ParameterSetName = AddAutoscaleProfileRecurrenceParamGroup, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The schedule minutes for the recurrence")]
         [ValidateNotNull]
-        public List<int> ScheduleMinutes { get; set; }
+        public List<int?> ScheduleMinutes { get; set; }
```
shahabhijeet: >?> [](start = 23, length = 2)

Nullable type after NotNull validator

shahabhijeet:
@gucalder This is a breaking change and from what we discussed in office hours, we need to give enough time to our customers so you can introduce breaking changes. So at this stage, I would suggest, you abandon this PR, as it is getting hard to understand what is new code and which one is reverted code to avoid breaking change.
I would suggest:
1) Please add deprecation warnings for your breaking changes that will be coming next release 
2) Have clear documentation of upcoming breaking changes in the release notes/changelog

PR #3008, opened 2016-09-23
Remove aggregation by Computer

lwang2016:
This pull request is to remove verification on the query field format in NewAzureOperationalInsightsComputerGroupCommand.

The verification is removed because of a recent new feature in our service backend. Previously, a computer group query must end with either "measure count() by Computer" or "distinct Computer". We recently added a new feature that allows customers to specify a computer simply by a list of comma-separated values, such as "C1, C2, C3". With the change, the original verification is no longer needed and thus removed in this change.

### DIFF
--- a/src/ResourceManager/OperationalInsights/Commands.OperationalInsights/Search/NewAzureOperationalInsightsComputerGroupCommand.cs
+++ b/src/ResourceManager/OperationalInsights/Commands.OperationalInsights/Search/NewAzureOperationalInsightsComputerGroupCommand.cs
@@ -73,11 +73,6 @@ namespace Microsoft.Azure.Commands.OperationalInsights
                 Tags = new List<Tag>() { new Tag() { Name = "Group", Value = "Computer" } }
             };
 
-            if (!SearchCommandHelper.IsListOfComputers(this.Query))
-            {
-                throw new PSArgumentException("Query is not a list of computers. Please use aggregations such as: distinct Computer or measure count() by Computer.");
-            }
-
             WriteObject(OperationalInsightsClient.CreateOrUpdateSavedSearch(ResourceGroupName, WorkspaceName, SavedSearchId, properties, Force, ConfirmAction), true);
         }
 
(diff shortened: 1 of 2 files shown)


azurecla:
Hi __@lwang2016__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (lewang). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

shahabhijeet:
Please clean up the description area and provide what is this PR is about. Please use the template by editing it, so that it can help to review your PR.

src/ResourceManager/OperationalInsights/Commands.OperationalInsights/Search/NewAzureOperationalInsightsComputerGroupCommand.cs
```diff
@@ -73,11 +73,6 @@ protected override void ProcessRecord()
                 Tags = new List<Tag>() { new Tag() { Name = "Group", Value = "Computer" } }
             };
 
-            if (!SearchCommandHelper.IsListOfComputers(this.Query))
```
lwang2016: The verification is removed because of a recent new feature in our service backend. Previously, a computer group query must end with either "measure count() by Computer" or "distinct Computer". We recently added a new feature that allows customers to specify a computer simply by a list of comma-separated values, such as "C1, C2, C3". With the change, the original verification is no longer needed and thus removed in this change.

lwang2016:
Updated description.

src/ResourceManager/OperationalInsights/Commands.OperationalInsights/Search/NewAzureOperationalInsightsComputerGroupCommand.cs
shahabhijeet: I don't see any test coverage for the new query format? Also is there any tests that make sure the old query will still work. Also can a user combine the new format with the old? If yes would be great to add that in the test coverage.

lwang2016: IMHO, I don't think extra test coverage is necessary in this change. What does the cmdlet do is to save the computer group Query to a table in OperationalInsights. The cmdlet doesn't care the format inside the query as long as it is saved to the table succesfully, which existing test cases in azure-powershell already covered.

There are test coverage on OperationalInsights side to make sure both the new and old query syntaxes work as expected. So the query functionalities are also covered.

shahabhijeet:
LGTM

PR #3014, opened 2016-09-26
Adding documentation for new test framework. Consolidating few powers…

shahabhijeet:
1. Adding documentation for new test framework. 
2. Consolidating few PowerShell scripts under repo-tasks.psm1, now you can
   Set-TestEnvironment ==> Will set connection string for running tests
   Get-BuildScopes ==> Will get list of scopes available that can be used to do build for specific scopes
   Start-RepoBuild ==> will either do a full build or when specified scope, will only build that particular scope
   Invoke-CheckinTests ==> run checkin tests
   Invoke-MockedScenarioTests ==> Will build, and run Mocked Scenario tests
   Install-VSTestProjectTemplates ==> Will install VS Test Project templates on the user machine
3. Adding shortcut to start VS command prompt combined with PowerShell (PS-VSPrompt.lnk)
4. Also adding a way to load user functions using user PowerShell scripts (this only affects when user uses PS-VSPrompt.lnk)
5. Issue [https://github.com/Azure/azure-powershell/issues/3039]


azurecla:
Hi __@shahabhijeet__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (abhishah). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

tools/Tasks.ps1
```diff
@@ -0,0 +1,200 @@
...
+                        "Run-MockedScenarioTests ==> Run Mocked Scenario Tests", `
+                        "Get-TaskHelp ==> Display supported tasks"
+    
+    #$tableFormat = @{Expression={$taskCount};width=50}, @{Expression={$_}};
+    #$tasks | Format-Table $tableFormat
+    $tasks |foreach { Write-Host ([string]::Format("{0}) {1}", $taskCount++, $_))}
+    Write-Host ""
+}
+
+Function Setup-TestEnvironment()
```
markcowl: If we define this using a cmdlet binding, we cna let powershell prompt for parmaters they don't provide.  Let's discuss.  Let's make this an actual script module, which will enable us to use the powershell help facility, powershell parameter checks, enforsing different parameter set mandatory parameters, etc.  I want to do this alowly with all of the scripts we are using int he tools folder.

tools/Tasks.ps1
```diff
@@ -0,0 +1,200 @@
...
+
+    #$appUri = [string]::Format("https://{0}/{1}", $tenantName, $appName)
+
+    #$app = AzureRmADApplication -DisplayName $appName -HomePage $appUri -IdentifierUris $appUri -Password $pwd
+
+    #New-AzureRmADServicePrincipal -ApplicationId $app.ApplicationId
+    #New-AzureRmRoleAssignment -RoleDefinitionName Reader -ServicePrincipalName $app.ApplicationId.Guid
+}
+
+Function Print-ConnectionString([string]$subId, [string]$aadTenant, [string]$spn, [string]$spnSecret, [string]$recordMode, [string]$uris)
```
markcowl: Should we have it actually set the connection string too?  Thinking that setting the connection string and launching VS is what most people are goint to want to do.  This functionbality for priting the connection string is definitely necessary, however.

shahabhijeet: It does set the connection string by executing $env:TEST_CSM_ORGID_AUTHENTICATION

tools/Tasks.ps1
shahabhijeet: Converted to a powershell module and got rid of all of the Read-Host calls for prompts.

shahabhijeet:
Splitting into three separate PRs. Closing this one.

PR #3020, opened 2016-09-27
[Release 3.0.0] Fix error when publishing modules to gallery

cormacpayne:
NuGet throws an error when trying to publish modules if there is a space in any of the tags. This PR fixes a tag in the AzureRM.Storage module that contains a space.

Deleting PR template since this is a fix for the release process.


markcowl:
LGTM once the build passes

PR #3050, opened 2016-10-06
Update Batch Management Lib - rename/reorganize files to better adhere to ARM patterns

jasper-schneider:
## Overall Changes
  Updated to latest version of the Batch management lib, which included some renames of the RP operation classes and methods to better adhere to ARM patterns.  
  The Subscription directory was renamed to Location to more accurately describe the scope.
  The Accounts directory was renamed to BatchAccounts to match the resource name.
  The application package cmdlets were moved into their own directory, but names/parameters remained unchanged.
  Renamed the Subscription directory to Location
  Renamed the Accounts directory to BatchAccounts
  Moved application package tests into their own folder.
- [x] [Add/remove/edit cmdlet(s)](#cmdlet-signature)
  Renamed the Get-AzureRmBatchSubscriptionQuotas cmdlet into Get-AzureRmBatchLocationQuotas

### DIFF
--- a/src/ResourceManager/AzureBatch/Commands.Batch/Locations/GetBatchLocationQuotasCommand.cs
+++ b/src/ResourceManager/AzureBatch/Commands.Batch/Locations/GetBatchLocationQuotasCommand.cs
@@ -19,6 +19,7 @@ using Constants = Microsoft.Azure.Commands.Batch.Utils.Constants;
 namespace Microsoft.Azure.Commands.Batch
 {
     [Cmdlet(VerbsCommon.Get, Constants.AzureRmBatchLocationQuotas), OutputType(typeof(PSBatchLocationQuotas))]
+    // This alias was added in 10/2016 for backwards compatibility
     [Alias("Get-AzureRmBatchSubscriptionQuotas")]
     public class GetBatchLocationQuotasCommand : BatchCmdletBase
     {


src/ResourceManager/AzureBatch/Commands.Batch/Locations/GetBatchLocationQuotasCommand.cs
```diff
@@ -18,8 +18,8 @@
 
 namespace Microsoft.Azure.Commands.Batch
 {
-    [Cmdlet(VerbsCommon.Get, Constants.AzureRmBatchSubscriptionQuotas), OutputType(typeof(PSBatchSubscriptionQuotas))]
-    public class GetBatchSubscriptionQuotasCommand : BatchCmdletBase
+    [Cmdlet(VerbsCommon.Get, Constants.AzureRmBatchLocationQuotas), OutputType(typeof(PSBatchLocationQuotas))]
```
cormacpayne: @jasper-schneider this is a breaking change; you are changing the name of a cmdlet without adding an alias to the original name. Existing scripts that use the cmdlet `Get-AzureRmBatchSubscriptionQuotas` will no longer work.

Also, the output type has changed, which is another breaking change.

jasper-schneider: @cormacpayne 
Yes, the management lib had breaking changes, which results in breaking changes here.
I'll add an alias for the old name, but can you elaborate on what you're looking for with respect to the output type?

markcowl: @jasper-schneider The policy is to warn the user of impending breaking changes at least a sprint before making the breaking change.  The advice is to mitigate breaks with custom powershell code, not to remove properties from ourput types or from existing parameter sets.

jasper-schneider: @markcowl, thanks for the info.  I apologize though, because I'm still kind of fuzzy on what exactly you're asking me to do.  Can you elaborate a bit on what the ask is here?
The output type has changed names, but no properties or parameter sets have been altered.  Scripts that use the cmdlet (with the alias added) will not be broken.  This will still work:
$quotas = Get-AzureRmBatchSubscriptionQuotas "westus";
Write-Host "My account quota in westus is $quotas.AccountQuota";

There's no cmdlet that accepts this object as an input parameter, so users won't be instantiating their own.  Quota changes are currently handled via support ticket only, so this API is only about giving users info on what their current quota value is.  There's no quota-related APIs that would use this object.

Feel free to ping me if it would be easier.  Thanks.

PR #3052, opened 2016-10-06
Consolidating basic command repository tasks into a script module. Ad…

shahabhijeet:
Consolidating few PowerShell scripts under repo-tasks.psm1, now you can
Set-TestEnvironment ==> Will set connection string for running tests
Get-BuildScopes ==> Will get list of scopes available that can be used to do build for specific scopes
Start-RepoBuild ==> will either do a full build or when specified scope, will only build that particular scope
Invoke-CheckinTests ==> run checkin tests
Invoke-MockedScenarioTests ==> Will build, and run Mocked Scenario tests
Adding shortcut to start VS command prompt combined with PowerShell (PS-VSPrompt.lnk)
Also adding a way to load user functions using user PowerShell scripts (this only affects when user uses PS-VSPrompt.lnk)

### DIFF
--- a/documentation/Repo-Tasks-Module.md
+++ b/documentation/Repo-Tasks-Module.md
@@ -1,0 +1,23 @@
+# Repo-Tasks Module #
+
+###### Usage:
+
+1. Start PS-VSPrompt.lnk (shortcut), this will start VS Dev Prompt in powershell
+2. Import-Module ./Repo-Tasks.psm1
+	1. During import, we allow to load additional functions that users might want to use it in their session.
+	2. If you have any userPreference.ps1 file under %userprofile%/psFiles directory, the module will try to load it by dot sourcing it.
+	2. It will also honor environment variable $env:psuserpreferences and load .ps1 files from the location that is pointed by $env:psuserpreferences
+	3. As long as you have exported all the functions that you need from your ps1 file using export-modulemember -function <name of function>. We deliberately do this to avoid polluting list of commands available (when you use Get-Command)
+3. Currently Repo-Tasks module supports following tasks:
+	1. Set-TestEnvironment
+		1. Will allow you create a test connection string required to setup test environment in order to run tests. More information about Test environment can be found [here](https://github.com/Azure/azure-powershell/blob/dev/documentation/Using-Azure-TestFramework.md "here")
+	2. Start-Build
+		1. Will allow you to kick off full build
+		2. Or will allow you build for a particular scope (e.g. Start-Build -BuildScope ResourceManagment\Compute)
+	3. Get-BuildScopes
+		1. Will allow you to query and find existing build scopes that can be used to build.
+	4. Invoke-CheckinTests
+		1. Will build and run existing tests.
+
+###Note:
+If you do not start your powershell session using PS-VSPrompt shortcut, you will not have access to all the environment variables that are set as part of VS Dev Command prompt.


azurecla:
Hi __@shahabhijeet__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (abhishah). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

tools/Repo-Tasks.psm1
```diff
@@ -0,0 +1,253 @@
...
+In order to successfully run a test, you will need SubscriptionId, TenantId
+This cmdlet will only prompt you for Subscription and Tenant information, rest all other parameters are optional
+
+#>
+    param(
+        [parameter(Mandatory=$true, Position=0, HelpMessage='SubscriptionId you would like to use')]
+        [ValidateNotNullOrEmpty()]
+        [string]$SubscriptionId,
+
+        [parameter(Mandatory=$true, Position=1, HelpMessage='AADTenant/TenantId you would like to use:')]
```
markcowl: Here are the parameter sets:

(1) Service Principal With Secret (TenantId = required (Position 1), PSCredential ServicePrincipalCredentials = required, [Position = 2])
(2) UserID [default parameter set] PSCredential UserCredential = optional

The following parameters belong to all parameter sets:
SubscriptionId=required
Environment=optional ValidateSet=Prod| Dogfood| Current | Next, Default=Prod)
RecordMode=optional, ValidateSet=(Record, None, Playback, default = Playback)

tools/Repo-Tasks.psm1
```diff
@@ -0,0 +1,253 @@
...
+#>
+
+    Write-Host "Below are available scopes you can specify for building specific projects"
+    Write-Host ""    
+    Get-ChildItem -path "$env:repoRoot\src\ResourceManager" -dir | Format-Wide -Column 5 | Format-Table -Property Name    
+    Get-ChildItem -path "$env:repoRoot\src\ServiceManagement" -dir | Format-Wide -Column 5 | Format-Table -Property Name
+}
+
+[CmdletBinding]
+Function Start-RepoBuild
```
markcowl: Why not just Start-Build?

tools/Repo-Tasks.psm1
```diff
@@ -0,0 +1,253 @@
...
+{
+<#
+.SYNOPSIS
+Runs all the check in tests
+#>
+    Write-Host "cmdline Args: msbuild.exe $env:repoRoot\build.proj /t:Test"
+    msbuild.exe "$env:repoRoot\build.proj" /t:Test
+}
+
+[cmdletBinding]
```
markcowl: Let's add this with the project templates

tools/Repo-Tasks.psm1
```diff
@@ -0,0 +1,253 @@
...
+            Write-Host "Missing templates to install, make sure you have project templates available in the repo under $env:repoRoot\tools\ProjectTemplates\"
+        }
+    }
+    else
+    {
+        Write-Host "Unsupported VS Version detected. Visual Studio 2015 is the only supported version for current set of project templates"
+    }
+}
+
+Function Create-ServicePrincipal-Help()
```
markcowl: Let's remove until this actually creates a service principal.  Once it does, we will need to add a module manifest, as we will have some dependencies.

tools/Repo-Tasks.psm1
```diff
@@ -0,0 +1,253 @@
...
+    $env:TEST_CSM_ORGID_AUTHENTICATION=$formattedConnStr
+    Write-Host ""
+
+    # Retrieve the environment variable
+    Write-Host "Setting up below connection string. Start Visual Studio by typing devenv" -ForegroundColor Green
+    [Environment]::GetEnvironmentVariable($envVariableName)
+    Write-Host ""
+    
+    Write-Host "If your needs demand you to set connection string differently, for all the supported Key/Value pairs in connection string"
+    Write-Host "Please visit https://github.com/Azure/azure-powershell/blob/dev/documentation/Using-Azure-TestFramework.md"
```
markcowl: Nice

tools/Repo-Tasks.psm1
```diff
@@ -0,0 +1,253 @@
...
+
+    #$appUri = [string]::Format("https://{0}/{1}", $tenantName, $appName)
+
+    #$app = AzureRmADApplication -DisplayName $appName -HomePage $appUri -IdentifierUris $appUri -Password $pwd
+
+    #New-AzureRmADServicePrincipal -ApplicationId $app.ApplicationId
+    #New-AzureRmRoleAssignment -RoleDefinitionName Reader -ServicePrincipalName $app.ApplicationId.Guid
+}
+
+Function Print-ConnectionString([string]$subId, [string]$aadTenant, [string]$spn, [string]$spnSecret, [string]$recordMode, [string]$targetEnvironment, [string]$uris)
```
markcowl: Doesn't looks like this is used anywhere, or exported, please remove

tools/Repo-Tasks.psm1
kilasuit: Other modules are likely to use Start-Build so I would suggest a rename to Start-AzureRMBuild

shahabhijeet: @kilasuit Start-AzureRMBuild might be misleading as we also have non-RM sources in this repo.
I like @markcowl suggestion of Start-Build. Will change the name to Start-Build

tools/Repo-Tasks.psm1
```diff
@@ -0,0 +1,245 @@
...
+        [string]$TenantId,
+
+        [ValidateSet("Playback", "Record", "None")]
+        [string]$RecordMode='Playback',
+
+        [ValidateSet("Prod", "Dogfood", "Current", "Next")]
+        [string]$TargetEnvironment='Prod'
+    )
+
+    [string]$uris="https://management.azure.com/;AADAuthEndpoint=https://login.windows.net/;GraphUri=https://graph.windows.net/"
```
markcowl: Only the BaseUri is needed in this case, the rest of these (AADAuthENdpoint, GraphUri) will override whatever values are in Environment, so we should remove these

tools/Repo-Tasks.psm1
```diff
@@ -0,0 +1,245 @@
...
+#>
+    param(
+    [parameter(Mandatory=$false, Position=0, HelpMessage='BuildScope that you would like to use. For list of build scopes, run List-BuildScopes')]
+    [string]$BuildScope
+    )    
+    
+    if([string]::IsNullOrEmpty($BuildScope) -eq $true)
+    {
+       Write-Host "Starting Full build"
+       msbuild.exe "$env:repoRoot\build.proj" /t:Build
```
markcowl: Same comment:  We need to make sure we are using the correct version of MSBuild, as in here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1243/ We should expand this script to include later versions of VS, and ensure the correct environment is used

tools/Repo-Tasks.psm1
```diff
@@ -0,0 +1,245 @@
...
+    Write-Host ""
+    Write-Host "Below connection string was set. Start Visual Studio by typing devenv" -ForegroundColor Green
+    [Environment]::GetEnvironmentVariable($envVariableName)
+    Write-Host ""
+    
+    Write-Host "If your needs demand you to set connection string differently, for all the supported Key/Value pairs in connection string"
+    Write-Host "Please visit https://github.com/Azure/azure-powershell/blob/dev/documentation/Using-Azure-TestFramework.md" -ForegroundColor Yellow
+}
+
+Function Print-ConnectionString([string]$uid, [string]$pwd, [string]$subId, [string]$aadTenant, [string]$spn, [string]$spnSecret, [string]$recordMode, [string]$targetEnvironment, [string]$uris)
```
markcowl: I thought you were removing this function

tools/Repo-Tasks.psm1
markcowl: But you are just referring to msbuild in the path.  We want to make sure we get the version of MS build from the Program Files directory, and not the version in the .Net 4.x directory.  I don't see anywhere where twe are controllign the path where the msbuild exe comes from.

tools/Repo-Tasks.psm1
shahabhijeet: I actually liked your earlier suggestion that, this will be an helpful function for users to look at their connection string.

PR #3076, opened 2016-10-13
Run PS Smoke tests with service principal

shahabhijeet:
This will allow to run our smoke tests to run without manual login.
This change enables to 
1) login in AzureRm using ServicePrincipal
2) login in ASM using publishsettings files
3) Stores all credentials in azure keyvault

Workflow
1) First time you run this tests, checks if certificate is installed
2) If not asks you to log in interactively so that Azure KeyVault can be accessed
3) Get's all the credentials/certs that are required to login
4) Logs in AzureRm and ASM
5) Once the test is done, removes subscription as well as any files that were downloaded.

This script has also the capability of create ServicePrincipal as well as new certificates if expired.
The logic to detect expired certificate is not implemented, but all the code that is required to
1) Create new certificates and store in Azure KeyVault
2) Create new SPN
3) Upload new PublishSettings file (if required)

### DIFF
--- a/documentation/RunSmokeTests.md
+++ b/documentation/RunSmokeTests.md
@@ -1,0 +1,40 @@
+# Running Tests with Auto Login #
+
+This script will allow to login as long as you have required certificate installed on your machine
+
+## WorkFlow #1 ##
+
+1. Will Check if the certificate is installed
+2. If not will ask you to login (one time) to access KeyVault
+3. Download Certificate, install certificate on the machine (will prompt to set certificate password)
+4. Login using the recently installed certificate
+5. Set the context for Test Subscription
+6. Download PublishSettings file from keyVault to enable log into ASM
+7. Run Tests
+8. Clean up - Delete any locally downloaded certificates or publishsettings file
+
+## WorkFlow #2 ##
+
+1. Check if certificate is installed
+2. If Yes, log in using certificate
+3. Download publishsettings file from KeyVault
+4. Log into ASM using publish settings file
+5. Run Tests
+6. Clean up
+
+
+The script that kick starts running test is RunInstallationTests.ps1
+This script takes two parameters
+
+1. uninstallLocalCert
+	1. $true: will uninstalled existing certificate
+	2. $false: will not uninstall existing certificate
+
+1. runOnCIMachine
+	1. $true:
+		1. this mode will allow to fail fast, especially if it detects local certificate is not installed and needs manual login, it will fail
+		2. this will also set the DebugPreference = "Continue", this allows you to get more verbose output of the various cmdlets that are executed as part of the test
+	2. $false: will check if the local certificate is installed, if not manually will prompt user to log in and install the certificate from KeyVault
+
+
+
--- a/tools/InstallationTests/RunInstallationTests.ps1
+++ b/tools/InstallationTests/RunInstallationTests.ps1
@@ -106,8 +106,6 @@ Run-TestProtected { Test-SetAzureStorageBlobContent } "Test-SetAzureStorageBlobC
 
 Run-TestProtected { Test-UpdateStorageAccount } "Test-UpdateStorageAccount"
 
-Exit
-
 $serviceCommands | % { Run-TestProtected $_  $_.ToString() }
 Write-Host -ForegroundColor Green "STARTING RESOURCE MANAGER TESTS"
 


tools/InstallationTests/Assert.ps1
```diff
@@ -1,4 +1,4 @@
-﻿# ----------------------------------------------------------------------------------
```
markcowl: Please revert this change

tools/InstallationTests/AutoTestLogin.ps1
```diff
@@ -0,0 +1,490 @@
...
+    if($cert -eq $null)
+    {
+        Log-Info "Trying to find certificate in LocalMachine"
+        $cert = Get-ChildItem 'Cert:\LocalMachine\My' | Where-Object {$_.Subject -eq $Global:gLocalCertSubjectName}
+    }
+
+    return $cert
+}
+
+Function Get-ADAppForAutoTest()
```
markcowl: This should be a cmdlet, so you can implement SHouldProcess/ShouldContinue

tools/InstallationTests/AutoTestLogin.ps1
```diff
@@ -0,0 +1,490 @@
...
+        $certPwd = Get-LocalCertificatePassword
+        $secCertPwd = ConvertTo-SecureString -String $certPwd -AsPlainText -Force
+        $pfxLocalFilePath = [System.IO.Path]::Combine($global:localPfxDirPath, $global:gPfxLocalFileName)
+        $localCert = New-Object System.Security.Cryptography.X509Certificates.X509Certificate($pfxLocalFilePath, $secCertPwd)        
+
+        $localCert.GetEffectiveDateString()
+        $localCert.GetExpirationDateString()
+        #>
+
+        $kvCertObj = Get-AzureKeyVaultCertificate -VaultName $global:gVaultName -Name $global:kvSecKey_PsAutoTestCertNameKey
```
markcowl: This should be protected with ShouldProcess/ShouldContinue

tools/InstallationTests/AutoTestLogin.ps1
markcowl: Also, wondering about the name, this should be something like Ensure, rather than get.  It should return the ADApp, so that you don't have to retrieve it again later

tools/InstallationTests/AutoTestLogin.ps1
```diff
@@ -0,0 +1,490 @@
...
+    $installedCert = Get-LocalCertificate
+    if($installedCert -eq $null)
+    {
+        throw [System.ApplicationException] "Unable to retrieve installed certificate for running automation test"
+    }
+    else
+    {
+        Log-Info "Successfully installed certificate to '$global:gLocalCertStore'"
+    }
+}
```
markcowl: Remove the certificate (pfx file) when done

tools/InstallationTests/AutoTestLogin.ps1
```diff
@@ -0,0 +1,490 @@
...
+
+Function Login-AzureRMWithCertificate()
+{
+    $appId = $global:gPsAutoTestADAppId
+
+    $localCert = Get-LocalCertificate
+    if($localCert -eq $null)
+    {
+        #we could not find certificate installed locally, get it from KeyVault
+        Login-InteractivelyAndSelectTestSubscription
```
markcowl: This should be protected by SHouldContinue, so it will just fail if run as part of a ccript.  This means the function should eb a cmdlet

tools/InstallationTests/AutoTestLogin.ps1
```diff
@@ -0,0 +1,490 @@
...
+    $global:gLoggedInCtx = Select-AzureRmSubscription -SubscriptionId $global:gPsAutoTestSubscriptionId
+
+    return $global:gLoggedInCtx
+}
+
+Function Login-Azure([bool]$deleteLocalCertificate=$false)
+{
+    Remove-AllSubscriptions
+
+    Delete-LocalCertificate $deleteLocalCertificate
```
markcowl: This is kind of odd, why not use if ($deleteLoclCertificate)?

tools/InstallationTests/AutoTestLogin.ps1
```diff
@@ -0,0 +1,490 @@
...
+
+    Login-AzureRMWithCertificate
+    Select-AzureRmSubscription -SubscriptionId $global:gPsAutoTestSubscriptionId -TenantId $global:gTenantId
+
+    Login-AzureWithCertificate
+    Select-AzureSubscription -SubscriptionId $global:gPsAutoTestSubscriptionId -Current
+}
+
+<#
+Function Test-CleanUp()
```
markcowl: Remove commented code

tools/InstallationTests/AutoTestLogin.ps1
```diff
@@ -0,0 +1,490 @@
...
+        $certCollection2 = New-Object System.Security.Cryptography.X509Certificates.X509Certificate2Collection
+        $certCollection2.Import($kvSecretBytes, $null, [System.Security.Cryptography.X509Certificates.X509KeyStorageFlags]::Exportable)
+        
+        #Save Pfx locally
+        $pwd = Get-LocalCertificatePassword
+        $certPwdProtectedInBytes = $certCollection2.Export([System.Security.Cryptography.X509Certificates.X509ContentType]::Pkcs12, $pwd)
+        $pfxLocalFilePath = [System.IO.Path]::Combine($global:localPfxDirPath, $global:gPfxLocalFileName)
+        [System.IO.File]::WriteAllBytes($pfxLocalFilePath, $certPwdProtectedInBytes)
+
+        #We are not going the .cer route for now, rather use publishSettings file
```
markcowl: Remove commented code

tools/InstallationTests/AutoTestLogin.ps1
```diff
@@ -0,0 +1,490 @@
...
+# limitations under the License.
+# ----------------------------------------------------------------------------------
+
+This script will allow to login as long as you have required certificate installed on your machine
+
+WorkFlow# 1
+===========
+1) Will Check if the certificate is installed
+2) If not will ask you to login (one time) to access KeyVault
+3) Download Certificate, install certificate on the machine (will prompt to set certificate password)
```
markcowl: Any prompts should be ShouldContinue prompts, so that they will fail when this is being run as part of script execution.  The idea is to allow extra bootstrapping as a convenience to manual users, but allowing automation to fail rapidly if the machine is not properly set up.

Also, just to be clear.  A properly set up machine should require no downloads from KeyVault, I think

tools/InstallationTests/AutoTestLogin.ps1
```diff
@@ -0,0 +1,490 @@
...
+    {
+        $certPath = $cert.PSPath
+        Log-Info "Deleting local certificate $certPath"
+        Remove-Item $cert.PSPath
+    }
+}
+
+Function CleanUpSPN()
+{
+    $psAutoADApp = Get-AzureRmADApplication -DisplayNameStartWith $global:gPsAutoTestADAppUsingSPNKey
```
markcowl: Why do we remove the application?

tools/InstallationTests/AutoTestLogin.ps1
```diff
@@ -0,0 +1,490 @@
...
+}
+
+Function Get-KVSecretValue([string] $kvSecretName)
+{
+    return Get-AzureKeyVaultSecret -VaultName $global:gVaultName -Name $kvSecretName
+}
+
+Function Delete-TestCertificateFromKv([string] $vaultName, [string] $certName)
+{
+    Remove-AzureKeyVaultCertificate -VaultName $vaultName -Name $certName
```
markcowl: ??

tools/InstallationTests/Common.ps1
```diff
@@ -193,14 +193,28 @@ function Test-Setup
 #############################
 function Test-Cleanup
```
markcowl: Please ensure that these are the latest versions from the PowerShell share.  The files here are quite out of sync with the scripts we have been executing.

tools/InstallationTests/AutoTestLogin.ps1
shahabhijeet: ShouldContinue will be implemented when we decide to make the entire login script to a module. Currently added a switch that will allow to fail fast on CI machine.

tools/InstallationTests/AutoTestLogin.ps1
shahabhijeet: Extracted one time machine setup function to a separate psm1 file which will eventually a script module.

tools/InstallationTests/AutoTestLogin.ps1
shahabhijeet: Added Try/Finally to remove all downloaded resources

tools/InstallationTests/AutoTestLogin.ps1
shahabhijeet: Fixed the logic

tools/InstallationTests/AutoTestLogin.ps1
shahabhijeet: Moved code to a separate file.

tools/InstallationTests/Common.ps1
```diff
@@ -193,14 +193,33 @@ function Test-Setup
...
+     #$global:VerbosePreference = $global:oldVerbosePreference
+     #$global:WarningPreference = $global:oldWarningPreference
+     #$global:WhatIfPreference = $global:oldWhatIfPreference
+
+    Delete-DownloadedCertAndPubSetting
+    
+    Remove-AllSubscriptions
+}
+
+Function Delete-DownloadedCertAndPubSetting
```
markcowl: Seems like this belongs in the same file as the code that downloads the cert and publishsettings

shahabhijeet: Moved to the same that creates these files.

PR #3099, opened 2016-10-19
Using new Swagger spec for the SDK, avoiding breaking changes, and adding deprecation messages for future changes

gucalder:
## Description

---

This checklist is used to make sure that common guidelines for a pull request are followed. You can find a more complete discussion of PowerShell cmdlet best practices [here](https://msdn.microsoft.com/en-us/library/dd878270%28v=vs.85%29.aspx).
- [x ] **I have read the [contribution guidelines](https://github.com/Azure/azure-powershell/blob/dev/CONTRIBUTING.md).**
- [ ] Cmdlet specifies `OutputType` attribute if any output is produced - if the cmdlet produces no output, it should implement a `PassThrough` parameter.

### DIFF
--- a/src/ResourceManager/Insights/Commands.Insights/Commands.Insights.csproj
+++ b/src/ResourceManager/Insights/Commands.Insights/Commands.Insights.csproj
@@ -41,15 +41,6 @@
     <DelaySign>true</DelaySign>
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
-  <PropertyGroup>
-    <SignAssembly>false</SignAssembly>
-  </PropertyGroup>
-  <PropertyGroup>
-    <AssemblyOriginatorKeyFile>MSSharedLibKey.snk</AssemblyOriginatorKeyFile>
-  </PropertyGroup>
-  <PropertyGroup>
-    <DelaySign>false</DelaySign>
-  </PropertyGroup>
   <ItemGroup>
     <Reference Include="Hyak.Common">
       <SpecificVersion>False</SpecificVersion>
--- a/src/ResourceManager/Sql/Commands.Sql.Test/ScenarioTests/ElasticPoolCrudTests.ps1
+++ b/src/ResourceManager/Sql/Commands.Sql.Test/ScenarioTests/ElasticPoolCrudTests.ps1
@@ -171,10 +171,9 @@ function Test-GetElasticPoolMetric
     Assert-NotNull $ep1
     
     # Get pool metrics with all values
-    # **** NOTE: this section of the test has been commented out by gucalder because the current Get-AzureRmEmtric has changed, it requires different parameters and the recorded session is not longer compatible with it.
-    # $metrics = $ep1 | Get-AzureRmMetric -TimeGrain "0:5:0" -StartTime "2015-04-22T16:00:00Z" -EndTime "2015-04-22T17:00:00Z"
-    # Assert-NotNull $metrics
-    # Assert-True { $metrics.Count -gt 0 }
+    $metrics = $ep1 | Get-AzureRmMetric -TimeGrain "0:5:0" -StartTime "2015-04-22T16:00:00Z" -EndTime "2015-04-22T17:00:00Z"
+    Assert-NotNull $metrics
+    Assert-True { $metrics.Count -gt 0 }
 }
 
 


src/ResourceManager/Insights/Commands.Insights/Metrics/GetAzureRmMetricCommand.cs
```diff
@@ -127,16 +129,22 @@ protected string ProcessParameters()
...
         protected override void ProcessRecordInternal()
         {
+            WriteWarning("API deprecation: The use of the legacy metrics API will be discontinued in the next release. This implies a change in the call and the output of this cmdlet.");
+
             string queryFilter = this.ProcessParameters();
             bool fullDetails = this.DetailedOutput.IsPresent;
 
-            // Call the proper API methods to return a list of raw records.
-            // If fullDetails is present full details of the records displayed, otherwise only a summary of the values is displayed
-            MetricListResponse response = this.InsightsClient.MetricOperations.GetMetricsAsync(resourceUri: this.ResourceId, filterString: queryFilter, cancellationToken: CancellationToken.None).Result;
```
shahabhijeet: @gucalder seems the output will change looking at this code change.
Is this breaking change?

gucalder: No, response is still a MetricListResponse.

Here I am consuming the old SDK code, that I just added to the project, instead of the code in the new SDK library.
I had to add the using block around the call to create an old client and be able to call the exact same method I used to call in the previous (SDK).
This change is only to avoid a breaking change. The call and the output remain identical between the previous and this release.

gucalder: No, response is still a MetricListResponse.

Here I am consuming the old SDK code, that I just added to the project, instead of the code in the new SDK library.
I had to add the using block around the call to create an old client and be able to call the exact same method I used to call in the previous (SDK).
This change is only to avoid a breaking change. The call and the output remain identical between the previous and this release.

src/ResourceManager/Insights/Commands.Insights/Commands.Insights.csproj
```diff
@@ -41,23 +41,29 @@
     <DelaySign>true</DelaySign>
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
+  <PropertyGroup>
```
markcowl: Why is this section added?  If you want seoparate settings for the debug build,  they should be placed inside a conditional, so they don't apply to debug and release.

src/ResourceManager/Insights/Commands.Insights/OnetimeGenerated/Insights/MetricDefinitionOperations.cs
```diff
@@ -0,0 +1,505 @@
+//
```
markcowl: Why are these files in a different folder than the rest of the geernerated client operations

src/ResourceManager/Sql/Commands.Sql.Test/ScenarioTests/ElasticPoolCrudTests.ps1
```diff
@@ -171,9 +171,10 @@ function Test-GetElasticPoolMetric
     Assert-NotNull $ep1
     
     # Get pool metrics with all values
-    $metrics = $ep1 | Get-AzureRmMetric -TimeGrain "0:5:0" -StartTime "2015-04-22T16:00:00Z" -EndTime "2015-04-22T17:00:00Z"
-    Assert-NotNull $metrics
-    Assert-True { $metrics.Count -gt 0 }
+    # **** NOTE: this section of the test has been commented out by gucalder because the current Get-AzureRmEmtric has changed, it requires different parameters and the recorded session is not longer compatible with it.
+    # $metrics = $ep1 | Get-AzureRmMetric -TimeGrain "0:5:0" -StartTime "2015-04-22T16:00:00Z" -EndTime "2015-04-22T17:00:00Z"
```
markcowl: ?? Please verify that the parameter set used here is still valiud.  The only valid reason to comment this out would be for a different wire request/response pattern, due to the change in management library

src/ResourceManager/Insights/Commands.Insights/Commands.Insights.csproj
gucalder: Section removed.

src/ResourceManager/Insights/Commands.Insights/OnetimeGenerated/Insights/MetricDefinitionOperations.cs
gucalder: There were several folders for metrics in the previous SDK intended to modify and/or extend the metrics functionality beyond what we were able to specify in Hydra.
What I did here was just to copy all files and folders related to mertics into this project so that I can use them to keep metrics working as before. I kept the structure of the files and folders untouched since this is a temporary change.

src/ResourceManager/Sql/Commands.Sql.Test/ScenarioTests/ElasticPoolCrudTests.ps1
gucalder: Test section re-enabled. Sorry, I forgot to do this when I made Metric look like before.

markcowl:
on demand run here:  http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1249/

LGTM once the build passes

Issue #3026, opened 2016-09-28
Add PowerShell Version to user agent string for powershell cmdlets


markcowl:
### Description

Add PowerShell Version to user agent string for powershell cmdlets
- [x] Add the PowerShell version to the UserAgent additions in ClientFactory
- [x] Add unit test for the existence of the UserAgent in created clients

shahabhijeet:
PR https://github.com/Azure/azure-powershell/pull/3154

PR #3122, opened 2016-10-25
Adding PSversion to userAgent for every powershell cmdlet

shahabhijeet:
Adding PsVersion to userAgent
Adding test for userAgent
Fixing file misspell

### DIFF
--- a/src/Common/Commands.Common/AzurePSCmdlet.cs
+++ b/src/Common/Commands.Common/AzurePSCmdlet.cs
@@ -36,6 +36,7 @@ namespace Microsoft.WindowsAzure.Commands.Utilities.Common
     public abstract class AzurePSCmdlet : PSCmdlet, IDisposable
     {
         private const string PSVERSION = "PSVersion";
+        private const string DEFAULT_PSVERSION = "3.0.0.0";
 
         public ConcurrentQueue<string> DebugMessages { get; private set; }
 
@@ -79,8 +80,8 @@ namespace Microsoft.WindowsAzure.Commands.Utilities.Common
                     }
                     else
                     {
-                        System.Management.Automation.PowerShell ps = System.Management.Automation.PowerShell.Create();
-                        _psVersion = ps.Runspace.Version.ToString();
+                        //We are doing this for perf. reasons. This code will execute during tests and so reducing the perf. overhead while running tests.
+                        _psVersion = DEFAULT_PSVERSION;
                     }
                 }
 


src/Common/Commands.Common/AzurePSCmdlet.cs
```diff
@@ -243,9 +267,10 @@ protected virtual void TearDownDebuggingTraces()
 
         protected virtual void SetupHttpClientPipeline()
         {
-            ProductInfoHeaderValue userAgentValue = new ProductInfoHeaderValue(
-                ModuleName, string.Format("v{0}", ModuleVersion));
-            AzureSession.ClientFactory.UserAgents.Add(userAgentValue);
+            //We do not want to add a prefix 'v' to version, because when data will be read out of the telemetry, it will again have to be parsed to extract the actualy version
+            AzureSession.ClientFactory.UserAgents.Add(new ProductInfoHeaderValue(ModuleName, ModuleVersion));
```
markcowl: This is removing the 'v' from the version value

src/Common/Commands.Common/AzurePSCmdlet.cs
```diff
@@ -58,6 +60,28 @@ protected virtual bool IsErrorMetricEnabled
...
+        /// Indicates installed PowerShell version
+        /// </summary>
+        private string _psVersion;
+
+        /// <summary>
+        /// Get PsVersion returned from PowerShell.Runspace instance
+        /// </summary>
+        protected string PSVersion
+        {
+            get
```
markcowl: This value is available from Cmdlet.CommandRuntime.Host.Version.  You will need to make sure that Host is not null.

shahabhijeet:
Address issue
https://github.com/Azure/azure-powershell/issues/3026

src/Common/Commands.Common/AzurePSCmdlet.cs
```diff
@@ -73,8 +73,15 @@ protected string PSVersion
...
                 {   
-                    System.Management.Automation.PowerShell ps = System.Management.Automation.PowerShell.Create();
-                    _psVersion = ps.Runspace.Version.ToString();
+                    if(this.Host != null)
+                    {
+                        _psVersion = this.Host.Version.ToString();
+                    }
+                    else
+                    {
+                        System.Management.Automation.PowerShell ps = System.Management.Automation.PowerShell.Create();
```
markcowl: I would say we can return a default value in this case.  This should only occur in tests

shahabhijeet:
PS OnDemand
http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1271/

markcowl:
closing in favor of the new PR

PR #3162, opened 2016-11-01
Need to check for null when creating Tags.

athipp:
If a server comesback with null for the Tags object. then the Get comand will fail creating a dictionary object with a null reference.

Fix is to check for null and use a different constructor instead.

Same for sku

### DIFF
--- a/src/ResourceManager/AnalysisServices/Commands.AnalysisServices/Models/AzureAnalysisServicesServer.cs
+++ b/src/ResourceManager/AnalysisServices/Commands.AnalysisServices/Models/AzureAnalysisServicesServer.cs
@@ -39,6 +39,8 @@ namespace Microsoft.Azure.Commands.AnalysisServices.Models
 
         public string Location { get; set; }
 
+        public string ServerFullName { get; private set; }
+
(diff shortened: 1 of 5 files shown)


azurecla:
Hi __@athipp__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (athipp). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

cormacpayne:
@athipp please re-target this PR to the dev branch

src/ResourceManager/AnalysisServices/Commands.AnalysisServices/Models/AzureAnalysisServicesServer.cs
```diff
@@ -52,16 +52,16 @@ internal static AzureAnalysisServicesServer FromAnalysisServicesServer(AnalysisS
```
markcowl: @athipp please add test coverage, as we discussed

PR #3207, opened 2016-11-18
Breaking change tool for Azure PowerShell

cormacpayne:
## Description

Added tool to Static Analysis tool that will detect breaking changes.

The list of breaking changes handled by this tool can be found [here](https://github.com/Azure/azure-powershell/blob/dev/documentation/breaking-changes/breaking-changes-definition.md).

---

- [x] Cmdlet specifies `OutputType` attribute if any output is produced - if the cmdlet produces no output, it should implement a `PassThrough` parameter.

### DIFF
--- a/tools/StaticAnalysis/SerializedCmdlets/Microsoft.Azure.Commands.Compute.dll.json
+++ b/tools/StaticAnalysis/SerializedCmdlets/Microsoft.Azure.Commands.Compute.dll.json
@@ -6520,8 +6520,14 @@
             "Name": "AutoBackupSettings",
             "AssemblyQualifiedName": "Microsoft.Azure.Commands.Compute.AutoBackupSettings, Microsoft.Azure.Commands.Compute, Version=2.4.0.0, Culture=neutral, PublicKeyToken=null",
             "Properties": {
+              "BackupScheduleType": "System.String",
+              "BackupSystemDbs": "System.Nullable`1[System.Boolean]",
               "Enable": "System.Boolean",
               "EnableEncryption": "System.Boolean",
+              "FullBackupFrequency": "System.String",
+              "FullBackupStartTime": "System.Nullable`1[System.Int32]",
+              "FullBackupWindowHours": "System.Nullable`1[System.Int32]",
+              "LogBackupFrequency": "System.Nullable`1[System.Int32]",
               "Password": "System.String",
               "RetentionPeriod": "System.Int32",
               "StorageAccessKey": "System.String",
@@ -6658,6 +6664,102 @@
           },
           "ValidateSet": [],
           "ValidateNotNullOrEmpty": true
+        },
+        {
+          "Name": "BackupSystemDbs",
+          "AliasList": [],
+          "Type": {
+            "Namespace": "System.Management.Automation",
+            "Name": "SwitchParameter",
+            "AssemblyQualifiedName": "System.Management.Automation.SwitchParameter, System.Management.Automation, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35",
+            "Properties": {},
+            "ElementType": null,
+            "GenericTypeArguments": []
+          },
+          "ValidateSet": [],
+          "ValidateNotNullOrEmpty": false
+        },
+        {
+          "Name": "BackupScheduleType",
+          "AliasList": [],
+          "Type": {
+            "Namespace": "System",
+            "Name": "String",
+            "AssemblyQualifiedName": "System.String, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089",
+            "Properties": {},
+            "ElementType": null,
+            "GenericTypeArguments": []
+          },
+          "ValidateSet": [
+            "Manual",
+            "Automated"
+          ],
+          "ValidateNotNullOrEmpty": false
+        },
+        {
+          "Name": "FullBackupFrequency",
+          "AliasList": [],
+          "Type": {
+            "Namespace": "System",
+            "Name": "String",
+            "AssemblyQualifiedName": "System.String, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089",
+            "Properties": {},
+            "ElementType": null,
+            "GenericTypeArguments": []
+          },
+          "ValidateSet": [
+            "Daily",
+            "Weekly"
+          ],
+          "ValidateNotNullOrEmpty": false
+        },
+        {
+          "Name": "FullBackupStartHour",
+          "AliasList": [],
+          "Type": {
+            "Namespace": "System",
+            "Name": "Nullable",
+            "AssemblyQualifiedName": "System.Nullable`1[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]], mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089",
+            "Properties": {},
+            "ElementType": null,
+            "GenericTypeArguments": [
+              "System.Int32"
+            ]
+          },
+          "ValidateSet": [],
+          "ValidateNotNullOrEmpty": false
+        },
+        {
+          "Name": "FullBackupWindowInHours",
+          "AliasList": [],
+          "Type": {
+            "Namespace": "System",
+            "Name": "Nullable",
+            "AssemblyQualifiedName": "System.Nullable`1[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]], mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089",
+            "Properties": {},
+            "ElementType": null,
+            "GenericTypeArguments": [
+              "System.Int32"
+            ]
+          },
+          "ValidateSet": [],
+          "ValidateNotNullOrEmpty": false
+        },
+        {
+          "Name": "LogBackupFrequencyInMinutes",
+          "AliasList": [],
+          "Type": {
+            "Namespace": "System",
+            "Name": "Nullable",
+            "AssemblyQualifiedName": "System.Nullable`1[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]], mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089",
+            "Properties": {},
+            "ElementType": null,
+            "GenericTypeArguments": [
+              "System.Int32"
+            ]
+          },
+          "ValidateSet": [],
+          "ValidateNotNullOrEmpty": false
         }
       ],
       "ParameterSets": [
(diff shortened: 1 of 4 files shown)


tools/StaticAnalysis/BreakingChangeAnalyzer/BreakingChangeAnalyzer.cs
```diff
@@ -0,0 +1,968 @@
...
+            // For each of the parameters in the old parameter set assembly, find the parameter
+            // in the new parameter set assembly that either has the same name, or an alias to
+            // the original name
+            foreach (var oldParameter in oldParameterSet.Parameters)
+            {
+                bool found = false;
+
+                foreach (var newParameter in newParameterSet.Parameters)
+                {
+                    if (oldParameter.ParameterMetadata.Name.Equals(newParameter.ParameterMetadata.Name) ||
```
markcowl: I think we should be explicit about comparison, we need to use StringComparison.OrdinalIgnoreCase when comparing strings.

tools/StaticAnalysis/BreakingChangeAnalyzer/BreakingChangeAnalyzer.cs
```diff
@@ -0,0 +1,968 @@
...
+            // assembly, and if so, do the necessary breaking change checks
+            foreach (var oldCmdlet in oldCmdlets)
+            {
+                // If the cmdlet is in the new assembly, carry out the breaking change checks
+                if (cmdletMap.ContainsKey(oldCmdlet.Name))
+                {
+                    // Get the cmdlet metadata from the new assembly
+                    var newCmdlet = cmdletMap[oldCmdlet.Name];
+
+                    CheckForRemovedCmdletAlias(oldCmdlet, newCmdlet, issueLogger);
```
markcowl: This feels like too many methods in one class and that the methods aretoo large.  I would consider refactoring the methods so that they are extensions over the metadata types.  So you could have CmdletMetadata.CheckForBreakingChanges, which took the old version as an input.  This leaves the metdata types as poco types, but allows you to access their private data, and encapsulates the logic that knows about parameters in the parameter metadata

tools/StaticAnalysis/BreakingChangeAnalyzer/BreakingChangeAnalyzer.cs
```diff
@@ -0,0 +1,968 @@
...
+            // output in the new cmdlet assembly
+            foreach (var oldOutput in oldCmdlet.OutputTypes)
+            {
+                bool found = false;
+
+                foreach (var newOutput in newCmdlet.OutputTypes)
+                {
+                    // If the corresponding output has been found in the new assembly,
+                    // look for through all of the properties
+                    if (oldOutput.Type.Name.Equals(newOutput.Type.Name))
```
markcowl: Hmm, so I think this would be a little smoother and easier to follow  if:

(1) You had a dictionary of output types, keyed by output type name.  You could then just look for the key in the dictionary 
(2) If there is no match, use Linq to cimplify your code - if you have a simple predocate that does recursive property patching, you can just call Any<t>(fiulter) on the list of values to find if any outpout types match
(3) If you put the matching function into the output types themselves, it will separate concers a little better and clean up the look of this code.

tools/StaticAnalysis/BreakingChangeAnalyzer/BreakingChangeAnalyzer.cs
```diff
@@ -0,0 +1,968 @@
...
+            {
+                return;
+            }
+
+            ParameterSetMetadata oldDefaultParameterSet = null;
+            ParameterSetMetadata newDefaultParameterSet = null;
+
+            // Get the ParameterSetMetadata object corresponding to the
+            // default parameter set for the old assembly
+            foreach (var oldParameterSet in oldCmdlet.ParameterSets)
```
markcowl: Use linq instead to get the default parameter set using FirstOrDefault<T>(filterFuncrion)

tools/StaticAnalysis/BreakingChangeAnalyzer/BreakingChangeAnalyzer.cs
```diff
@@ -0,0 +1,968 @@
...
+                if (oldParameterSet.Name.Equals(oldCmdlet.DefaultParameterSetName))
+                {
+                    oldDefaultParameterSet = oldParameterSet;
+                    break;
+                }
+            }
+
+            // Get the ParameterSetMetadata object corresponding to the
+            // default parameter set for the new assembly
+            foreach (var newParameterSet in newCmdlet.ParameterSets)
```
markcowl: same comment

tools/StaticAnalysis/BreakingChangeAnalyzer/BreakingChangeAnalyzer.cs
```diff
@@ -0,0 +1,968 @@
...
+        /// <param name="issueLogger"></param>
+        private void CompareTypes(
+            CmdletBreakingChangeMetadata cmdlet,
+            TypeMetadata oldType,
+            TypeMetadata newType,
+            ReportLogger<BreakingChangeIssue> issueLogger)
+        {
+            // For each property in the old assembly type, find the corresponding
+            // property in the new assembly type
+            foreach (var oldProperty in oldType.Properties.Keys)
```
markcowl: what happens with primitive types?

cormacpayne:
@markcowl fixed issue with build failing due to generics, just need to address review comments and other items we talked about previously.

tools/StaticAnalysis/BreakingChangeAnalyzer/BreakingChangeAnalyzer.cs
cormacpayne: Added helper class for each metadata type that knows how to compare the corresponding metadata and log appropriate issues.

tools/StaticAnalysis/BreakingChangeAnalyzer/BreakingChangeAnalyzer.cs
```diff
@@ -0,0 +1,968 @@
...
+        }
+
+        /// <summary>
+        /// Serialize the cmdlets so they can be compared to change modules later
+        /// </summary>
+        /// <param name="fileName"></param>
+        /// <param name="cmdlets"></param>
+        private void SerializeCmdlets(string fileName, List<CmdletBreakingChangeMetadata> cmdlets)
+        {
+            string json = JsonConvert.SerializeObject(cmdlets, Formatting.Indented);
```
cormacpayne: The dictionary of type metadata is now serialized.

tools/StaticAnalysis/BreakingChangeAnalyzer/BreakingChangeAnalyzer.cs
cormacpayne: Used dictionary to simplify code

tools/StaticAnalysis/BreakingChangeAnalyzer/BreakingChangeAnalyzer.cs
cormacpayne: Used dictionary to simplify code

tools/StaticAnalysis/BreakingChangeAnalyzer/BreakingChangeAnalyzer.cs
cormacpayne: Used dictionary to simplify code

tools/StaticAnalysis/BreakingChangeAnalyzer/BreakingChangeAnalyzer.cs
cormacpayne: In `ModuleMetadata`, the type dictionary is initialized with all of the primitive types, so they will never be passed through as arguments to this method

cormacpayne:
On-demand: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1315/

markcowl:
on demand run here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1317/

cormacpayne:
On demand: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1322/

tools/StaticAnalysis/BreakingChangeAnalyzer/TypeMetadata.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// Allow easy conversion between types and type metadata objects.
+        /// </summary>
+        /// <param name="typeToProcess"></param>
+        /// <returns>The capture type metadata</returns>
+        public static implicit operator TypeMetadata(Type typeToProcess)
+        {
+            return new TypeMetadata(typeToProcess);
+        }
+
+        public TypeMetadata()
```
markcowl: empty default constructor is implied - no need to explicitly code it

tools/StaticAnalysis/BreakingChangeAnalyzer/TypeMetadataHelper.cs
```diff
@@ -0,0 +1,481 @@
...
+                                // For each element in the generic type arguments list, make sure that the types
+                                // are the same
+                                for (int idx = 0; idx < oldTypeMetadata.GenericTypeArguments.Count; idx++)
+                                {
+                                    if (oldTypeMetadata.GenericTypeArguments[idx].Equals(newTypeMetadata.GenericTypeArguments[idx], StringComparison.OrdinalIgnoreCase))
+                                    {
+                                        // If we have not previously seen this generic type argument,
+                                        // run this method on the type
+                                        if (!_typeSet.Contains(oldTypeMetadata.GenericTypeArguments[idx]))
+                                        {
```
markcowl: the extra spacing makes the code a bit less readable.  There should only be vertical space after close braces

tools/StaticAnalysis/BreakingChangeAnalyzer/TypeMetadataHelper.cs
```diff
@@ -0,0 +1,481 @@
...
+        /// </summary>
+        /// <param name="cmdlet">The cmdlet metadata currently being checked.</param>
+        /// <param name="oldType">The type metadata from the old (serialized) assembly.</param>
+        /// <param name="newType">The type metadata from the new assembly.</param>
+        /// <param name="issueLogger">ReportLogger that will keep track of issues found.</param>
+        public void CompareTypeMetadata(
+            CmdletBreakingChangeMetadata cmdlet,
+            TypeMetadata oldType,
+            TypeMetadata newType,
+            ReportLogger<BreakingChangeIssue> issueLogger)
```
markcowl: This method is a bit too large.  I feel that splitting it up into separate methods for each case would make the code easier to read and maintain.

tools/StaticAnalysis/BreakingChangeAnalyzer/TypeMetadataHelper.cs
```diff
@@ -0,0 +1,481 @@
...
+                                            _typeSet.Add(oldTypeMetadata.GenericTypeArguments[idx]);
+
+                                            var oldElementType = _oldTypeDictionary[oldTypeMetadata.GenericTypeArguments[idx]];
+                                            var newElementType = _newTypeDictionary[newTypeMetadata.GenericTypeArguments[idx]];
+
+                                            CompareTypeMetadata(cmdlet, oldElementType, newElementType, issueLogger);
+                                        }
+                                    }
+                                    // If the generic type arguments aren't the same, log an issue
+                                    else
```
markcowl: so, if they aren't the same, don't we want to compare the properties recirsively?

tools/StaticAnalysis/BreakingChangeAnalyzer/TypeMetadataHelper.cs
```diff
@@ -0,0 +1,481 @@
...
+        /// <summary>
+        /// Checks if the type of the parameter is an array or a generic, and makes sure there are no breaking changes.
+        /// If the type is not an array or a generic, it proceeds with the normal type checking with CompareTypeMetadata.
+        /// </summary>
+        /// <param name="cmdlet">The cmdlet whose parameter metadata is being checked for breaking changes.</param>
+        /// <param name="parameter">The parameter whose type metadata is being checked for breaking changes.</param>
+        /// <param name="oldTypeMetadata">The type metadata from the old (serialized) assembly.</param>
+        /// <param name="newTypeMetadata">The type metadata from the new assembly.</param>
+        /// <param name="issueLogger">ReportLogger that will keep track of issues found.</param>
+        public void CheckParameterType(
```
markcowl: Same comment.  If we create separate methods for array checks versus geenric type checks, it will make the code easier to read.  Also, this lokks like a lot of duplication with the previus method.  If we had separate methods for the individual types of comparison, we might be able to reuse them in both places.

tools/StaticAnalysis/BreakingChangeAnalyzer/TypeMetadataHelper.cs
```diff
@@ -0,0 +1,481 @@
...
+
+        /// <summary>
+        /// Checks if the type of the output is an array or a generic, and makes sure there are no breaking changes.
+        /// If the type is not an array or a generic, it proceeds with the normal type checking with CompareTypeMetadata.
+        /// </summary>
+        /// <param name="cmdlet">The cmdlet whose output metadata is being checked for breaking changes.</param>
+        /// <param name="oldTypeMetadata">The type metadata from the old (serialized) assembly.</param>
+        /// <param name="newTypeMetadata">The type metadata from the new assembly.</param>
+        /// <param name="issueLogger">ReportLogger that will keep track of issues found.</param>
+        public void CheckOutputType(
```
markcowl: same comment - this should be broken down into the individual types of comparison, and we should look for opportunities to reuse code so there is less duplication

tools/StaticAnalysis/BreakingChangeAnalyzer/TypeMetadataHelper.cs
markcowl: So, we are just comparing the type names for generic types as far as I can tell.  Just as we do with non-generic types, we should conside comparing the generic type arguments property by property.

For example, List<NewObject>  may contain no breaking changes from List<OldObject>, but this check will flag it as an issue because the type name of the generic type arguments don't match. LMK if I have misinterpreted the logic or this is not clear.

PR #3211, opened 2016-11-18
Repo tasks add spn

shahabhijeet:
Added cmdlet to create Service Principal.
Refactored Repo-Tasks to include multiple modular modules (e.g. Build-Tasks, TestFx-Tasks)


tools/Modules/TestFx-Tasks.psm1
```diff
@@ -0,0 +1,436 @@
...
+        $azAdApp = Get-ADApp -ADAppDisplayName $ADAppDisplayName
+        $appId = $azAdApp.ApplicationId
+    }
+    
+    if([string]::IsNullOrEmpty($ADAppDisplayName) -eq $true)
+    {
+        $appId = $ADAppId
+    }
+
+    $azSpn = Get-AzureRmADServicePrincipal -ServicePrincipalName $appId
```
markcowl: Normally you have to create a service principal after creating the appi, see: https://gist.github.com/devigned/dae74a7ca54000f7b714#file-sp-w-cert-azps-1-0-ps1-L30

tools/Modules/TestFx-Tasks.psd1
```diff
@@ -0,0 +1,126 @@
...
+# Module manifest for module 'PSGet_TestFx-Tasks'
+#
+# Generated by: ShahAbhijeet
+#
+# Generated on: 11/18/2016
+#
+
+@{
+
+# Script module or binary module file associated with this manifest.
```
markcowl: I'm not sure I understand this script.  Where is the issue that prompted this work, and what do we want this to do?  Is it just about making the Azure cmdlets easier to use?

shahabhijeet: @markcowl almost every time when people try to use our test framework they ask how can I create service principal. So able to create service principal when someone is trying to use our module to create connection string is something that would be beneficial.
This is again part of the various tasks that would help our RPs to use our test infrastructure. There is not task item for this particular PR.

PR #3363, opened 2017-01-06
ADL Updates for Jan release

begoldsm:
## Description
This change set includes the following.
1. Fix TraceLogging for unstructured requests and responses to truncate at 10KB of data to protect against very large payloads being sent or received and bogging down memory
2. Add support for Top to Get-AdlJob
3. Fix Get-AdlJob to, by default, sort by the newest jobs when displaying
4. Re-enable serviceClientTracing for upload and download, now that TraceLogging does not use as much memory. The next step will be to include custom trace logging (to a file) instead of to debug output for these commands
5. Add explicit support for "None" as an encryption option for New-AdlStore and added warnings that, by default, encryption will be enabled.
6. Full help updates for all ADLA and ADLS cmdlets to be platyPS compliant and accurate.
7. As a result of the help update, cleaned up incorrect output types/return types and enabled some new passthru options.
8. Add ability to select product tier during new-adlstore, set-adlstore and new-adlanalyticsaccount and set-adlanalyticsaccount

---

- [x] Cmdlet specifies `OutputType` attribute if any output is produced - if the cmdlet produces no output, it should implement a `PassThrough` parameter.

### DIFF
--- a/src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Microsoft.Azure.Commands.DataLakeStore.dll-help.xml
+++ b/src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Microsoft.Azure.Commands.DataLakeStore.dll-help.xml
@@ -552,7 +552,7 @@
 <command:parameterValue required="true" variableLength="false">LogLevel</command:parameterValue>
 <dev:type><maml:name>LogLevel</maml:name>
 <maml:uri /></dev:type>
-<dev:defaultValue>None</dev:defaultValue>
+<dev:defaultValue>Error</dev:defaultValue>
 </command:parameter>
 <command:parameter required="true" variableLength="true" globbing="false" pipelineInput="True (ByPropertyName)" position="named" aliases="none"><maml:name>DiagnosticLogPath</maml:name>
 <maml:Description><maml:para>Specifies the path for the diagnostic log to record events to during the file or folder import.
@@ -615,7 +615,7 @@
 <command:parameterValue required="true" variableLength="false">LogLevel</command:parameterValue>
 <dev:type><maml:name>LogLevel</maml:name>
 <maml:uri /></dev:type>
-<dev:defaultValue>None</dev:defaultValue>
+<dev:defaultValue>Error</dev:defaultValue>
 </command:parameter>
 <command:parameter required="true" variableLength="true" globbing="false" pipelineInput="True (ByPropertyName)" position="named" aliases="none"><maml:name>DiagnosticLogPath</maml:name>
 <maml:Description><maml:para>Specifies the path for the diagnostic log to record events to during the file or folder import.
@@ -1798,7 +1798,7 @@
 <command:parameterValue required="true" variableLength="false">LogLevel</command:parameterValue>
 <dev:type><maml:name>LogLevel</maml:name>
 <maml:uri /></dev:type>
-<dev:defaultValue>None</dev:defaultValue>
+<dev:defaultValue>Error</dev:defaultValue>
 </command:parameter>
 <command:parameter required="true" variableLength="true" globbing="false" pipelineInput="True (ByPropertyName)" position="named" aliases="none"><maml:name>DiagnosticLogPath</maml:name>
 <maml:Description><maml:para>Specifies the path for the diagnostic log to record events to during the file or folder import.
@@ -1861,7 +1861,7 @@
 <command:parameterValue required="true" variableLength="false">LogLevel</command:parameterValue>
 <dev:type><maml:name>LogLevel</maml:name>
 <maml:uri /></dev:type>
-<dev:defaultValue>None</dev:defaultValue>
+<dev:defaultValue>Error</dev:defaultValue>
 </command:parameter>
 <command:parameter required="true" variableLength="true" globbing="false" pipelineInput="True (ByPropertyName)" position="named" aliases="none"><maml:name>DiagnosticLogPath</maml:name>
 <maml:Description><maml:para>Specifies the path for the diagnostic log to record events to during the file or folder import.


src/Common/Commands.Common/GeneralUtilities.cs
```diff
@@ -318,7 +321,9 @@ public static string FormatString(string content)
             }
             else
             {
-                return content;
+                return content.Length <= GeneralUtilities.StreamCutOffSize ?
```
markcowl: This is a quick, elegant solution, I like it.

cormacpayne:
On-demand: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1375/

src/ResourceManager/DataLakeAnalytics/Commands.DataLakeAnalytics/Commands/GetAzureRmDataLakeAnalyticsJob.cs
```diff
@@ -87,6 +87,12 @@ public class GetAzureDataLakeAnalyticsJob : DataLakeAnalyticsCmdletBase
         [ValidateNotNullOrEmpty]
         public JobResult[] Result { get; set; }
 
+        [Parameter(ParameterSetName = BaseParameterSetName, ValueFromPipelineByPropertyName = true,
+            Mandatory = false, HelpMessage = "An optional value which indicates the number of jobs to return. Default value is 500")]
+        [ValidateNotNullOrEmpty]
+
```
cormacpayne: @begoldsm please remove this space

src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Commands/RemoveAzureRmDataLakeStoreItemAcl.cs
```diff
@@ -47,40 +47,37 @@ public class RemoveAzureDataLakeStoreItemAcl : DataLakeStoreFileSystemCmdletBase
...
+                        DataLakeStoreFileSystemClient.RemoveDefaultAcl(Path.TransformedPath, Account);
+                    }
+                    else
+                    {
+                        DataLakeStoreFileSystemClient.RemoveAcl(Path.TransformedPath, Account);
+                    }
+
+                    if (PassThru)
+                    {
+                        WriteObject(true);
```
cormacpayne: @begoldsm if the user doesn't want output from the cmdlet, shouldn't the `bool` represent if the removal was successful or not, not just a hard-coded value?

src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Commands/RemoveAzureRmDataLakeStoreItemAclEntry.cs
```diff
@@ -84,7 +90,13 @@ public override void ExecuteCmdlet()
             ConfirmAction(
                 string.Format(Resources.RemoveDataLakeStoreItemAcl, string.Empty, Path.OriginalPath),
                 Path.OriginalPath,
-                () => { DataLakeStoreFileSystemClient.RemoveAclEntries(Path.TransformedPath, Account, aclSpec); });
+                () => {
+                    DataLakeStoreFileSystemClient.RemoveAclEntries(Path.TransformedPath, Account, aclSpec);
+                    if (PassThru)
+                    {
+                        WriteObject(true);
```
cormacpayne: @begoldsm same comment

src/ResourceManager/DataLakeAnalytics/Commands.DataLakeAnalytics/Commands/SetAzureRmDataLakeAnalyticsCatalogSecret.cs
```diff
@@ -65,15 +65,16 @@ public class SetAzureDataLakeAnalyticsCatalogSecret : DataLakeAnalyticsCmdletBas
...
         {
+            WriteWarning(Resources.IncorrectOutputTypeWarning);
             if (Uri != null && Uri.Port <= 0)
             {
                 WriteWarning(string.Format(Resources.NoPortSpecified, Uri));
             }
 
             var toUse = Uri ?? new Uri(string.Format("https://{0}:{1}", DatabaseHost, Port));
 
-            WriteObject(DataLakeAnalyticsClient.UpdateSecret(Account, DatabaseName, Secret.UserName,
```
begoldsm: Good catch. I will update this to return an empty object, since the SDK correctly returns null now, I will need to write out a nothing object.

src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Commands/RemoveAzureRmDataLakeStoreItemAcl.cs
begoldsm: Its a good question. In this case, failure will result in an exception, so no false will be returned. If they want an actual value returned from this cmdlet (which otherwise would not return anything) the result will always be either true or exception. If there is a better way that we should be handling this I am happy to do so, though.

src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Commands/RemoveAzureRmDataLakeStoreItemAclEntry.cs
begoldsm: same reply :)

src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Commands/RemoveAzureRmDataLakeStoreItemAcl.cs
cormacpayne: Discussed offline - this is the intended pattern

src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Commands/RemoveAzureRmDataLakeStoreItemAclEntry.cs
cormacpayne: Discussed offline

cormacpayne:
On-demand: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1376/

src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Commands/ImportAzureRmDataLakeStoreItem.cs
```diff
@@ -82,49 +85,113 @@ public int ConcurrentFileCount
...
         {
             var powerShellSourcePath = SessionState.Path.GetUnresolvedProviderPathFromPSPath(Path);
+            string diagnosticPath = string.Empty;
+            if (!string.IsNullOrEmpty(DiagnosticLogPath))
+            {
+                diagnosticPath = SessionState.Path.GetUnresolvedProviderPathFromPSPath(DiagnosticLogPath);
+            }
+
+            // Setting -Debug overwrites LogLevel to debug.
+            if (MyInvocation.BoundParameters.ContainsKey("Debug") && (SwitchParameter)MyInvocation.BoundParameters["Debug"])
```
markcowl: This is equivalent to $DebugPreference="Continue", if you really want to use this, you should abide by both the preference variable and the flag.  Also, this does not detect ```-Debug:$false``` which is also a possible use of the flag.

markcowl: For now, I would suggest using your own explicit log level, and thinking about how to use powershell settings for Verbose, Warning, Information, and Debug streams intelligently ina  future release.

src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Models/DataLakeStoreTraceLogger.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        public readonly IServiceClientTracingInterceptor SdkTracingInterceptor;
+
+        public DataLakeStoreTraceLogger(Cmdlet commandToLog, string logFilePath = null, LogLevel logLevel = LogLevel.Information)
+        {
+            LogFilePath = logFilePath;
+            LogLevel = logLevel;
+            if (string.IsNullOrEmpty(LogFilePath))
+            {
+                LogFilePath = string.Format(@"{0}\ADLDataTransfer\ADLDataTransfer_{1}.log",
```
markcowl: Do you really want to log if this is not specified?  This would be surprising to the user, I think, unless this default location is well documented.

markcowl: Also, this is guaranteed to blow up when the cmdlets move to .Net core (later this year), instead, I would require a log location

src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Models/DataLakeStoreTraceLogger.cs
```diff
@@ -0,0 +1,161 @@
...
+                    Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData),
+                    DateTime.Now.ToString("MM-dd-yyyy.HH.mm"));
+            }
+            else if(Directory.Exists(LogFilePath)) // the user passed in a directory instead of a file
```
markcowl: Suggestion: You should consider usign IDataStore for file system operations - this allows you to mock the file system in tests.  You would need to augment it to allow returning a stream to support this entire scenario.

src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Commands/ExportAzureRmDataLakeStoreItem.cs
```diff
@@ -95,20 +116,59 @@ public override void ExecuteCmdlet()
                         throw new CloudException(string.Format(Resources.InvalidExportPathType, Path.TransformedPath));
                     }
 
-                    if (type == FileType.FILE)
+                    DataLakeStoreTraceLogger logger = null;
```
markcowl: Suggestion: For the future, you'll need to think about how this would need to be augmented to support muyltiple cmdlet executions simultaneously in an AppDomain.  In this case, you would want to differentiate which traces go to which log file.

src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Commands/SetAzureRmDataLakeStoreItemAclEntry.cs
```diff
@@ -79,8 +80,21 @@ public class SetAzureDataLakeStoreItemAclEntry : DataLakeStoreFileSystemCmdletBa
...
+        [Parameter(ValueFromPipelineByPropertyName = true, Mandatory = false,
+            HelpMessage =
+                "Indicates the resulting ACL should be returned.",
+            ParameterSetName = BaseParameterSetName
+            )]
+        [Parameter(ValueFromPipelineByPropertyName = true, Mandatory = false,
+            HelpMessage =
+                "Indicates the resulting ACL should be returned.",
+            ParameterSetName = SpecificAceParameterSetName
+            )]
```
shahabhijeet: @begoldsm this parameter does not need any parameterset names. This parameter will be part of global parameter set

begoldsm: Ok great, I will remove it for passthru.

src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Commands/ImportAzureRmDataLakeStoreItem.cs
begoldsm: Fair enough. I have removed anything involving the -Debug flag for now, and I agree that it would be nice to have some way to enable some kind of logging (to a file or otherwise) with the -debug flag, or be able to use some other kind of interceptor. We can discuss this more during PowerShell office hours, since I would like to do something that can benefit everyone and not just these two cmdlets (ideally!)

src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Commands/ImportAzureRmDataLakeStoreItem.cs
```diff
@@ -82,49 +85,113 @@ public int ConcurrentFileCount
...
-                            Destination.TransformedPath,
-                            Account,
-                            powerShellSourcePath,
-                            CmdletCancellationToken,
-                            PerFileThreadCount,
-                            Force,
-                            Resume,
-                            ForceBinary,
-                            cmdletRunningRequest: this);
+                        if (DiagnosticLogLevel != LogLevel.None)
```
begoldsm: Agreed. I have moved everything into the constructor and dispose() methods of the logger for now. I think in the future (based on your suggestions) this can be refactored further.

src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Models/DataLakeStoreTraceLogger.cs
begoldsm: Fair enough, I have added a Diagnostic parameter set and removed the default log location and updated this to be a required parameter in the diagnostic parameter set. Seems reasonable to me.

src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Models/DataLakeStoreTraceLogger.cs
begoldsm: This is a good suggestion. I am open to doing this in the next release to add to testability. I will work with you to make sure I understand the pattern and can implement it nicely.

src/ResourceManager/DataLakeStore/Commands.DataLakeStore/Commands/ExportAzureRmDataLakeStoreItem.cs
begoldsm: Agreed. This trace logging is a "first step" to enable some of our internal and external partners to get initial heuristics about upload and download operations (mostly to help us with performance tuning). As this matures we will definitely need to update this and make it more robust. Ideally, we are able to integrate this into core AzurePowerShell functionality so that everyone can take advantage of it.

cormacpayne:
On-demand: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1379/

PR #3383, opened 2017-01-12
Fix sign build break

shahabhijeet:
Adding ServiceManagement Common project to RedisCache solution because it is not getting build with the same configuration during Signed build.

Why this happens:
1) Msbuild has a bug where when trying to create a project build map (meta project) it only get's the project that are listed in the solution
2) In this particular case, RedisCacheTest solution added project reference to Commands.Storage
3) But Commands.Storage has project reference to ServiceManagement.Common, this causes msbuild to miss this project in it's map and hence defaults to project default properties (in this case debug)
resulting in SM.Common project getting build as debug and hence not delayed signed and hence the entire build fails.

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -45,7 +45,7 @@
     <Scope Condition=" $(Scope) == '' " >all</Scope>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


markcowl:
LGTM

PR #3476, opened 2017-02-07
Policy parameters support in New-AzureRmPolicyDefinition and New-AzureRmPolicyAssignment

luanshixia:
## Description
Added support for policy parameters to the 2 cmdlets to use the feature provided by API version 2016-12-01.

---

### DIFF
--- a/src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
+++ b/src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
@@ -74,7 +74,7 @@ namespace Microsoft.Azure.Commands.ResourceManager.Cmdlets.Implementation
         /// Gets or sets the policy assignment policy parameter object.
         /// </summary>
         [Parameter(ParameterSetName = PolicyParameterObjectParameterSetName,
-            Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The policy parameter object.")]
+            Mandatory = true, ValueFromPipelineByPropertyName = false, HelpMessage = "The policy parameter object.")]
         public Hashtable PolicyParameterObject { get; set; }
 
         /// <summary>
@@ -168,13 +168,13 @@ namespace Microsoft.Azure.Commands.ResourceManager.Cmdlets.Implementation
                         var dp = new RuntimeDefinedParameter
                         {
                             Name = param.Name,
-                            ParameterType = typeString.Equals("array", StringComparison.OrdinalIgnoreCase) ? typeof(object[]) : typeof(string)
+                            ParameterType = typeString.Equals("array", StringComparison.OrdinalIgnoreCase) ? typeof(string[]) : typeof(string)
                         };
                         dp.Attributes.Add(new ParameterAttribute
                         {
                             ParameterSetName = ParameterlessPolicyParameterSetName,
                             Mandatory = true,
-                            ValueFromPipelineByPropertyName = true,
+                            ValueFromPipelineByPropertyName = false,
                             HelpMessage = helpString
                         });
                         this.dynamicParameters.Add(param.Name, dp);


azurecla:
Hi __@luanshixia__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (wayan). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -54,6 +61,19 @@ public class NewAzurePolicyAssignmentCmdlet : PolicyAssignmentCmdletBase
         public PSObject PolicyDefinition { get; set; }
 
         /// <summary>
+        /// Gets or sets the policy assignment policy parameter object.
+        /// </summary>
+        [Parameter(Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The pollicy parameter object.")]
```
maphad: Typo in the help me

maphad: Typo in the help message "pollicy" should be "policy"

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -54,6 +61,19 @@ public class NewAzurePolicyAssignmentCmdlet : PolicyAssignmentCmdletBase
...
+        public Hashtable PolicyParameterObject { get; set; }
+
+        /// <summary>
+        /// Gets or sets the policy assignment policy parameter file or string.
+        /// </summary>
+        [Parameter(Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The pollicy parameter file or string.")]
```
maphad: Typo in the help message "pollicy"

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -112,11 +132,88 @@ private JToken GetResource()
                 {
                     DisplayName = this.DisplayName ?? null,
                     PolicyDefinitionId = this.PolicyDefinition.Properties["policyDefinitionId"].Value.ToString(),
-                    Scope = this.Scope
+                    Scope = this.Scope,
+                    Parameters = GetParameters()
```
maphad: we usually scope the function calls, so this should be this.GetParameters()

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -54,6 +61,19 @@ public class NewAzurePolicyAssignmentCmdlet : PolicyAssignmentCmdletBase
         public PSObject PolicyDefinition { get; set; }
 
         /// <summary>
+        /// Gets or sets the policy assignment policy parameter object.
+        /// </summary>
+        [Parameter(Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The pollicy parameter object.")]
+        public Hashtable PolicyParameterObject { get; set; }
```
maphad: We should have validation so that user cannot specify both PolicyParameterObject and PolicyParameters.
This can be implemented using parameter sets.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -54,6 +61,19 @@ public class NewAzurePolicyAssignmentCmdlet : PolicyAssignmentCmdletBase
...
+
+        /// <summary>
+        /// Gets or sets the policy assignment policy parameter file or string.
+        /// </summary>
+        [Parameter(Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The pollicy parameter file or string.")]
+        [ValidateNotNullOrEmpty]
+        public string PolicyParameters { get; set; }
```
maphad: Should we rename this to PolicyParametersFile?

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -112,11 +132,88 @@ private JToken GetResource()
...
+
+            return dynamicParameters;
+        }
+
+        protected JObject GetParameters()
+        {
+            // Load parameters from local file or literal
+            if (PolicyParameters != null)
+            {
+                string policyParameterFilePath = this.TryResolvePath(PolicyParameters);
```
maphad: I think it is confusing to use a single parameter as maybe a file name or maybe a string representing parameters.
We should explicitly make them two separate parameters. So in total we should have three parameters:
PolicyParameters
PolicyParametersFile
PolicyParametersObject.
Each should be in its own ParameterSet so that user can only specify one of them.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -112,11 +132,88 @@ private JToken GetResource()
...
+                return parameterObject;
+            }
+
+            // Load dynamic parameters
+            var parameters = PowerShellUtilities.GetUsedDynamicParameters(dynamicParameters, MyInvocation);
+            if (parameters.Count() == 0)
+            {
+                return null;
+            }
+            else
```
maphad: No need of else here since you are returning control from if

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -112,11 +132,88 @@ private JToken GetResource()
...
+                    Parameters = GetParameters()
                 }
             };
 
             return policyassignmentObject.ToJToken();
         }
+
+        object IDynamicParameters.GetDynamicParameters()
+        {
+            if (PolicyDefinition != null)
```
maphad: scope it as this.PolicyDefinition, everywhere

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyDefinition.cs
```diff
@@ -113,7 +120,8 @@ private JToken GetResource()
                 {
                     Description = this.Description ?? null,
                     DisplayName = this.DisplayName ?? null,
-                    PolicyRule = JObject.Parse(GetPolicyRuleObject().ToString())
+                    PolicyRule = JObject.Parse(GetPolicyRuleObject().ToString()),
+                    Parameters = this.Parameters == null ? null : JObject.Parse(GetParametersObject().ToString())
```
maphad: this.Parameters ?? JObject.Parse(GetParametersObject().ToString())

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyDefinition.cs
```diff
@@ -131,5 +139,17 @@ private JToken GetPolicyRuleObject()
                 ? JToken.FromObject(FileUtilities.DataStore.ReadFileAsText(policyFilePath))
                 : JToken.FromObject(this.Policy);
         }
+
+        /// <summary>
+        /// Gets the parameters object
+        /// </summary>
+        private JToken GetParametersObject()
+        {
+            string parametersFilePath = this.TryResolvePath(this.Parameters);
```
maphad: As above I think we shouldn't use the same parameter as maybe a filePath or a parameter. It would be cleaner to have two separate parameters in two different parameter sets.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -112,11 +132,88 @@ private JToken GetResource()
...
-                    Scope = this.Scope
+                    Scope = this.Scope,
+                    Parameters = GetParameters()
                 }
             };
 
             return policyassignmentObject.ToJToken();
         }
+
+        object IDynamicParameters.GetDynamicParameters()
```
maphad: who calls this method and can u point me to how dynamic parameters work?

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyDefinition.cs
maphad: this.GetParametersObject

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
maphad: For how to use ParameterSets u can refer to this documentation https://msdn.microsoft.com/en-us/library/dd878348(v=vs.85).aspx. Also several of our existing cmdlets use parameter sets.

maphad: For how to use ParameterSets u can refer to this documentation https://msdn.microsoft.com/en-us/library/dd878348(v=vs.85).aspx. Also several of our existing cmdlets use parameter sets.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Entities/Policy/PolicyAssignmentProperties.cs
```diff
@@ -15,6 +15,7 @@
 namespace Microsoft.Azure.Commands.ResourceManager.Cmdlets.Entities.Policy
 {
     using Newtonsoft.Json;
```
maphad: This looks like it has 3 commits. We should squash changes to a single commit, so that it is easy to revert the entire change if required.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyDefinition.cs
```diff
@@ -131,5 +139,17 @@ private JToken GetPolicyRuleObject()
...
+
+            return File.Exists(parametersFilePath)
+                ? JToken.FromObject(FileUtilities.DataStore.ReadFileAsText(parametersFilePath))
+                : JToken.FromObject(this.Policy);
+        }
     }
```
maphad: You can add scenario tests once you have addressed the PR comments.
Also get this PR reviewed by vivsrini.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
luanshixia: This interface method is called by the PowerShell engine (thus a callback), so PowerShell can know what parameters are defined dynamically in the cmdlet and allow users to provide values for them.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
luanshixia: Just confirmed with Mike that it is the expected behavior.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
luanshixia: For the 1st one: Just confirmed with Mike that it is the expected behavior.
For the 2nd one: Yes will do that.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyDefinition.cs
luanshixia: Just confirmed with Mike that it is the expected behavior.

cormacpayne:
@azuresdkci add to whitelist

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -112,11 +137,85 @@ private JToken GetResource()
...
-                    Scope = this.Scope
+                    Scope = this.Scope,
+                    Parameters = this.GetParameters()
                 }
             };
 
             return policyassignmentObject.ToJToken();
         }
+
+        object IDynamicParameters.GetDynamicParameters()
```
vivsriaus: This doesn't look like it belongs here. Can we move this out to a common/util file?

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -17,15 +17,25 @@ namespace Microsoft.Azure.Commands.ResourceManager.Cmdlets.Implementation
...
+    using System;
+    using System.Linq;
+    using System.Collections;
 
     /// <summary>
     /// Creates a policy assignment.
     /// </summary>
     [Cmdlet(VerbsCommon.New, "AzureRmPolicyAssignment"), OutputType(typeof(PSObject))]
-    public class NewAzurePolicyAssignmentCmdlet : PolicyAssignmentCmdletBase
+    public class NewAzurePolicyAssignmentCmdlet : PolicyAssignmentCmdletBase, IDynamicParameters
```
vivsriaus: Please add scenario tests/moq tests for both these cmdlets

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Entities/Policy/PolicyDefinitionProperties.cs
```diff
@@ -39,5 +39,11 @@ public class PolicyDefinitionProperties
         /// </summary>
         [JsonProperty(Required = Required.Always)]
         public JObject PolicyRule { get; set; }
+
+        /// <summary>
+        /// The parameters declaration.
+        /// </summary>
+        [JsonProperty(Required = Required.Default)]
+        public JObject Parameters { get; set; }
```
vivsriaus: Can you please confirm that there are no changes required for other policy cmdlets? i.e.,
1. Get/set/remove policy definition works with and without parameters
2. Get/set/remove policy assignment works with and without parameters

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyDefinition.cs
```diff
@@ -59,6 +59,13 @@ public class NewAzurePolicyDefinitionCmdlet : PolicyDefinitionCmdletBase
         public string Policy { get; set; }
 
         /// <summary>
+        /// Gets or sets the parameters parameter
+        /// </summary>
+        [Parameter(Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The parameters for policy definition. This can either be a path to a file name containing the parameters declaration, or the parameters declaration as string.")]
+        [ValidateNotNullOrEmpty]
+        public string Parameters { get; set; }
```
vivsriaus: PolicyDefinitionParameters?

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -50,10 +60,25 @@ public class NewAzurePolicyAssignmentCmdlet : PolicyAssignmentCmdletBase
...
+            Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The policy parameters object.")]
+        public Hashtable PolicyParametersObject { get; set; }
+
+        /// <summary>
+        /// Gets or sets the policy assignment policy parameters file path or policy parameters string.
+        /// </summary>
+        [Parameter(ParameterSetName = PolicyParametersStringParameterSetName, 
+            Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The policy parameters file path or policy parameters string.")]
+        [ValidateNotNullOrEmpty]
+        public string PolicyParameters { get; set; }
```
vivsriaus: PolicyAssignmentParameters?

TBH, the names are confusing as-is (what is the difference between parameters in policy def vs policy assignment?)

vivsriaus: Also, please ensure the help files are updated accordingly

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyDefinition.cs
```diff
@@ -59,6 +59,13 @@ public class NewAzurePolicyDefinitionCmdlet : PolicyDefinitionCmdletBase
         public string Policy { get; set; }
 
         /// <summary>
+        /// Gets or sets the parameters parameter
```
vivsriaus: :)

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
vivsriaus: No api-version change for policy cmdlets?

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
luanshixia: For 1: So the parameters in policy def are declarations and those in policy assignment are actual values. I know the names are confusing, but I have confirmed with Mike about these names.
For 2: Will do that.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
luanshixia: This is the implementation of the interface IDynamicParameters.

cormacpayne:
@luanshixia the build is failing since you are introducing parameters `PolicyParameters` and `Parameters`, which do not follow the PowerShell convention of parameters having a singular noun

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Components/Constants.cs
```diff
@@ -63,7 +63,7 @@ public static class Constants
         /// <summary>
         /// The default policy API version.
         /// </summary>
-        public static readonly string PolicyApiVersion = "2016-04-01";
+        public static readonly string PolicyApiVersion = "2016-12-01";
```
maphad: just curious as to why the new API version date is set to December instead of a more recent date?

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Entities/Policy/PolicyAssignmentProperties.cs
```diff
@@ -38,5 +39,11 @@ public class PolicyAssignmentProperties
         /// </summary>
         [JsonProperty(Required = Required.Always)]
         public string PolicyDefinitionId { get; set; }
+
+        /// <summary>
+        /// The parameter values.
+        /// </summary>
+        [JsonProperty(Required = Required.Default)]
+        public JObject Parameters { get; set; }
```
maphad: You can rename this to Parameter to be consistent with the PolicyDefinitionProperties class

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -50,10 +60,25 @@ public class NewAzurePolicyAssignmentCmdlet : PolicyAssignmentCmdletBase
         /// <summary>
         /// Gets or sets the policy assignment policy definition parameter.
         /// </summary>
-        [Parameter(Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The pollicy definition object.")]
+        [Parameter(Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The policy definition object.")]
```
maphad: For the parameter set to work correctly all other parameters should belong to both parameter sets

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -112,11 +137,86 @@ private JToken GetResource()
...
+                    ? JObject.Parse(FileUtilities.DataStore.ReadFileAsText(policyParameterFilePath))
+                    : JObject.Parse(this.PolicyParameter);
+            }
+
+            // Load from PS object
+            JObject parameterObject = null;
+            if (this.PolicyParameterObject != null)
+            {
+                parameterObject = new JObject();
+                foreach (string paramKey in this.PolicyParameterObject.Keys)
```
maphad: This logic can be moved to a Util class in a method called HashTableToJToken.
Similarly you can create another Util method for the logic on line 209 DynamicParamToJToken.
If there isnt already a utility class for common Json operations like this you can create a new one and move this re-usable code there.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Components/Constants.cs
luanshixia: Per the resource schema seen on Azure Portal, the latest version for policy is 2016-12-01:
```json
        {
            "resourceType": "policyDefinitions",
            "locations": [],
            "apiVersions": [
                "2016-12-01",
                "2016-04-01",
                "2015-10-01-preview"
            ]
        },
        {
            "resourceType": "policyAssignments",
            "locations": [],
            "apiVersions": [
                "2016-12-01",
                "2016-04-01",
                "2015-10-01-preview"
            ]
        },
```

cormacpayne:
@luanshixia the build is failing because of an issue in the BreakingChangeAnalyzer. I am working on a fix for it right now.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Entities/Policy/PolicyAssignmentProperties.cs
luanshixia: This can be a 1-click when merging on this page, though I don't have permission to do that now.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
luanshixia: As we discussed, I fixed this according to NewAzureRmRGDeployment.

luanshixia:
@cormacpayne  Thanks so much for the help. I will rollback my workaround once you have a fix.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
luanshixia: Definitely agree. The reason I was not doing that was just I was not sure whether I should modify the util classes or not. So now I will do.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Entities/Policy/PolicyAssignmentProperties.cs
luanshixia: Unfortunately this property cannot be renamed because that should match the REST API. (I also only realized this when I caught an error after trying to keep consistency). Of course we can match names using JsonProperty attribute, but no need.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Entities/Policy/PolicyDefinitionProperties.cs
luanshixia: AFAIK, no impact at this time.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Entities/Policy/PolicyAssignmentProperties.cs
maphad: ok please squash before merging then.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Extensions/JsonExtensions.cs
```diff
@@ -212,5 +213,31 @@ public static bool TryConvertTo<TType>(this JToken jobject, out TType result)
...
+            foreach (string paramKey in keys)
+            {
+                var valueObject = new JObject();
+                valueObject.Add("value", dict[paramKey].ToJToken());
+                parameterObject.Add(paramKey, valueObject);
+            }
+            return parameterObject;
+        }
     }
 }
```
maphad: Add a newline at end.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Extensions/JsonExtensions.cs
```diff
@@ -212,5 +213,31 @@ public static bool TryConvertTo<TType>(this JToken jobject, out TType result)
...
+            if (dict == null)
+            {
+                return null;
+            }
+            if (keys == null)
+            {
+                keys = dict.Keys;
+            }
+            var parameterObject = new JObject();
+            foreach (string paramKey in keys)
```
maphad: you can use var instead of string as is the convention.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Extensions/JsonExtensions.cs
```diff
@@ -212,5 +213,31 @@ public static bool TryConvertTo<TType>(this JToken jobject, out TType result)
...
             return false;
         }
+
+        /// <summary>
+        /// Converts an <see cref="IDictionary"/> object to <see cref="JObject"/> with an extra level labeled as "value".
+        /// </summary>
+        /// <param name="dict">The <see cref="IDictionary"/> object.</param>
+        /// <param name="keys">The key set to extract from the <see cref="IDictionary"/> object. Default is all keys.</param>
+        /// <returns>The conversion result.</returns>
+        public static JObject ToParameterObject(this IDictionary dict, IEnumerable keys = null)
```
maphad: The method name ToParameterObject is specific to its use in the policy commandlet.
Since this is a utility method it can be used from elsewhere. Can we use a more generic name like ToJsonObject or ToJObject?

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -50,10 +61,31 @@ public class NewAzurePolicyAssignmentCmdlet : PolicyAssignmentCmdletBase
         /// <summary>
         /// Gets or sets the policy assignment policy definition parameter.
         /// </summary>
-        [Parameter(Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The pollicy definition object.")]
+        [Parameter(Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The policy definition object.")]
```
maphad: Is this line required?

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -112,11 +144,73 @@ private JToken GetResource()
...
+            var parameters = PowerShellUtilities.GetUsedDynamicParameters(dynamicParameters, MyInvocation);
+            if (parameters.Count() > 0)
+            {
+                return MyInvocation.BoundParameters.ToParameterObject(parameters.Select(p => p.Name));
+            }
+
+            return null;
+        }
     }
 }
```
maphad: Add newline at end of file.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyDefinition.cs
```diff
@@ -113,7 +120,8 @@ private JToken GetResource()
                 {
                     Description = this.Description ?? null,
                     DisplayName = this.DisplayName ?? null,
-                    PolicyRule = JObject.Parse(GetPolicyRuleObject().ToString())
+                    PolicyRule = JObject.Parse(GetPolicyRuleObject().ToString()),
+                    Parameters = this.Parameter == null ? null : JObject.Parse(GetParametersObject().ToString())
```
maphad: scope local method calls as this.GetParametersObject()

src/ResourceManager/Resources/Commands.Resources.Test/ScenarioTests/PolicyTests.ps1
```diff
@@ -78,3 +78,88 @@ function Test-PolicyAssignmentCRUD
...
+
+<#
+.SYNOPSIS
+Tests Policy definition creation with parameters
+#>
+function Test-PolicyDefinitionWithParameters
+{
+	# Test
+	$actual = New-AzureRMPolicyDefinition -Name testPDWP -Policy "$TestOutputRoot\SamplePolicyDefinitionWithParameters.json" -Parameter "$TestOutputRoot\SamplePolicyDefinitionParameters.json"
+	$expected = Get-AzureRMPolicyDefinition -Name testPDWP
```
maphad: I dont think it is a good idea to use another commandlet to give you the expected value.
If there is a bug that affects both the commandlets then it is likely it will go uncaught.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
luanshixia: So @cormacpayne is working on a fix for BreakingChangeAnalyzer which prevents this PR from passing the build (the reason: the analyzer compares the parameter set with the previous build. In previous build this property belongs to the default parameter set, so it enforces this for later builds). This line is a workaround which can be removed after his fix.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Extensions/JsonExtensions.cs
luanshixia: Here "keys" is of type IEnumerable not IEnumerable<string> so paramKey will not be inferred as string unless declared specific.

luanshixia:
@cormacpayne I have already got two peer approvals but I don't have the permission to merge it. Could you please merge it so that it can make the 2/15 deadline? Thanks!

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
cormacpayne: @luanshixia if this parameter is required in every single parameter set, why make this change? Previously it was a globally required parameter, so it was required in any parameter set you used. The same applies now with the changes you are trying to make here.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -17,15 +17,26 @@ namespace Microsoft.Azure.Commands.ResourceManager.Cmdlets.Implementation
...
-    [Cmdlet(VerbsCommon.New, "AzureRmPolicyAssignment"), OutputType(typeof(PSObject))]
-    public class NewAzurePolicyAssignmentCmdlet : PolicyAssignmentCmdletBase
+    [Cmdlet(VerbsCommon.New, "AzureRmPolicyAssignment", DefaultParameterSetName = ParameterlessPolicyParameterSetName), OutputType(typeof(PSObject))]
+    public class NewAzurePolicyAssignmentCmdlet : PolicyAssignmentCmdletBase, IDynamicParameters
     {
+        protected RuntimeDefinedParameterDictionary dynamicParameters = new RuntimeDefinedParameterDictionary();
+
+        protected const string PolicyParameterObjectParameterSetName = "Policy assignment with parameters via policy parameter object";
+        protected const string PolicyParameterStringParameterSetName = "Policy assignment with parameters via policy parameter string";
+        protected const string ParameterlessPolicyParameterSetName = "Policy assignment without parameters";
```
cormacpayne: @luanshixia the name of this parameter set is misleading - `Name` and `Scope` (and `PolicyDefinition` based on its implementation) parameters are globally required parameters, so you are always going to need to provide a value for them, meaning you'll never have a parameter set that is "parameterless"

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -112,11 +144,73 @@ private JToken GetResource()
...
-                    Scope = this.Scope
+                    Scope = this.Scope,
+                    Parameters = this.GetParameters()
                 }
             };
 
             return policyassignmentObject.ToJToken();
         }
+
+        object IDynamicParameters.GetDynamicParameters()
```
cormacpayne: @luanshixia why do you need to add dynamic parameters to this cmdlet? We try to avoid using them as much as possible since they cannot be picked up by our StaticAnalysis tooling and don't allow us to easily check any public interface changes

src/ResourceManager/Resources/Commands.Resources/help/New-AzureRmPolicyAssignment.md
```diff
@@ -175,6 +190,32 @@ Accept pipeline input: False
 Accept wildcard characters: False
 ```
 
+### -PolicyParameter
+The policy parameter file path or policy parameter string.```yaml
```
cormacpayne: @luanshixia please move "```yaml" to a new line since it is causing errors with the displayed markdown

src/ResourceManager/Resources/Commands.Resources/help/New-AzureRmPolicyAssignment.md
```diff
@@ -175,6 +190,32 @@ Accept pipeline input: False
...
+
+Required: True
+Position: Named
+Default value: None
+Accept pipeline input: True (ByPropertyName)
+Accept wildcard characters: False
+```
+
+### -PolicyParameterObject
+The policy parameter object.```yaml
```
cormacpayne: @luanshixia same comment

src/ResourceManager/Resources/Commands.Resources/help/New-AzureRmPolicyDefinition.md
```diff
@@ -171,6 +171,19 @@ Accept pipeline input: False
 Accept wildcard characters: False
 ```
 
+### -Parameter
+The parameters declaration for policy definition. This can either be a path to a file name containing the parameters declaration, or the parameters declaration as string.```yaml
```
cormacpayne: @luanshixia same comment

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
luanshixia: @cormacpayne No, "parameter" here refers to the policy parameters, not the cmdlet parameters.. This is similar to what is done for template parameters in [NewAzureRmResourceGroupDeployment](https://github.com/Azure/azure-powershell/blob/dev/src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/CmdletBase/ResourceWithParameterCmdletBase.cs).

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
luanshixia: @cormacpayne So there are 3 usages:
* PolicyDefinition only
* PolicyDefinition + PolicyParameterObject
* PolicyDefinition + PolicyParameter
We don't want users to set both PolicyParameterObject and PolicyParameter, so we use two different sets for them. But we also need to allow users to set none of PolicyParameterObject and PolicyParameter, which comes the third set. Note we have to set Mandatory=true for PolicyParameterObject and PolicyParameter or we will get parameter set resolving conflict.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
luanshixia: @cormacpayne Because we want to allow users to specify the policy parameters through cmdlet parameters, so we have to dynamically parse the policy definition and extract a list of defined parameters. That is from the requirements. In the future, we also want to add IntelliSense for the dynamic parameters, just like what we did for template deployment.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
cormacpayne: @luanshixia I am fine with the three different parameter sets, but there is no need to change `PolicyDefinition` from a global parameter to a parameter that's in each of the three parameter sets (that is what is implied by being a global parameter).

cormacpayne:
@vivsriaus can you approve the review you have started? You commented with LGTM, but it still says you have changes requested.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -112,11 +144,73 @@ private JToken GetResource()
...
+                    if (parameters != null)
+                    {
+                        foreach (var param in (parameters.Value as PSObject).Properties)
+                        {
+                            var type = (param.Value as PSObject).Properties["type"];
+                            var typeString = type != null ? type.Value.ToString() : string.Empty;
+                            var dp = new RuntimeDefinedParameter
+                            {
+                                Name = param.Name,
+                                ParameterType = typeString.Equals("array", StringComparison.OrdinalIgnoreCase) ? typeof(object[]) : typeof(string)
```
cormacpayne: @luanshixia can you leverage `typeString` to include a switch case for types you would want to define (_e.g._, `int`, `bool`, etc.)? It seems like a bad idea to have the default type of non-arrays as `string`

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -112,11 +144,73 @@ private JToken GetResource()
...
+                            };
+                            dp.Attributes.Add(new ParameterAttribute
+                            {
+                                Mandatory = false,
+                                ValueFromPipelineByPropertyName = true,
```
cormacpayne: @luanshixia what kind of scenario would lend users to use the `ValueFromPipelineByPropertyName` for these dynamic parameters?

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -112,11 +144,73 @@ private JToken GetResource()
...
+                                HelpMessage = param.Name
```
cormacpayne: @luanshixia this is not good help for users, it's just repeating the parameter name

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -112,11 +144,73 @@ private JToken GetResource()
...
+                        foreach (var param in (parameters.Value as PSObject).Properties)
+                        {
+                            var type = (param.Value as PSObject).Properties["type"];
+                            var typeString = type != null ? type.Value.ToString() : string.Empty;
+                            var dp = new RuntimeDefinedParameter
+                            {
+                                Name = param.Name,
+                                ParameterType = typeString.Equals("array", StringComparison.OrdinalIgnoreCase) ? typeof(object[]) : typeof(string)
+                            };
+                            dp.Attributes.Add(new ParameterAttribute
```
cormacpayne: @luanshixia the dynamic parameters will need to be in the `Parameterless` parameter set, or else users will be able to use them with the provided hashtable or json, which will defeat the purpose of having the dynamic parameters

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -50,10 +61,31 @@ public class NewAzurePolicyAssignmentCmdlet : PolicyAssignmentCmdletBase
...
         /// Gets or sets the policy assignment policy definition parameter.
         /// </summary>
-        [Parameter(Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The pollicy definition object.")]
+        [Parameter(Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The policy definition object.")]
+        [Parameter(ParameterSetName = ParameterlessPolicyParameterSetName,
+            Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The policy definition object.")]
+        [Parameter(ParameterSetName = PolicyParameterObjectParameterSetName,
+            Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The policy definition object.")]
+        [Parameter(ParameterSetName = PolicyParameterStringParameterSetName,
+            Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The policy definition object.")]
```
cormacpayne: @luanshixia how will users use this `ValueFromPipelineByPropertyName` attribute for this parameter? Will there be an object they can pass through the pipeline to this cmdlet that contains the `PolicyDefinition` property? Will it also include the below parameters you are trying to add since they also have the `ValueFromPipelineByPropertyName` attribute as well?

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
luanshixia: @cormacpayne Per [Policy doc](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-policy#parameters), "the type of a parameter can be either string or array".

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
luanshixia: @cormacpayne Good catch.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
luanshixia: @cormacpayne I checked the existing code and found pipeline usage is supported across all Resources cmdlets. So the user can do like: @{"PolicyDefinition": @{...} ...} | New-AzureRmPolicyAssignment

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
luanshixia: @cormacpayne Similarly, the user can put dynamic parameters in an object (say $param) and say: $param | New-AzureRmPolicyAssignment

cormacpayne:
On-demand: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1397/

cormacpayne:
LGTM once on-demand passes

PR #3485, opened 2017-02-09
PsAzStack Signing issues

shahabhijeet:
1. Addressing Signing bugs. Now we sign all dll with Sha2 with Sha1, MSI with Sha1, all scripts with Sha2.
2. Removing all .js as well as json files from layouts.
3. Adding Setup tests. Regenerating wix files.
4. Fixing reference for ASM-ExpressRoute causing it to be part of the layout.
5. Removing iisnode.dll, iisnode_schema.xml from layout


shahabhijeet:
@azuresdkci test this please

PR #3676, opened 2017-03-28
Support for RecoveryServices Vault upgrade

neetus15:
## Description
Adding support for Vault upgrade for site recovery and backup vaults.
Migrating customers from RDFE vaults to ARM.
As part of this, introduced two new cmdlets- one for checking prerequisites and other for triggering vault upgrade job.

Because of some issues, couldn't generate test JSONs. Will update the PR with them by EOD today.

---

- [Yes ] **I have read the [contribution guidelines](https://github.com/Azure/azure-powershell/blob/dev/CONTRIBUTING.md).**
- [ Yes] **If changes were made to any cmdlet, the XML help was regenerated using the [platyPSHelp module](https://github.com/Azure/azure-powershell/blob/dev/documentation/platyPSHelp-documentation.md).**
- [Yes ] **If any large changes are made to a service, they are reflected in the respective [change log](https://github.com/Azure/azure-powershell/blob/dev/CONTRIBUTING.md#updating-the-change-log).**

- [Yes ] Title of the pull request is clear and informative.
- [Yes] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR. For more information on cleaning up the commits in your PR, [see this page](https://github.com/Azure/azure-powershell/blob/dev/documentation/cleaning-up-commits.md).
- [Yes ] The pull request does not introduce [breaking changes](https://github.com/Azure/azure-powershell/blob/dev/documentation/breaking-changes.md) (unless a major version change occurs in the assembly and module).

- [Yes ] Pull request includes test coverage for the included changes.
- [Yes] PowerShell scripts used in tests should do any necessary setup as part of the test or suite setup, and should not use hard-coded values for locations or existing resources.

- [Yes ] New cmdlets that make changes or have side effects should implement `ShouldProcess` and have `SupportShouldProcess=true` specified in the cmdlet attribute. You can find more information on `ShouldProcess` [here](https://gist.github.com/markcowl/338e16fe5c8bbf195aff9f8af0db585d#what-is-the-change).
- [Yes ] Cmdlet specifies `OutputType` attribute if any output is produced - if the cmdlet produces no output, it should implement a `PassThru` parameter.

- [ Yes] Parameter types should not expose types from the management library - complex parameter types should be defined in the module.
- [ Yes] Complex parameter types are discouraged - a parameter type should be simple types as often as possible. If complex types are used, they should be shallow and easily creatable from a constructor or another cmdlet.
- [ Yes] Cmdlet parameter sets should be mutually exclusive - each parameter set must have at least one mandatory parameter not in other parameter sets.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,10 +1,4 @@
-﻿## 2017.03.31 - Version 3.8.0
-* RecoveryServices
-    * Added new cmdlets to support Recovery Services vault upgrade
-        - Test-AzureRecoveryServicesVaultUpgrade
-        - Invoke-AzureRecoveryServicesVaultUpgrade
-
-## 2017.03.09 - Version 3.7.0
+﻿## 2017.03.09 - Version 3.7.0
 * ApiManagement
     * Added new cmdlets to manage Backend entity
         - New-AzureRmApiManagementBackend
--- a/src/ServiceManagement/Services/Commands.Utilities/ChangeLog.md
+++ b/src/ServiceManagement/Services/Commands.Utilities/ChangeLog.md
@@ -22,6 +22,10 @@
 
 * Add Maintenance Status field to Get-AzureVM response.
 
+* Added new cmdlets to support Recovery Services vault upgrade
+    - Test-AzureRecoveryServicesVaultUpgrade
+    - Invoke-AzureRecoveryServicesVaultUpgrade
+
 ## Version 3.7.0
 * Update the output object of migration cmdlets (Move-AzureService, Move-AzureStorageAccount, Move-AzureVirtualNetwork, Move-AzureNetworkSecurityGroup, Move-AzureReservedIP, Move-AzureRouteTable):
     - ValidationMessages contain "Information" and "Warning" messages in addition to "Error" messages.


shahabhijeet:
@azuresdkci test this please

shahabhijeet:
@azuresdkci add to whitelist

setup/azurecmdfiles.wxi
```diff
@@ -3984,9 +3984,6 @@
                         <Component Id="cmpF7ADD37722F45985B5323202F35958F2" Guid="*">
                             <File Id="fil0FA91D9C31AF07DCD03BFB842FE2B77E" KeyPath="yes" Source="$(var.sourceDir)\ServiceManagement\Azure\Compute\Microsoft.WindowsAzure.Commands.ServiceManagement.PlatformImageRepository.dll" />
                         </Component>
-                        <Component Id="cmp9D64A1111722F25BC100121E73BE039E" Guid="*">
```
shahabhijeet: @neetus15 any reasons these items are removed from the installer. This is not related to your change.
We are really warry of installer changes, most of the time your additions should not be removing any of the other components from the installer.
Please make sure you have a clean build before you generate the installer.
git status (make sure this is clean)
git clean -xdf
from the root
msbuild build.proj
After a successful build, please run Generate.ps1 -buildConfig Debug

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/PSRecoveryServicesClient/PSRecoveryServicesClient.cs
```diff
@@ -259,6 +261,44 @@ private SiteRecoveryManagementClient GetSiteRecoveryClient()
...
+        /// <param name="location">Resource location.</param>
+        /// <param name="resourceName">Resource Name.</param>
+        /// <param name="resourceType">Resource Type.</param>
+        /// <returns>Vault Upgrade Management client.</returns>
+        private RecoveryServicesVaultUpgradeManagementClient GetVaultUpgradeClient(
+            string location,
+            string resourceName,
+            string resourceType)
+        {
+            string resourceNamespace = Environment.GetEnvironmentVariable("RESOURCE_NAMESPACE");
```
shahabhijeet: @neetus15 do you expect users to set this environment variable? If yes, what is the reason for that?
Also it is quite possible, I might have a environment variable "RESOURCE_NAMESPACE" and I have no intention to use it for VaultUpgrade and if I call this cmdlet, this code will end up using it.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,238 @@
...
+
+        /// <summary>
+        /// Gets or sets vault type.
+        /// </summary>
+        [Parameter(Mandatory = true)]
+        [ValidateSet(
+            Constants.ASRVaultType,
+            Constants.BackupVault,
+            IgnoreCase = true)]
+        public string ResourceType { get; set; }
```
shahabhijeet: @neetus15 RecoverServiceClient needs Namespace, why is that VaultUpgrade does not have any parameter that asks users to provide resource namespace?
So if I understand this design, the only way a user can pass in the resourceNamespace is by creating an environment variable=RESOURCE_NAMESPACE

Also you create RecoverServiceClient twice for each cmdlet (Test-* Invoke-*) the namespace can be different between two clients due to the presence of RESOURCE_NAMESPACE env. variable

setup/azurecmdfiles.wxi
neetus15: I tried the build without these changes. But the build started failing saying couldn't find these files. Then I reverted the code to this and build worked.

And I generated it using the exact steps you have given here. I just finally did "git add ., then again build "

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/PSRecoveryServicesClient/PSRecoveryServicesClient.cs
neetus15: Actually this has been added for internal testing only. 

But as you said it's quite possible that because of customer's environment, this gets used. Instead of the prod values. One check I can put here is "the value should be either of the two allowed values(WABackupBVTD2/HRMBvtd2ToDogfood)". The chances that he would be using the same value, unless intentionally, is very less.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
neetus15: Resource_Namespace Environment variable is just for internal testing. By default, resource namespace variable will pick the prod value.
So, it will first check if there is an environment variable with name "resource_namepsace". If yes, take that value. If no, pick the production value.

Now i will add the check that in case the env variable "resource_namespace" contains the value, it should be either of the two allowed values, just to be on the safe side. 

We didn't want to give customers  the option to choose the resource namespace since we already had resource type and we can easily conclude the resource namespace based on that. We didn't want redundancy.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/PSRecoveryServicesClient/PSRecoveryServicesClient.cs
praveennet: probably add comments around this why this is used/ present in code. also suggest renaming the variable to "RecoveryServices_VaultUpgrade_ResourceNamespace" to be very explicit

shahabhijeet: @neetus15 I think this was mentioned in the review to Praveen, if you need to pass in some custom object/data, you can use HttpHandler and pass it to ServiceClient. All classes are partial, so you should be able to extend any of the classes
Extend you service client and pass in an HttpHandler that will setup you custom value in your request.
Please understand, you cannot start adding test hooks because it's easy to implement.
Bottom line, are you guys comfortable with this implementation?
If you still think you want to go ahead with this implementation, please open an issue that you would be fixing this and the time frame when this will be fixed.
We can make exception, but we need a plan as to how you will be fixing it.

setup/azurecmdfiles.wxi
shahabhijeet: @neetus15 I checked it on my side with a fresh clone, with this set of given changes, all the changes that are now shown in setup should not be there.
One possibility is:
1) You have not pulled from upstream in a while (if not do a pull and get the latest changes)
2) rebuild again and then build the installer using generate.ps1
Remember, all you should see with your set of changes is an additional dll getting added to the setup.
In a nutshell if you diff the output directory (\azure-powershell\src\Package\Debug\ServiceManagement) prior to your change and after your change, the only difference should be an additional .dll, that is exactly what setup update should be (unless you have added additional resource files or other files that needs to be installed on the users machines)
But apart from those changes, no files should be deleted or nothing additional should be added.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,238 @@
...
+                ExceptionDetails details =
+                    this.RecoveryServicesClient.HandleVaultUpgradeException(exception);
+                if (!string.IsNullOrEmpty(details.WarningDetails))
+                {
+                    this.WriteWarning(details.WarningDetails);
+                }
+
+                if (!string.IsNullOrEmpty(details.ErrorDetails))
+                {
+                    throw new InvalidOperationException(
```
shahabhijeet: @neetus15 consider using WriteError(new InvalidOperationException()), instead of throwing. You get two advantages, your ErrorCategory can align with your Exception (in this case the ErrorCategory.InvalidOperation)
And users who are using your cmdlet in a script have set $ErrorView="CategoryView"
e.g.
WriteError(new ErrorRecord(ex, string.Empty, ErrorCategory.InvalidOperation, null));

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/TestAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,110 @@
...
+                ExceptionDetails details =
+                    this.RecoveryServicesClient.HandleVaultUpgradeException(exception);
+                if (!string.IsNullOrEmpty(details.WarningDetails))
+                {
+                    this.WriteWarning(details.WarningDetails);
+                }
+
+                if (!string.IsNullOrEmpty(details.ErrorDetails))
+                {
+                    throw new InvalidOperationException(
```
shahabhijeet: @neetus15 consider using WriteError() as pointed above

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,238 @@
...
+                    this.ResourceType);
+
+                elapsedSeconds = DateTime.Now.Subtract(startTime).TotalSeconds;
+                record.PercentComplete = (int)elapsedSeconds * 100 / (int)taskTimeoutInSeconds;
+                this.WriteProgress(record);
+            }
+            while (response.OperationStatus == Constants.InProgress &&
+                elapsedSeconds < taskTimeoutInSeconds &&
+                !this.StopProcessingFlag);
+
```
markcowl: It is a good idea to write a completion record to the progress stream when complete (this is just setting the recordtype parameter in your ProgressRecord).

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,238 @@
...
+            try
+            {
+                this.RecoveryServicesClient.TestVaultUpgradePrerequistes(
+                    this.VaultName,
+                    this.Location,
+                    this.ResourceType,
+                    this.TargetResourceGroupName,
+                    this.Profile.Context.Subscription.Id.ToString());
+
+                this.WriteObject(Properties.Resources.CheckPrereqSucceded);
```
markcowl: You should only write one object to the output stream.  If you want to print details of the validation, the should be written to the progress, verbose, or debug streams

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,238 @@
...
+            return response;
+        }
+
+        /// <summary>
+        /// Writes content to the screen.
+        /// </summary>
+        /// <param name="contents">Data to be printed on the screen.</param>
+        private void WriteResponse(string contents)
+        {
+            this.WriteObject(Environment.NewLine);
```
markcowl: This method should not write to the Output stream - the Output stream contains the return value of the cmdlet.  This should write to the progress, debug, or verbose streams

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/TestAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                this.RecoveryServicesClient.TestVaultUpgradePrerequistes(
+                    this.VaultName,
+                    this.Location,
+                    this.ResourceType,
+                    this.TargetResourceGroupName,
+                    this.Profile.Context.Subscription.Id.ToString());
+
+                this.WriteResponse(Properties.Resources.CheckPrereqSucceded);
+                this.WriteObject(Environment.NewLine);
```
markcowl: The output stream is the return type of the cmdlet.  You should not write strings there.  Information about progress and the like should be written to one of the informational streams (progress, verbose, or debug)

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/TestAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,110 @@
...
+                            details.ErrorDetails));
+                }
+            }
+        }
+
+        /// <summary>
+        /// Writes content to the screen.
+        /// </summary>
+        /// <param name="contents">Data to be printed on the screen.</param>
+        private void WriteResponse(string contents)
```
markcowl: First - why implement this method twice?  Second, the output stream is the return value of your cmdlet, this method should write to one of the other streams as previously mentioned.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/TestAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,110 @@
...
+using System;
+using System.Management.Automation;
+using Microsoft.Azure.Commands.RecoveryServices.SiteRecovery;
+
+namespace Microsoft.Azure.Commands.RecoveryServices
+{
+    /// <summary>
+    /// Used to validate vault upgrade prerequisites.
+    /// </summary>
+    [Cmdlet(VerbsDiagnostic.Test, "AzureRecoveryServicesVaultUpgrade")]
```
markcowl: Need an OutputType attribute, containing the output type of the cmdlet.  In this case, this could be validation details, or a boolean

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
neetus15: Taken..

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,238 @@
...
+using System.Threading;
+using Microsoft.Azure.Commands.RecoveryServices.SiteRecovery;
+using Microsoft.WindowsAzure.Management.RecoveryServicesVaultUpgrade.Models;
+
+namespace Microsoft.Azure.Commands.RecoveryServices
+{
+    /// <summary>
+    /// Used to initiate vault upgrade operation.
+    /// </summary>
+    [Cmdlet(VerbsLifecycle.Invoke, "AzureRecoveryServicesVaultUpgrade", SupportsShouldProcess = true)]
```
markcowl: Need an OutputType attribute for the cmdlet.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,238 @@
...
+                new ProgressRecord(
+                    0,
+                    Properties.Resources.VaultUpgradeInProgress,
+                    Properties.Resources.WaitingForCompletion);
+
+            this.WriteResponse(Properties.Resources.VaultUpgradeInProgress);
+
+            do
+            {
+                Thread.Sleep(PSRecoveryServicesClient.TimeToSleepBeforeFetchingJobDetailsAgain);
```
markcowl: Please use this method: https://github.com/Azure/azure-powershell/blob/dev/src/Common/Commands.Common/TestMockSupport.cs#L38  instead of calling Thread.Sleep directly.  This will keep mocked tests that run this cmdlet from having to wait.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,238 @@
...
+                    this.Location,
+                    this.ResourceType,
+                    this.TargetResourceGroupName,
+                    this.Profile.Context.Subscription.Id.ToString());
+
+                this.WriteObject(Properties.Resources.CheckPrereqSucceded);
+            }
+            catch (Exception exception)
+            {
+                this.WriteResponse(Properties.Resources.CheckPrereqFailed);
```
neetus15: So all of this has been done to improve readability.
All of our existing auxportal customers have to use powershell to upgrade the vault and we are  pretty sure that many of them will not be even familiar to powershell's way of doing things. We want the ux to be as easy and as readable to the customer as possible.

Second point is it's just a one  time thing for all the customer: just to move the vault from auxportal to ibiza.
Writeojects are written in a way to update the customer with the current status of the upgrade.
Aggregating some or all, will either introduce inconsistency or compromise with readability.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/TestAzureRecoveryServicesVaultUpgrade.cs
neetus15: Done

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/TestAzureRecoveryServicesVaultUpgrade.cs
neetus15: Done

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
neetus15: done

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/TestAzureRecoveryServicesVaultUpgrade.cs
neetus15: Removed twice occurrence. 

So the justification of using this is 
Most of our auxportal customers are not even familiar with powershell's way of doing things. Most of them will be using the powershell cmdlets for the first time. So considering that we want to make the ui as close to user as possible(readable). So, I have used the write-object function in a way that it will literally update the customer with the current status of vault upgrade. It will keep updating him after every major point so that he knows what's happening and the thing that migration is not stucked and it's still in progress.

a) **We are not expecting the customers to use the return value of this cmdlet in any way.  
b) It's just a "one time thing" for the customer: once the migration from auxportal to ibiza happens, it's done. People are not going to use it over and over again.**
c) We DO want to show the user all the intermediate steps without even -debug and -verbose. WriteProgress doesn't solbve my purpose as i don't have any progress records. I just have statements.
I saw WriteInformation as well in the base class but it's not implemented both in srevicemanagement and arm code. I am not sure why and whether even this will solve the issue?

If the bold text makes sense, let me know. If you have some better option, we can consider that but please note that "we do want to show the intermediate steps to the client, otherwise the experience would be nothing happening on the powershell screen from past 10-15 mins, which is a really bad experience."

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/TestAzureRecoveryServicesVaultUpgrade.cs
neetus15: Please find the complete justification for this in the next comment.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/TestAzureRecoveryServicesVaultUpgrade.cs
neetus15: Done

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
neetus15: Given the justification in the second last comment.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
neetus15: Given justification/purpose/requirement in the second last comment.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
neetus15: So once the control gets out of while loop:there can be four situations a) upgrade passed b) upgrade failed c) upgrade terminated d) upgrade timed out

I don't think i should set the progressrecord to completion for the last two situations(it would be a misnomer). for first two, i will add.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
neetus15: done

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/PSRecoveryServicesClient/PSRecoveryServicesVaultUpgradeClient.cs
```diff
@@ -0,0 +1,357 @@
...
+                .RecoveryServicesVaultUpgrade.TrackResourceUpgrade(this.GetCustomRequestHeaders(false));
+        }
+
+        /// <summary>
+        /// Vault upgrade exception handler.
+        /// </summary>
+        /// <param name="ex">Exception to handle.</param>
+        /// <returns>Object to be returned after categorizing all the errors properly
+        /// received as part of vault upgrade operations.</returns>
+        public ExceptionDetails HandleVaultUpgradeException(Exception ex)
```
neetus15: done based on what I understood.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/PSRecoveryServicesClient/PSRecoveryServicesVaultUpgradeClient.cs
```diff
@@ -0,0 +1,357 @@
...
+
+            return exceptionMessage;
+        }
+
+        /// <summary>
+        /// Parse the exception thrown from vault upgrade operations.
+        /// </summary>
+        /// <param name="error">Instance of VaultUpgradeError.</param>
+        /// <returns>Exception message to be shown to customer.</returns>
+        private StringBuilder ParseError(
```
neetus15: same as above :)

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
neetus15: Mark, I initially added the code. But on testing, found out that the progress bar no longer shows up. I tried the operation again without this change and things again started working. So, holding this comment as of now.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,238 @@
...
+        [Parameter(Mandatory = true)]
+        [ValidateNotNullOrEmpty]
+        [Alias("TargetRG", "TargetRGName", "RG")]
+        public string TargetResourceGroupName { get; set; }
+
+        /// <summary>
+        /// Gets or sets switch parameter. On passing, command does not ask for confirmation.
+        /// </summary>
+        [Parameter(Mandatory = false)]
+        public SwitchParameter Force { get; set; }
```
neetus15: a) I copy pasted the exact code: if (ShouldProcess(target, message) && (!Force.IsPresent || ShouldContinue(query, caption))) { ... cmdlet execution }

In my scenario, the expected behavior is with force, it shouldn't ask for confirmations. Without force, it should.
Whereas the opposite was happening here: with force, it was asking for confirmation. 

Also since anyways we want to expose force parameter, can i continue with the approach I am using currently?

neetus15: ConfirmAction(targetm, message, Action): Customer can, at any time, by pass the checks by setting$confirmlevel >= $confirmpreference.
In the approach I am using currently, customer would be knowing that he is passing the force parameter intentionally. Whereas in this scenario, it might happen that he has changed powershell settings for some other cmdlet and during the course of action, might end up running this cmdlet and hence unknowingly by passing the check.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,238 @@
...
+            bool actionConfirmed = false;
+            ResourceUpgradeDetails response = null;
+            string message = result == CheckVaultUpgradePrerequisitesResponse.Succeeded ?
+                Properties.Resources.PrerequisitesCheckPassed :
+                Properties.Resources.PrerequisitesCheckReturnedWarnings;
+
+            try
+            {
+                this.ConfirmAction(
+                    this.Force.IsPresent,
```
neetus15: same explanation as above.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/RecoveryServicesCmdletBase.cs
```diff
@@ -133,6 +135,117 @@ public void HandleException(Exception ex)
...
+                                {
+                                    return this.recoveryServicesClient.ParseErrorDetails(error.Details, clientRequestIdMsg);
+                                }
+                                else
+                                {
+                                    StringBuilder exceptionMessage = new StringBuilder();
+                                    exceptionMessage.AppendLine(Properties.Resources.VaultUpgradeExceptionDetails).AppendLine(" ");
+                                    exceptionMessage.Append(this.recoveryServicesClient.ParseError(error));
+                                    exceptionMessage.AppendLine(clientRequestIdMsg);
+                                    this.ThrowTerminatingError(
```
shahabhijeet: @neetus15 I had forgotten to click on submit. But this can be ignore as just add it to the list of things that needs to be done after the merge into a separate task.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/RecoveryServicesCmdletBase.cs
```diff
@@ -157,6 +270,16 @@ public void WaitForJobCompletion(string jobId)
         }
 
         /// <summary>
+        /// Writes content to the screen.
+        /// </summary>
+        /// <param name="contents">Data to be printed on the screen.</param>
+        public void WriteResponse(string contents)
+        {
+            this.WriteObject(Environment.NewLine);
```
markcowl: This is not changed as asked.  Please change this to not write to the output stream

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
markcowl: List<string> is not an acceptable output type for this.  Your cmdlet should output an instance with details of the upgrade request, a boolean, or nothing

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
markcowl: As I said, if you are going to have a Force parameter, you cannot use ConfirmAction.  The ConfirmAction overload that takes a 'Force' parameter in the service management cmdlets does not implement confirmation correctly for the -Confirm and WhatIf parameters, instead implement as follows:

```
if (ShouldProcess(target, message) && (Force.IsPresent || ShouldContinue( query, caption)) {... cmdlet execution }
```

This is a requirement for new cmdlets.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
markcowl: This should make the progress bar disappear after upgrade, which is what you want.  If yoiu are running mocked, then this will mean no progress is displayedm, because you are not actually waiting any time between polls

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,234 @@
...
+                record.PercentComplete = (int)elapsedSeconds * 100 / (int)taskTimeoutInSeconds;
+                this.WriteProgress(record);
+            }
+            while (response.OperationStatus == Constants.InProgress &&
+                elapsedSeconds < taskTimeoutInSeconds &&
+                !this.StopProcessingFlag);
+
+            if (response.OperationStatus == Constants.Completed)
+            {
+                this.WriteResponse(
```
markcowl: This is not acceptable.  Do not write strings to the output stream

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,234 @@
...
+                    string.Format(
+                        Properties.Resources.TrackVaultUpgradeJobResult,
+                        response.OperationResult,
+                        response.OperationStatus));
+
+                if (response.OperationResult == Constants.Succeeded)
+                {
+                    this.WriteResponse(
```
markcowl: Same comment

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,234 @@
...
+                        string.Format(
+                            Properties.Resources.VaultUpgradeSucceded,
+                            this.ResourceType,
+                            this.VaultName));
+                }
+                else
+                {
+                    this.WriteResponse(Properties.Resources.VaultUpgradeNotSucceded);
```
markcowl: Same comment

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,234 @@
...
+                }
+            }
+            else if (this.StopProcessingFlag)
+            {
+                this.WriteResponse(Properties.Resources.VaultUpgradeTerminated);
```
markcowl: same comment

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,234 @@
...
+            }
+            else
+            {
+                this.WriteResponse(Properties.Resources.VaultUpgradeTimedOut);
```
markcowl: same comment

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,234 @@
...
+            try
+            {
+                this.RecoveryServicesClient.TestVaultUpgradePrerequistes(
+                    this.VaultName,
+                    this.Location,
+                    this.ResourceType,
+                    this.TargetResourceGroupName,
+                    this.Profile.Context.Subscription.Id.ToString());
+
+                this.WriteObject(Properties.Resources.CheckPrereqSucceeded);
```
markcowl: You should only ever write one object to the output stream in the execution of ProcessRecord

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/TestAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,108 @@
...
+    [OutputType(typeof(List<string>))]
+    public class TestAzureRecoveryServicesVaultUpgrade : RecoveryServicesCmdletBase
```
markcowl: List<string> is not an acceptable output type.  You should write a single object  that describes the result of the validation, a boolean, or nothing

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/TestAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,108 @@
...
+        #endregion Parameters
+
+        /// <summary>
+        /// ProcessRecord of the command.
+        /// </summary>
+        public override void ExecuteCmdlet()
+        {
+            try
+            {
+                this.WriteResponse(Properties.Resources.StartingPrerequisitesCheck);
```
markcowl: No, you cannot write strings like this to the output stream

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/TestAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,108 @@
...
+                this.RecoveryServicesClient.TestVaultUpgradePrerequistes(
+                    this.VaultName,
+                    this.Location,
+                    this.ResourceType,
+                    this.TargetResourceGroupName,
+                    this.Profile.Context.Subscription.Id.ToString());
+
+                this.WriteObject(Properties.Resources.CheckPrereqSucceeded);
```
markcowl: You should only write one object to the output stream as a result of thsi cmdlet, you need to decide which one.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/TestAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,108 @@
...
+                this.WriteObject(Environment.NewLine);
```
markcowl: Same comment, do not write strings to the output stream

shahabhijeet:
@neetus15 forgot to mention, please add changelog as below
https://github.com/Azure/azure-powershell/blob/dev/src/ServiceManagement/Services/Commands.Utilities/ChangeLog.md

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
neetus15: Done

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
neetus15: Taken!!!

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
neetus15: Taken!!!

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/TestAzureRecoveryServicesVaultUpgrade.cs
neetus15: Taken!!!  Created a new object and passed the response to that.

src/ServiceManagement/RecoveryServices/Commands.RecoveryServices/Service/InvokeAzureRecoveryServicesVaultUpgrade.cs
```diff
@@ -0,0 +1,244 @@
...
+            bool actionConfirmed = false;
+            ResourceUpgradeDetails response = null;
+            string message = result == CheckVaultUpgradePrerequisitesResponse.Succeeded ?
+                Properties.Resources.PrerequisitesCheckPassed :
+                Properties.Resources.PrerequisitesCheckReturnedWarnings;
+
+            try
+            {
+                if (this.ShouldProcess(this.VaultName, Properties.Resources.StartVaultUpgradeWhatIfMessage) &&
+                    (this.Force.IsPresent || this.ShouldContinue(string.Format(message), string.Empty)))
```
markcowl: For the future, you may consider only printing the SHouldContinue prompt if the vault upgrade test returned warnings

Issue #3642, opened 2017-03-17
FIPS mode breaks AzureRM


pburkholder:
### Cmdlet(s)

Almost all

### PowerShell Version

5.1.14409.1005

### Module Version

2.7.0      AzureRM.profile

3.7.0      AzureRM.Resources

### OS Version

Windows 10
Windows 2012R2

### Description

This works:  `Add-AzureRmAccount -EnvironmentName AzureUSGovernment`

Then pretty much every other command fails like this:

```
PS C:\Users\peterb> Get-AzureRmResourceGroup
Get-AzureRmResourceGroup : Exception has been thrown by the target of an invocation.
At line:1 char:1
+ Get-AzureRmResourceGroup
+ ~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Get-AzureRmResourceGroup], TargetInvocationException
    + FullyQualifiedErrorId : System.Reflection.TargetInvocationException,Microsoft.Azure.Commands.ResourceManager.Cmd
   lets.Implementation.GetAzureResourceGroupCmdlet
```

### Debug Output
 Using `$DebugPreference="Continue"` or `-Verbose` or `-Debug` results in the same output

### Script/Steps for Reproduction

```
Set-ItemProperty HKLM:\System\CurrentControlSet\Control\Lsa\FipsAlgorithmPolicy -Name Enabled -Value 1
Restart-Computer -Force
```

New session:

```
# confirm FIPS setting:
Get-ItemProperty HKLM:\System\CurrentControlSet\Control\Lsa\FipsAlgorithmPolicy

# Login -- this works
Add-AzureRmAccount -EnvironmentName AzureUSGovernment

# Try any other AzureRM command
> Get-AzureRmResourceGroup
Get-AzureRmResourceGroup : Exception has been thrown by the target of an invocation.
At line:1 char:1
+ Get-AzureRmResourceGroup
+ ~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Get-AzureRmResourceGroup], TargetInvocationException
    + FullyQualifiedErrorId : System.Reflection.TargetInvocationException,Microsoft.Azure.Commands.ResourceManager.Cmd
   lets.Implementation.GetAzureResourceGroupCmdlet
```

cormacpayne:
@pburkholder Hey Peter, thanks for filing this issue - we will look into this.

[...]

PR #3683, opened 2017-03-29
Update hash function used by telemetry cmdlets for FIPS compliance

markcowl:
## Description

Fixes the specific issue in [#3642]

---

### DIFF
--- a/src/Common/Commands.Common/MetricHelper.cs
+++ b/src/Common/Commands.Common/MetricHelper.cs
@@ -207,6 +207,11 @@ namespace Microsoft.WindowsAzure.Commands.Common
         /// <returns></returns>
         public static string GenerateSha256HashString(string originInput)
         {
+            if (string.IsNullOrWhiteSpace(originInput))
+            {
+                throw new ArgumentNullException("originInput");
+            }
+
             using (var sha256 = new SHA256CryptoServiceProvider())
             {
                 var bytes = sha256.ComputeHash(Encoding.UTF8.GetBytes(originInput));
--- a/src/ResourceManager/Profile/Commands.Profile.Test/Commands.Profile.Test.csproj
+++ b/src/ResourceManager/Profile/Commands.Profile.Test/Commands.Profile.Test.csproj
@@ -187,6 +187,7 @@
     <Compile Include="ProfileController.cs" />
     <Compile Include="RPRegistrationDelegatingHandlerTests.cs" />
     <Compile Include="SubscriptionCmdletTests.cs" />
+    <Compile Include="TelemetryTests.cs" />
     <Compile Include="TenantCmdletTests.cs" />
     <Compile Include="LoginCmdletTests.cs" />
     <Compile Include="ContextCmdletTests.cs" />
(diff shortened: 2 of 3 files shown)


src/Common/Commands.Common/MetricHelper.cs
```diff
@@ -207,9 +207,11 @@ public void FlushMetric()
         /// <returns></returns>
         public static string GenerateSha256HashString(string originInput)
         {
-            SHA256 sha256 = SHA256.Create();
-            var bytes = sha256.ComputeHash(Encoding.UTF8.GetBytes(originInput));
-            return BitConverter.ToString(bytes);
+            using (var sha256 = new SHA256CryptoServiceProvider())
```
shahabhijeet: @markcowl Can you please describe, how changing the algorithm class removes the exception.
Maybe I am not able to understand what was causing the exception in the first place.

markcowl: @shahabhijeet There are a set of providers that are FIPS compliant, and a set that are not.  If you use one of the FIPS-compliant providers, then there is no exception when FIPS mode is turned on: https://blogs.msdn.microsoft.com/icumove/2009/01/31/working-with-fips-in-net-c/

src/Common/Commands.Common/MetricHelper.cs
```diff
@@ -207,9 +207,11 @@ public void FlushMetric()
         /// <returns></returns>
         public static string GenerateSha256HashString(string originInput)
```
shahabhijeet: @markcowl is null check necessary here? or is it getting null checked before calling this function.
Would be good to skip if "originInput" is null

pburkholder: @shahabhijeet It seems that Win10AnniversaryEdition does not properly enforce FIPS-mode. See https://gist.github.com/pburkholder/01214219f89c794a4740c7a59d1846ff#file-findings-md

markcowl: It is a public function, so the .Net guidelines say we should always do a null check

markcowl: And yes, this is supported in .net 4.5 on all of our supported platforms

markcowl:
@shahabhijeet Responded to feedback, and added unit tests

src/Common/Commands.Common/MetricHelper.cs
markcowl: @pburkholder This:  https://technet.microsoft.com/en-us/library/security/cc750357.aspx#_Microsoft_FIPS_140  says that FIPS is not enforced by the OS, and it is up to each application to abide by the FIPS security policy

pburkholder: https://web-beta.archive.org/web/20141129025434/https://technet.microsoft.com/en-us/library/security/cc750357.aspx which was last updated May 2014 has the same language:

> Note – There is no enforcement of the FIPS policy by the operating system or the validated cryptographic modules. Instead, each individual application must check this flag and enforce 

It seems .NET at 4.6.2 is supposed to throw an exception when using SHA256Managed at https://referencesource.microsoft.com/#mscorlib/system/security/cryptography/sha256managed.cs,35, but it doesn't throw it. As I see it, .Net has been written "to abide by the FIPS security policy" but doesn't actually abide by it.

Or is my logic wrong here? Help appreciated! Thanks....

pburkholder: @markcowl I know I'm seeing this on 4.6.0, but I don't have the source for that, so don't know if that's changed since.

Issue #1530, opened 2015-12-21
Get-AzureRmContext should not throw when no context exists and "-ErrorAction SilentlyContinue" is used


slapointe:
**Context**
- Windows Server 2012 R2
- PowerShell v5.0.10514.6
- AzureRM v1.4

**Steps to reproduce**
In a new PowerShell session, execute:

``` powershell
$context = Get-AzureRmContext -ErrorAction SilentlyContinue
```

**Expected**
$context should be $null and no errors should be thrown about the need to run Login-AzureRmAccount prior to my call.

**Intended scenario(s)**
Enable us to determine if a context is present or not without encountering an error. Could be used to support a clean persisted/restore profile mechanism. 

**Actual behavior**

``` powershell
PS C:\Windows\system32> Get-AzureRmContext -ErrorAction SilentlyContinue
Get-AzureRmContext : Run Login-AzureRmAccount to login.
At line:1 char:1
+ Get-AzureRmContext -ErrorAction SilentlyContinue
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Get-AzureRmContext], PSInvalidOperationException
    + FullyQualifiedErrorId : InvalidOperation,Microsoft.Azure.Commands.Profile.GetAzureRMContextCommand
```

smartnose:
+1 Seeing similar issue here.

haocs:
@slapointe To run Get-AzureRmContext requires user login(context must be present). If it is not present PowerShell has to terminate execution and throw exception. And this is a **terminating error** by design which means it will be thrown anyway. 
In your scenario, my suggestion is to add a try-catch block for the cmdlet. Then you are able to determine if context is present by the checking this exception.

slapointe:
A good alternative would be to introduce a Test-AzureRmContext, what do you think?

markcowl:
@slapointe @haocs   In general, not having a context is a terminating error - meaning that any cmdlet pipeline should be stopped immediately regardless of error settings.  In this particular case, since the work of the cmdlet is just getting the context, I agree that ErrorActionPreference should apply in this narrow case.  @haocs This probably sounds different from our previous communication, let's discuss.

blumu:
[...]

PR #3690, opened 2017-03-30
Fix error when calling Get-AzureRmContext with no context

cormacpayne:
## Description

Fix for issue https://github.com/Azure/azure-powershell/issues/1530

Previously, if a user was not signed in and tried to run `Get-AzureRmContext`, an error was thrown telling them to "Run Login-AzureRmAccount to login.", which could not be suppressed based on the `ErrorActionPreference`.

Made a change that writes to the error stream instead of throwing a terminating error, which allows users to silently continue through the call and receive a `null` context.

---

### DIFF
--- a/src/ResourceManager/Profile/Commands.Profile.Test/ContextCmdletTests.cs
+++ b/src/ResourceManager/Profile/Commands.Profile.Test/ContextCmdletTests.cs
@@ -89,10 +89,6 @@ namespace Microsoft.Azure.Commands.ResourceManager.Profile.Test
                 cmdlt.ExecuteCmdlet();
                 cmdlt.InvokeEndProcessing();
             }
-            catch
-            {
-                
-            }
             finally
             {
                 AzureRmProfileProvider.Instance.Profile = profile;
--- a/src/ResourceManager/Profile/Commands.Profile/Context/GetAzureRMContext.cs
+++ b/src/ResourceManager/Profile/Commands.Profile/Context/GetAzureRMContext.cs
@@ -45,7 +45,7 @@ namespace Microsoft.Azure.Commands.Profile
 
         public override void ExecuteCmdlet()
         {
-            var context = (PSAzureContext)AzureRmProfileProvider.Instance.Profile.Context;
+            var context = (PSAzureContext)DefaultContext;
             if (context == null)
             {
                 WriteError(new ErrorRecord(


src/ResourceManager/Profile/Commands.Profile/Context/GetAzureRMContext.cs
```diff
@@ -26,6 +27,26 @@ namespace Microsoft.Azure.Commands.Profile
     [OutputType(typeof(PSAzureContext))]
     public class GetAzureRMContextCommand : AzureRMCmdlet
     {
+        /// <summary>
+        /// Gets the current default context.
+        /// </summary>
+        protected override AzureContext DefaultContext
```
shahabhijeet: @cormacpayne are you planning to add any tests for this?
Can you also point what version of PS you have it on your machine?
The reason I ask is, I was trying to find the enum on my machine that you have used 'AuthenticationError', I could not find that enum.

src/ResourceManager/Profile/Commands.Profile/Context/GetAzureRMContext.cs
```diff
@@ -26,6 +27,26 @@ namespace Microsoft.Azure.Commands.Profile
...
+                if (DefaultProfile == null || DefaultProfile.Context == null)
+                {
+                    WriteError(new ErrorRecord(
+                        new PSInvalidOperationException("Run Login-AzureRmAccount to login."),
+                        string.Empty,
+                        ErrorCategory.AuthenticationError,
+                        null));
+                }
+
+                return DefaultProfile.Context;
```
shahabhijeet: @cormacpayne also is there a reason, we are not replacing the throw Exception in the base class with WriteError.
Do you know under any circumstances if we simply have it in the base class won't solve this issue?

cormacpayne: @shahabhijeet here is a comment Mark left on the original issue regarding this:

_In general, not having a context is a terminating error - meaning that any cmdlet pipeline should be stopped immediately regardless of error settings. In this particular case, since the work of the cmdlet is just getting the context, I agree that ErrorActionPreference should apply in this narrow case._

src/ResourceManager/Profile/Commands.Profile/Context/GetAzureRMContext.cs
cormacpayne: @shahabhijeet it's found in [`System.Management.Automation`](https://msdn.microsoft.com/en-us/library/system.management.automation.errorcategory(v=vs.85).aspx)

I will add a test as well

src/ResourceManager/Profile/Commands.Profile/Context/GetAzureRMContext.cs
shahabhijeet: @cormacpayne ok make sense

src/ResourceManager/Profile/Commands.Profile/Context/GetAzureRMContext.cs
markcowl: Looks good - add the test and this should be good - you should be able to test this with a unit test mocking out ICommandRuntime

src/ResourceManager/Profile/Commands.Profile/Context/GetAzureRMContext.cs
markcowl: This will still throw an exception if DefaultProfile is null.  I think the right way to handle this is do have the DefaultContext return null, and write the error inside ExecuteCmdlet

src/ResourceManager/Profile/Commands.Profile/Context/GetAzureRMContext.cs
```diff
@@ -26,9 +27,34 @@ namespace Microsoft.Azure.Commands.Profile
...
+            }
+        }
+
         public override void ExecuteCmdlet()
         {
-            WriteObject((PSAzureContext)AzureRmProfileProvider.Instance.Profile.Context);
+            var context = (PSAzureContext)AzureRmProfileProvider.Instance.Profile.Context;
```
markcowl: You can just use DefaultContext here

src/ResourceManager/Profile/Commands.Profile.Test/ContextCmdletTests.cs
```diff
@@ -73,6 +73,42 @@ public void GetAzureContext()
...
+            AzureRmProfileProvider.Instance.Profile = new AzureRMProfile();
+
+            try
+            {
+                // Act
+                cmdlt.InvokeBeginProcessing();
+                cmdlt.ExecuteCmdlet();
+                cmdlt.InvokeEndProcessing();
+            }
+            catch
```
markcowl: Is this catch needed?

PR #3697, opened 2017-03-30
 Obsoleting Profile cmdlets to use 'Context' noun, and changing 'Sele…

markcowl:
…ct' verb to 'Import'

## Description

---

### DIFF
--- a/src/ResourceManager/Profile/ChangeLog.md
+++ b/src/ResourceManager/Profile/ChangeLog.md
@@ -20,7 +20,8 @@
 ## Current Release
 * *Obsolete*: Save-AzureRmProfile is renamed to Save-AzureRmContext, there is an alias to the old cmdlet name, the alias will be removed in the next release.
 * *Obsolete*: Select-AzureRmProfile is renamed to Import-AzureRmContext, there is an alias to the old cmdlet name, the alias will be removed in the next release.
-* The output types of all profile cmdlets will be changed in the next release.
+* The PSAzureContext and PSAzureProfile output types of profile cmdlets will be changed in the next release.
+* The Save-AzureRmContext cmdlet will have no OutputType in the next release.
 * Fix bug in cmdlet common code to use FIPS-compliant algorithm for data hashes: https://github.com/Azure/azure-powershell/issues/3651 
 
 ## Version 2.7.0


src/ResourceManager/Profile/Commands.Profile/Context/ImportAzureRMContext.cs
```diff
@@ -25,8 +25,10 @@ namespace Microsoft.Azure.Commands.Profile
     /// <summary>
     /// Selects Microsoft Azure profile.
     /// </summary>
-    [Cmdlet(VerbsCommon.Select, "AzureRmProfile"), OutputType(typeof(PSAzureProfile))]
-    public class SelectAzureRMProfileCommand : AzureRMCmdlet
+    [Cmdlet(VerbsData.Import, "AzureRmContext"), OutputType(typeof(PSAzureProfile))]
+    [Alias("Select-AzureRmProfile")]
+    [Obsolete("Select-AzureRmProfile will be renamed to Import-AzureRmContext in the next release.")]
```
shahabhijeet: @markcowl any reason we are not choosing the new name as Select-AzureRmContext?

src/ResourceManager/Profile/Commands.Profile/Context/SaveAzureRMContext.cs
```diff
@@ -26,8 +26,10 @@ namespace Microsoft.Azure.Commands.Profile
     /// <summary>
     /// Saves Microsoft Azure profile.
     /// </summary>
-    [Cmdlet(VerbsData.Save, "AzureRmProfile", SupportsShouldProcess = true), OutputType(typeof(PSAzureProfile))]
-    public class SaveAzureRMProfileCommand : AzureRMCmdlet
+    [Cmdlet(VerbsData.Save, "AzureRmContext", SupportsShouldProcess = true), OutputType(typeof(PSAzureProfile))]
+    [Alias("Save-AzureRmProfile")]
+    [Obsolete("Save-AzureRmProfile will be renamed to Save-AzureRmContext in the next release.", false)]
+    public class SaveAzureRMContextCommand : AzureRMCmdlet
```
shahabhijeet: @markcowl is this the reason it's failing in breaking change analyzer?
What is the reason to change the class name now? This change can be made when we actually introduce Profile commands correct? I am trying to understand why the class names have be changed now, this is not going to cause any conflicts.
Aren't we going to break if people have tests against the actual classes as we do?

markcowl: No, the analyzer is complaining about the ValidateNotNullOrEmpty attribute added to the path parameter.  Since I added an alias to the old cmdlet name, the cmdlet name change is not a breaking change (and the tool doesn't consider it to be one).

shahabhijeet: @markcowl I was just wondering about static analyzer.
But my main concern was class name changes, I know its not a breaking change, but tests will be broken like ours?

src/ResourceManager/Profile/Commands.Profile/Context/ImportAzureRMContext.cs
```diff
@@ -0,0 +1,100 @@
...
+                        throw new PSArgumentException(string.Format(
+                            Microsoft.Azure.Commands.Profile.Properties.Resources.FileNotFound,
+                            Path));
+                    }
+
+                    AzureRmProfileProvider.Instance.Profile = new AzureRMProfile(Path);
+                    executionComplete = true;
+                });
+            }
+            else
```
shahabhijeet: for clarity/readability, can we do this
else if(MyInvovation.BoundParameters.ContainsKey("AzureContext")

src/ResourceManager/Profile/Commands.Profile/Context/SaveAzureRMContext.cs
```diff
@@ -65,7 +67,7 @@ public override void ExecuteCmdlet()
                         throw new ArgumentException(Resources.AzureProfileMustNotBeNull);
                     }
 
-                    if (!AzureSession.DataStore.FileExists(Path) || Force ||
+                    if (!AzureSession.DataStore.FileExists(Path) || Force.IsPresent ||
```
shahabhijeet: Why are we not using the same pattern as previous cmdlet
if (MyInvocation.BoundParameters.ContainsKey("Profile"))?
Also can we get away from writing verbose messages here?
Also any reason we are not executing WriteObject(ProfileObject) at the end?

src/ResourceManager/Profile/Commands.Profile/Context/ImportAzureRMContext.cs
```diff
@@ -0,0 +1,100 @@
...
+        [Parameter(ParameterSetName = InMemoryProfileParameterSet, Mandatory = true, Position = 0, ValueFromPipelineByPropertyName = true)]
+        [ValidateNotNull]
+        [Alias("Profile")]
+        public AzureRMProfile AzureContext { get; set; }
+
+        [Parameter(ParameterSetName = ProfileFromDiskParameterSet, Mandatory = true, Position = 0, ValueFromPipelineByPropertyName = true)]
+        [ValidateNotNullOrEmpty]
+        public string Path { get; set; }
+
+        protected override void BeginProcessing()
```
markcowl: This is not new, this is the existing cmdlet implementation.  It shows up as new because the name changed.

src/ResourceManager/Profile/Commands.Profile/Context/ImportAzureRMContext.cs
```diff
@@ -0,0 +1,100 @@
...
+        public override void ExecuteCmdlet()
+        {
+            bool executionComplete = false;
+            if (MyInvocation.BoundParameters.ContainsKey("Path"))
+            {
+                ConfirmAction(string.Format(Resources.ProcessImportContextFromFile, Path), Resources.ImportContextTarget, () =>
+                {
+                    if (!Common.Authentication.AzureSession.DataStore.FileExists(Path))
+                    {
+                        throw new PSArgumentException(string.Format(
```
markcowl: I don't want this to be a terminating error, it will be caught by the ProcessRecord implementation and written to the error stream

src/ResourceManager/Profile/Commands.Profile/Context/ImportAzureRMContext.cs
markcowl: There is no other parameter set, so this code would be redundant.  And, again, this is existing code that is simply renamed.

src/ResourceManager/Profile/Commands.Profile/Context/ImportAzureRMContext.cs
```diff
@@ -0,0 +1,100 @@
...
+            }
+
+            if (executionComplete)
+            {
+                if (AzureRmProfileProvider.Instance.Profile == null)
+                {
+                    WriteExceptionError( new ArgumentException(Resources.AzureProfileMustNotBeNull));
+                }
+
+                if (AzureRmProfileProvider.Instance.Profile.Context != null &&
```
markcowl: We don't have access to this operator, because we are not using that version of the compiler.  And again, this is just moved existing code.

src/ResourceManager/Profile/Commands.Profile/Context/SaveAzureRMContext.cs
markcowl: Well, because there is only one parameter set and File is always provided.  We are nit checking for a null path, we are checking for file existence to prompt the user for overwrite.

Why would I want to avoid verbose messages? The verbose message is entirely appropriate - it is printed if the user is printing more detailed messages, and is on the verbose stream regardelsss.  

The cmdlet id designed to return nothing ot throw, this is a typical PowerShell pattern for action cmdlets.  Any change in the output would be a breakign change in any case.

src/ResourceManager/Profile/Commands.Profile/Context/SaveAzureRMContext.cs
markcowl: What tests are you referring to?  We own the tests for these cmdlets, and any tests written by others would use the cmdlet name, not the class name.

src/ResourceManager/Profile/Commands.Profile/Context/ImportAzureRMContext.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.Management.Automation;
+
+namespace Microsoft.Azure.Commands.Profile
+{
+    /// <summary>
+    /// Selects Microsoft Azure profile.
+    /// </summary>
+    [Cmdlet(VerbsData.Import, "AzureRmContext", SupportsShouldProcess = true), OutputType(typeof(PSAzureProfile))]
+    [Alias("Select-AzureRmProfile")]
+    [Obsolete("Select-AzureRmProfile will be renamed to Import-AzureRmContext in the next release.")]
```
markcowl: That was the change I was going to make, but looking at the verb defintions, Import is more approrpiate.  Since we are making a reakign change, in the next release, I thought we should use the most appropriate verb and noun

src/ResourceManager/Profile/Commands.Profile/Context/SaveAzureRMContext.cs
markcowl: Any tests outside our unit tests would use the cmdlet name, which is covered by the alias.

src/ResourceManager/Profile/Commands.Profile/Context/ImportAzureRMContext.cs
markcowl: I thought about that, but looking at the verb definitions, Import is more appropriate, and since the cmdlet name is changing anyway, we should take this opportunity to use the appropriate verb

src/ResourceManager/Profile/Commands.Profile/Context/SaveAzureRMContext.cs
markcowl: Because this is not a similar situtation.  In this situation, we are not trying to determien the parameter set used, becasue there is only one parameter set.  Instead, we are checking file existence to give an overwte prompt, if a file will be overwritten

markcowl: Verbose messages are entirely appropriate.  This is the recommended practivce for these kinds of messages in PowerShell cmdlets.

It is also a standard pattern to return nothing if successful (or throw if not) for PowerShell action cmdlets.

Issue #3719, opened 2017-04-03
Some cmdlets return List objects which does not pass through the pipeline 'cleanly'


lfshr:
### Cmdlet(s)
Get-AzureRmDisk
Get-AzureRmStorageAccountKey

### PowerShell Version
PSVersion                      5.1.14393.953
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.14393.953
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1

### Module Version
AzureRM.Automation          2.7.0
AzureRM.Batch               2.7.0
AzureRM.Compute             2.8.0
AzureRM.DataFactories       2.7.0
AzureRM.DataLakeAnalytics   2.7.0
AzureRM.Insights            2.7.0
AzureRM.Network             3.6.0
AzureRM.OperationalInsights 2.7.0
AzureRM.Profile             2.7.0
AzureRM.Resources           3.7.0
AzureRM.Tags                2.7.0

### OS Version
10.0.14393.953

### Description
Some cmdlets return a List of objects which doesn't pass through the pipeline without wrapping the command in parenthesis. I believe this list should be enumerated, and the underlying objects should be written to the pipeline instead.

### Script/Steps for Reproduction
```
Get-AzureRmDisk | Where-Object Name -eq 'GenericDiskName'
```
```
Get-AzureRmStorageAccountKey -ResourceGroupName 'GenericResourceGroupName' -Name 'GenericName' | Select-Object -ExpandProperty Key2
```

shahabhijeet:
@lfshr I believe you were wondering about Get-AzureDisk and not Get-AzureRmDisk (typo maybe?)

lfshr:
@shahabhijeet this refers to AzureRm.Storage\Get-AzureRmDisk. Not the classic portal equivalent. 

https://docs.microsoft.com/en-gb/powershell/module/azurerm.compute/get-azurermdisk?view=azurermps-3.7.0

lfshr:
Beg your pardon. After searching, it's in Compute, not Storage. https://github.com/Azure/azure-powershell/blob/6732a078940ffda7e2a7b47ca7699af925a5deb7/src/ResourceManager/Compute/Commands.Compute/Generated/Disk/DiskGetMethod.cs

lfshr:
[...]

PR #3742, opened 2017-04-08
Fixed pipelining issue with Get-AzureRmStorageAccountKey

lfshr:
## Description
Partial fix for issue #3719 
Currently Get-AzureRmStorageAccountKey returns a List<StorageAccountKey> object, which is passed down the pipeline (as a List) which was not handled correctly. The following command will not work:
```
Get-AzureRmStorageAccountKey -ResourceGroupName <ResourceGroupName> -Name <StorageAccountName> | ? KeyName -eq 'Key1'
```
The List object gets passed through the pipeline instead of the StorageAccountKey objects, so this query would return null. The current work around is to create 2 pipelines:
```
(Get-AzureRmStorageAccountKey -ResourceGroupName <ResourceGroupName> -Name <StorageAccountName>) | ? KeyName -eq 'Key1'
```

This PR strips out the List and writes the objects directly to the pipeline.

---

### DIFF
--- a/src/ResourceManager/Storage/Commands.Management.Storage/StorageAccount/GetAzureStorageAccountKey.cs
+++ b/src/ResourceManager/Storage/Commands.Management.Storage/StorageAccount/GetAzureStorageAccountKey.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Azure.Commands.Management.Storage
                  this.ResourceGroupName,
                  this.Name).Keys;
 
-            WriteObject(storageKeys);
+            WriteObject(storageKeys, true);
         }
     }
 }


markcowl:
@lfshr why not just set enumerate = true on the WriteObject call?

lfshr:
@markcowl worked a charm! Thanks 😄

markcowl:
@azuresdkci add to whitelist

@lfshr I like this, but we will need to get signoff from the storage team, and we will also need to move this to the preview branch

@blueww can you take a look, I think we want to take this

blueww:
My only concern for this change is the output type is changed for List to Array, will that be a breaking change?
I have verfied, Get Key with $key[0].Value, and $key[1].Value still work, so it the function still work well.
@markcowl If the Ourput Type change is OK, I think we can accept it.

lfshr:
@markcowl thanks very much!  I'll keep an eye out for any other cmdlet returning Lists to the pipeline. 😄 

@blueww In my opinion, this is the way all Cmdlets should work. A single object should be written directly to the output stream as soon as it becomes available, otherwise the pipeline becomes a little pointless. The only time I feel an array/list should go into the pipeline is if there's an array of arrays/lists and the use-case makes sense.
You are right though, this is potentially a breaking change. If a user is using any of the List methods that aren't contained in an array, there scripts won't work. But I feel the affected users will be minimal, and it's a change that is necessary for usability and consistency.

markcowl:
@blueww Thanks for taking a look - the output for the cmdlet is still correct, enumerating collections should be the standard

markcowl:
@lfshr  I have duplicated your fix here, since the dev branch is not open for check-ins.  This will make it in for the next release : https://github.com/Azure/azure-powershell/pull/3767

Meanwhile, closing this one

PR #3782, opened 2017-04-17
Add PowerShell cmdlets for Azure Container Registry

djyou:
- New-AzureRmContainerRegistry
 - Get-AzureRmContainerRegistry
 - Update-AzureRmContainerRegistry
 - Remove-AzureRmContainerRegistry
 - Get-AzureRmContainerRegistryCredential
 - Update-AzureRmContainerRegistryCredential
 - Test-AzureRmContainerRegistryNameAvailability

## Description

---

- [ ] **I have read the [contribution guidelines](https://github.com/Azure/azure-powershell/blob/dev/CONTRIBUTING.md).**


setup/azurecmdfiles.wxi
```diff
@@ -686,6 +686,9 @@
                         <Component Id="cmp0BDDC10BCAFF9EC25710B20367203609" Guid="*">
                             <File Id="fil1B9798D9FA9E8F9FC7A4235BCFAB93E4" KeyPath="yes" Source="$(var.sourceDir)\ResourceManager\AzureResourceManager\AzureRM.Compute\AzureRM.Tags.psd1" />
                         </Component>
+                        <Component Id="cmpA3F52D8426DFD1BC61236CCBAAB1BFD1" Guid="*">
+                            <File Id="fil87FE6B06F22A47039BF4D2B02510152A" KeyPath="yes" Source="$(var.sourceDir)\ResourceManager\AzureResourceManager\AzureRM.Compute\AzureRmNetworkStartup.ps1" />
+                        </Component>
```
djyou: @markcowl These items kept showing up in this file even if I started from a clean repo. Are they missed from previous commits/PRs or is there another way to clean them up (besides manually)?

cormacpayne:
@azuresdkci add to whitelist

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/GetAzureContainerRegistry.cs
```diff
@@ -0,0 +1,84 @@
...
+                while (!string.IsNullOrEmpty(registries.NextPageLink))
+                {
+                    registries = RegistryClient.ListRegistriesUsingNextLink(ResourceGroupName, registries.NextPageLink);
+
+                    list = new List<PSContainerRegistry>();
+                    foreach (Registry registry in registries)
+                    {
+                        list.Add(new PSContainerRegistry(registry));
+                    }
+                    WriteObject(list, true);
```
cormacpayne: @djyou I'm not 100% sure what happens when you write an enumerated collection to the pipeline a page at a time, but I would recommend doing one of the following:

1) Populate the `registries` list with every page of the results, and then write to the pipeline _once_ after the while loop has terminated.

2) For the individual objects to the pipeline rather than a page at a time. For example, the while loop would instead look like the following (as well as the section before the while loop):

```cs
while (!string.IsNullOrEmpty(registries.NextPageLink))
{
    registries = RegistryClient.ListRegistriesUsingNextLink(ResourceGroupName, registries.NextPageLink);
    foreach (Registry registry in registries)
    {
        WriteObject(new PSContainerRegistry(registry);
    }
}
```

Both of these scenarios try to avoid the possibility of writing groups of objects to the pipeline and causing unusual behavior. It would be better to write each object as you get them, or write all objects at once.

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/TestAzureContainerRegistryNameAvailability.cs
```diff
@@ -0,0 +1,37 @@
...
+// limitations under the License.
+// ----------------------------------------------------------------------------------
+
+using System.Management.Automation;
+using Microsoft.Azure.Management.ContainerRegistry.Models;
+
+namespace Microsoft.Azure.Commands.ContainerRegistry
+{
+    [Cmdlet(VerbsDiagnostic.Test, ContainerRegistryNameAvailabilityNoun), OutputType(typeof(RegistryNameStatus))]
+    public class GetAzureContainerRegistryNameAvailability : ContainerRegistryCmdletBase
```
cormacpayne: @djyou class name is incorrect

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/TestAzureContainerRegistryNameAvailability.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        [Parameter(
+            Position = 1,
```
cormacpayne: @djyou position should be 0 since it's the first (and only) parameter

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/NewAzureContainerRegistry.cs
```diff
@@ -0,0 +1,106 @@
...
+            HelpMessage = "Container Registry Location. Default to the location of the resource group.")]
+        [ValidateNotNullOrEmpty]
+        public string Location { get; set; }
+
+        [Parameter(
+            Mandatory = false,
+            HelpMessage = "Indicates whether the admin user is enabled.")]
+        [ValidateNotNull]
+        [Alias(AdminEnabledAlias)]
+        public bool? AdminUserEnabled { get; set; }
```
cormacpayne: @djyou we're still waiting to hear back on the correct pattern for this issue

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/NewAzureContainerRegistry.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            if (Location == null)
+            {
+                Location = ResourceManagerClient.GetResourceGroupLocation(ResourceGroupName);
+            }
+
+            DeploymentExtended result = ResourceManagerClient.CreateRegistry(
+                ResourceGroupName, Name, Location, Sku, AdminUserEnabled, StorageAccountName, tags);
+
+            if (result.Properties.ProvisioningState == DeploymentState.Succeeded.ToString())
```
cormacpayne: @djyou do we want to print anything if the create was unsuccessful?

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/NewAzureContainerRegistry.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.Collections;
+using System.Management.Automation;
+using Microsoft.Azure.Management.ContainerRegistry.Models;
+using Microsoft.Azure.Management.ResourceManager.Models;
+using Microsoft.Azure.Commands.ResourceManager.Common.Tags;
+using DeploymentState = Microsoft.Azure.Commands.ResourceManager.Cmdlets.Entities.ProvisioningState;
+
+namespace Microsoft.Azure.Commands.ContainerRegistry
+{
+    [Cmdlet(VerbsCommon.New, ContainerRegistryNoun), OutputType(typeof(PSContainerRegistry))]
```
cormacpayne: @djyou this will need to implement `SupportsShouldProcess` for this cmdlet since it is creating a resource in Azure

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/RemoveAzureContainerRegistry.cs
```diff
@@ -0,0 +1,47 @@
...
+        [Parameter(
+            Position = 1,
+            Mandatory = true,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Container Registry Name.")]
+        [Alias(ContainerRegistryNameAlias, RegistryNameAlias, ResourceNameAlias)]
+        [ValidateNotNullOrEmpty]
+        public string Name { get; set; }
+
+        public override void ExecuteCmdlet()
```
cormacpayne: @djyou there should be a parameter set for the registry container object that you plan on removing, which can be passed through by value in the pipeline, enabling the following scenario:

```powershell
Get-AzureRmContainerRegistry | Remove-AzureRmContainerRegistry
```

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/UpdateAzureContainerRegistry.cs
```diff
@@ -0,0 +1,75 @@
...
+// ----------------------------------------------------------------------------------
+
+using System.Collections;
+using System.Management.Automation;
+using Microsoft.Azure.Commands.ResourceManager.Common.Tags;
+
+namespace Microsoft.Azure.Commands.ContainerRegistry
+{
+    [Cmdlet(VerbsData.Update, ContainerRegistryNoun), OutputType(typeof(PSContainerRegistry))]
+    public class UpdateAzureContainerRegistry : ContainerRegistryCmdletBase
```
cormacpayne: @djyou this cmdlet should implement `SupportsShouldProcess` since it is making changes to an object in Azure

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/UpdateAzureContainerRegistry.cs
```diff
@@ -0,0 +1,75 @@
...
+            HelpMessage = "Container Registry Tags.")]
+        [ValidateNotNull]
+        [Alias(TagsAlias)]
+        public Hashtable Tag { get; set; }
+
+        [Parameter(
+            Mandatory = false,
+            HelpMessage = "The name of an existing storage account.")]
+        public string StorageAccountName { get; set; }
+
```
cormacpayne: @djyou same comment about adding a parameter to allow passing in the container registry object through the pipeline

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/UpdateAzureContainerRegistryCredential.cs
```diff
@@ -0,0 +1,58 @@
...
+        public string Name { get; set; }
+
+        [Parameter(
+            Position = 2,
+            Mandatory = true,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "The name of password to regenerate. " + AllowedPasswordNames)]
+        [ValidateNotNullOrEmpty]
+        public PasswordName PasswordName { get; set; }
+
```
cormacpayne: @djyou same comment about adding a parameter to allowing passing in a container registry object from the pipeline

src/ResourceManager/ContainerRegistry/ChangeLog.md
```diff
@@ -0,0 +1,30 @@
...
+    * Overview of change #4
+        - Additional information about change #4
+
+    ## YYYY.MM.DD - Version X.Y.Z (Previous Release)
+    * Overview of change #1
+        - Additional information about change #1
+-->
+## Current Release
+
+## Version 1.0.0
```
djyou: Without a version number? It doesn't seem to be the pattern other modules follow, can you confirm?

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands.ContainerRegistry.csproj
```diff
@@ -0,0 +1,179 @@
...
+    </ProjectReference>
+    <ProjectReference Include="..\..\..\Common\Commands.Common\Commands.Common.csproj">
+      <Project>{5ee72c53-1720-4309-b54b-5fb79703195f}</Project>
+      <Name>Commands.Common</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\..\Common\Commands.ResourceManager.Common\Commands.ResourceManager.Common.csproj">
+      <Project>{3819d8a7-c62c-4c47-8ddd-0332d9ce1252}</Project>
+      <Name>Commands.ResourceManager.Common</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\..\Resources\Commands.ResourceManager\Cmdlets\Commands.Resources.Rest.csproj">
```
djyou: We take dependency on template deployment and polling logic provided therein.

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/NewAzureContainerRegistry.cs
djyou: The polling logic of template deployment will print the error message if it is unsuccessful. This if block here is just another GET to print the created registry object if it is successful.

src/ResourceManager/ContainerRegistry/ChangeLog.md
cormacpayne: @djyou that's our pattern - any changes going out in the next release will be under the "Current Release" header, and changes that have gone out in prior releases are documented in the respective version header. We run a script during the release process that will move the information from the "Current Release" section into their own version section.

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/RemoveAzureContainerRegistry.cs
djyou: I am not getting what will be achieved by adding another parameter set here. This piping already works now. Is there a scenario where ValueFromPipelineByPropertyName doesn't work, or do you have an example from another command module that does this?

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/UpdateAzureContainerRegistry.cs
djyou: This is a PATCH operation and I thought we wouldn't enable passing in a container registry object as discussed?

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/RemoveAzureContainerRegistry.cs
cormacpayne: @djyou ahh yes, you are right 😄 ignore this comment

src/ResourceManager/ContainerRegistry/ChangeLog.md
djyou: It seems odd to have change logs without saying what version it is (current doesn't say version though). Is this something new since I didn't find another module doing what you were suggesting?

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/UpdateAzureContainerRegistry.cs
cormacpayne: @djyou ignore this comment as well 😄

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/UpdateAzureContainerRegistryCredential.cs
djyou: A general question, I am not quite clear on whether ValueFromPipelineByPropertyName and piping-by-object should both be implemented, or either one should be implemented. Is there a scenario where ValueFromPipelineByPropertyName doesn't work but piping-by-object works, or vice versa? Can you offer an example or documentation on passing objects? Thanks.

djyou:
@cormacpayne Thanks a lot for the detailed comments. I will fix the ones I didn't comment and the ones after clarification from you. /cc @SajayAntony

src/ResourceManager/ContainerRegistry/ChangeLog.md
cormacpayne: @djyou I suppose the header should say "Upcoming Release" instead of "Current Release". Regardless, this is the pattern that we have followed for every RP:

- Service makes change(s) to their cmdlet(s)
- Change log is updated (under the "Current Release" section) to update these changes
- During release, module versions are updated, and "Current Release" header is changed to reflect the released version, and a new "Current Release" header is created
- Customers can view that 1.0.0 of ContainerRegistry was just released and can view the notes about the changes under the "Version 1.0.0" section of the change log (or in the release notes on PowerShell Gallery)

djyou: Makes sense. Thank you.

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/UpdateAzureContainerRegistryCredential.cs
cormacpayne: @djyou we try to push towards piping by object instead of by property name as often as possible. A good example would be if you had a complex object with a large number of properties that you were interesting in knowing (possibly for a PUT update cmdlet). It wouldn't be reasonable to have a parameter for _every single property_ that you were interested in and giving them the `ValueFromPipelineByPropertyName` attribute; this would be a bad experience for users who weren't piping the object, and it would take a little longer to execute as PowerShell would need to resolve the parameter bindings. Instead, if you accepted the complex object from the pipeline, you would only need one parameter, and it would take less time to resolve. Let me know if that answers your question.

djyou: So if this command supports piping by object, should `ValueFromPipelineByPropertyName = true` be removed from `ResourceGroupName` and `Name`? It seems that if the default parameter set is using `ResourceGroupName` and `Name`, piping will prioritize them over object.

cormacpayne: @djyou for all of your cmdlets (including this one), because you only need two parameters (`ResourceGroupName` and `Name`) to determine the registry container that you are performing an action on, using `ValueFromPipelineByPropertyName` for those two parameters instead of introducing a new parameter that will have the `ValueFromPipeline` attribute will definitely work. We try to push people towards passing objects into a cmdlet from the pipeline when they can, but in this case, because the number of parameters required is so small, it works just to use those instead of the object.

src/ResourceManager/ContainerRegistry/AzureRM.ContainerRegistry.psd1
```diff
@@ -0,0 +1,132 @@
...
+# Generated on: 3/8/2017
+#
+
+@{
+
+# Script module or binary module file associated with this manifest.
+# RootModule = ''
+
+# Version number of this module.
+ModuleVersion = '1.0.0'
```
cormacpayne: @djyou can we start the module version at 0.1.0?

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry.Test/ScenarioTests/ContainerRegistryTests.ps1
```diff
@@ -0,0 +1,167 @@
...
+<#
+.SYNOPSIS
+Test New-AzureRmContainerRegistry, Get-AzureRmContainerRegistry, Update-AzureRmContainerRegistry, and Remove-AzureRmContainerRegistry.
+#>
+function Test-AzureContainerRegistry
+{
+    # Setup
+    $resourceGroupName = Get-RandomResourceGroupName
+    $registryName = Get-RandomRegistryName
+    $location = "southcentralus"
```
cormacpayne: @djyou any reason you don't get the location from the `Common.ps1` script? We want to avoid hard-coding location values in tests as it makes it difficult for other users to re-record them in the future

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands.ContainerRegistry.csproj
```diff
@@ -0,0 +1,179 @@
...
+    <Compile Include="Models\DeploymentTemplate.cs" />
+    <Compile Include="Models\PSContainerRegistry.cs" />
+    <Compile Include="Models\ContainerRegistryClient.cs" />
+    <Compile Include="Models\ContainerRegistryCmdletBase.cs" />
+    <Compile Include="Models\PSContainerRegistryCredential.cs" />
+    <Compile Include="Models\ResourceManagerClient.cs" />
+    <Compile Include="Properties\AssemblyInfo.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <Content Include="Microsoft.Azure.Commands.ContainerRegistry.dll-Help.xml">
```
cormacpayne: @djyou we no longer have dll-Help.xml files in the repository, so you can remove this after pulling the latest changes from the `preview` branch

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/NewAzureContainerRegistry.cs
```diff
@@ -0,0 +1,110 @@
...
+            HelpMessage = "Container Registry Location. Default to the location of the resource group.")]
+        [ValidateNotNullOrEmpty]
+        public string Location { get; set; }
+
+        [Parameter(
+            Mandatory = false,
+            HelpMessage = "Indicates whether the admin user is enabled.")]
+        [ValidateNotNull]
+        [Alias(AdminEnabledAlias)]
+        public bool? AdminUserEnabled { get; set; }
```
cormacpayne: @djyou we talked to the team and decided the following:

- `New-*` will have a SwitchParameter for `EnableAdminUser`
- `Update-*` will have two SwitchParameters for `EnableAdminUser` and `DisableAdminUser` that are in separate parameter sets

Let me know if you have any questions about this design

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/UpdateAzureContainerRegistry.cs
```diff
@@ -0,0 +1,79 @@
...
+        [Alias(ContainerRegistryNameAlias, RegistryNameAlias, ResourceNameAlias)]
+        [ValidateNotNullOrEmpty]
+        public string Name { get; set; }
+
+        [Parameter(
+            Mandatory = false,
+            HelpMessage = "Indicates whether the admin user is enabled.")]
+        [ValidateNotNull]
+        [Alias(AdminEnabledAlias)]
+        public bool? AdminUserEnabled { get; set; }
```
cormacpayne: @djyou see previous comment

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands/NewAzureContainerRegistry.cs
djyou: This is fine to us. Just a minor question, are they in separate parameter sets so that they couldn't be specified together?

cormacpayne: @djyou exactly, they will be in separate parameter sets where one will be mandatory, and the other will be optional. This gives your cmdlet the following three scenarios:
- neither Enable or Disable
- only Enable
- only Disable

Also, I said previously that `New-*` will have a SwitchParameter for `EnableAdminUser` - I don't remember what is set by default, so if by default you want admin user to be _disabled_, then add the switch for `EnableAdminUser`, but if admin user is _enabled_ by default, then add a switch for `DisableAdminUser`

djyou: Thank you. It is disabled by default so `EnableAdminUser` is good.

src/ResourceManager/ContainerRegistry/Commands.ContainerRegistry/Commands.ContainerRegistry.csproj
```diff
@@ -0,0 +1,179 @@
...
+    <ProjectReference Include="..\..\Common\Commands.ResourceManager.Common\Commands.ResourceManager.Common.csproj">
+      <Project>{3819d8a7-c62c-4c47-8ddd-0332d9ce1252}</Project>
+      <Name>Commands.ResourceManager.Common</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\..\Resources\Commands.ResourceManager\Cmdlets\Commands.Resources.Rest.csproj">
+      <Project>{8058d403-06e3-4bed-8924-d166ce303961}</Project>
+      <Name>Commands.Resources.Rest</Name>
+    </ProjectReference>
+  </ItemGroup>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
```
cormacpayne: @djyou one additional note, you will need to add the following snippet to this file so the markdown files get copied over to the output directory: https://github.com/Azure/azure-powershell/blob/preview/src/ResourceManager/Profile/Commands.Profile/Commands.Profile.csproj#L193-L198

cormacpayne:
Sign job: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-sign/988/

LGTM once completed and package successfully built and signed in Build Artifacts

markcowl:
on demand run here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1495/

PR #3787, opened 2017-04-17
adding support for backing up and restoring of KeyVault secrets

dragav:
## Description

---

- [X] **I have read the [contribution guidelines](https://github.com/Azure/azure-powershell/blob/dev/CONTRIBUTING.md).**
- [X] **If changes were made to any cmdlet, the XML help was regenerated using the [platyPSHelp module](https://github.com/Azure/azure-powershell/blob/dev/documentation/platyPSHelp-documentation.md).**
- [X] **If any large changes are made to a service, they are reflected in the respective [change log](https://github.com/Azure/azure-powershell/blob/dev/CONTRIBUTING.md#updating-the-change-log).**

- [X] Title of the pull request is clear and informative.
- [X] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR. For more information on cleaning up the commits in your PR, [see this page](https://github.com/Azure/azure-powershell/blob/dev/documentation/cleaning-up-commits.md).
- [X] The pull request does not introduce [breaking changes](https://github.com/Azure/azure-powershell/blob/dev/documentation/breaking-changes.md) (unless a major version change occurs in the assembly and module).

- [X] Pull request includes test coverage for the included changes.
- [X] PowerShell scripts used in tests should do any necessary setup as part of the test or suite setup, and should not use hard-coded values for locations or existing resources.

- [X] New cmdlets that make changes or have side effects should implement `ShouldProcess` and have `SupportShouldProcess=true` specified in the cmdlet attribute. You can find more information on `ShouldProcess` [here](https://gist.github.com/markcowl/338e16fe5c8bbf195aff9f8af0db585d#what-is-the-change).
- [X] Cmdlet specifies `OutputType` attribute if any output is produced - if the cmdlet produces no output, it should implement a `PassThru` parameter.

- [X] Parameter types should not expose types from the management library - complex parameter types should be defined in the module.
- [X] Complex parameter types are discouraged - a parameter type should be simple types as often as possible. If complex types are used, they should be shallow and easily creatable from a constructor or another cmdlet.
- [X] Cmdlet parameter sets should be mutually exclusive - each parameter set must have at least one mandatory parameter not in other parameter sets.

### DIFF
--- a/src/ResourceManager/KeyVault/AzureRM.KeyVault.psd1
+++ b/src/ResourceManager/KeyVault/AzureRM.KeyVault.psd1
@@ -98,7 +98,9 @@ CmdletsToExport = 'Add-AzureKeyVaultCertificate',
                'Set-AzureKeyVaultSecretAttribute', 
                'Get-AzureKeyVaultCertificatePolicy', 
                'New-AzureKeyVaultCertificateAdministratorDetails', 
-               'New-AzureKeyVaultCertificateOrganizationDetails'
+               'New-AzureKeyVaultCertificateOrganizationDetails',
+               'Backup-AzureKeyVaultSecret',
+               'Restore-AzureKeyVaultSecret'
 
 # Variables to export from this module
 # VariablesToExport = @()
--- a/src/ResourceManager/KeyVault/Commands.KeyVault/Commands/BackupAzureKeyVaultSecret.cs
+++ b/src/ResourceManager/KeyVault/Commands.KeyVault/Commands/BackupAzureKeyVaultSecret.cs
@@ -22,61 +22,61 @@ namespace Microsoft.Azure.Commands.KeyVault
     /// <summary>
     /// Requests that a backup of the specified key be downloaded and stored to a file
     /// </summary>
-    [Cmdlet(VerbsData.Backup, "AzureKeyVaultKey",
-        SupportsShouldProcess = true, 
-        HelpUri = Constants.KeyVaultHelpUri)]
-    [OutputType(typeof(String))]
-    public class BackupAzureKeyVaultKey : KeyVaultCmdletBase
+    [Cmdlet( VerbsData.Backup, "AzureKeyVaultSecret",
+        SupportsShouldProcess = true,
+        HelpUri = Constants.KeyVaultHelpUri )]
+    [OutputType( typeof( String ) )]
+    public class BackupAzureKeyVaultSecret : KeyVaultCmdletBase
     {
         #region Input Parameter Definitions
 
         /// <summary>
         /// Vault name
         /// </summary>
-        [Parameter(Mandatory = true,
+        [Parameter( Mandatory = true,
                    Position = 0,
                    ValueFromPipelineByPropertyName = true,
-                   HelpMessage = "Vault name. Cmdlet constructs the FQDN of a vault based on the name and currently selected environment.")]
-        [ValidateNotNullOrEmpty]        
+                   HelpMessage = "Vault name. Cmdlet constructs the FQDN of a vault based on the name and currently selected environment." )]
+        [ValidateNotNullOrEmpty]
         public string VaultName { get; set; }
 
         /// <summary>
         /// Key name
         /// </summary>
-        [Parameter(Mandatory = true,
+        [Parameter( Mandatory = true,
                    Position = 1,
                    ValueFromPipelineByPropertyName = true,
-                   HelpMessage = "Key name. Cmdlet constructs the FQDN of a key from vault name, currently selected environment and key name.")]
-        [ValidateNotNullOrEmpty]        
-        [Alias(Constants.KeyName)]
+                   HelpMessage = "Secret name. Cmdlet constructs the FQDN of a secret from vault name, currently selected environment and secret name." )]
+        [ValidateNotNullOrEmpty]
+        [Alias( Constants.SecretName )]
         public string Name { get; set; }
 
         /// <summary>
         /// The output file in which the backup blob is to be stored
         /// </summary>
-        [Parameter(Mandatory = false,
+        [Parameter( Mandatory = false,
                    Position = 2,
                    ValueFromPipelineByPropertyName = true,
-                   HelpMessage = "Output file. The output file to store the backed up key blob in. If not present, a default filename is chosen.")]
+                   HelpMessage = "Output file. The output file to store the backed up secret blob in. If not present, a default filename is chosen." )]
         [ValidateNotNullOrEmpty]
         public string OutputFile { get; set; }
 
         #endregion Input Parameter Definition
 
-        public override void ExecuteCmdlet()
+        public override void ExecuteCmdlet( )
         {
-            if (ShouldProcess(Name, Properties.Resources.BackupKey))
+            if ( ShouldProcess( Name, Properties.Resources.BackupSecret ) )
             {
-                if (string.IsNullOrEmpty(OutputFile))
+                if ( string.IsNullOrEmpty( OutputFile ) )
                 {
                     OutputFile = GetDefaultFileForOperation("backup", VaultName, Name);
                 }
 
-                var filePath = ResolvePathFromFilename(OutputFile, throwOnPreExisting: true, errorMessage: KeyVaultProperties.Resources.BackupKeyFileAlreadyExists);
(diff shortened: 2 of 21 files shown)


src/ResourceManager/KeyVault/AzureRM.KeyVault.psd1
```diff
@@ -98,7 +98,9 @@ CmdletsToExport = 'Add-AzureKeyVaultCertificate',
                'Set-AzureKeyVaultSecretAttribute', 
                'Get-AzureKeyVaultCertificatePolicy', 
                'New-AzureKeyVaultCertificateAdministratorDetails', 
-               'New-AzureKeyVaultCertificateOrganizationDetails'
+               'New-AzureKeyVaultCertificateOrganizationDetails',
```
markcowl: Please add these cmdlet changes in your Changelog for KeyVault.

Also, please update the minor version in your assemblyinfo.cs and in your psd1 file

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/BackupAzureKeyVaultSecret.cs
```diff
@@ -0,0 +1,83 @@
...
+        public override void ExecuteCmdlet( )
+        {
+            if ( ShouldProcess( Name, Properties.Resources.BackupSecret ) )
+            {
+                if ( string.IsNullOrEmpty( OutputFile ) )
+                {
+                    OutputFile = GetDefaultFileForOperation("backup", VaultName, Name);
+                }
+
+                var filePath = ResolvePathFromFilename(OutputFile, throwOnPreExisting: true, errorMessage: KeyVaultProperties.Resources.BackupSecretFileAlreadyExists);
```
markcowl: Two things:

(1) You should be using AzureSession.DataStore when interacting with the file system (checking for existence and the like), as this is mockable in tests
(2) The normal standrad would be to prompt for overwite if the file exists, which would mena having a ShouldContinue prompt asking if the user wants to overwrite and a Force parameter.  Is there a reason why you wouldn't want to allow the user to overwrite the file?

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/BackupAzureKeyVaultSecret.cs
```diff
@@ -0,0 +1,83 @@
...
+
+namespace Microsoft.Azure.Commands.KeyVault
+{
+    /// <summary>
+    /// Requests that a backup of the specified key be downloaded and stored to a file
+    /// </summary>
+    [Cmdlet( VerbsData.Backup, "AzureKeyVaultSecret",
+        SupportsShouldProcess = true,
+        HelpUri = Constants.KeyVaultHelpUri )]
+    [OutputType( typeof( String ) )]
```
markcowl: Not sure the OutputType is useful.  You should probably return nothing and implement a passthrough parameter so you can return true/false for success / failure

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/BackupAzureKeyVaultSecret.cs
```diff
@@ -0,0 +1,83 @@
...
+    public class BackupAzureKeyVaultSecret : KeyVaultCmdletBase
+    {
+        #region Input Parameter Definitions
+
+        /// <summary>
```
markcowl: You should consider having an alternate parameter set that takes a KeyVaultSecret so that users can more easily pipe get-AzureKeyVaultSecret to this cmdlet.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/RestoreAzureKeyVaultSecret.cs
```diff
@@ -0,0 +1,76 @@
...
+
+                var restoredSecret = this.DataServiceClient.RestoreSecret(VaultName, filePath);
+
+                this.WriteObject( restoredSecret );
+            }
+        }
+
+        private string ResolvePath( string filePath )
+        {
+            FileInfo secretFile = new FileInfo(this.GetUnresolvedProviderPathFromPSPath(filePath));
```
markcowl: Please use the data store, as outlined above

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/RestoreAzureKeyVaultSecret.cs
```diff
@@ -0,0 +1,76 @@
...
+
+namespace Microsoft.Azure.Commands.KeyVault
+{
+    /// <summary>
+    /// Restores the backup secret into a vault 
+    /// </summary>
+    [Cmdlet( VerbsData.Restore, "AzureKeyVaultSecret",
+        SupportsShouldProcess = true,
+        HelpUri = Constants.KeyVaultHelpUri )]
+    [OutputType( typeof( Secret ) )]
```
markcowl: It looks like you are returnign the secret FullName, isn't that a string?

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/RestoreAzureKeyVaultSecret.cs
```diff
@@ -0,0 +1,76 @@
...
+        public string VaultName { get; set; }
+
+        /// <summary>
+        /// The input file in which the backup blob is stored
+        /// </summary>
+        [Parameter( Mandatory = true,
+                   Position = 1,
+                   HelpMessage = "Input file. The input file containing the backed-up blob" )]
+        [ValidateNotNullOrEmpty]
+        public string InputFile { get; set; }
```
markcowl: It would be nice to take a list of files in this parameter, and process each one

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/RestoreAzureKeyVaultSecret.cs
dragav: Restore returns the newly reinstated secret (or key, for existing cmdlets). It is similar in that sense with the Get command.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/RestoreAzureKeyVaultSecret.cs
dragav: Will add this to the backlog for a future release :)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/BackupAzureKeyVaultSecret.cs
dragav: I don't quite see the value of this scenario - the piped-in object would only serve to extract the identifier, as that's what the server expects. The service must find/load/encrypt the object before creating a backup - the client can't upload an object to be backed up. 

Another thing to consider is that the secret can be retrieved by id (ie a specific version) whereas the backup acts on the entire family (head secret + all versions). I'm concerned that the user expectation (on passing in a specific secret object) will be that an individual backup of that object will be generated.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/RestoreAzureKeyVaultSecret.cs
dragav: We are returning the newly restored secret (essentially saving the user a Get call).

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/RestoreAzureKeyVaultSecret.cs
markcowl: Sounds good, this would be a non-breaking change

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/BackupAzureKeyVaultSecret.cs
dragav: The output of this cmdlet is the file path of the newly generated backup file.

cormacpayne:
@dragav please pull the latest changes from the `preview` branch - these changes remove dll-Help.xml files from the repository, so you will no longer need to update this file in your PR

dragav:
Addressed feedback and ported these changes to the corresponding preview branch. Will resubmit the PR shortly.

PR #3791, opened 2017-04-18
Update appveyor.yml to use external script

qinezh:
@markcowl

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,7 +7,6 @@ environment:
     secure: VMFbecLLHzDq/09YDPbcM0VDDSwwgY57vr5GXK6cZZ4Ti/Xs5RZoylzV8MMr1350
 
 before_build:
-  - set path=%programfiles(x86)%\Microsoft SDKs\TypeScript\2.2;%path%
   - ps: Install-Product node $env:nodejs_version
   - git config --global credential.helper store
   - ps: ac "$env:USERPROFILE\.git-credentials" "https://$($env:github_access_token):x-oauth-basic@github.com`n"
@@ -16,8 +15,8 @@ before_build:
   - git clone -q https://github.com/MicrosoftDocs/azure-psdocs-ciscripts azpsci
   - cd azpsci
   - npm install
-  - npm install -g typescript
-  - tsc
+  - npm install typescript
+  - node ./node_modules/typescript/bin/tsc
   - cd ..
 build_script:
   - node azpsci/out/ci0.js src c:\projects\_output


appveyor.yml
```diff
@@ -2,61 +2,30 @@ branches:
...
+  - cd ..
+build_script:
+  - node azpsci/out/ci0.js src c:\projects\_output
+
+test: off
+on_success:
   - git clone -q --branch=%target_branch% %content_repo% %TEMP%\Azure
   - cd %TEMP%\Azure
-  - ps: ls $global:output -dir | % { copy $_.FullName (ls -dir | select -First 1) -Recurse -Force }
+  - ps: ls c:\projects\_output -dir | % { copy $_.FullName (ls -dir | select -First 1) -Recurse -Force }
```
twitchax: Why change this away from a build variable?

markcowl: Cannot assume this will be on c: drive

appveyor.yml
```diff
@@ -2,61 +2,30 @@ branches:
...
-          {
-            robocopy $help "*.md" (Join-Path $global:output $folder | Join-Path -ChildPath (gi $psd1).BaseName | Join-Path -ChildPath ("v" + $matches[0]))
-          }
-        }
-      }
-test: off
-shallow_clone: true
-on_success:
+before_build:
+  - set path=%programfiles(x86)%\Microsoft SDKs\TypeScript\2.2;%path%
```
markcowl: This is making all kinds of assumptions about the environment it is running in.  What are the required environment variables, and why is it assumign it is runnign on a 64-bit system?

appveyor.yml
```diff
@@ -2,61 +2,30 @@ branches:
...
+  - ps: Install-Product node $env:nodejs_version
   - git config --global credential.helper store
   - ps: ac "$env:USERPROFILE\.git-credentials" "https://$($env:github_access_token):x-oauth-basic@github.com`n"
   - git config --global user.email %email%
   - git config --global user.name %name%
+  - git clone -q https://github.com/MicrosoftDocs/azure-psdocs-ciscripts azpsci
+  - cd azpsci
+  - npm install
+  - npm install -g typescript
```
markcowl: Why install this globally?

appveyor.yml
markcowl: This cannot assume it's working on the c drive

qinezh:
There's an assumption, the script is for AppVeyor. Based on the assumption, the script is running on the c drive according to https://www.appveyor.com/docs/appveyor-yml.

appveyor.yml
qinezh: @markcowl  I think it works even if the script not working on the c drive.

qinezh: @twitchax Beacuse it doesn't need the variable anymore.

PR #3826, opened 2017-04-26
Fix, update, and add tests for ResourceManager SQL Failover Group cmdlets

dlemMSFT:
## Description
This pull request contains a number of changes to the Resource Manager SQL Failover Group cmdlets. There are several breaking changes -- they are removals of parameters that were previously declared obsolete.

Here is a list of the changes. This is mostly copied from the ChangeLog update.

- Remove 'Tag' parameters (previously obsoleted)
- Remove 'PartnerResourceGroupName' and 'PartnerServerName' parameters from Remove-AzureRmSqlDatabaseFailoverGroup cmdlet (previously obsoleted)
- Add 'GracePeriodWithDataLossHours' parameter to New- and Set- cmdlets, which shall eventually replace 'GracePeriodWithDataLossHour'
- Documentation has been fleshed out and updated
- Change formatting of returned objects and fix some bugs where fields were not always populated
- Add 'DatabaseNames' and 'PartnerLocation' properties to Failover Group object
- Fix bug causing Switch- cmdlet to return immediately rather than waiting for operation to complete
- Fix integer overflow bug when high grace period values are used
- Adjust grace period to a minimum of 1 hour and warn user if a lower one is provided
- Add and record scenario tests for Failover Group cmdlets

---

- [x] (No new cmdlets) New cmdlets that make changes or have side effects should implement `ShouldProcess` and have `SupportShouldProcess=true` specified in the cmdlet attribute. You can find more information on `ShouldProcess` [here](https://gist.github.com/markcowl/338e16fe5c8bbf195aff9f8af0db585d#what-is-the-change).
- [x] (No new cmdlets) Cmdlet specifies `OutputType` attribute if any output is produced - if the cmdlet produces no output, it should implement a `PassThru` parameter.

### DIFF
--- a/src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/AzureSqlFailoverGroupCmdletBase.cs
+++ b/src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/AzureSqlFailoverGroupCmdletBase.cs
@@ -42,15 +42,10 @@ namespace Microsoft.Azure.Commands.Sql.FailoverGroup.Cmdlet
         {
             int? gracePeriod = null;
 
-            // Respect both parameters, but the non-obselete one takes precedence.
-            foreach (var parameterName in new[] { "GracePeriodWithDataLossHours", "GracePeriodWithDataLossHour" })
+            object parameterValue;
+            if (MyInvocation.BoundParameters.TryGetValue("GracePeriodWithDataLossHours", out parameterValue))
             {
-                object parameterValue;
-                if (MyInvocation.BoundParameters.TryGetValue(parameterName, out parameterValue))
-                {
-                    gracePeriod = (int)parameterValue;
-                    break;
-                }
+                gracePeriod = (int)parameterValue;
             }
 
             if (!gracePeriod.HasValue && policy == FailoverPolicy.Automatic)
--- a/src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/NewAzureSqlFailoverGroup.cs
+++ b/src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/NewAzureSqlFailoverGroup.cs
@@ -83,17 +83,6 @@ namespace Microsoft.Azure.Commands.Sql.FailoverGroup.Cmdlet
         public int GracePeriodWithDataLossHours { get; set; }
 
         /// <summary>
-        /// Gets or sets the grace period with data loss for the Sql Azure Failover Group.
-        /// </summary>
-        [Parameter(Mandatory = false,
-            HelpMessage = "Interval before automatic failover is initiated if an outage occurs on the primary server and failover cannot be completed without data loss.")]
-        [ValidateNotNullOrEmpty]
-        [ValidateRange(0, int.MaxValue)]
-        [PSDefaultValue(Help = "1")]
-        [Obsolete("This parameter will be deprecated in the next release.")]
-        public int GracePeriodWithDataLossHour { get; set; }
-
-        /// <summary>
         /// Gets or sets the failover policy for read only endpoint of the Sql Azure Failover Group.
         /// </summary>
         [Parameter(Mandatory = false,
--- a/src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/SetAzureSqlFailoverGroup.cs
+++ b/src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/SetAzureSqlFailoverGroup.cs
@@ -70,17 +70,6 @@ namespace Microsoft.Azure.Commands.Sql.FailoverGroup.Cmdlet
         public int GracePeriodWithDataLossHours { get; set; }
 
         /// <summary>
-        /// Gets or sets the grace period with data loss for the Sql Azure Failover Group.
-        /// </summary>
-        [Parameter(Mandatory = false,
-            HelpMessage = "Interval before automatic failover is initiated if an outage occurs on the primary server and failover cannot be completed without data loss.")]
-        [ValidateNotNullOrEmpty]
-        [ValidateRange(0, int.MaxValue)]
-        [PSDefaultValue(Help = "1")]
-        [Obsolete("This parameter will be deprecated in the next release.")]
-        public int GracePeriodWithDataLossHour { get; set; }
-
-        /// <summary>
         /// Gets or sets the failover policy for read only endpoint of the Sql Azure Failover Group.
         /// </summary>
         [Parameter(Mandatory = false,
--- a/src/ResourceManager/Sql/Commands.Sql.Test/ScenarioTests/FailoverGroupTests.ps1
+++ b/src/ResourceManager/Sql/Commands.Sql.Test/ScenarioTests/FailoverGroupTests.ps1
@@ -122,7 +122,7 @@ function Test-FailoverGroup()
 
 		# Create with default values
 		$fgName = Get-FailoverGroupName
-		$fg = New-AzureRmSqlDatabaseFailoverGroup -ResourceGroupName $server.ResourceGroupName -ServerName $server.ServerName -PartnerServerName $partnerServer.ServerName -FailoverGroupName $fgName -FailoverPolicy Automatic -GracePeriodWithDataLossHour 1 -AllowReadOnlyFailoverToPrimary Enabled
+		$fg = New-AzureRmSqlDatabaseFailoverGroup -ResourceGroupName $server.ResourceGroupName -ServerName $server.ServerName -PartnerServerName $partnerServer.ServerName -FailoverGroupName $fgName -FailoverPolicy Automatic -GracePeriodWithDataLossHours 1 -AllowReadOnlyFailoverToPrimary Enabled
 		Validate-FailoverGroup $server $partnerServer $fgName Primary Automatic 1 Enabled @() $fg
 
 		# Alter all properties
@@ -284,7 +284,7 @@ function Test-SetFailoverGroup-AutomaticWithGracePeriodZero()
 	Handle-FailoverGroupTestWithFailoverGroup {
 		Param($fg)
 
-		$newFg = $fg | Set-AzureRmSqlDatabaseFailoverGroup -FailoverPolicy Automatic -GracePeriodWithDataLossHour 0 -AllowReadOnlyFailoverToPrimary Disabled
+		$newFg = $fg | Set-AzureRmSqlDatabaseFailoverGroup -FailoverPolicy Automatic -GracePeriodWithDataLossHours 0 -AllowReadOnlyFailoverToPrimary Disabled
 		Assert-FailoverGroupsEqual $fg $newFg -failoverPolicy Automatic -gracePeriod 1 -readOnlyFailoverPolicy Disabled
 		Validate-FailoverGroupWithGet $newFg
 	} -failoverPolicy Manual
--- a/src/ResourceManager/Sql/documentation/current-breaking-changes.md
+++ b/src/ResourceManager/Sql/documentation/current-breaking-changes.md
@@ -44,24 +44,26 @@ The following cmdlets were affected this release:
 
 **New-AzureRmSqlDatabaseFailoverGroup**
 - Tag parameter was removed
+- 'GracePeriodWithDataLossHour' parameter was renamed to 'GracePeriodWithDataLossHours'
 
 ```powershell
 # Old
-New-AzureRmSqlDatabaseFailoverGroup -ResourceGroupName rg -ServerName server1 -FailoverGroupName fg -PartnerServerName server2 -Tag @{ Environment="Test" }
+New-AzureRmSqlDatabaseFailoverGroup -ResourceGroupName rg -ServerName server1 -FailoverGroupName fg -PartnerServerName server2 -FailoverPolicy Automatic -GracePeriodWithDataLossHour 1 -Tag @{ Environment="Test" }
 
 # New
-New-AzureRmSqlDatabaseFailoverGroup -ResourceGroupName rg -ServerName server1 -FailoverGroupName fg -PartnerServerName server2
+New-AzureRmSqlDatabaseFailoverGroup -ResourceGroupName rg -ServerName server1 -FailoverGroupName fg -PartnerServerName server2 -FailoverPolicy Automatic -GracePeriodWithDataLossHours 1
 ```
 
 **Set-AzureRmSqlDatabaseFailoverGroup**
 - Tag parameter was removed
+- 'GracePeriodWithDataLossHour' parameter was renamed to 'GracePeriodWithDataLossHours'
 
 ```powershell
 # Old
-Set-AzureRmSqlDatabaseFailoverGroup -ResourceGroupName rg -ServerName server1 -FailoverGroupName fg -FailoverPolicy Automatic -Tag @{ Environment="Test" }
+Set-AzureRmSqlDatabaseFailoverGroup -ResourceGroupName rg -ServerName server1 -FailoverGroupName fg -FailoverPolicy Automatic -GracePeriodWithDataLossHour 1 -Tag @{ Environment="Test" }
 
 # New
-Set-AzureRmSqlDatabaseFailoverGroup -ResourceGroupName rg -ServerName server1 -FailoverGroupName fg -FailoverPolicy Automatic
+Set-AzureRmSqlDatabaseFailoverGroup -ResourceGroupName rg -ServerName server1 -FailoverGroupName fg -FailoverPolicy Automatic -GracePeriodWithDataLossHours 1
 ```
 
 **Add-AzureRmSqlDatabaseToFailoverGroup**


src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/AzureSqlFailoverGroupCmdletBase.cs
```diff
@@ -43,5 +33,49 @@ protected override AzureSqlFailoverGroupAdapter InitModelAdapter(AzureSubscripti
...
+                    gracePeriod = (int)parameterValue;
+                    break;
+                }
+            }
+
+            if (!gracePeriod.HasValue && policy == FailoverPolicy.Automatic)
+            {
+                // If the policy is Automatic, the grace period must be non-null. If this is an update and the grace
+                // period was non-null, use the existing grace period. Otherwise, use a default of 1.
+                gracePeriod = originalGracePeriod ?? 1;
```
akromm: Is it safe to assume a default value of 1 hour?  Would it be better to return an error requesting the user to set the grace period?

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/NewAzureSqlFailoverGroup.cs
```diff
@@ -30,62 +30,73 @@ namespace Microsoft.Azure.Commands.Sql.FailoverGroup.Cmdlet
...
+            HelpMessage = "The failover policy of the Azure SQL Database Failover Group.")]
         [ValidateNotNullOrEmpty]
         public FailoverPolicy FailoverPolicy { get; set; }
 
         /// <summary>
         /// Gets or sets the grace period with data loss for the Sql Azure Failover Group.
         /// </summary>
         [Parameter(Mandatory = false,
-            HelpMessage = "The window of grace period that we tolerate with data loss during a failover operation for the failover group.")]
+            HelpMessage = "The grace period during outage before automatic failover with data loss of the Failover Group is triggered.")]
```
akromm: Please re-word this.  It is a bit tricky to comprehend.

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/NewAzureSqlFailoverGroup.cs
```diff
@@ -30,62 +30,73 @@ namespace Microsoft.Azure.Commands.Sql.FailoverGroup.Cmdlet
...
         [ValidateNotNullOrEmpty]
-        [Obsolete("This parameter will be deprecated in the next release.")]
-        public int GracePeriodWithDataLossHour { get; set; }
+        [ValidateRange(0, int.MaxValue)]
+        public int GracePeriodWithDataLossHours { get; set; }
```
akromm: Since this refers to a time duration I can't help but think that a TimeSpan would be a much better type than int.  Please consider updating the APIs to take a timespan rather than int in the future.

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/NewAzureSqlFailoverGroup.cs
```diff
@@ -30,62 +30,73 @@ namespace Microsoft.Azure.Commands.Sql.FailoverGroup.Cmdlet
...
-        /// Gets or sets the tag associated with the Azure SQL Database Failover Group
+        /// Gets or sets the failover policy for read only endpoint of the Sql Azure Failover Group.
         /// </summary>
-        /// 
-        [Obsolete("This parameter will be deprecated in the next release.")]
-        [Parameter(Mandatory = false,
-            HelpMessage = "The tag to associate with the Azure SQL Database Failover Group")]
-        public Hashtable Tag { get; set; }
+        [Parameter(Mandatory = false, DontShow = true,
+            HelpMessage = "Whether an outage on the secondary server should trigger automatic failover of the read-only endpoint. This feature is not yet supported.")]
```
akromm: If the feature is not yet supported why are you adding a parameter for it?

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/NewAzureSqlFailoverGroup.cs
```diff
@@ -111,9 +122,7 @@ protected override IEnumerable<AzureSqlFailoverGroupModel> GetEntity()
             }
 
             // The database already exists
```
akromm: *Failover Group

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/SetAzureSqlFailoverGroup.cs
```diff
@@ -44,36 +54,38 @@ public class SetAzureSqlFailoverGroup : AzureSqlFailoverGroupCmdletBase
...
+            HelpMessage = "The failover policy of the Azure SQL Database Failover Group.")]
         [ValidateNotNullOrEmpty]
         public FailoverPolicy FailoverPolicy { get; set; }
 
         /// <summary>
         /// Gets or sets the grace period with data loss for the Sql Azure Failover Group.
         /// </summary>
         [Parameter(Mandatory = false,
-            HelpMessage = "The grace period for failover with data loss of the failover group. This property defines how big of the window we tolerate for data loss during failover operation")]
+            HelpMessage = "The grace period during outage before automatic failover with data loss of the Failover Group is triggered.")]
```
akromm: please reword to make it easier to understand.

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/SetAzureSqlFailoverGroup.cs
```diff
@@ -44,36 +54,38 @@ public class SetAzureSqlFailoverGroup : AzureSqlFailoverGroupCmdletBase
...
+
+        /// <summary>
+        /// Gets or sets the grace period with data loss for the Sql Azure Failover Group.
+        /// </summary>
+        [Parameter(Mandatory = false, DontShow = true,
+            HelpMessage = "The grace period during outage before automatic failover with data loss of the Failover Group is triggered.")]
         [ValidateNotNullOrEmpty]
+        [ValidateRange(0, int.MaxValue)]
         [Obsolete("This parameter will be deprecated in the next release.")]
         public int GracePeriodWithDataLossHour { get; set; }
```
akromm: You have removed this parameter from other cmdlets. Can it not be removed from here also?

src/ResourceManager/Sql/Commands.Sql/Failover Group/Services/AzureSqlFailoverGroupAdapter.cs
```diff
@@ -111,19 +134,14 @@ internal AzureSqlFailoverGroupModel UpsertFailoverGroup(AzureSqlFailoverGroupMod
...
 
-            if (model.FailoverWithDataLossGracePeriodHours.HasValue && !string.Equals(model.ReadWriteFailoverPolicy, FailoverPolicy.Manual.ToString()))
-            {
-                readWriteEndpoint.FailoverWithDataLossGracePeriodMinutes = model.FailoverWithDataLossGracePeriodHours * 60;
-            }
-            else
+            if (model.FailoverWithDataLossGracePeriodHours.HasValue)
             {
-                readWriteEndpoint.FailoverWithDataLossGracePeriodMinutes = null;
+                readWriteEndpoint.FailoverWithDataLossGracePeriodMinutes = checked(model.FailoverWithDataLossGracePeriodHours * 60);
```
akromm: If the API uses minutes why does the cmdlet require hours to be passed in?

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/AddAzureSqlDatabaseToFailoverGroup.cs
```diff
@@ -44,19 +54,11 @@ public class AddAzureSqlDatabaseToFailoverGroup : AzureSqlFailoverGroupCmdletBas
         /// </summary>
         [Parameter(Mandatory = true,
             ValueFromPipeline = true,
-            HelpMessage = "The Azure SQL Database to be added to the secondary server.")]
+            HelpMessage = "One or more Azure SQL Databases on the Failover Group's primary server to be added to the Failover Group.")]
         [ValidateNotNullOrEmpty]
         public List<AzureSqlDatabaseModel> Database { get; set; }
```
jaredmoo: Why is this database model? Shouldn't it just be database name?

cormacpayne:
@azuresdkci add to whitelist

src/ResourceManager/Sql/Commands.Sql.Test/ScenarioTests/FailoverGroupTests.cs
```diff
@@ -27,12 +27,118 @@ public FailoverGroupTests(ITestOutputHelper output)
...
 
-        // Currently the test runs too long to be marked as a check-in test.
-        [Fact(Skip = "Hydra TestFramework Issue, cannot login to authenticate in order to record the tests")]
+        [Fact]
         public void TestFailoverGroup()
         {
             RunPowerShellTest("Test-FailoverGroup");
         }
 
+        [Fact]
```
cormacpayne: @dlemMSFT you will need to add the `CheckIn` trait to each test for them to be run during the build. For example:

```cs
[Fact]
[Trait(Category.AcceptanceType, Category.CheckIn)]
public void TestCreateFailoverGroup_Named()
{
    RunPowerShellTest("Test-CreateFailoverGroup-Named");
}
```

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/NewAzureSqlFailoverGroup.cs
jaredmoo: How do you create timespan in powershell? `[TimeSpan]::FromHours` (etc) is imo not beginner-friendly, so I think Hours is fine.

jaredmoo: What I'm trying to say is that cmdlet params should generally be int, string, or similar types that can be created as a literal value (e.g. double, decimal, etc) any data type more complex than those makes the cmdlet much harder to use for powershell beginners if there is no alternative simpler parameter.

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/AzureSqlFailoverGroupCmdletBase.cs
```diff
@@ -25,16 +25,6 @@ namespace Microsoft.Azure.Commands.Sql.FailoverGroup.Cmdlet
     public abstract class AzureSqlFailoverGroupCmdletBase : AzureSqlCmdletBase<IEnumerable<AzureSqlFailoverGroupModel>, AzureSqlFailoverGroupAdapter>
     {
         /// <summary>
-        /// Gets or sets the name of the server to use.
-        /// </summary>
-        [Parameter(Mandatory = true,
```
cormacpayne: @dlemMSFT any reason for removing this parameter from the base cmdlet if you are just going to add it back to each cmdlet that inherits from it?

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/NewAzureSqlFailoverGroup.cs
cormacpayne: @dlemMSFT @akromm if we're renaming this cmdlet, what do you guys think about renaming it to `GracePeriodWithDataLossInHours`? 

I say this because I know StaticAnalysis is going to throw an exception because parameter nouns should be singular (but nouns such as `InHours`, `InMinutes`, etc. are suppressed and won't trigger an exception). Also, I think `InHours` sounds better with the rest of the parameter noun than `Hours`. Let me know what you think.

src/ResourceManager/Sql/Commands.Sql/Failover Group/Services/AzureSqlFailoverGroupAdapter.cs
```diff
@@ -47,6 +47,29 @@ public class AzureSqlFailoverGroupAdapter
...
         /// <summary>
+        /// Dummy URI used in parsing server ID.
+        /// </summary>
+        private readonly Uri _dummyUri = new Uri("https://localhost");
+
+        /// <summary>
+        /// SQL server URI template used for parsing.
+        /// </summary>
+        private readonly UriTemplate _serverUriTemplate =
+            new UriTemplate("subscriptions/{subscriptionId}/resourcegroups/{resourceGroup}/providers/Microsoft.Sql/servers/{serverName}");
```
jaredmoo: Shouldn't this have leading `/` ? i.e. `/subscriptions/....` ?

src/ResourceManager/Sql/Commands.Sql/Failover Group/Services/AzureSqlFailoverGroupAdapter.cs
```diff
@@ -284,10 +292,46 @@ private AzureSqlFailoverGroupModel CreateFailoverGroupModelFromResponse(string r
             model.FailoverWithDataLossGracePeriodHours = failoverGroup.Properties.ReadWriteEndpoint.FailoverWithDataLossGracePeriodMinutes == null ?
                                                         null : failoverGroup.Properties.ReadWriteEndpoint.FailoverWithDataLossGracePeriodMinutes / 60;
 
-            model.Tags = TagsConversionHelper.CreateTagDictionary(TagsConversionHelper.CreateTagHashtable(failoverGroup.Tags), false);
-            model.Location = failoverGroup.Location;;
+            model.Id = failoverGroup.Id;
+            model.Location = failoverGroup.Location;
+            model.DatabaseNames = RetrieveDatabaseNames(failoverGroup);
+
+            UriTemplateMatch match = _failoverGroupUriTemplate.Match(_dummyUri, new Uri(_dummyUri, failoverGroup.Id));
```
jaredmoo: can you comment why dummy uri is needed?

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/NewAzureSqlFailoverGroup.cs
```diff
@@ -30,62 +30,73 @@ namespace Microsoft.Azure.Commands.Sql.FailoverGroup.Cmdlet
...
+        [ValidateRange(0, int.MaxValue)]
+        public int GracePeriodWithDataLossHours { get; set; }
 
         /// <summary>
-        /// Gets or sets the failover policy for read only endpoint of theSql Azure Failover Group.
+        /// Gets or sets the grace period with data loss for the Sql Azure Failover Group.
         /// </summary>
-        [Parameter(Mandatory = false,
-            HelpMessage = "The failover policy for read only endpoint of the failover group.")]
+        [Parameter(Mandatory = false, DontShow = true,
```
cormacpayne: @dlemMSFT why are you hiding this parameter? You renamed it above, so there's no need to re-add it and obsolete it.

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/SetAzureSqlFailoverGroup.cs
```diff
@@ -44,36 +54,38 @@ public class SetAzureSqlFailoverGroup : AzureSqlFailoverGroupCmdletBase
...
 
         /// <summary>
         /// Gets or sets the grace period with data loss for the Sql Azure Failover Group.
         /// </summary>
         [Parameter(Mandatory = false,
-            HelpMessage = "The grace period for failover with data loss of the failover group. This property defines how big of the window we tolerate for data loss during failover operation")]
+            HelpMessage = "The grace period during outage before automatic failover with data loss of the Failover Group is triggered.")]
+        [ValidateNotNullOrEmpty]
+        [ValidateRange(0, int.MaxValue)]
+        public int GracePeriodWithDataLossHours { get; set; }
```
cormacpayne: @dlemMSFT @akromm same comment about `InHours`

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/SetAzureSqlFailoverGroup.cs
```diff
@@ -44,36 +54,38 @@ public class SetAzureSqlFailoverGroup : AzureSqlFailoverGroupCmdletBase
...
+
+        /// <summary>
+        /// Gets or sets the grace period with data loss for the Sql Azure Failover Group.
+        /// </summary>
+        [Parameter(Mandatory = false, DontShow = true,
```
cormacpayne: @dlemMSFT same comment about `DontShow`

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/AzureSqlFailoverGroupCmdletBase.cs
dlemMSFT: I wanted to improve the documentation to refer to "primary" and "secondary" server rather than just server, as the distinction is important for these cmdlets.

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/NewAzureSqlFailoverGroup.cs
dlemMSFT: We thought this would be less disruptive to customers. Unlike the other removed parameters, we expect that customers will actually be using this one. We obsoleted it during the previous release, but we provided no alternative. We'd like to remove it after customers have had a chance to move to the alternative (in the next breaking release).

Would you suggest that we remove immediately?

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/NewAzureSqlFailoverGroup.cs
dlemMSFT: The parameter is already added and cannot be removed since it is not obsoleted. The github diff doesn't seem to be doing a good job of clearly showing the changes.

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/SetAzureSqlFailoverGroup.cs
dlemMSFT: This wasn't removed from the other cmdlet -- the github diff is obscuring the fact that it existed previously in both.

src/ResourceManager/Sql/Commands.Sql/Failover Group/Services/AzureSqlFailoverGroupAdapter.cs
dlemMSFT: 'Hours' is representative of what we actually support. We think making it 'Hours' in PowerShell regardless of the API is the best option to avoid misleading users. We intend to change the API.

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/AzureSqlFailoverGroupCmdletBase.cs
dlemMSFT: This is a subjective choice we made in favor of user convenience. What exactly do you mean by "safe?" Avoiding data loss unless the user very explicitly opts in? Also, this would arguably be a breaking change since previously the command would have worked without specifying the grace period.

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/AddAzureSqlDatabaseToFailoverGroup.cs
dlemMSFT: The reason is that it makes it possible to do things like "Get-AzureRmSqlDatababase <...> | Add-AzureRmSqlDatabaseToFailoverGroup <...>" without the need to write any other PowerShell code. It needs to be a single command -- adding the databases one by one would take too long.

I think an argument could be made that we should also have a 'DatabaseName' parameter, but in my opinion that is redundant since this one already exists.

jaredmoo: Sure, but PowerShell beginners do not use piping. Piping is cool, I use it all the time, but I have watched people use PowerShell for the first time and it takes a while to figure this out. In fact most powershell beginner just type the command name `Add-AzureRmSqlDatabaseToFailoverGroup` and then follow the prompts for required parameters. 

This means in order to make your cmdlet approachable there needs to be a param set like `Add-AzureRmSqlDatabaseToFailoverGroup -ServerName myserver -FailoverGroupName mygroup -DatabaseName mydatabase`. Yes it's redundant once you've learned the more advanced piping technique, but we often make redundancies in order to improve UX.

In fact, the piping scenario would work if you had `DatabaseName` parameter with `ValueFromPipelineByPropertyName=true`. So it's actually `DatabaseName` that makes `Database` redundant, not the other way around :)

dlemMSFT: Sorry for being (very) unclear in my examples -- the important part is not that we are piping, it is that we are adding multiple databases at the same time. Using a model parameter, you can list all databases on a server, or list all databases in an elastic pool, and add them to a failover group at the same time (without code to select DatabaseName). Adding takes a long time, so adding many at a time is much better than adding one-by-one.

Your point about 'DatabaseName' being better for beginners or people doing tab-completion is well taken, though. We can do that in the future, but I don't want to put us at risk for the next release by making more changes in this pull request.

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/NewAzureSqlFailoverGroup.cs
dlemMSFT: @cormacpayne I don't disagree with your points -- but the competing argument is that we have already documented the signatures in other places, and shared them with customers. It would be be more work than just refreshing this PR to make that change.

cormacpayne: @dlemMSFT sure, no worries. Just wanted to throw it out as a suggestion.

src/ResourceManager/Sql/Commands.Sql.Test/ScenarioTests/FailoverGroupTests.cs
dlemMSFT: Fixed.

src/ResourceManager/Sql/Commands.Sql/Failover Group/Services/AzureSqlFailoverGroupAdapter.cs
dlemMSFT: It's either this or string.Split. Actually, after looking at how other services do it, I've converted this to use string.Split. The code is much cleaner.

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/NewAzureSqlFailoverGroup.cs
dlemMSFT: Fixed.

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/NewAzureSqlFailoverGroup.cs
dlemMSFT: Fixed.

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/SetAzureSqlFailoverGroup.cs
dlemMSFT: Done.

src/ResourceManager/Sql/Commands.Sql/Failover Group/Cmdlet/AddAzureSqlDatabaseToFailoverGroup.cs
jaredmoo: That's fine, please just consider it for next release :)

markcowl: Note that, for future plans, we will be suggesting using models like this for piping within modules, but ValueFromPipelineByPropertyName for piping between modules (from generic resource cmdlets)

As Jared indicates, it is best to have a piping parameter set like this, and a parameter set more friendly to interactive input.

tools/StaticAnalysis/Exceptions/SignatureIssues.csv
```diff
@@ -1362,7 +1362,10 @@
...
 "Microsoft.Azure.Commands.Sql.dll","Microsoft.Azure.Commands.Sql.FailoverGroup.Cmdlet.SetAzureSqlFailoverGroup","Set-AzureRmSqlDatabaseFailoverGroup","1","8100","Set-AzureRmSqlDatabaseFailoverGroup Does not support ShouldProcess but the cmdlet verb Set indicates that it should.","Determine if the cmdlet should implement ShouldProcess and if so determine if it should implement Force / ShouldContinue"
+"Microsoft.Azure.Commands.Sql.dll","Microsoft.Azure.Commands.Sql.FailoverGroup.Cmdlet.SetAzureSqlFailoverGroup","Set-AzureRmSqlDatabaseFailoverGroup","2","8210","Set-AzureRmSqlDatabaseFailoverGroup does not have a Force parameter but the cmdlet verb 'Set' indicates that it may perform destructive actions under certain circumstances. Consider whether the cmdlet should have a Force parameter anduse ShouldContinue under some circumstances. ","Consider wether the cmdlet should have a Force parameter and use ShouldContinue under some circumstances. "
```
markcowl: Note that you do not need to add suppressions for severity 2 issues

markcowl:
On demand run here:  http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1512/

PR #3847, opened 2017-04-27
Set-AzureRmVirtualNetworkGatewayConnection parameters

henry416:
## Description
+ added parameters to Set-AzureRmVirtualNetworkGatewayConnection for convenience
+ normalized the UsePolicyBasedTS parameter

---

### DIFF
--- a/src/ResourceManager/Network/Commands.Network.Test/ScenarioTests/VirtualNetworkGatewayTests.ps1
+++ b/src/ResourceManager/Network/Commands.Network.Test/ScenarioTests/VirtualNetworkGatewayTests.ps1
@@ -462,8 +462,8 @@ function Test-VirtualNetworkGatewayBgpRouteApi
 		$gw1 = New-AzureRmVirtualNetworkGateway -ResourceGroupName $rgname -name $gwname1 -location $location -IpConfigurations $vnetIpConfig1 -GatewayType Vpn -VpnType RouteBased -GatewaySku Standard -Asn 1337
 		$gw1 = Get-AzureRmVirtualNetworkGateway -ResourceGroupName $rgname -name $gwname1
 
-		New-AzureRmVirtualNetworkGatewayConnection -ResourceGroupName $rgname -name $connectionName -location $location -VirtualNetworkGateway1 $gw -VirtualNetworkGateway2 $gw1 -ConnectionType Vnet2Vnet -SharedKey chocolate -EnableBgp true
-		New-AzureRmVirtualNetworkGatewayConnection -ResourceGroupName $rgname -name $connectionName1 -location $location -VirtualNetworkGateway1 $gw1 -VirtualNetworkGateway2 $gw -ConnectionType Vnet2Vnet -SharedKey chocolate -EnableBgp true
+		New-AzureRmVirtualNetworkGatewayConnection -ResourceGroupName $rgname -name $connectionName -location $location -VirtualNetworkGateway1 $gw -VirtualNetworkGateway2 $gw1 -ConnectionType Vnet2Vnet -SharedKey chocolate -EnableBgp $true
+		New-AzureRmVirtualNetworkGatewayConnection -ResourceGroupName $rgname -name $connectionName1 -location $location -VirtualNetworkGateway1 $gw1 -VirtualNetworkGateway2 $gw -ConnectionType Vnet2Vnet -SharedKey chocolate -EnableBgp $true
 
 		$bgpPeerStatus = Get-AzureRmVirtualNetworkGatewayBGPPeerStatus -ResourceGroupName $rgname -VirtualNetworkGatewayName $gwname
 		$bgpLearnedRoutes = Get-AzureRmVirtualNetworkGatewayLearnedRoute -ResourceGroupName $rgname -VirtualNetworkGatewayName $gwname


henry416:
Is there a method to generate the .msi or PS installer so I can run some flexible command tests?

src/ResourceManager/Network/Commands.Network/VirtualNetworkGatewayConnection/NewAzureVirtualNetworkGatewayConnectionCommand.cs
```diff
@@ -134,7 +134,7 @@ public class NewAzureVirtualNetworkGatewayConnectionCommand : VirtualNetworkGate
         [Parameter(
             Mandatory = false,
             HelpMessage = "Use policy-based traffic selectors for a S2S connection")]
-        public SwitchParameter UsePolicyBasedTrafficSelectors { get; set; }
+        public string UsePolicyBasedTrafficSelectors { get; set; }
```
cormacpayne: @henry416 why was this change made? You should **never** try to parse a string to a bool.

It looks like `vnetGatewayConnection.UsePolicyBasedTrafficSelectors` is false by default, so if a user wants to make it true, then they just use the SwitchParameter that is currently offered. There should be no reason to change this functionality and introduce a breaking change.

src/ResourceManager/Network/Commands.Network/VirtualNetworkGatewayConnection/UpdateAzureVirtualNetworkGatewayConnectionCommand.cs
```diff
@@ -31,6 +32,23 @@ public class SetAzureVirtualNetworkGatewayConnectionCommand : VirtualNetworkGate
...
+        [Parameter(
+            Mandatory = false,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Whether to establish a BGP session over a S2S VPN tunnel")]
+        public string EnableBgp { get; set; }
+
+        [Parameter(
+            Mandatory = false,
+            HelpMessage = "Use policy-based traffic selectors for a S2S connection")]
+        public string UsePolicyBasedTrafficSelectors { get; set; }
```
cormacpayne: @henry416 if this cmdlet is a PATCH operation, this parameter should be a bool, if it's a PUT operation, then it should be a SwitchParameter. Same comment about not parsing a string to a bool.

src/ResourceManager/Network/Commands.Network/VirtualNetworkGatewayConnection/UpdateAzureVirtualNetworkGatewayConnectionCommand.cs
```diff
@@ -31,6 +32,23 @@ public class SetAzureVirtualNetworkGatewayConnectionCommand : VirtualNetworkGate
...
+
+        [Parameter(
+             Mandatory = false,
+             ValueFromPipelineByPropertyName = true,
+             HelpMessage = "A list of IPSec policies.")]
+        public List<PSIpsecPolicy> IpsecPolicies { get; set; }
```
cormacpayne: @henry416 this parameter name should be `IpsecPolicy`. From the [Strongly Encouraged Developer Guidelines](https://msdn.microsoft.com/en-us/library/dd878270(v=vs.85).aspx):

_Avoid using plural names for parameters whose value is a single element. This includes parameters that take arrays or lists because the user might supply an array or list with only one element._

src/ResourceManager/Network/Commands.Network/VirtualNetworkGatewayConnection/UpdateAzureVirtualNetworkGatewayConnectionCommand.cs
```diff
@@ -31,6 +32,23 @@ public class SetAzureVirtualNetworkGatewayConnectionCommand : VirtualNetworkGate
             ValueFromPipeline = true,
             HelpMessage = "The VirtualNetworkGatewayConnection")]
         public PSVirtualNetworkGatewayConnection VirtualNetworkGatewayConnection { get; set; }
+        
+        [Parameter(
+            Mandatory = false,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Whether to establish a BGP session over a S2S VPN tunnel")]
+        public string EnableBgp { get; set; }
```
cormacpayne: @henry416 if this cmdlet is a PATCH operation, this parameter should be a bool, if it's a PUT operation, then it should be a SwitchParameter. Same comment about not parsing a string to a bool.

src/ResourceManager/Network/Commands.Network/VirtualNetworkGatewayConnection/NewAzureVirtualNetworkGatewayConnectionCommand.cs
henry416: UsePolicyBasedTrafficSelectors has not been documented and exposed to customer so this should be an okay change. We are normalizing the parameter to the same use as our other Boolean flags. See the below comments.

src/ResourceManager/Network/Commands.Network/VirtualNetworkGatewayConnection/UpdateAzureVirtualNetworkGatewayConnectionCommand.cs
henry416: @cormacpayne This is not the desired behavior. Lack of a switch parameter will mean that BGP will be disabled, which will break existing use cases.

src/ResourceManager/Network/Commands.Network/VirtualNetworkGatewayConnection/UpdateAzureVirtualNetworkGatewayConnectionCommand.cs
henry416: @cormacpayne same as above.

src/ResourceManager/Network/Commands.Network/VirtualNetworkGatewayConnection/UpdateAzureVirtualNetworkGatewayConnectionCommand.cs
henry416: @cormacpayne This is the naming used in the REST spec and NRP

henry416:
@cormacpayne Let me know if there are more questions and I will address asap

yushwang:
@cormacpayne 

Hi Cormac, unfortunately, pretty much all VPN and ExpressRoute cmdlets follow the convention of using $True/$False when turning features on/off. Changing this behavior now on one parameter will be very confusing to customers. That's why I asked Henry to change.

Could you unblock this PR?

Thanks,
Yushun

yushwang:
@henry416 The type should be Boolean right? Not string (or I am just not familiar with the types).

markcowl:
@yushwang This is unfortunate howerver, the right thing to do is to conform with best practices in these cmdlets.  Please make the changes as specified.

yushwang:
@markcowl 

Hi Mark,

Here's the problem, even for the "New" cmdlet for many of our resources, there were implicit default values where the absence of an optional parameter is still implying certain default value associated with the parameter. In this case, we are discussing "Set". The fact that the underlying API is PUT is somewhat hidden for the customers because we did not implement PATCH, and the absence of a particular parameter is considered to be unchanged, rather than $False or "Disabled". 

And in this case, the parameters (Traffic Selector and BGP, former for IPsec security and latter for routing) have critical consequences if customers do not understand the best practices. We are asking for an exception for an optional behavior. Could you guys unblock?

Thanks,
Yushun

markcowl:
@yushwang The standard for this is to use two switch parameters (enable/disable) for a single pool - this way enable means enable, disable means disable, and unprovided means not changed, with parameter sets to make these mutually exclusive.

If you have more than one boolean parameter, then you can use a bool in this instance, as we do not wnat parameter sets to proliferate.

Also, if this is acting like a patch then the cmdlet needs to be called 'Update' and this is the release in which to do it.

cormacpayne:
@henry416 please resolve the merge conflicts

Also, you will need to update the `tools/StaticAnalysis/Exceptions/BreakingChangeIssues.csv` file with the exceptions that you are getting from the build (located in the **Build Artifacts** folder under `src/Package/BreakingChangeIssues.csv`) in order for the build to pass

src/ResourceManager/Network/Commands.Network/VirtualNetworkGatewayConnection/UpdateAzureVirtualNetworkGatewayConnectionCommand.cs
```diff
@@ -36,13 +36,24 @@ public class SetAzureVirtualNetworkGatewayConnectionCommand : VirtualNetworkGate
...
-            HelpMessage = "Use policy-based traffic selectors for a S2S connection")]
-        public string UsePolicyBasedTrafficSelectors { get; set; }
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Disable a BGP session over a S2S VPN tunnel")]
+        public SwitchParameter DisableBgp { get; set; }
+
+        [Parameter(
+            Mandatory = false,
+            HelpMessage = "Enable policy-based traffic selectors for a S2S connection")]
+        public SwitchParameter EnablePolicyBasedTrafficSelectors { get; set; }
```
markcowl: Since there are multiple switches here, please change this to bool parameters, or set up parameter sets to make these switch parameters mutually exclusive, per the advies in the email about this.  Either would be acceptable

src/ResourceManager/Network/Commands.Network/VirtualNetworkGatewayConnection/NewAzureVirtualNetworkGatewayConnectionCommand.cs
```diff
@@ -172,22 +172,14 @@ private PSVirtualNetworkGatewayConnection CreateVirtualNetworkGatewayConnection(
...
-            if (!string.IsNullOrEmpty(this.EnableBgp))
-            {
-                vnetGatewayConnection.EnableBgp = bool.Parse(this.EnableBgp);
-            }
-            else
-            {
-                vnetGatewayConnection.EnableBgp = false;
-            }
+
+            vnetGatewayConnection.EnableBgp = this.EnableBgp;
```
markcowl: Now you are stting a value regardless of whether the user set this value, which is different than before, where you only set the value if the user provided it.  Is this intentional?

Note that you can get the same behavior usign bool parameters if you use MyInvocation.BoundParameters to see if the user provided a value

henry416: @markcowl  Is this MyInvocation.BoundParameters a dictionary of provided parameters? (key = parametername, value = parametervalue)? I was not aware of this capability - this might change my entire design

src/ResourceManager/Network/Commands.Network/VirtualNetworkGatewayConnection/UpdateAzureVirtualNetworkGatewayConnectionCommand.cs
henry416: @markcowl this might be changed, see below

src/ResourceManager/Network/Commands.Network/VirtualNetworkGatewayConnection/NewAzureVirtualNetworkGatewayConnectionCommand.cs
henry416: Default values for the Booleans are false anyways, so directly setting in New-cmd should be good. I am interested in using the MyInvocation for the Set-cmd since it can allow us to keep the same parameter type across both new and set.

henry416:
FYI changes made, pending session tests

cormacpayne:
On-demand: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1567/

markcowl:
signing run here:http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-sign/1016/

on demand run here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1568/

henry416:
@markcowl Issues were fixed - merge should be able to proceed

PR #4003, opened 2017-05-19
AzureRm.Bootstrapper: Enable setting & removing default profile

### LINKED ISSUE #3308: Allow setting Default Profile through PowerShell default parameters
_From @markcowl on November 14, 2016 22:57_

### Description

- [x] Design a mechanism usin the DefaultParameters feature with profiles to set up a default profile version
- [x] Design cmdlets for settign and removing these defautl aprameter values

### Cost: 4

_Copied from original issue: markcowl/azure-powershell#13_

viananth:
## Description

---
1. Added Set-AzureRmDefaultProfile and Remove-AzureRmDefaultProfile cmdlets
2. Refactored retry commands to a single command with scriptblock parameter for retry operations (Invoke-CommandWithRetry)
3. Removed RetryGetContent and RemoveWithRetry functions
4. Updated unit tests
5. Updated help files 
6. Closes #3308  

### DIFF
--- a/tools/AzureRM.BootStrapper/AzureRM.BootStrapper.Tests.ps1
+++ b/tools/AzureRM.BootStrapper/AzureRM.BootStrapper.Tests.ps1
@@ -1399,7 +1399,8 @@ Describe "Remove-AzureRmDefaultProfile" {
         Mock Remove-Module -Verifiable {}
         Mock Remove-ProfileSetting -Verifiable {}
         Mock Test-Path -Verifiable { $true }
-        
+        Mock Get-Module -Verifiable { "AzureRm" }
+
         Context "Default profile presents in the profile file & default variable" {
             It "Should successfully remove default profile from shell & profile file" {
                 Remove-AzureRmDefaultProfile -Force
@@ -1417,17 +1418,15 @@ Describe "Remove-AzureRmDefaultProfile" {
         }
 
         Context "Profile files do not exist" {
-            Mock Invoke-CommandWithRetry {}
             Mock Test-Path -Verifiable { $false }
             It "Should not invoke remove script" {
                 Remove-AzureRmDefaultProfile -Force
-                Assert-MockCalled Invoke-CommandWithRetry -Exactly 0
+                Assert-MockCalled Remove-ProfileSetting -Exactly 0
             }
         }
 
         Context "Remove default profile in admin mode" {
             Mock Remove-Module -verifiable {}
-            Mock Invoke-CommandWithRetry -Verifiable {}
             $Script:IsAdmin = $true
             It "Should remove setting in AllUsersAllHosts and CurrentUserAllHosts profiles" {
                 Remove-AzureRmDefaultProfile -Force
@@ -1443,7 +1442,6 @@ Describe "Remove-AzureRmDefaultProfile" {
             It "Should remove setting in CurrentUserAllHosts profile" {
                 Remove-AzureRmDefaultProfile -Force
                 Assert-MockCalled Test-Path -Exactly 1
-               
             }
         }
     }    


tools/AzureRM.BootStrapper/AzureRM.Bootstrapper.psm1
```diff
@@ -1055,3 +1084,112 @@ function Set-BootstrapRepo
...
+    
+    $defaultProfile = $Global:PSDefaultParameterValues["*-AzureRmProfile:Profile"]
+    if ($defaultProfile -ne $armProfile)
+    {
+      # Set DefaultProfile for this session
+      $Global:PSDefaultParameterValues["*-AzureRmProfile:Profile"]="$armProfile"
+
+      # Edit the profile content
+      $profileContent = @"
+          `$PSDefaultParameterValues["*-AzureRmProfile:Profile"]="$armProfile"
```
markcowl: I am not following what the intention is here.  This just sets the default for AzureRmProfile cmdlets.  This does not set the default version of cmdlets that will be loaded when PowerShell is opened.

If we wanted to set the default cmdlet version loaded at runtime, we would need to change the default parameter for ```Import-Module```, probably with a script block, selecting the version based on the module name

bganapa: right now, the default is being passed only for the *profile* cmdlets. agree that we need to change the default parameter for Import-Module as well. just talked to vishnu. it will be part of the next revision of this PR

tools/AzureRM.BootStrapper/AzureRM.Bootstrapper.psm1
```diff
@@ -1055,3 +1123,126 @@ function Set-BootstrapRepo
...
+          $profilePath = Select-Profile -Scope $Scope
+
+          # Remove previous setting if exists
+          Write-Verbose "Updating default profile value to $armProfile"
+          Write-Debug "Removing previous setting if exists"
+          $RemoveSettingScriptBlock = {
+            (Get-Content -Path $profilePath -ErrorAction Stop) | Select-String -pattern 'AzureRmProfile:Profile' -notmatch | Set-Content -path  $profilePath -ErrorAction Stop
+            (Get-Content -Path $profilePath -ErrorAction Stop) | Select-String -pattern 'Import-Module:RequiredVersion' -notmatch | Set-Content -path  $profilePath -ErrorAction Stop
+            (Get-Content -Path $profilePath -ErrorAction Stop) | Select-String -pattern 'MyInvocation.Line.ToLower' -notmatch | Set-Content -path  $profilePath -ErrorAction Stop
+            (Get-Content -Path $profilePath -ErrorAction Stop) | Select-String -pattern '}' -notmatch | Set-Content -path  $profilePath -ErrorAction Stop
```
viananth: I'm not sure if this is the right way to do it. This might remove '}' that was not intended to be removed. It would be great if you have any other mechanism for removing this scriptblock.

tools/AzureRM.BootStrapper/AzureRM.Bootstrapper.psm1
```diff
@@ -611,6 +562,128 @@ function Invoke-InstallModule
...
+  foreach ($module in $Modules)
+  {
+    if ($module.ToLower() -eq $RollUpModule.ToLower())
+    {
+      continue
+    }
+    $versions = $ProfileMap.$armProfile.$module
+    $version = Get-LatestModuleVersion -versions $versions
+    $module = $module.ToLower()
+    $ScriptBlock += " elseif (`$MyInvocation.Line.ToLower().Contains(`"$module`")) { `"$version`" } `r`n"
```
markcowl: Why do we need this?  This is the default behavior anyway

tools/AzureRM.BootStrapper/AzureRM.Bootstrapper.psm1
```diff
@@ -611,6 +562,128 @@ function Invoke-InstallModule
...
+
+  $ProfileMap = (Get-AzProfile)
+  $Modules = ($ProfileMap.$armProfile | Get-Member -MemberType NoteProperty).Name
+
+  # In scriptblock first cmd should be "if" and others "elseif". So handle first element separately
+  $ScriptBlock = @()
+  $firstModule = $modules[0].ToLower()
+  $versions = $ProfileMap.$armProfile.$firstModule
+  $version = Get-LatestModuleVersion -versions $versions
+  $ScriptBlock += "if (`$MyInvocation.Line.ToLower().Contains(`"$firstModule`")) { `"$version`" } `r`n"
```
markcowl: I think this is the wrong approach, because it means we are writing a lot of stuff to the profile that will be difficult to remove, and it is not robust over modules being deleted.  I think instead the right thing is to:

(1) Write a function that takes the versionProfile string and $MyInvocation.Line as parameters, gets the correct version, checks for the module version on disk and returns the appropriate value.  This could be impemented in a function in the bootstrapper module, and the Profile function would just check for the bootstrapper module and call the bootstrapper function, which would make this a lot easier to delete.

(2) write this function to the profile, and just invoke the fucntion in the default parameter

The diea is that, if the user removed modules or even removes the bootstrapper, the function will actually do nothing.

tools/AzureRM.BootStrapper/AzureRM.Bootstrapper.psm1
markcowl: See my comments above.  I think if we refactor how this is implemented in the profile, it will make this much simpler.

tools/AzureRM.BootStrapper/AzureRM.Bootstrapper.psm1
```diff
@@ -1055,3 +1158,119 @@ function Set-BootstrapRepo
...
+          # Check Profile existence and choose proper profile
+          $profilePath = Select-Profile -Scope $Scope
+
+          # Set DefaultProfile for this session
+          $Global:PSDefaultParameterValues["*-AzureRmProfile:Profile"]="$armProfile"
+          $Global:PSDefaultParameterValues["Import-Module:RequiredVersion"]={ Get-ModuleVersion -armProfile $Global:PSDefaultParameterValues["*-AzureRmProfile:Profile"] -invocationLine $MyInvocation.Line }
+
+          # Edit the profile content
+          $profileContent = @"
+`$PSDefaultParameterValues["*-AzureRmProfile:Profile"]="$armProfile" `r`n
```
markcowl: See the comment above, I think that if we put content markers before an after the profile script, this will make it easier to detect and remove.

tools/AzureRM.BootStrapper/AzureRM.Bootstrapper.psm1
```diff
@@ -1055,3 +1158,119 @@ function Set-BootstrapRepo
...
+
+          # Get Script to be added to the $profile path
+          $profileContent += Get-ScriptBlock -ProfilePath $profilePath
+
+          Write-Verbose "Updating default profile value to $armProfile"
+          Write-Debug "Removing previous setting if exists"
+          $RemoveSettingScriptBlock = {
+            (Get-Content -Path $profilePath -ErrorAction Stop) | Select-String -pattern 'AzureRmProfile:Profile' -notmatch | Set-Content -path  $profilePath -ErrorAction Stop
+            (Get-Content -Path $profilePath -ErrorAction Stop) | Select-String -pattern 'Import-Module:RequiredVersion' -notmatch | Set-Content -path  $profilePath -ErrorAction Stop
+          }
```
markcowl: It doesn't look like we are removing the new function definition.  I think the removal logic should be abstracted here, so we can use it both in Set and Remove default profile cmdlets

tools/AzureRM.BootStrapper/AzureRM.BootStrapper.Tests.ps1
```diff
@@ -1291,3 +1265,132 @@ Describe "Set-BootstrapRepo" {
         }
     }
 }
+
+Describe "Set-AzureRmDefaultProfile" {
+    InModuleScope AzureRM.Bootstrapper {
+        $sb = {
```
markcowl: Should add tests for:

(1) Setting a default profile twice (verify that file is not altered twice

tools/AzureRM.BootStrapper/AzureRM.BootStrapper.Tests.ps1
```diff
@@ -1291,3 +1322,129 @@ Describe "Set-BootstrapRepo" {
...
+}
+
+Describe "Remove-AzureRmDefaultProfile" {
+    InModuleScope AzureRM.Bootstrapper {
+        Mock Remove-Module -Verifiable {}
+        Mock Remove-ProfileSetting -Verifiable {}
+        Mock Test-Path -Verifiable { $true }
+        
+        Context "Default profile presents in the profile file & default variable" {
+            It "Should successfully remove default profile from shell & profile file" {
```
markcowl: Failing in CI for not calling the RemoveModule mock

tools/AzureRM.BootStrapper/AzureRM.BootStrapper.Tests.ps1
```diff
@@ -1291,3 +1322,129 @@ Describe "Set-BootstrapRepo" {
...
+                Remove-AzureRmDefaultProfile -Force
+                $Global:PSDefaultParameterValues["*-AzureRmProfile:Profile"] | Should Be $null
+                Assert-VerifiableMocks
+            }
+        }
+        
+        Context "Default profile is not set previously or was removed" {
+            It "Should return null for default profile" {
```
markcowl: Same

PR #4058, opened 2017-05-31
Add Azure Stack Environment using ARM Endpoint

### LINKED ISSUE #3836: Add-AzureRmEnvironment: New Parameter set
### Cmdlet(s)
Add-AzureRmEnvironment

### PowerShell Version
5

### Description
Add-AzureRmEnvironment should support parameter "ArmEndpoint" and should register the environment using it.

### Script/Steps for Reproduction
Add-AzureRmEnvironment -Name "MyAzure" **-ArmEndpoint** "EndpointValue"

viananth:
## Description

Enables users to add azurermenvironment using ResourceManager url
Closes  #3836
---

### DIFF
--- a/src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
+++ b/src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
@@ -142,9 +142,14 @@ namespace Microsoft.Azure.Commands.Profile
                 }
                 catch (AggregateException ae)
                 {
-                    foreach (Exception ex in ae.Flatten().InnerExceptions)
+                    if (ae.Flatten().InnerExceptions.Count > 1)
                     {
-                        throw ex;
+                        throw;
+                    }
+
+                    if (ae.InnerException != null)
+                    {
+                        throw ae.InnerException;
                     }
                 }
             }
--- a/src/ResourceManager/Profile/Commands.Profile/Environment/SetAzureRMEnvironment.cs
+++ b/src/ResourceManager/Profile/Commands.Profile/Environment/SetAzureRMEnvironment.cs
@@ -151,9 +151,14 @@ namespace Microsoft.Azure.Commands.Profile
                 }
                 catch (AggregateException ae)
                 {
-                    foreach (Exception ex in ae.Flatten().InnerExceptions)
+                    if (ae.Flatten().InnerExceptions.Count > 1)
                     {
-                        throw ex;
+                        throw;
+                    }
+
+                    if (ae.InnerException != null)
+                    {
+                        throw ae.InnerException;
                     }
                 }
             }
--- a/src/ResourceManager/Profile/Commands.Profile/Utilities/EnvironmentHelper.cs
+++ b/src/ResourceManager/Profile/Commands.Profile/Utilities/EnvironmentHelper.cs
@@ -59,30 +59,19 @@ namespace Microsoft.Azure.Commands.Profile.Utilities
                 response = JsonConvert.DeserializeObject<MetadataResponse>(responseJson);
             }
 
-            if ((null == response) || (CheckIfAnyPropertyIsNull(response)))
+            if ((null == response) || string.IsNullOrEmpty(response.GalleryEndpoint) || string.IsNullOrEmpty(response.GraphEndpoint)
+                || string.IsNullOrEmpty(response.PortalEndpoint))
             {
-                throw new CloudException("An error occurred while trying to retrieve metadata endpoints. Please try again later.");    
+                throw new CloudException("An error occurred while trying to retrieve metadata endpoints. Please try again later.");
             }
 
-            bool isAnyPropertyNull = response.authentication
-                                            .GetType()
-                                            .GetProperties()
-                                            .Select(pi => pi.GetValue(response.authentication))
-                                            .Any(value => (value == null) || (string.IsNullOrEmpty(value.ToString())));
-            if (isAnyPropertyNull)
+            if (null == response.authentication || string.IsNullOrEmpty(response.authentication.LoginEndpoint)
+                || (response.authentication.Audiences.Any(string.IsNullOrEmpty)))
             {
                 throw new CloudException("An error occurred while trying to retrieve metadata endpoints. Please try again later.");
             }
 
             return response;
-        }
-
-        internal static bool CheckIfAnyPropertyIsNull(MetadataResponse response)
-        {
-            return response.GetType()
-                           .GetProperties()
-                           .Select(pi => pi.GetValue(response))
-                           .Any(value => (value == null) || (string.IsNullOrEmpty(value.ToString())));
         }
     }
 }


src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
```diff
@@ -14,92 +14,101 @@
...
     /// <summary>
     /// Cmdlet to add Azure Environment to Profile.
     /// </summary>
-    [Cmdlet(VerbsCommon.Add, "AzureRmEnvironment")]
+    [Cmdlet(VerbsCommon.Add, "AzureRmEnvironment", DefaultParameterSetName = EnvironmentNameParameterSet)]
     [OutputType(typeof(PSAzureEnvironment))]
     public class AddAzureRMEnvironmentCommand : AzureRMCmdlet
     {
-        [Parameter(Position = 0, Mandatory = true, ValueFromPipelineByPropertyName = true)]
+        private const string ArmParameterSet = "ResourceManagerEndpoint";
```
bganapa: call this as  private const string MetadataParameterSet = "MetadataParameterSet ";

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
```diff
@@ -14,92 +14,101 @@
...
+        private const string EnvironmentNameParameterSet = "Name";
```
bganapa: I don't think there is a need to specifically call out the second parameter set, can you please check

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
```diff
@@ -14,92 +14,101 @@
...
+
+        [Parameter(ParameterSetName = EnvironmentNameParameterSet, Position = 0, Mandatory = true, ValueFromPipelineByPropertyName = true)]
+        [Parameter(ParameterSetName = ArmParameterSet, Position = 0, Mandatory = false, ValueFromPipelineByPropertyName = true)]
```
bganapa: Name should be mandatory for ArmParameterSet

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
```diff
@@ -14,92 +14,101 @@
...
         public string StorageEndpoint { get; set; }
 
-        [Parameter(Position = 5, Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The URI for the Active Directory service for this environment")]
+        [Parameter(ParameterSetName = EnvironmentNameParameterSet, Position = 5, Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The URI for the Active Directory service for this environment")]
         [Alias("AdEndpointUrl", "ActiveDirectory", "ActiveDirectoryAuthority")]
         public string ActiveDirectoryEndpoint { get; set; }
 
-        [Parameter(Position = 6, Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The cloud service endpoint")]
+        [Parameter(ParameterSetName = EnvironmentNameParameterSet, Position = 6, Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The cloud service endpoint")]
+        [Parameter(ParameterSetName = ArmParameterSet, Position = 1, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The cloud service endpoint")]
```
bganapa: HelpMessage= Azure Resource Manager endpoint

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
```diff
@@ -112,12 +121,31 @@ protected override void BeginProcessing()
...
+            var result = profileClient.ListEnvironments(Name).Select(s => (PSAzureEnvironment)s).ToList();
+            if (result.Count != 0)
+            {
+                throw new InvalidOperationException("The specified environment already exists. Use a different name to create a new environment.");
+            }
 
             var newEnvironment = new AzureEnvironment
+                                 {
+                                     Name = Name,
+                                     OnPremise = EnableAdfsAuthentication
```
bganapa: We should override OnPremise detecting whether authority url ends with ADFS

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
```diff
@@ -112,12 +121,31 @@ protected override void BeginProcessing()
         public override void ExecuteCmdlet()
         {
             var profileClient = new RMProfileClient(AzureRmProfileProvider.Instance.Profile);
+            var result = profileClient.ListEnvironments(Name).Select(s => (PSAzureEnvironment)s).ToList();
```
bganapa: we don't need this check. its a breaking change. Add-AzureRmEnvironment providing the same name will update the existing properties.. we should keep that behavious as is

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
```diff
@@ -112,12 +121,31 @@ protected override void BeginProcessing()
...
+                                     Name = Name,
+                                     OnPremise = EnableAdfsAuthentication
+                                 };
+
+            if (this.ParameterSetName.Equals(ArmParameterSet, StringComparison.Ordinal))
             {
-                Name = Name,
-                OnPremise = EnableAdfsAuthentication
-            };
+                MetadataResponse metadataEndpoints = new MetadataResponse();
```
bganapa: no need of constructing the object

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
```diff
@@ -112,12 +121,31 @@ protected override void BeginProcessing()
...
+                metadataEndpoints = EnvironmentHelper.RetrieveMetaDataEndpoints(ResourceManagerEndpoint).Result;
+                string domain = EnvironmentHelper.RetrieveDomain(metadataEndpoints.PortalEndpoint);
+                ActiveDirectoryEndpoint = metadataEndpoints.authentication.LoginEndpoint.TrimEnd('/') + '/';
+                ActiveDirectoryServiceEndpointResourceId = metadataEndpoints.authentication.Audiences[0];
+                GalleryEndpoint = metadataEndpoints.GalleryEndpoint;
+                GraphEndpoint = metadataEndpoints.GraphEndpoint;
+                AzureKeyVaultDnsSuffix = $"vault.{domain}".ToLowerInvariant();
```
bganapa: $ syntax seems wrong

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
```diff
@@ -136,7 +164,11 @@ public override void ExecuteCmdlet()
             newEnvironment.Endpoints[AzureEnvironment.Endpoint.AzureDataLakeStoreFileSystemEndpointSuffix] = AzureDataLakeStoreFileSystemEndpointSuffix;
             newEnvironment.Endpoints[AzureEnvironment.Endpoint.AdTenant] = AdTenant;
             newEnvironment.Endpoints[AzureEnvironment.Endpoint.GraphEndpointResourceId] = GraphAudience;
-            WriteObject((PSAzureEnvironment)profileClient.AddOrSetEnvironment(newEnvironment));
+
+            if (ShouldProcess(Name, "Adding new environment"))
```
bganapa: why shouldprocess, that's a breaking change as well

src/ResourceManager/Profile/Commands.Profile/Utilities/EnvironmentHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+    using Newtonsoft.Json;
+
+    internal static class EnvironmentHelper
+    {
+        /// <summary>
+        /// Gets the valid endpoint.
+        /// </summary>
+        /// <param name="armEndpoint">The arm endpoint.</param>
+        /// <returns></returns>
+        internal static string GetValidEndpoint(string armEndpoint)
```
bganapa: we don't need this, if needed please validate with System.Uri

src/ResourceManager/Profile/Commands.Profile/Utilities/EnvironmentHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            return armEndpoint;
+        }
+
+        /// <summary>
+        /// Retrieves the domain.
+        /// </summary>
+        /// <param name="portalEndpoint">The portal endpoint.</param>
+        /// <returns>Domain</returns>
+        internal static string RetrieveDomain(string portalEndpoint)
```
bganapa: resourceManagerEndpoint instead of portalEndpoint ?

src/ResourceManager/Profile/Commands.Profile/Utilities/EnvironmentHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            return portalEndpoint.Replace(portalEndpoint.Split('.')[0], "").TrimEnd('/').TrimStart('.');
```
bganapa: providing an example format and stating the assumption here would help

src/ResourceManager/Profile/Commands.Profile/Utilities/EnvironmentHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+
+                        if (statusCode == HttpStatusCode.OK)
+                        {
+                            string responseContent = await httpResponse.Content.ReadAsStringAsync().ConfigureAwait(false);
+                            response = JsonConvert.DeserializeObject<MetadataResponse>(responseContent);
+                        }
+                    }
+                    finally
+                    {
+                        httpResponse?.Dispose();
```
bganapa: with using, you don't need dispose

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
viananth: We need to match the parameter name with the set name; this seems to be the practice in this project. But we could change the variable name as `private const string MetadataParameterSet = "ResourceManagerEndpoint"`

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
viananth: fixed.

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
viananth: fixed.

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
viananth: This was also a design discussion change from Mark.

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
viananth: Could you please clarify? It works as expected in my tests.

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
viananth: This is a requirement from the design discussion we had with Mark.

src/ResourceManager/Profile/Commands.Profile/Utilities/EnvironmentHelper.cs
viananth: Both gives the same result. Portal: `portal.azure.com`; Arm: `management.azure.com`. Both will return `azure.com `as domain.

src/ResourceManager/Profile/Commands.Profile/Utilities/EnvironmentHelper.cs
viananth: Added.

src/ResourceManager/Profile/Commands.Profile/Utilities/EnvironmentHelper.cs
viananth: Using only disposes httpclient in this case; not httprequest & httpresponse objects.

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
viananth: fixed.

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
markcowl: Yes, this is needed.  Can we please name these constants 'MetadataParameterSet' and 'EnvironmentPropetiesParameterSet'.  The values of the constants just need to be unique

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
```diff
@@ -14,92 +14,101 @@
...
-    [Cmdlet(VerbsCommon.Add, "AzureRmEnvironment")]
+    [Cmdlet(VerbsCommon.Add, "AzureRmEnvironment", DefaultParameterSetName = EnvironmentNameParameterSet)]
     [OutputType(typeof(PSAzureEnvironment))]
     public class AddAzureRMEnvironmentCommand : AzureRMCmdlet
     {
-        [Parameter(Position = 0, Mandatory = true, ValueFromPipelineByPropertyName = true)]
+        private const string ArmParameterSet = "ResourceManagerEndpoint";
+        private const string EnvironmentNameParameterSet = "Name";
+
+        [Parameter(ParameterSetName = EnvironmentNameParameterSet, Position = 0, Mandatory = true, ValueFromPipelineByPropertyName = true)]
```
markcowl: Since these have identical properties, you can use a single parameter attribute, without a parameter set name

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
markcowl: This is using c#  6  features (string interpolation) which may not be available in the ci compiler.

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
markcowl: SHouldProcess is not a breaking change - default execution is unaffected.  Can we please add the environment settings we are using to the SHoudlProcess string?

src/ResourceManager/Profile/Commands.Profile/Environment/SetAzureRMEnvironment.cs
```diff
@@ -25,82 +26,88 @@ namespace Microsoft.Azure.Commands.Profile
     /// <summary>
     /// Cmdlet to set Azure Environment in Profile.
     /// </summary>
-    [Cmdlet(VerbsCommon.Set, "AzureRmEnvironment")]
+    [Cmdlet(VerbsCommon.Set, "AzureRmEnvironment", SupportsShouldProcess = true, DefaultParameterSetName = EnvironmentNameParameterSet)]
     [OutputType(typeof(PSAzureEnvironment))]
     public class SetAzureRMEnvironmentCommand : AzureRMCmdlet
     {
-        [Parameter(Position = 0, Mandatory = true, ValueFromPipelineByPropertyName = true)]
+		private const string ArmParameterSet = "ResourceManagerEndpoint";
```
markcowl: The same comments apply here

src/ResourceManager/Profile/Commands.Profile/Models/MetadataResponse.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// <summary>
+    /// Class to receive meta data endpoints json as objects
+    /// </summary>
+    public class MetadataResponse
+    {
+        private string _galleryEndpoint;
+        private string _graphEndpoint;
+        private string _portalEndpoint;
+
+        public string GalleryEndpoint
```
markcowl: Why not just use automatic properties?

src/ResourceManager/Profile/Commands.Profile/Utilities/EnvironmentHelper.cs
markcowl: I would prefer less string parsing.  Why not start with Uri.Host, and then get the last two elements when splitting by '.'  For example, what if a uri is https://mydomain.com/portal?.  Host is 'mydomain.com' and we want to use 'mydomain.com'

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
```diff
@@ -113,6 +122,21 @@ public override void ExecuteCmdlet()
...
+                MetadataResponse metadataEndpoints = null;
+                metadataEndpoints = EnvironmentHelper.RetrieveMetaDataEndpoints(ResourceManagerEndpoint).Result;
+                string domain = EnvironmentHelper.RetrieveDomain(metadataEndpoints.PortalEndpoint);
+                ActiveDirectoryEndpoint = metadataEndpoints.authentication.LoginEndpoint.TrimEnd('/') + '/';
+                ActiveDirectoryServiceEndpointResourceId = metadataEndpoints.authentication.Audiences[0];
+                GalleryEndpoint = metadataEndpoints.GalleryEndpoint;
+                GraphEndpoint = metadataEndpoints.GraphEndpoint;
+                AzureKeyVaultDnsSuffix = $"vault.{domain}".ToLowerInvariant();
+                AzureKeyVaultServiceEndpointResourceId = ($"https://vault.{domain}".ToLowerInvariant());
+                StorageEndpoint = domain;
```
markcowl: SO, does this work?  The storage suffix for Azure is '.core.windows.net'

src/ResourceManager/Profile/Commands.Profile/Utilities/EnvironmentHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// </summary>
+        /// <param name="url">The URL.</param>
+        /// <returns></returns>
+        internal static async Task<MetadataResponse> RetrieveMetaDataEndpoints(string url)
+        {
+            if (!Uri.IsWellFormedUriString(url, UriKind.Absolute))
+            {
+                throw new ArgumentException("The ResourceManagement Endpoint provided was invalid.");
+            }
+            url = String.Concat(url.ToLower(), "/metadata/endpoints?api-version=1.0");
```
markcowl: vertical space after ```}```

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
```diff
@@ -113,6 +122,21 @@ public override void ExecuteCmdlet()
         {
             var profileClient = new RMProfileClient(AzureRmProfileProvider.Instance.Profile);
 
+            if (this.ParameterSetName.Equals(ArmParameterSet, StringComparison.Ordinal))
+            {
+                MetadataResponse metadataEndpoints = null;
+                metadataEndpoints = EnvironmentHelper.RetrieveMetaDataEndpoints(ResourceManagerEndpoint).Result;
```
markcowl: We need to check the returned value is not null before dereferencing it (and throw a meaningful error if it is)

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
```diff
@@ -113,6 +122,21 @@ public override void ExecuteCmdlet()
...
+                string domain = EnvironmentHelper.RetrieveDomain(metadataEndpoints.PortalEndpoint);
+                ActiveDirectoryEndpoint = metadataEndpoints.authentication.LoginEndpoint.TrimEnd('/') + '/';
```
markcowl: again, make sure this value is not null or empty before we use it.

src/ResourceManager/Profile/Commands.Profile.Test/EnvironmentCmdletTests.cs
```diff
@@ -72,10 +76,40 @@ public void AddsAzureEnvironment()
             Assert.Equal(env.Endpoints[AzureEnvironment.Endpoint.ManagementPortalUrl], cmdlet.ManagementPortalUrl);
             Assert.Equal(env.Endpoints[AzureEnvironment.Endpoint.Gallery], "http://galleryendpoint.com");
         }
-        
+
         [Fact]
         [Trait(Category.AcceptanceType, Category.CheckIn)]
-        public void AddsEnvironmentMultipleTimes()
```
markcowl: I think we should have a recorded mocked test or a test marked 'liveonly' for this, but it is fine to add this later

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
viananth: No, but I guess we agreed upon adding this for stack-dev branch for now. We will have to revisit during stack-dev to dev branch merge.

viananth: I meant it works only in stack. For azure, it returns 'azure.com' which is not the correct storage suffix.

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
bganapa: lets use string.Format here. It is not breaking in the CI, but my local with VS2013 breaks

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
markcowl: should check metadataendpoints.authentication is not null as well as metadataendpoints.authentication.LoginEndpoint

src/ResourceManager/Profile/Commands.Profile/Environment/SetAzureRMEnvironment.cs
```diff
@@ -131,17 +130,27 @@ public override void ExecuteCmdlet()
...
-                MetadataResponse metadataEndpoints = new MetadataResponse();
+                MetadataResponse metadataEndpoints = null;
                 metadataEndpoints = EnvironmentHelper.RetrieveMetaDataEndpoints(ResourceManagerEndpoint).Result;
+                if (null == metadataEndpoints)
+                {
+                    throw new Exception("An error occurred while trying to retrieve metadata endpoints. Please try again later.");
+                }
+
                 string domain = EnvironmentHelper.RetrieveDomain(metadataEndpoints.PortalEndpoint);
                 ActiveDirectoryEndpoint = metadataEndpoints.authentication.LoginEndpoint.TrimEnd('/') + '/';
```
markcowl: Nulkl check not just metadataendpoints, but all the proeprties we derefernece

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
```diff
@@ -14,92 +14,100 @@
...
-        [Parameter(Position = 5, Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The URI for the Active Directory service for this environment")]
+        [Parameter(ParameterSetName = EnvironmentPropertiesParameterSet, Position = 5, Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The URI for the Active Directory service for this environment")]
         [Alias("AdEndpointUrl", "ActiveDirectory", "ActiveDirectoryAuthority")]
         public string ActiveDirectoryEndpoint { get; set; }
 
-        [Parameter(Position = 6, Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The cloud service endpoint")]
+        [Parameter(ParameterSetName = EnvironmentPropertiesParameterSet, Position = 6, Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The cloud service endpoint")]
+        [Parameter(ParameterSetName = MetadataParameterSet, Position = 1, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The Azure Resource Manager endpoint")]
         [Alias("ResourceManager", "ResourceManagerUrl")]
         public string ResourceManagerEndpoint { get; set; }
```
viananth: There is ambiguity in selecting `ParameterSet `name using `ResourceManagerEndpoint `for both parameter sets. We might want to introduce new parameter unique to `MetadataParameterSet`. This helps to resolve the parameter sets correctly in my tests.

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
```diff
@@ -113,6 +123,34 @@ public override void ExecuteCmdlet()
...
+                    GalleryEndpoint = metadataEndpoints.GalleryEndpoint;
+                    GraphEndpoint = metadataEndpoints.GraphEndpoint;
+                    AzureKeyVaultDnsSuffix = string.Format("vault.{0}", domain).ToLowerInvariant();
+                    AzureKeyVaultServiceEndpointResourceId = string.Format("https://vault.{0}", domain).ToLowerInvariant();
+                    StorageEndpoint = domain;
+                    EnableAdfsAuthentication = metadataEndpoints.authentication.LoginEndpoint.TrimEnd('/').EndsWith("/adfs", System.StringComparison.OrdinalIgnoreCase);
+                }
+                catch (Exception ex)
+                {
+                    if (ex.InnerException != null)
```
markcowl: ??  why would we unwrap the inner exception in this case

src/ResourceManager/Profile/Commands.Profile/Environment/SetAzureRMEnvironment.cs
```diff
@@ -124,10 +132,39 @@ public override void ExecuteCmdlet()
                 }
             }
 
+            if (this.ParameterSetName.Equals(MetadataParameterSet, StringComparison.Ordinal))
+            {
+                MetadataResponse metadataEndpoints = null;
+                ResourceManagerEndpoint = ARMEndpoint;
+                try
+                {
+                    metadataEndpoints = EnvironmentHelper.RetrieveMetaDataEndpoints(ResourceManagerEndpoint).Result;
```
markcowl: identical comments.

src/ResourceManager/Profile/Commands.Profile/Utilities/EnvironmentHelper.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// Retrieves the domain.
+        /// </summary>
+        /// <param name="portalEndpoint">The portal endpoint.</param>
+        /// <returns>Domain</returns>
+        internal static string RetrieveDomain(string portalEndpoint)
+        {
+            // Todo: Revisit this when azure metadata endpoint supports keyvault suffix and storage endpoints
+            // Example format:: portal endpoint: "management.azure.com"; returns: "azure.com"
+            string domainHost = new Uri(portalEndpoint).Host;
+            return domainHost.Replace(domainHost.Split('.')[0], "").TrimEnd('/').TrimStart('.');
```
markcowl: You should not need the ```TrimEnd('/')``` call, as the Host fiels will not contain any part of the path

src/ResourceManager/Profile/Commands.Profile/Utilities/EnvironmentHelper.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <param name="url">The URL.</param>
+        /// <returns></returns>
+        internal static async Task<MetadataResponse> RetrieveMetaDataEndpoints(string url)
+        {
+            if (!Uri.IsWellFormedUriString(url, UriKind.Absolute))
+            {
+                throw new ArgumentException("The ResourceManagement Endpoint provided was invalid.");
+            }
+
+            url = String.Concat(url.ToLower(), "/metadata/endpoints?api-version=1.0");
```
markcowl: You should trim off any trailing '/' in the url befoire concatenating, to prevent ```//```

src/ResourceManager/Profile/Commands.Profile/Utilities/EnvironmentHelper.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            if (CheckIfAnyPropertyIsNull(response.authentication))
+            {
+                throw new CloudException("An error occurred while trying to retrieve metadata endpoints. Please try again later.");
+            }
+
+            return response;
+        }
+
+        internal static bool CheckIfAnyPropertyIsNull(object response)
```
markcowl: This is a well-defined type, and it is possible that in the future, some proeprty may be null, I would suggest *not* using reflection in this case, but doing specific null checks for each checked property.

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
viananth: It throws aggregate exception if I don't unwrap this. The ArgumentException or CloudException thrown from RetrieveMetaDataEndpoints function gets converted  into AggregateException. In PowerShell, AggregateException did not give much info about the exception (any error message). Should we still throw as it is?

src/ResourceManager/Profile/Commands.Profile/Environment/AddAzureRMEnvironment.cs
```diff
@@ -113,6 +123,32 @@ public override void ExecuteCmdlet()
...
+                    AzureKeyVaultDnsSuffix = string.Format("vault.{0}", domain).ToLowerInvariant();
+                    AzureKeyVaultServiceEndpointResourceId = string.Format("https://vault.{0}", domain).ToLowerInvariant();
+                    StorageEndpoint = domain;
+                    EnableAdfsAuthentication = metadataEndpoints.authentication.LoginEndpoint.TrimEnd('/').EndsWith("/adfs", System.StringComparison.OrdinalIgnoreCase);
+                }
+                catch (AggregateException ae)
+                {
+                    foreach (Exception ex in ae.Flatten().InnerExceptions)
+                    {
+                        throw ex;
```
markcowl: this is only going to throw one exception.  WIll leave for now, but you shoudlk throw the aggregate if there is more than one exception

src/ResourceManager/Profile/Commands.Profile/Utilities/EnvironmentHelper.cs
```diff
@@ -0,0 +1,88 @@
...
+                string responseJson = await client.GetStringAsync(url).ConfigureAwait(false);
+                response = JsonConvert.DeserializeObject<MetadataResponse>(responseJson);
+            }
+
+            if ((null == response) || (CheckIfAnyPropertyIsNull(response)))
+            {
+                throw new CloudException("An error occurred while trying to retrieve metadata endpoints. Please try again later.");    
+            }
+
+            bool isAnyPropertyNull = response.authentication
```
markcowl: This is still using reflection, please just check the actual proeprties we have specified in the type above for null / empty

that is

```if (response.authentication == null || string.IsNullOrEmpty(response.authentication.LoginEndpoint) || ...```

markcowl: Using the type properties is fragile

PR #4061, opened 2017-06-01
Key Vault: Powershell cmdlets for Managed Key Vault Storage Accounts …

kunalcha:
Feature: Key Vault powershell cmdlets for Managed Key Vault Storage Accounts

## Description

This feature in key vault will allow customers to associate an azure storage account with key vault to have its keys managed by key vault. 
Key Vault will auto regenerate alternate storage account keys per the information provided by the customer. This will come with a prerequisite that customer grants Key Vault identity permissions to perform list and regenerate actions using RBAC cmdlets. The account keys managed by key vault are never returned to the customer. Additionally, key vault will allow customers an ability to define sas parameters as named resources. This will allow customers to retrieve sas tokens as key vault secrets per the sas definitions set by the customer. 

---

### DIFF
--- a/src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/UpdateAzureKeyVaultManagedStorageAccountKey.cs
+++ b/src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/UpdateAzureKeyVaultManagedStorageAccountKey.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Azure.Commands.KeyVault
 {
     [Cmdlet( VerbsData.Update, CmdletNoun.AzureKeyVaultManagedStorageAccountKey,
         SupportsShouldProcess = true,
-        ConfirmImpact = ConfirmImpact.High,
+        ConfirmImpact = ConfirmImpact.Medium,
         HelpUri = Constants.KeyVaultHelpUri )]
     [OutputType( typeof( ManagedStorageAccount ) )]
     public class UpdateAzureKeyVaultManagedStorageAccountKey : KeyVaultCmdletBase


cormacpayne:
@azuresdkci add to whitelist

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageAccountAttribute.cs
```diff
@@ -0,0 +1,107 @@
...
+using System.Collections;
+using System.Management.Automation;
+
+namespace Microsoft.Azure.Commands.KeyVault
+{
+    [Cmdlet( VerbsCommon.Set, CmdletNoun.AzureKeyVaultManagedStorageAccountAttribute,
+        SupportsShouldProcess = true,
+        HelpUri = Constants.KeyVaultHelpUri )]
+    [OutputType( typeof( ManagedStorageAccount ) )]
+    public class SetAzureKeyVaultManagedStorageAccountAttribute : KeyVaultCmdletBase
```
cormacpayne: @kunalcha same comment about PUT vs PATCH

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageAccountAttribute.cs
```diff
@@ -0,0 +1,107 @@
...
+                          "account remains unchanged" )]
+        [ValidateNotNull]
+        public TimeSpan? RegenerationPeriod { get; set; }
+
+        [Parameter( Mandatory = false,
+            HelpMessage = "If present, enables a use of a managed storage account key for sas token generation if value is true. Disables use of a managed " +
+                          "storage account key for sas token generation if value is false. If not specified, the existing value of the storage account's " +
+                          "enabled/disabled state remains unchanged." )]
+        [ValidateNotNull]
+        public bool? Enable { get; set; }
```
cormacpayne: @kunalcha same comment about a more descriptive name

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageSasDefinition.cs
```diff
@@ -0,0 +1,361 @@
...
+using System.Xml;
+
+namespace Microsoft.Azure.Commands.KeyVault
+{
+    [Cmdlet( VerbsCommon.Set, CmdletNoun.AzureKeyVaultManagedStorageSasDefinition,
+        SupportsShouldProcess = true,
+        HelpUri = Constants.KeyVaultHelpUri,
+        DefaultParameterSetName = ParameterSetRawSas )]
+    [OutputType( typeof( ManagedStorageSasDefinition ) )]
+    public partial class SetAzureKeyVaultManagedStorageSasDefinition : KeyVaultCmdletBase
```
cormacpayne: @kunalcha same comment about PUT vs PATCH

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageSasDefinition.cs
```diff
@@ -0,0 +1,361 @@
...
+            Position = 3,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Sas definition parameters that will be used to create the sas token.",
+            ParameterSetName = ParameterSetRawSas )]
+        [ValidateNotNull]
+        public Hashtable Parameter { get; set; }
+
+        [Parameter( Mandatory = false,
+            HelpMessage = "Disables the use of sas definition for generation of sas token." )]
+        public SwitchParameter Disable { get; set; }
```
cormacpayne: @kunalcha same comment about a more descriptive parameter name

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/GetAzureKeyVaultManagedStorageAccount.cs
```diff
@@ -0,0 +1,96 @@
...
+        private const string ByAccountNameParameterSet = "ByAccountName";
+
+        #endregion
+
+        #region Input Parameter Definitions
+
+        /// <summary>
+        /// Vault name
+        /// </summary>
+        [Parameter( Mandatory = true,
```
kunalcha: Yes, a vault name is always required. A Storage account cannot be managed by Key Vault without a vault in first place. That's the paradigm with other objects inside the vault (such as keys, secret and certificates) as well.

---
In reply to: [119717241](https://github.com/Azure/azure-powershell/pull/4061#discussion_r119717241) [](ancestors = 119717241)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/NewAzureKeyVaultManagedStorageAccountKey.cs
```diff
@@ -0,0 +1,87 @@
...
+        public string KeyName { get; set; }
+
+        /// <summary>
+        /// If present, do not ask for confirmation
+        /// </summary>
+        [Parameter( Mandatory = false, HelpMessage = "Do not ask for confirmation." )]
+        public SwitchParameter Force { get; set; }
+
+        [Parameter( Mandatory = false,
+            HelpMessage = "Cmdlet does not return an object by default. If this switch is specified, cmdlet returns the managed storage account that was deleted." )]
```
kunalcha: Thanks for the catch. It should simply say "cmdlet returns the managed storage account"

---
In reply to: [119721727](https://github.com/Azure/azure-powershell/pull/4061#discussion_r119721727) [](ancestors = 119721727)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/RemoveAzureKeyVaultManagedStorageSasDefinition.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        /// <summary>
+        /// If present, do not ask for confirmation
+        /// </summary>
+        [Parameter( Mandatory = false,
+            HelpMessage = "Do not ask for confirmation." )]
+        public SwitchParameter Force { get; set; }
+
+        [Parameter( Mandatory = false,
+            HelpMessage = "Cmdlet does not return an object by default. If this switch is specified, cmdlet returns the managed storage account that was deleted." )]
```
kunalcha: Thanks for the catch. Fixed.

---
In reply to: [119723676](https://github.com/Azure/azure-powershell/pull/4061#discussion_r119723676) [](ancestors = 119723676)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageAccount.cs
```diff
@@ -0,0 +1,101 @@
...
+        [Parameter( Mandatory = false,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Regeneration period. If auto regenerate key is enabled, this value specifies the timespan after which managed storage account's inactive key" +
+                          "gets auto regenerated and becomes the new active key." )]
+        [ValidateNotNull]
+        public TimeSpan? RegenerationPeriod { get; set; }
+
+        [Parameter( Mandatory = false, 
+            HelpMessage = "Disables the use of managed storage account's key for generation of sas tokens." )]
+        public SwitchParameter Disable { get; set; }
```
kunalcha: This parameter is named along the lines of similar parameter in Set-AzureKeyVaultSecret and Set-AzureKeyVaultKey and carries similar meaning of disabling use of the key vault object. 

---
In reply to: [119725609](https://github.com/Azure/azure-powershell/pull/4061#discussion_r119725609) [](ancestors = 119725609)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageAccount.cs
```diff
@@ -0,0 +1,101 @@
...
+// ----------------------------------------------------------------------------------
+
+using Microsoft.Azure.Commands.KeyVault.Models;
+using System;
+using System.Collections;
+using System.Management.Automation;
+
+namespace Microsoft.Azure.Commands.KeyVault
+{
+    [Cmdlet( VerbsCommon.Set, CmdletNoun.AzureKeyVaultManagedStorageAccount,
```
kunalcha: This is a PUT behavior. KeyVault has used Set***Attribute cmdlet for PATCH and this new feature follows the same design for consistency.

---
In reply to: [119727250](https://github.com/Azure/azure-powershell/pull/4061#discussion_r119727250) [](ancestors = 119727250)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageSasDefinition.cs
kunalcha: This parameter is named along the lines of similar parameter in Set-AzureKeyVaultSecret and Set-AzureKeyVaultKey and carries similar meaning of disabling use of the key vault object. 

---
In reply to: [119732397](https://github.com/Azure/azure-powershell/pull/4061#discussion_r119732397) [](ancestors = 119732397)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageSasDefinition.cs
kunalcha: This is a PUT.

---
In reply to: [119731517](https://github.com/Azure/azure-powershell/pull/4061#discussion_r119731517) [](ancestors = 119731517)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageAccountAttribute.cs
kunalcha: This is along the design of existing key vault cmdlet of SetAzureKeyVaultKeyAttribute and SetAzureKeyVaultSecretAttribute

---
In reply to: [119727653](https://github.com/Azure/azure-powershell/pull/4061#discussion_r119727653) [](ancestors = 119727653)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageAccountAttribute.cs
kunalcha: This is a PATCH and naming is similar to naming of existing SetAzureKeyVaultKeyAttribute and SetAzureKeyVaultSecretAttribute. Not all fields of object are editable once set, only certain attributes are editable.

---
In reply to: [119727599](https://github.com/Azure/azure-powershell/pull/4061#discussion_r119727599) [](ancestors = 119727599)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/RemoveAzureKeyVaultManagedStorageAccount.cs
```diff
@@ -0,0 +1,78 @@
...
+        public string VaultName { get; set; }
+
+        [Parameter( Mandatory = true,
+            Position = 1,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Key Vault managed storage account name. Cmdlet constructs the FQDN of a managed storage account name from vault name, currently " +
+                          "selected environment and manged storage account name." )]
+        [ValidateNotNullOrEmpty]
+        [Alias( Constants.StorageAccountName )]
+        public string AccountName { get; set; }
```
kunalcha: Can you please elaborate how piping gets impacted with alias of "Name". IMO, having an alias of Name might cause confusion as there are multiple names across this family of cmdlets - VaultName, Storage Account Name and Sas definition Name. Also for e.g. an existing cmdlet of New-AzureRmKeyVault -VaultName doesn't use an alias of Name for VaultName.

---
In reply to: [119722704](https://github.com/Azure/azure-powershell/pull/4061#discussion_r119722704) [](ancestors = 119722704)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageAccount.cs
```diff
@@ -0,0 +1,101 @@
...
+        public string VaultName { get; set; }
+
+        [Parameter( Mandatory = true,
+            Position = 1,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Key Vault managed storage account name. Cmdlet constructs the FQDN of a managed storage account name from vault name, currently " +
+                          "selected environment and manged storage account name." )]
+        [ValidateNotNullOrEmpty]
+        [Alias( Constants.StorageAccountName )]
+        public string AccountName { get; set; }
```
kunalcha: Answered on other comment.

---
In reply to: [119727542](https://github.com/Azure/azure-powershell/pull/4061#discussion_r119727542) [](ancestors = 119727542)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/RemoveAzureKeyVaultManagedStorageAccount.cs
cormacpayne: @kunalcha this is a pattern that we are trying to enforce across the board now for new cmdlets. It allows generic cmdlets to return objects with properties such as `Name`, `ResourceGroupName`, etc. and have those piped to this cmdlet by property name. I don't think it will cause confusion with the different cmdlets in the set since is an alias, so it will not show up when the user is tabbing through the possible parameters and in the help (`Name` will be mentioned in the properties part of the parameter as an alias). Let me know if you have any other questions.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageAccountAttribute.cs
markcowl: @kunalcha PATCH cmdlets should use the ```Update``` verb rather than a longer cmdlet noun

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/NewAzureKeyVaultManagedStorageAccountKey.cs
```diff
@@ -0,0 +1,87 @@
...
+        [Parameter( Mandatory = true,
+            Position = 2,
+            HelpMessage = "Name of storage account key to regenerate and make active." )]
+        [ValidateNotNullOrEmpty]
+        public string KeyName { get; set; }
+
+        /// <summary>
+        /// If present, do not ask for confirmation
+        /// </summary>
+        [Parameter( Mandatory = false, HelpMessage = "Do not ask for confirmation." )]
```
markcowl: So, this doesn't look like Force should be required for this cmdlet.  Force should only be used if there are special circumstances that could occur for a particular cmdlet invocation (for example, if you detected that you were going to overwrite a key that was in use).  In this case, the user should be able to exeucte the cmdlet without a prompt.

The user cna use ```WhatIf``` and ```Confirm``` and ```ConfirmPreference``` to control prompting.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/RemoveAzureKeyVaultManagedStorageAccount.cs
markcowl: @kunalcha @cormacpayne Since these are data plane cmdlets, this isn't necessary - the Name/ResourceGroupName piping only applies for those cmdlets that should accept pipeline input from generic resources from the generic resource cmdlets

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/RemoveAzureKeyVaultManagedStorageAccount.cs
```diff
@@ -0,0 +1,78 @@
...
+using Microsoft.Azure.Commands.KeyVault.Models;
+using System.Globalization;
+using System.Management.Automation;
+using KeyVaultProperties = Microsoft.Azure.Commands.KeyVault.Properties;
+
+namespace Microsoft.Azure.Commands.KeyVault
+{
+    [Cmdlet( VerbsCommon.Remove, CmdletNoun.AzureKeyVaultManagedStorageAccount,
+        SupportsShouldProcess = true,
+         ConfirmImpact = ConfirmImpact.High,
```
markcowl: You should not use ConfirmImpact.High

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/RemoveAzureKeyVaultManagedStorageSasDefinition.cs
```diff
@@ -0,0 +1,91 @@
...
+using Microsoft.Azure.Commands.KeyVault.Models;
+using System.Globalization;
+using System.Management.Automation;
+using KeyVaultProperties = Microsoft.Azure.Commands.KeyVault.Properties;
+
+namespace Microsoft.Azure.Commands.KeyVault
+{
+    [Cmdlet( VerbsCommon.Remove, CmdletNoun.AzureKeyVaultManagedStorageSasDefinition,
+        SupportsShouldProcess = true,
+         ConfirmImpact = ConfirmImpact.High,
```
markcowl: Do not use ConfirmImact.High

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageAccount.cs
```diff
@@ -0,0 +1,101 @@
...
+    [Cmdlet( VerbsCommon.Set, CmdletNoun.AzureKeyVaultManagedStorageAccount,
+        SupportsShouldProcess = true,
+        HelpUri = Constants.KeyVaultHelpUri)]
+    [OutputType(typeof(ManagedStorageAccount))]
+    public class SetAzureKeyVaultManagedStorageAccount : KeyVaultCmdletBase
+    {
+        #region Input Parameter Definitions
+        [Parameter(Mandatory = true,
+            Position = 0,
+            ValueFromPipelineByPropertyName = true,
```
markcowl: Reiterating the pieplien note form previously (this applies to all cmdlets).

Our advice is for piping within a module, you should *not* use ValueFromPipelineByPropertyName, and use ValueFromPipeline with a new parameter set that allows piping the output type from the cmdlet that creates the type you are piping.

i.e.:

```powershell
[Parameter(Mandatory=true, ValueFromPipeline=true, ParameterSetName="KeyVaultPipingParameterSet")]
public KeyVaultKey InputObject {get; set;}
```

Note that this is not a *required* change, but a suggested change

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageAccountAttribute.cs
```diff
@@ -0,0 +1,107 @@
...
+            Position = 1,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Key Vault managed storage account name. Cmdlet constructs the FQDN of a managed storage account name from vault name, currently " +
+                          "selected environment and manged storage account name." )]
+        [ValidateNotNullOrEmpty]
+        [Alias( Constants.StorageAccountName )]
+        public string AccountName { get; set; }
+
+        [Parameter( Mandatory = false,
+            ValueFromPipelineByPropertyName = true,
```
markcowl: I don't think piping makes sense for these values for a PATCH scenario - the entire idea of PATCH is to just provide those proeprties that you want to change, and not bing the other properties.

Once again, the ```InputObject``` piping model from above would work here, to just allow piping the identifyign properties of the managed storage account, and not any of its behavioral properties.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageSasDefinition.Service.Share.File.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <summary>
+        /// container pipeline paremeter set name with policy
+        /// </summary>
+        private const string ParameterSetStoredPolicyServiceFileSas = "StoredPolicyServiceFileSas";
+
+        private const string PathHelpMessage = "Path to the cloud file to generate sas token against.";
+        [Parameter( Mandatory = true, HelpMessage = PathHelpMessage, ParameterSetName = ParameterSetAdhocServiceFileSas )]
+        [Parameter( Mandatory = true, HelpMessage = PathHelpMessage, ParameterSetName = ParameterSetStoredPolicyServiceFileSas )]
+
+        [ValidateNotNullOrEmpty]
```
markcowl: ? is this intended as a parameter?

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageSasDefinition.cs
```diff
@@ -0,0 +1,361 @@
...
+        [Alias( Constants.SasDefinitionName )]
+        public string Name { get; set; }
+
+        [Parameter( Mandatory = true,
+            Position = 3,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Sas definition parameters that will be used to create the sas token.",
+            ParameterSetName = ParameterSetRawSas )]
+        [ValidateNotNull]
+        public Hashtable Parameter { get; set; }
```
markcowl: I wonder if we can't do  a bit better using an IStorageContext for the storage account connections tring.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageSasDefinition.cs
```diff
@@ -0,0 +1,361 @@
...
+        [Parameter( Mandatory = false,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "A hashtable representing tags of sas definition." )]
+        [Alias( Constants.TagsAlias )]
+        public Hashtable Tag { get; set; }
+
+        #region Common SAS Arguments
+
+        private const string TargetStorageVersionHelpMessage = "Specifies the signed storage service version to use to authenticate requests made with the SAS token.";
+        [Parameter( HelpMessage = TargetStorageVersionHelpMessage, ParameterSetName = ParameterSetAdhocAccountSas )]
```
markcowl: Can we get a validateset for this?

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/SetAzureKeyVaultAccessPolicy.cs
```diff
@@ -198,6 +212,24 @@ public class SetAzureKeyVaultAccessPolicy : KeyVaultManagementCmdletBase
...
+        [Parameter( Mandatory = false,
+            ParameterSetName = ByServicePrincipalName,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Specifies managed storage account and sas definition operation permissions to grant to a user or service principal." )]
+        [Parameter( Mandatory = false,
+            ParameterSetName = ByUserPrincipalName,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Specifies managed storage account and sas definition operation permissions to grant to a user or service principal." )]
+        [ValidateSet( "get", "list", "delete", "set", "update", "regeneratekey", "getsas", "listsas", "deletesas", "setsas", "all" )]
+        public string[] PermissionsToStorage { get; set; }
```
markcowl: Why not ```StoragePermissions```

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/GetAzureKeyVaultManagedStorageSasDefinition.cs
```diff
@@ -0,0 +1,111 @@
...
+using System.Collections.Generic;
+using System.Management.Automation;
+using KeyVaultProperties = Microsoft.Azure.Commands.KeyVault.Properties;
+
+namespace Microsoft.Azure.Commands.KeyVault
+{
+    [Cmdlet( VerbsCommon.Get, CmdletNoun.AzureKeyVaultManagedStorageSasDefinition,
+        DefaultParameterSetName = ByAccountNameParameterSet,
+        HelpUri = Constants.KeyVaultHelpUri )]
+    [OutputType( typeof( List<ManagedStorageSasDefinitionListItem> ), typeof( ManagedStorageSasDefinition ) )]
```
kunalcha: Yes, one is result of Get on collection (list) and other is result of get on resource.

---
In reply to: [119998218](https://github.com/Azure/azure-powershell/pull/4061#discussion_r119998218) [](ancestors = 119998218)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/NewAzureKeyVaultManagedStorageAccountKey.cs
```diff
@@ -0,0 +1,87 @@
...
+using Microsoft.Azure.Commands.KeyVault.Models;
+using System.Globalization;
+using System.Management.Automation;
+using KeyVaultProperties = Microsoft.Azure.Commands.KeyVault.Properties;
+
+namespace Microsoft.Azure.Commands.KeyVault
+{
+    [Cmdlet( VerbsCommon.New, CmdletNoun.AzureKeyVaultManagedStorageAccountKey,
+        SupportsShouldProcess = true,
+         ConfirmImpact = ConfirmImpact.High,
```
kunalcha: This can be a destructive change as it regenerates the storage account key (like existing powershell cmdlet of New-AzureRmStorageAccountKey)

---
In reply to: [119998290](https://github.com/Azure/azure-powershell/pull/4061#discussion_r119998290) [](ancestors = 119998290)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/NewAzureKeyVaultManagedStorageAccountKey.cs
kunalcha: There is no way for this cmdlet to know which key is in use as it issues a POST request to service.  Hence, this can be a case where user regenerates the key that is in use. So your statement of use of Force seems applicable.

---
In reply to: [119998322](https://github.com/Azure/azure-powershell/pull/4061#discussion_r119998322) [](ancestors = 119998322)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/GetAzureKeyVaultManagedStorageAccount.cs
```diff
@@ -0,0 +1,96 @@
...
+            Position = 0,
+            ValueFromPipelineByPropertyName = true,
+            ParameterSetName = ByAccountNameParameterSet,
+            HelpMessage = "Vault name. Cmdlet constructs the FQDN of a vault based on the name and currently selected environment." )]
+        [Parameter( Mandatory = true,
+            Position = 0,
+            ValueFromPipelineByPropertyName = true,
+            ParameterSetName = ByVaultNameParameterSet,
+            HelpMessage = "Vault name. Cmdlet constructs the FQDN of a vault based on the name and currently selected environment." )]
```
kunalcha: Thanks for letting us know. This will impact existing cmdlets as well and perhaps can be treated as bug to change this and existing cmdlets.

---
In reply to: [119998211](https://github.com/Azure/azure-powershell/pull/4061#discussion_r119998211) [](ancestors = 119998211)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/GetAzureKeyVaultManagedStorageSasDefinition.cs
```diff
@@ -0,0 +1,111 @@
...
+            Position = 1,
+            ValueFromPipelineByPropertyName = true,
+            ParameterSetName = ByDefinitionNameParameterSet,
+                        HelpMessage = "Key Vault managed storage account name. Cmdlet constructs the FQDN of a managed storage account name from vault name, currently " +
+                          "selected environment and manged storage account name." )]
+        [ValidateNotNullOrEmpty]
+        [Alias( Constants.StorageAccountName )]
+        public string AccountName { get; set; }
+
+        [Parameter( Mandatory = false,
```
kunalcha: This was reuse of code pattern from Key and secret cmdlets. One parameter set is for GET and other for LIST (Get on collection). What's the advantage of using what you suggest over this? I'd prefer code consistency with existing cmdlets and treat this as a bug to change the pattern if there is advantage.

---
In reply to: [119998282](https://github.com/Azure/azure-powershell/pull/4061#discussion_r119998282) [](ancestors = 119998282)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/NewAzureKeyVaultManagedStorageAccountKey.cs
markcowl: It is not appropriate to set ConfirmImpact to high in this case, it violates PowerShell norms.  Please change this

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/NewAzureKeyVaultManagedStorageAccountKey.cs
markcowl: No, that is precisely the point.  Asking for confirmation on every invocation does not provide the user with any useful information about the effects of that invocation.  Therefore, you should just implement ShouldProcess and *not* have a Force parameter.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageSasDefinition.cs
markcowl: The idea is - the user can provide (1) A storage context (which is generated by storage data plane cmdlets); (2) A Hashtable of the connection string properties; (3) The specific parameters you provide elsewhere

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageSasDefinition.Service.Share.File.cs
markcowl: remove extra vertical space so the grouping of attributes is clear

markcowl: Please remove line 35 so this is clearer

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageSasDefinition.cs
markcowl: This is a suggestion, not a required change

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/NewAzureKeyVaultManagedStorageAccountKey.cs
dragav: Changed this to Medium; by merely creating a managed storage account key in KV, the customer cannot cause an immediate disruption. To cause a rollover of the storage account key, the customer has to either set a policy (with a minimum of a 1 day delay from creation) or run a subsequent command to roll the key.

dragav: Upon further deliberation, this command should be renamed to a Set, as the key is guaranteed to exist (it's an intrinsic property of the storage account); the command's effect is to roll the key over, and therefore the High impact and presence of Force are both warranted. Agree those parameters would not make sense for a New-type command.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/SetAzureKeyVaultAccessPolicy.cs
dragav: This reflects existing naming pattern (PermissionsToSecrets/Keys/Certificates) and will be retained for consistency.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/NewAzureKeyVaultManagedStorageAccountKey.cs
dragav: This cmdlet has been renamed to Update-AKVManagedStorageAccountKey, and triggers the immediate rollover of that account's key. The key is guaranteed to exist (by the existence of the storage account) and so an Update operation is effectively an overwrite.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/RemoveAzureKeyVaultManagedStorageAccount.cs
dragav: Lowered to Medium; please note, however, that removal of this entity from the vault can cause immediate disruption of service - callers will not be able to obtain access to storage via KV anymore.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/RemoveAzureKeyVaultManagedStorageSasDefinition.cs
dragav: Lowered to Medium.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageAccount.cs
dragav: Given that an Update-AKVManagedStorageAccount cmdlet is being introduced with these changes, I'm renaming the Set- cmdlet to Add-. That way the PUT/add semantics are clearer.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageAccount.cs
dragav: Will address this in the subsequent release.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/SetAzureKeyVaultManagedStorageSasDefinition.Service.Share.File.cs
dragav: (addressed)

dragav:
@markcowl, @cormacpayne: your feedback was addressed with commit 4538b66. We won't be making any further changes - any unaddressed feedback should be considered "won't fix". Please review at your convenience and let us know (@kunalcha and me) if this will be taken in the June release. Many thanks!

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/UpdateAzureKeyVaultManagedStorageAccount.cs
```diff
@@ -0,0 +1,106 @@
...
+            Position = 1,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Key Vault managed storage account name. Cmdlet constructs the FQDN of a managed storage account name from vault name, currently " +
+                          "selected environment and manged storage account name." )]
+        [ValidateNotNullOrEmpty]
+        [Alias( Constants.StorageAccountName, Constants.Name )]
+        public string AccountName { get; set; }
+
+        [Parameter( Mandatory = false,
+            ValueFromPipelineByPropertyName = true,
```
markcowl: For the future - for PATCH scenarios only identity parameters should be able to get their value from the pipeline

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/UpdateAzureKeyVaultManagedStorageAccountKey.cs
```diff
@@ -0,0 +1,87 @@
...
+using Microsoft.Azure.Commands.KeyVault.Models;
+using System.Globalization;
+using System.Management.Automation;
+using KeyVaultProperties = Microsoft.Azure.Commands.KeyVault.Properties;
+
+namespace Microsoft.Azure.Commands.KeyVault
+{
+    [Cmdlet( VerbsData.Update, CmdletNoun.AzureKeyVaultManagedStorageAccountKey,
+        SupportsShouldProcess = true,
+        ConfirmImpact = ConfirmImpact.High,
```
markcowl: Why doies this have high confirm impact?

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/UpdateAzureKeyVaultManagedStorageAccountKey.cs
```diff
@@ -0,0 +1,87 @@
...
+        [Parameter( Mandatory = true,
+            Position = 2,
+            HelpMessage = "Name of storage account key to regenerate and make active." )]
+        [ValidateNotNullOrEmpty]
+        public string KeyName { get; set; }
+
+        /// <summary>
+        /// If present, do not ask for confirmation
+        /// </summary>
+        [Parameter( Mandatory = false, HelpMessage = "Do not ask for confirmation." )]
```
markcowl: Why does this have a Force parameter?

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/UpdateAzureKeyVaultManagedStorageAccountKey.cs
dragav: @markcowl: this cmdlet in effect changes the account key on the storage account, and so it can cause immediate disruption of service. As I understand, there were background discussions with you and Kunal re: finding a more appropriate verb. We must retain the 'high impact' semantics of this cmdlet, and the ability to force a change.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/UpdateAzureKeyVaultManagedStorageAccountKey.cs
dragav: As above, this replaces the previous New-AKVMSAK cmdlet, and was renamed specifically to avoid the restrictions on 'new' verbs; this command rolls over the storage account key with immediate effect and is therefore overwriting an existing value. Force allows the user to apply the changes.

markcowl:
on demand run here: http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1655/

PR #4105, opened 2017-06-09
Removing the 2 second polling and matching latest azure behavour for resource group deployment

bganapa:
This matches the latest azure behaviour in polling intervals

### DIFF
--- a/src/ResourceManager/Resources/Commands.Resources/Models.ResourceGroups/ResourceClient.cs
+++ b/src/ResourceManager/Resources/Commands.Resources/Models.ResourceGroups/ResourceClient.cs
@@ -313,16 +313,29 @@ namespace Microsoft.Azure.Commands.Resources.Models
         {
             DeploymentExtended deployment;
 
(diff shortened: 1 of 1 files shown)


src/ResourceManager/Resources/Commands.Resources/Models.ResourceGroups/ResourceClient.cs
```diff
@@ -313,16 +313,29 @@ public static List<string> ParseDetailErrorMessage(string statusMessage)
         {
             DeploymentExtended deployment;
 
+            // Poll deployment state and deployment operations after RetryAfter.
+            // If no RetryAfter provided: In phase one, poll every 5 seconds. Phase one 
+            // takes 400 seconds. In phase two, poll every 60 seconds.
```
cormacpayne: @bganapa out of curiosity, does phase two have a set length? Phase one is 400 seconds, but it doesn't look like phase two has a length and would run indefinitely. Is that what is expected?

markcowl:
This change is in another PR< closing as per the discussion there

Issue #635, opened 2015-07-21
ARM 0.9.5 :: Get-AzureSqlServer requires -ResourceGroupName parameter


pcgeek86:
# Symptom

The `Get-AzureSqlServer` command requires the `-ResourceGroupName` parameter.
# Expected Results

The `Get-AzureSqlServer` command should not require the `-ResourceGroupName` parameter.

Cheers,
Trevor Sullivan
Microsoft MVP: PowerShell

devigned:
It's been tagged and it's being triaged. Thank you for the issue report.

akromm:
@pcgeek86 Its been a while, but I've started working on this :)

akromm:
#4188 has been merged and will be in the next powershell release!

cormacpayne:
This fix can be found in [version 4.2.0](https://github.com/Azure/azure-powershell/releases/tag/v4.2.0-July2017) of Azure PowerShell.

PR #4188, opened 2017-06-23
Making -ResourceGroupName parameter optional for Get-AzureRmSqlServer cmdlet

akromm-zz:
## Description

---

- [ ] **I have read the [contribution guidelines](https://github.com/Azure/azure-powershell/blob/preview/CONTRIBUTING.md).**

- [ ] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR. For more information on cleaning up the commits in your PR, [see this page](https://github.com/Azure/azure-powershell/blob/preview/documentation/cleaning-up-commits.md).
- [ ] The pull request does not introduce [breaking changes](https://github.com/Azure/azure-powershell/blob/preview/documentation/breaking-changes/breaking-changes-definition.md) (unless a major version change occurs in the assembly and module).

### DIFF
--- a/src/ResourceManager/Sql/Commands.Sql.Test/Commands.Sql.Test.csproj
+++ b/src/ResourceManager/Sql/Commands.Sql.Test/Commands.Sql.Test.csproj
@@ -91,9 +91,8 @@
     <Reference Include="Microsoft.Azure.Test.Framework">
       <HintPath>..\..\..\packages\Microsoft.Azure.Test.Framework.1.0.6179.26854-prerelease\lib\net45\Microsoft.Azure.Test.Framework.dll</HintPath>
     </Reference>
-    <Reference Include="Microsoft.Azure.Test.HttpRecorder, Version=1.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
-      <HintPath>..\..\..\packages\Microsoft.Azure.Test.HttpRecorder.1.8.1\lib\net452\Microsoft.Azure.Test.HttpRecorder.dll</HintPath>
-      <Private>True</Private>
+    <Reference Include="Microsoft.Azure.Test.HttpRecorder">
+      <HintPath>..\..\..\packages\Microsoft.Azure.Test.HttpRecorder.1.6.7-preview\lib\net45\Microsoft.Azure.Test.HttpRecorder.dll</HintPath>
     </Reference>
     <Reference Include="Microsoft.Data.Edm">
       <SpecificVersion>False</SpecificVersion>
@@ -124,12 +123,11 @@
       <Private>True</Private>
     </Reference>
     <Reference Include="Microsoft.Rest.ClientRuntime.Azure.Authentication, Version=2.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
-      <HintPath>..\..\..\packages\Microsoft.Rest.ClientRuntime.Azure.Authentication.2.3.0\lib\net452\Microsoft.Rest.ClientRuntime.Azure.Authentication.dll</HintPath>
+      <HintPath>..\..\..\packages\Microsoft.Rest.ClientRuntime.Azure.Authentication.2.2.9-preview\lib\net45\Microsoft.Rest.ClientRuntime.Azure.Authentication.dll</HintPath>
       <Private>True</Private>
     </Reference>
-    <Reference Include="Microsoft.Rest.ClientRuntime.Azure.TestFramework, Version=1.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
-      <HintPath>..\..\..\packages\Microsoft.Rest.ClientRuntime.Azure.TestFramework.1.7.2\lib\net452\Microsoft.Rest.ClientRuntime.Azure.TestFramework.dll</HintPath>
-      <Private>True</Private>
+    <Reference Include="Microsoft.Rest.ClientRuntime.Azure.TestFramework">
+      <HintPath>..\..\..\packages\Microsoft.Rest.ClientRuntime.Azure.TestFramework.1.5.0-preview\lib\net45\Microsoft.Rest.ClientRuntime.Azure.TestFramework.dll</HintPath>
     </Reference>
     <Reference Include="Microsoft.Threading.Tasks, Version=1.0.12.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
       <SpecificVersion>False</SpecificVersion>
--- a/src/ResourceManager/Sql/Commands.Sql.Test/packages.config
+++ b/src/ResourceManager/Sql/Commands.Sql.Test/packages.config
@@ -11,7 +11,7 @@
   <package id="Microsoft.Azure.Management.Sql" version="1.5.0-preview" targetFramework="net452" />
   <package id="Microsoft.Azure.Management.Storage" version="2.4.0-preview" targetFramework="net45" />
   <package id="Microsoft.Azure.Test.Framework" version="1.0.6179.26854-prerelease" targetFramework="net45" />
-  <package id="Microsoft.Azure.Test.HttpRecorder" version="1.8.1" targetFramework="net452" />
+  <package id="Microsoft.Azure.Test.HttpRecorder" version="1.6.7-preview" targetFramework="net45" />
   <package id="Microsoft.Bcl" version="1.1.9" targetFramework="net45" />
   <package id="Microsoft.Bcl.Async" version="1.0.168" targetFramework="net45" />
   <package id="Microsoft.Bcl.Build" version="1.0.14" targetFramework="net45" />
@@ -27,7 +27,7 @@
   <package id="Microsoft.WindowsAzure.Management" version="4.1.1" targetFramework="net45" />
   <package id="Microsoft.WindowsAzure.Management.Storage" version="5.1.1" targetFramework="net45" />
   <package id="Moq" version="4.2.1510.2205" targetFramework="net45" />
-  <package id="Newtonsoft.Json" version="9.0.1" targetFramework="net452" />
+  <package id="Newtonsoft.Json" version="6.0.8" targetFramework="net45" />
   <package id="System.Spatial" version="5.6.4" targetFramework="net45" />
   <package id="WindowsAzure.Storage" version="5.0.0" targetFramework="net45" />
   <package id="xunit" version="2.1.0" targetFramework="net45" />


akromm:
#635

src/ResourceManager/Sql/Commands.Sql.Test/ScenarioTests/ServerCrudTests.ps1
```diff
@@ -85,29 +85,34 @@ function Test-GetServer
...
 		Assert-AreEqual $server2.SqlAdministratorLogin $resp2.SqlAdministratorLogin
-
+		
 		$all = Get-AzureRmSqlServer -ResourceGroupName $rg.ResourceGroupName
-		Assert-AreEqual $all.Count 2
+		Assert-AreEqual 2 $all.Count
+
+		# Test getting all servers in all resource groups
+		$all2 = Get-AzureRmSqlServer
+		Assert-AreEqual 3 $all2.Count
```
jaredmoo: This will fail if I run it in my subscription that already has some sql servers unrelated to PS test. Can you change this so that it checks that the list contains the 3 previous servers?

src/ResourceManager/Sql/Commands.Sql/Common/AzureSqlCmdletBaseBase.cs
```diff
@@ -0,0 +1,149 @@
...
+using Microsoft.Azure.Commands.Sql.Services;
+using System.Management.Automation;
+using Microsoft.Azure.Commands.Common.Authentication.Abstractions;
+
+namespace Microsoft.Azure.Commands.Sql.Common
+{
+    /// <summary>
+    /// The base class for all Azure Sql cmdlets
+    /// </summary>
+    public abstract class AzureSqlCmdletBaseBase<M, A> : AzureRMCmdlet
```
jaredmoo: haha okay ;)

src/ResourceManager/Sql/Commands.Sql/Server/Cmdlet/GetAzureSqlServer.cs
```diff
@@ -26,9 +29,19 @@ namespace Microsoft.Azure.Commands.Sql.Server.Cmdlet
     /// Defines the Get-AzureRmSqlServer cmdlet
     /// </summary>
     [Cmdlet(VerbsCommon.Get, "AzureRmSqlServer", ConfirmImpact = ConfirmImpact.None, SupportsShouldProcess = true)]
-    public class GetAzureSqlServer : AzureSqlServerCmdletBase, IModuleAssemblyInitializer
+    public class GetAzureSqlServer : AzureSqlCmdletBaseBase<IEnumerable<AzureSqlServerModel>, AzureSqlServerAdapter>, IModuleAssemblyInitializer
     {
         /// <summary>
+        /// Gets or sets the name of the resource group to use.
+        /// </summary>
+        [Parameter(Mandatory = false,
```
jaredmoo: can you <remarks> or some comment explaining why BaseBase when you have ResourceGroup param? I understand that's it's for `Mandatory=false` when looking at this change in isolation, but it won't be obvious when reading this code later

src/ResourceManager/Sql/Commands.Sql/Server/Cmdlet/GetAzureSqlServer.cs
```diff
@@ -40,21 +53,39 @@ public class GetAzureSqlServer : AzureSqlServerCmdletBase, IModuleAssemblyInitia
...
+                results = ModelAdapter.ListServersByResourceGroup(this.ResourceGroupName);
+            }
+            else if (!MyInvocation.BoundParameters.ContainsKey("ServerName"))
+            {
+                results = ModelAdapter.ListServers();
+            }
             else
             {
-                results = ModelAdapter.GetServers(this.ResourceGroupName);
+                throw new PSArgumentException("When specifying the serverName parameter the ResourceGroup parameter must also be used");
```
jaredmoo: Can you split this into 2 ParamGroups so that it's enforced by powershell logic instead of custom code?
Group 1: `Get-AzureRmSqlServer`
Group 2: `Get-AzureRmSqlServer -ResourceGroupName [- ServerName]`

jaredmoo: I might be wrong here, not sure if "no parameters" is possible to specify as a parameter set.

src/ResourceManager/Sql/Commands.Sql.Test/ScenarioTests/ServerCrudTests.ps1
akromm: Sure, I can do that.

src/ResourceManager/Sql/Commands.Sql/Common/AzureSqlCmdletBaseBase.cs
akromm: My creativity shining through :)

src/ResourceManager/Sql/Commands.Sql/Server/Cmdlet/GetAzureSqlServer.cs
akromm: added a comment.

src/ResourceManager/Sql/Commands.Sql/Server/Cmdlet/GetAzureSqlServer.cs
akromm: I thought a bit about this, but wasn't sure how to achieve it.  According to [MSDN-Parameter Set Requirements](https://msdn.microsoft.com/en-us/library/dd878348(v=vs.85).aspx#Anchor_3) a parameter set must have at least 1 parameter.

Another option is to allow specifying the ServerName without the resourceGroupName and then just retrieve the list of all servers and filter it down to just the one with the matching name (potentially bad perf)? The underlying API doesn't support retrieving a specific server without the resource group name.

src/ResourceManager/Sql/Commands.Sql.Test/ScenarioTests/ServerCrudTests.ps1
```diff
@@ -107,7 +107,17 @@ function Test-GetServer
 
 		# Test getting all servers in all resource groups
 		$all2 = Get-AzureRmSqlServer
-		Assert-AreEqual 3 $all2.Count
+
+		# It is possible that there were existing servers in the subscription when the test was recorded, so make sure
+		# that the servers that we created are retrieved and ignore the other ones.
```
jaredmoo: This works, but it's a bit confusing to read. Does something like this work?
`($server1, $server2, $server3) | foreach { Assert-True $_.ServerName -in $all2.ServerName }`

akromm: Interestingly that does work. I wasn't aware that trying to access a property on an array would return the property value for each object in the array.

src/ResourceManager/Sql/Commands.Sql/Common/AzureSqlCmdletBaseBase.cs
shahabhijeet: @akromm any reason there aren't any constraints for M, A?

src/ResourceManager/Sql/Commands.Sql/Common/AzureSqlCmdletBaseBase.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// <summary>
+        /// Default constructor.  Generates a request ID
+        /// </summary>
+        internal AzureSqlCmdletBaseBase()
+        {
+        }
+
+        protected virtual string GetResourceId(M model)
+        {
+            var serverProperty = model.GetType().GetProperty("ServerName");
```
shahabhijeet: @akromm maybe I am missing to see how this will work when model is of type IEnumerable<AzureSqlServerModel>

src/ResourceManager/Sql/Commands.Sql/Common/AzureSqlCmdletBase.cs
```diff
@@ -59,87 +36,5 @@ protected virtual string GetResourceId(M model)
             HelpMessage = "The name of the resource group.")]
         [ValidateNotNullOrEmpty]
         public string ResourceGroupName { get; set; }
```
akromm: @shahabhijeet is relaxing the requirements of parameters a breaking change?  I'm doing this change because there have been multiple github issues requesting it. eg: #635

cormacpayne: @akromm @shahabhijeet changing a parameter from mandatory to non-mandatory is not a breaking change; however, the other way around is a breaking change

src/ResourceManager/Sql/Commands.Sql/Common/AzureSqlCmdletBaseBase.cs
akromm: @shahabhijeet I have updated the method to work with IEnumerable and other single entity types. I have also added a corresponding unittest to ensure it works.

src/ResourceManager/Sql/Commands.Sql/Common/AzureSqlCmdletBase.cs
shahabhijeet: @cormacpayne thanks for correcting. I wrote that comment and then a day later realized that is not right what I wrote, but by that time I was not able to come back to GitHub to correct it.

src/ResourceManager/Sql/Commands.Sql/Common/AzureSqlCmdletBaseBase.cs
```diff
@@ -0,0 +1,169 @@
...
+        {
+            return Properties.Resources.BaseConfirmActionProcessMessage;
+        }
+
+        /// <summary>
+        /// Executes the cmdlet
+        /// </summary>
+        public override void ExecuteCmdlet()
+        {
+            ModelAdapter = InitModelAdapter(DefaultProfile.DefaultContext.Subscription);
```
akromm: @shahabhijeet InitModelAdapter is implemented in each individual cmdlet. [Database Cmdlet base example](https://github.com/Azure/azure-powershell/blob/preview/src/ResourceManager/Sql/Commands.Sql/Database/Cmdlet/AzureSqlDatabaseCmdletBase.cs#L42)

src/ResourceManager/Sql/Commands.Sql/Common/AzureSqlCmdletBaseBase.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// <summary>
+        /// Default constructor.  Generates a request ID
+        /// </summary>
+        internal AzureSqlCmdletBaseBase()
+        {
+        }
+
+        protected virtual string GetResourceId(M model)
+        {
+            object m = null;
```
jaredmoo: So that overriding methods in subclasses can use the specific type

src/ResourceManager/Sql/Commands.Sql/Common/AzureSqlCmdletBaseBase.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// <param name="subscription">The AzureSubscription in which the current execution is performed</param>
+        /// <returns>An initialized and ready to use ModelAdapter object</returns>
+        protected abstract A InitModelAdapter(IAzureSubscription subscription);
+
+        /// <summary>
+        /// Transforms the given model object to be an object that is written out
+        /// </summary>
+        /// <param name="model">The about to be written model object</param>
+        /// <returns>The prepared object to be written out</returns>
+        protected virtual object TransformModelToOutputObject(M model)
```
akromm: This is only used in a couple places.
[Get-AzureRmSqlDatabaseDataMaskingRule](https://github.com/Azure/azure-powershell/blob/preview/src/ResourceManager/Sql/Commands.Sql/Data%20Masking/Cmdlet/GetAzureSqlDatabaseDataMaskingRule.cs#L50) cmdlet uses it to filter the results before showing the user.
[SqlDatabaseDataMaskingRuleCmdletBase](https://github.com/Azure/azure-powershell/blob/preview/src/ResourceManager/Sql/Commands.Sql/Data%20Masking/Cmdlet/SqlDatabaseDataMaskingRuleCmdletBase.cs#L63) also uses it to do client side filtering of the results.
[New-AzureRmSqlDatabase](https://github.com/Azure/azure-powershell/blob/preview/src/ResourceManager/Sql/Commands.Sql/Database/Cmdlet/NewAzureSqlDatabase.cs#L203) Actually uses it to modify the response slightly.

That being said, all this code is pre-existing code that already existed. I'm just moving it to a new file.  I might have a different way to achieve my main goal without moving this stuff around.

src/ResourceManager/Sql/Commands.Sql/Common/AzureSqlCmdletBaseBase.cs
akromm: @shahabhijeet I think the right thing to do here is make  the method Abstract and force each cmdlet to implement it.

To answer your question. It is of type object because M can be IEnumerable<T> and I want an object of type T not M.

src/ResourceManager/Sql/Commands.Sql/Common/AzureSqlCmdletBaseBase.cs
```diff
@@ -0,0 +1,169 @@
...
+        protected string clientRequestId { get; set; }
+
+        /// <summary>
+        /// Default constructor.  Generates a request ID
+        /// </summary>
+        internal AzureSqlCmdletBaseBase()
+        {
+        }
+
+        protected virtual string GetResourceId(M model)
```
akromm: @shahabhijeet  This would be overridden per cmdlet type (Database, Server, DataMasking, ...) and as such the type M will be well known and there shouldn't be the complex logic I have.

akromm:
@cormacpayne @shahabhijeet @jaredmoo Finally got the tests to pass (issues with dependencies and NuGet package versions).  Can I please get this reviewed and merged now :)

Thanks.

shahabhijeet:
On-Demand
http://azuresdkci.cloudapp.net/view/1-AzurePowerShell/job/powershell-demand/1724/

cormacpayne:
@azuresdkci test this please

shahabhijeet:
@azuresdkci test this please

PR #4764, opened 2017-10-10
BootStrapper module updates for the release 1.2.11 and 4.4.1

bganapa:


tools/AzureRM.BootStrapper/AzureRM.BootStrapper.psd1
```diff
@@ -12,7 +12,7 @@
 RootModule = 'AzureRM.Bootstrapper.psm1'
 
 # Version number of this module.
-ModuleVersion = '0.2.0'
+ModuleVersion = '0.3.0'
```
viananth: Since this is a change to metadata and not the bootstrapper module, I guess we don't need to bump the version. It could be 0.2.1 or similar

Issue #4948, opened 2017-11-09
Set-AzureServiceDiagnosticsExtension fails InvalidArgument


huan086:
### PowerShell Version
5.1.14393.1770

### Module Version
5.0.0

### OS Version
10.0.14393.1770

### Description
Cannot use the overload of the cmdlet that takes -StorageContext 
See https://stackoverflow.com/questions/46607438/not-able-to-enable-diagnostics-for-web-role-using-powershell

### Debug Output
```powershell
Set-AzureServiceDiagnosticsExtension : Cannot bind parameter 'StorageContext'. Cannot convert the
"Microsoft.WindowsAzure.Commands.Storage.AzureStorageContext" value of type
"Microsoft.WindowsAzure.Commands.Storage.AzureStorageContext" to type
"Microsoft.WindowsAzure.Commands.Common.Storage.AzureStorageContext".
At line:1 char:54
+ ... reServiceDiagnosticsExtension -StorageContext $storageContext -Diagno ...
+                                                   ~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Set-AzureServiceDiagnosticsExtension], ParameterBindingException
    + FullyQualifiedErrorId : CannotConvertArgumentNoMessage,Microsoft.WindowsAzure.Commands.ServiceManagement.Extensi
   ons.SetAzureServiceDiagnosticsExtensionCommand
```

### Script/Steps for Reproduction
```powershell
$storageContext = New-AzureStorageContext $storageAccountName -StorageAccountKey $storageAccountKey.Primary
Set-AzureServiceDiagnosticsExtension -StorageContext $storageContext -DiagnosticsConfigurationPath $diagnosticsPath -ServiceName $cloudServiceName -Slot Staging -Role @('Example.Web', 'Example.WorkerRole')
```

markcowl:
@huan086 This is a very old RDFE-based cmdlet.  Is there a reason why you are not using the ARM VMs? You should also be able to work around this using the 'StorageAccountName, StorageAccountKey, and STorageAccountENdppoint parameters instead of the context.

markcowl:
@singhkays @huangpf  The fix for this would be to change this parameter to IStorageContext, so that it is independent of the particula storage version sued to create the context (as has been done for all of the ARM cmdlets).

singhkays:
@hyonholee Can you take a look?

yantang-msft:
[...]

PR #4981, opened 2017-11-13
Fix issue 4948. SetServiceDiagnosticsExtension should use IStorageContext

yantang-msft:
Fix issue https://github.com/Azure/azure-powershell/issues/4948. SetServiceDiagnosticsExtension and other related commands should user IStorageContext instead of AzureStorageContext, so it's irrelevant to the actual storage context type.


src/ServiceManagement/Compute/Commands.ServiceManagement/Common/DiagnosticsHelper.cs
```diff
@@ -551,7 +551,7 @@ private static bool TryGetStorageAccount(IStorageManagementClient storageClient,
             return storageAccount != null;
         }
 
-        private static AzureStorageContext CreateStorageContext(Uri blobEndpoint, Uri queueEndpoint, Uri tableEndpoint, Uri fileEndpoint,
```
markcowl: I don't think it's necessary to change this - AzureStorageContext implements IStorageContext - as long as this is not public, it should be fine to keep this internally.

src/ServiceManagement/Compute/Commands.ServiceManagement.Test/FunctionalTests/ServiceManagementCmdletTestHelper.cs
```diff
@@ -1200,17 +1199,17 @@ public ExtensionConfigurationInput NewAzureServiceDiagnosticsExtensionConfig
...
-            (string service, AzureStorageContext storageContext, string config = null, string[] roles = null, string slot = null)
+            (string service, IStorageContext storageContext, string config = null, string[] roles = null, string slot = null)
         {
             return RunPSCmdletAndReturnFirst<ManagementOperationContext>(new SetAzureServiceDiagnosticsExtensionCmdletInfo(service, storageContext, config, roles, slot));
         }
 
-        public ManagementOperationContext SetAzureServiceDiagnosticsExtension(string service, AzureStorageContext storageContext, X509Certificate2 cert, string config = null, string[] roles = null, string slot = null)
+        public ManagementOperationContext SetAzureServiceDiagnosticsExtension(string service, IStorageContext storageContext, X509Certificate2 cert, string config = null, string[] roles = null, string slot = null)
         {
             return RunPSCmdletAndReturnFirst<ManagementOperationContext>(new SetAzureServiceDiagnosticsExtensionCmdletInfo(service, storageContext, cert, config, roles, slot));
```
markcowl: Doesn't this generic method need to be changed to support IStorageContext?

src/ServiceManagement/Compute/Commands.ServiceManagement/Common/DiagnosticsHelper.cs
yantang-msft: You're right. But it's just a simple change and I think it's better to use this interface wherever we can, unless a specific storage type is really needed.

src/ServiceManagement/Compute/Commands.ServiceManagement.Test/FunctionalTests/ServiceManagementCmdletTestHelper.cs
yantang-msft: It doesn't need to be changed. Just want to clean up the places where it doesn't need to be a specific storage type.
I'll keep it like this unless you have strong preference.

src/ServiceManagement/Compute/Commands.ServiceManagement/Common/DiagnosticsHelper.cs
markcowl: What I'm getting at is that there are specific methods for getting the storage credentials form an IStorageContext.  Are those used in this implementation?:

yantang-msft: I just looked at the code, no, the credentials is not actually being used.
I believe the reason we create the AzureStorageContext (which need the credential) is simply reusing the logic of EndPointSuffix property of AzureStorageContext, since we don't want to maintain this piece of logic.

markcowl:
@yantang-msft What kind of end-to-end validation have you been able to do on this cmdlet.  It worries me that this can be broken without failing any of the automated validation that we have, so I am nervous about taking this fix without some additonal test validation - if you have done end-to-end live validation, we can defer test automation for now.  Please let me know.

yantang-msft:
@markcowl Yep, I tested this command from end to end and it works. :)

src/ServiceManagement/Compute/Commands.ServiceManagement/Extensions/Diagnostics/NewAzureServiceDiagnosticsExtensionConfig.cs
```diff
@@ -89,7 +89,7 @@ public override string StorageAccountEndpoint
         [Parameter(Position = 6, ValueFromPipelineByPropertyName = true, ParameterSetName = NewExtensionParameterSetName, HelpMessage = "Diagnostics Storage Name")]
         [Parameter(Position = 6, ValueFromPipelineByPropertyName = true, ParameterSetName = NewExtensionUsingThumbprintParameterSetName, HelpMessage = "Diagnostics Storage Name")]
         [ValidateNotNullOrEmpty]
-        public override AzureStorageContext StorageContext
```
markcowl: I will only feel comfprtable if there is an end-to-end test of this.  It is OK if you test this end-to-end by hand for this release, but ifd so, please file an issue for adding or altering test automation to cover thsi scenari (passing a storage context to this cmdlet).

markcowl:
on demand run here: https://azuresdkci.westus2.cloudapp.azure.com/job/powershell-demand/232/

Issue #4966, opened 2017-11-11
Add ability to avoid validation calls in CloudShell login


markcowl:
### Description

- When using the token parameter set in ```Add-AzureRmAccount```, add an optional switch -SkipValidation which does not list tenants or subscriptions
- Add end-to-end tests for the new parameter set

### Cost: 4

maddieclayton:
Fixed.

PR #5070, opened 2017-12-01
Add ability to avoid validation calls in CloudShell login

maddieclayton:
## Description
https://github.com/Azure/azure-powershell/issues/4966

---

- [ ] **I have read the [contribution guidelines](https://github.com/Azure/azure-powershell/blob/preview/CONTRIBUTING.md).**

- [ ] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR. For more information on cleaning up the commits in your PR, [see this page](https://github.com/Azure/azure-powershell/blob/preview/documentation/cleaning-up-commits.md).
- [ ] The pull request does not introduce [breaking changes](https://github.com/Azure/azure-powershell/blob/preview/documentation/breaking-changes/breaking-changes-definition.md) (unless a major version change occurs in the assembly and module).

### DIFF
--- a/src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
+++ b/src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
@@ -18,6 +18,7 @@ using Microsoft.Azure.Commands.Common.Authentication.Factories;
 using Microsoft.Azure.Commands.Common.Authentication.Models;
 using Microsoft.Azure.Commands.Common.Authentication.ResourceManager;
 using Microsoft.Azure.Commands.Profile.Models;
+using Microsoft.Azure.Commands.Profile.Properties;
 using Microsoft.Azure.Internal.Subscriptions;
 using Microsoft.Azure.Internal.Subscriptions.Models;
 using Microsoft.Rest;
@@ -122,7 +123,7 @@ namespace Microsoft.Azure.Commands.ResourceManager.Common
             {
                 if (string.IsNullOrEmpty(subscriptionId) || string.IsNullOrEmpty(tenantId))
                 {
-                    throw new PSInvalidOperationException("Must supply Subscription Id and Tenant Id when using -SkipValidation");
+                    throw new PSInvalidOperationException(Resources.SubscriptionOrTenantMissing);
                 }
 
                 newSubscription = new AzureSubscription
--- a/src/ResourceManager/Profile/Commands.Profile/Properties/Resources.Designer.cs
+++ b/src/ResourceManager/Profile/Commands.Profile/Properties/Resources.Designer.cs
@@ -754,6 +754,15 @@ namespace Microsoft.Azure.Commands.Profile.Properties {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to Must supply Subscription Id and Tenant Id when using -SkipValidation.
+        /// </summary>
+        internal static string SubscriptionOrTenantMissing {
+            get {
+                return ResourceManager.GetString("SubscriptionOrTenantMissing", resourceCulture);
+            }
+        }
+        
+        /// <summary>
         ///   Looks up a localized string similar to Please provide a valid tenant or a valid subscription..
         /// </summary>
         internal static string SubscriptionOrTenantRequired {
--- a/src/ResourceManager/Profile/Commands.Profile/Properties/Resources.resx
+++ b/src/ResourceManager/Profile/Commands.Profile/Properties/Resources.resx
@@ -400,4 +400,7 @@
   <data name="CreateResourceGroupMessage" xml:space="preserve">
     <value>Create new resource group '{0}' and set as default?</value>
   </data>
+  <data name="SubscriptionOrTenantMissing" xml:space="preserve">
+    <value>Must supply Subscription Id and Tenant Id when using -SkipValidation</value>
+  </data>
 </root>


maddieclayton:
@azuresdkci test this please

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
```diff
@@ -117,86 +118,118 @@ public bool TryRemoveContext(IAzureContext context)
...
                 {
-                    var tenant = tenants[i];
+                    var tenants = ListAccountTenants(account, environment, password, promptBehavior, promptAction)
+                        .Select(s => s.Id.ToString()).ToArray();
+                    account.SetProperty(AzureAccount.Property.Tenants, null);
+                    string accountId = null;
 
-                    IAzureTenant tempTenant;
-                    IAzureSubscription tempSubscription;
+                    for (int i = 0; i < tenants.Count(); i++)
```
twitchax: `tenants` is an array, so you should use `tenants.Length`. :)

Any time you have a concrete `List<T>` or `T[]`, you should use the O(1) operations on the concrete type.  For arrays, it is `.Length`, for lists, it is `.Count`.

However, since arrays and lists are both enumerable, they get all of the `IEnumberable` methods, which is why you will see `IEnumerable<T>.Count()` in the intellisense.  All methods on `IEnumerable` will re-enumerate the collection, so `.Count()` is an O(n) operation.

Also, I would actually do `.ToList()` for `tenants`.  `List<T>`s are [almost alwyas](https://stackoverflow.com/questions/434761/array-versus-listt-when-to-use-which) preferred over `T[]`s.  It's not a big deal either way, but it's just kind of a C# standard practice, at this point, to use lists unless an array is specifically required.

Lastly, you might prefer a `foreach` here, and I will make a note of how to address the `i` comparison below.

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
```diff
@@ -117,86 +118,118 @@ public bool TryRemoveContext(IAzureContext context)
...
-                        if (tempSubscription != null || i == (tenants.Count() - 1))
+                        if (token != null &&
+                            newTenant == null &&
+                            TryGetTenantSubscription(token, account, environment, tenant, subscriptionId, subscriptionName, out tempSubscription, out tempTenant))
                         {
-                            newTenant = tempTenant;
-                            newSubscription = tempSubscription;
+                            // If no subscription found for the given token/tenant 
+                            // discard tempTenant value unless current token/tenant is the last one.
+                            if (tempSubscription != null || i == (tenants.Count() - 1))
```
twitchax: Assuming you have a `foreach(var tenant in tenants)` above, this could be

```csharp
tenant == tenants[tenants.Count - 1]
```

twitchax: Also, this `if` can be merged with the enclosing `if` since there are no other statements in the enclosing block.

Unless, is the `out` param from `TryGetTenantSubscription` not available in the `if`-condition?

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
```diff
@@ -117,86 +118,118 @@ public bool TryRemoveContext(IAzureContext context)
...
 
-                    IAzureTenant tempTenant;
-                    IAzureSubscription tempSubscription;
+                    for (int i = 0; i < tenants.Count(); i++)
+                    {
+                        var tenant = tenants[i];
 
-                    IAccessToken token = null;
+                        IAzureTenant tempTenant;
+                        IAzureSubscription tempSubscription;
```
twitchax: This should not be required anymore in C# 7.

You can now call a method and declare the `out` parameter in one statement.  E.g.,

```csharp
Function(input1, input1, out var output1, out var output2);
```

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
maddieclayton: @twitchax Actually none of this code is mine, I just moved it into an else block to accommodate the -SkipValidation parameter. I'm happy to refactor though.

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
maddieclayton: Because C# 7 doesn't easily work in VS2015 (which we encourage people to use) I'm inclined not to put c# 7 code into the Profile module as this might cause a lot of people errors which we will then have to tell them how to fix.

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
maddieclayton: Done

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
maddieclayton: I would agree with whoever actually wrote this that having 5 conditions in the if statement is difficult to read, and that it seems logically to split the two because of the evaluation happening in TryGetTenantSubscription (also the additional comment).

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
```diff
@@ -117,86 +118,116 @@ public bool TryRemoveContext(IAzureContext context)
...
-                        if (tempSubscription != null || i == (tenants.Count() - 1))
+                        if (token != null &&
+                            newTenant == null &&
+                            TryGetTenantSubscription(token, account, environment, tenant, subscriptionId, subscriptionName, out tempSubscription, out tempTenant))
                         {
-                            newTenant = tempTenant;
-                            newSubscription = tempSubscription;
+                            // If no subscription found for the given token/tenant 
+                            // discard tempTenant value unless current token/tenant is the last one.
+                            if (tempSubscription != null || tenant.Equals(tenants[tenants.Count]))
```
markcowl: won't this be one past the end of the tenant list?

src/ResourceManager/Profile/Commands.Profile/Account/AddAzureRmAccount.cs
```diff
@@ -163,22 +167,27 @@ public override void ExecuteCmdlet()
 
             AzureAccount azureAccount = new AzureAccount();
 
-            switch(ParameterSetName)
+            if (!string.IsNullOrEmpty(AccessToken))
```
markcowl: I am worried about changing from a switch statement to an if based on the current parameters

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
twitchax: Yeah, you want `tenant == tenants[tenants.Count - 1]`.

Why do we lean towards `.Equals`?

src/ResourceManager/Profile/Commands.Profile/Account/AddAzureRmAccount.cs
maddieclayton: I'm still seeing the issue where ParameterSetName is null while running these tests. Let's discuss tomorrow.

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
maddieclayton: Thanks for the catch guys. @twitchax I lean towards .Equals because that's how you are supposed to compare strings in Java.  Is it different in C#?

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
twitchax: I wish there were a way to "dismiss" a comment without deleting it.

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
twitchax: Ahhh, so, it is a bit different in C#.  `==` is normally preferred, but you might want to make it clear that you _care_ about reference equality.  In that case, it may be best to make it explicit by doing `object.ReferenceEquals(tenant, tenants[tenants.Count - 1]`.

For strings, `==` is a value comparison, and `.Equals` is a reference compare.  Either will work, but you're right, the reference compare is better.

maddieclayton:
optimistic on demand run here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/powershell-demand/240/

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
```diff
@@ -117,86 +118,116 @@ public bool TryRemoveContext(IAzureContext context)
...
-                if (TryGetTenantSubscription(
-                    token, 
-                    account, 
-                    environment, 
-                    tenantId, 
-                    subscriptionId, 
-                    subscriptionName, 
-                    out newSubscription, 
-                    out newTenant))
+                if ((string.IsNullOrEmpty(subscriptionId) && string.IsNullOrEmpty(subscriptionName)) || string.IsNullOrEmpty(tenantId))
```
markcowl: We should require subscription ID - we need it for all of the subscription calls

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
```diff
@@ -117,86 +118,116 @@ public bool TryRemoveContext(IAzureContext context)
...
+                {
+                    throw new PSInvalidOperationException("Must supply Subscription Id/Name and Tenant Id when using -SkipValidation");
+                }
+
+                if (!string.IsNullOrEmpty(subscriptionId))
                 {
-                    account.SetOrAppendProperty(AzureAccount.Property.Tenants, new[] { newTenant.Id.ToString() });
+                    newSubscription = new AzureSubscription
```
markcowl: We should set up the subscription with both ID and Name (if set)

markcowl: We will also want to fill out the additional properties in Subscription (TenantId, accountId?)

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
maddieclayton: I cannot separate the AccessToken and SkipValidation parameter sets because there is no mandatory parameter in AccessToken that is not mandatory in SkipValidation (I end up getting parameter set cannot be resolved when trying to call the AccessToken parameter set).  That is why SubscriptionID can't be mandatory so I need this check.

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
maddieclayton: Subscription is just one parameter with the possibility of having either name or ID, so there is no way for the user to give us both.

maddieclayton: Second part is done.

maddieclayton:
on demand here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/powershell-demand/250/

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
```diff
@@ -117,86 +118,108 @@ public bool TryRemoveContext(IAzureContext context)
...
-                    environment, 
-                    tenantId, 
-                    subscriptionId, 
-                    subscriptionName, 
-                    out newSubscription, 
-                    out newTenant))
+                if (string.IsNullOrEmpty(subscriptionId) || string.IsNullOrEmpty(tenantId))
                 {
-                    account.SetOrAppendProperty(AzureAccount.Property.Tenants, new[] { newTenant.Id.ToString() });
+                    throw new PSInvalidOperationException("Must supply Subscription Id and Tenant Id when using -SkipValidation");
```
markcowl: Please put this into resources.

maddieclayton: done

maddieclayton:
on demand here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/powershell-demand/251/

maddieclayton:
@azuresdkci test this please

PR #5103, opened 2017-12-07
New MSI Generation

fearthecowboy:
This replaces the original MSI creation with a script that will 
- install the `Azure` and `AzureRM` modules and dependencies from the powershell gallery (override with `-repository` )
- tweaks the `PSGetModuleInfo.xml` file to remove the installed location (so it assumes the folder where it's contained)
- generates WiX include files 
- Compiles/Links WiX script to build x86 and x64 MSI files

Help for Script is included:
``` haskell
NAME
    generate.ps1

SYNOPSIS
    Generate the Azure Cmdlets MSI files.

SYNTAX
    C:\work\github\fresh\setup\generate.ps1 [[-version] <String>] [-force] [-noBuildNumber] [[-repository] <String>] [<CommonParameters>]


DESCRIPTION
    This script installs the azure cmdlets from the powershell gallery, tweaks them and generates MSI files for x86 and x64.
    When the MSI installs it will install the cmdlets to the same location the gallery would, and PowerShellGet treats them
    as if they were installed from the gallery

PARAMETERS
    -version <String>
        The version number for the generated MSI.

    -force [<SwitchParameter>]
        Forces a fresh installation of the Azure and AzureRm cmdlets from the gallery

    -noBuildNumber [<SwitchParameter>]
        Prevent a build number from being tacked on the end of the version number.

    -repository <String>
        Set the repository to pull packages from.
```


maddieclayton:
@azuresdkci Test this please

maddieclayton:
@fearthecowboy Can you resolve the merge conflicts?  It looks like they are too complex to do on GitHub.

fearthecowboy:
sure. It's all just a replacement  -- gimme a few minutes and I'll rebase it.

fearthecowboy:
FYI - running tests on this PR won't really do anything; the `generate.ps1` script isn't called from anywhere. i

maddieclayton:
Yup, testing the script itself now.

setup/PowerShellSetup.Test/MsiLayoutTests.cs
```diff
@@ -1,274 +0,0 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License. See License.txt in the project root for license information.
-
```
maddieclayton: You will need to remove all the code in build.proj that builds the old MSI (or uses anything that you deleted) in order for this not to break the build.  I would also think you'd want to add a line in build.proj that will build the new MSI.

setup/generate.ps1
```diff
@@ -0,0 +1,142 @@
...
+    Forces a fresh installation of the Azure and AzureRm cmdlets from the gallery
+.PARAMETER noBuildNumber
+    Prevent a build number from being tacked on the end of the version number.
+.PARAMETER repository
+    Set the repository to pull packages from.
+#>
+
+param(
+    [Parameter(HelpMessage="The version number for the generated MSI.")]
+    [string]$version = "6.0.0",
```
maddieclayton: I really don't think we should have a default version, we should have to manually enter it since 6.0.0 will only be right once.

setup/PowerShellSetup.Test/MsiLayoutTests.cs
fearthecowboy: I'm going to leave y'all to adjust the build process and signing. I create

maddieclayton: Okay sounds good. I’ll submit a change to your branch just to remove all of the old MSI stuff as it won’t build otherwise.

maddieclayton:
@fearthecowboy Discovered a bug: Install an older version of AzureRM using PSGallery, and update to a new version using the new MSI installer.  You will see that any module that was updated since the previous version of AzureRM will get a new side-by-side version of the module.  When you uninstall the program, not only will the newest versions of the updated modules will be deleted, but also the unchanged modules, leaving behind an empty folder in Program Files/WindowsPowerShell/Modules.  These modules should not be deleted.

Also let me know if you'd rather me just create a new issue for these kind of bugs or just tell you.

fearthecowboy:
O_o REAAALLLY.

I'm not seeing that. 

Steps:
```
> install-module azurerm -MaximumVersion 5.0.1 -Scope AllUsers

> get-installedmodule azurerm

Version    Name                                Repository           Description
-------    ----                                ----------           -----------
5.0.1      AzureRM                             PSGallery            Azure Resource Manager Module



# generate and install msi (has azurerm 5.1.1)

# what's installed
> get-installedmodule azurerm

Version    Name                                Repository           Description
-------    ----                                ----------           -----------
5.1.1      AzureRM                             PSGallery            Azure Resource Manager Module

# remove msi


> get-installedmodule azurerm

Version    Name                                Repository           Description
-------    ----                                ----------           -----------
5.0.1      AzureRM                             PSGallery            Azure Resource Manager Module

```

Now, I notice that I am seeing other modules that were installed alongsize `azurerm` got wrecked tho. 

Now, If they had a module installed manually before (ie, `AzureRM.TrafficManager`), and the MSI "installed" the same version, and then they remove the MSI, I'm not sure that I'd consider that the end of the world.

fearthecowboy:
I suspect that the behavior we're seeing is subject to the File Versioning Rules for Windows installer...

> Nonversioned Files (_ie, files that don't have an embedded version number?_) are User Data—If the Modified date is later than the Create date for the file on the computer, do not install the file because user customizations would be deleted. If the Modified and Create dates are the same, install the file. If the Create date is later than the Modified date, the file is considered unmodified, install the file.

So.. the files probably have Modified and Create dates  that are equal, which means the MSI is supposed to overwrite them, and when it's uninstalling, it removes files that it wrote.

maddieclayton:
@fearthecowboy You should have the two versions be the same (from PSGallery and MSI).  If you think it's the right behavior to delete the modules (which I'm not sure is true, but I do agree this a weird case) you should at least delete the folders in Program Files.

maddieclayton:
@azuresdkci test this please

maddieclayton:
@fearthecowboy Assigning this back to you to fix/address the one bug I found listed [here](https://github.com/Azure/azure-powershell/issues/5113).  All I have left to do is test on PowerShell 5.0/.Net <4.5.2.

setup/generate.ps1
```diff
@@ -0,0 +1,142 @@
...
+erase -ea 0 "$PSSCRIPTROOT/*.msi"
+
+# install modules locally.
+if ( -not (test-path $modulesDir))  {
+    Write-Host -fore green "Installing Modules..."
+    # create a folder for everything
+    $shh= mkdir -ea 0 $modulesDir 
+
+    # First install the modules locally
+    save-module azure -path $modulesDir -Repository $repository
```
maddieclayton: Let's delete this.

fearthecowboy: Delete which? 

the `mkdir`?

.gitignore
```diff
@@ -212,4 +212,8 @@ FakesAssemblies/
 *.GhostDoc.xml
 pingme.txt
 groupMapping*.json
-.vscode/
\ No newline at end of file
+
```
markcowl: please remove extra vertical space between entries

setup/azurecmd.wxs
```diff
@@ -1,119 +1,87 @@
...
-    <Condition Message="This setup requires the Windows PowerShell 3.0 or compatible version to be installed.">
-      <!-- This condition will break if PowerShell has version x3.0 and it is not compatible with 3.0 -->
-      <![CDATA[Installed or (PSCOMPATIBLEVERSION >< "3.0")]]>
-    </Condition>
-
-    <PropertyRef Id="NETFRAMEWORK45" />
-    <Condition Message="This setup requires the .NET Framework 4.5.2 or higher to be installed.">
-      <!-- NETFRAMEWORK45 contains a DWORD with installed version key to versions is here: -->
-      <!-- https://msdn.microsoft.com/en-us/library/ee942965(v=vs.110).aspx#detect_net -->
-      <![CDATA[Installed or (NETFRAMEWORK45 >= "#379893") or ISVSINSTALL]]>
```
markcowl: I believe that we still need to check .Net Framework version.  LMK if you verified that PowerShell 5 + .Net < 4.5.2 is possible

setup/generate.ps1
```diff
@@ -0,0 +1,142 @@
...
+    save-module azurerm -path $modulesDir -Repository $repository
+
+    Write-Host -fore green "Tweaking Modules"
+    cmd /c dir /a/s/b "$modulesDir\psgetmoduleinfo.xml" |% {
+        Write-Host -fore Gray " - Patching $_"
+        (gc $_ -raw ) -replace ".*<S N=.InstalledLocation.*S>",""  | Set-Content $_
```
markcowl: This is going to retain the information about which gallery the module came from, right?  Ideally we would set this to PSGallery, or allow the script writer to override, so that we could build official modules from build drops.

setup/generate.ps1
```diff
@@ -0,0 +1,142 @@
...
+    heat dir $modulesDir -out $includeFile -srd -sfrag -sreg -ag -g1 -cg "azurecmdfiles$arch" -dr "Modules$arch" -var var.modulesDir -indent 2 -nologo 
+    if( $LASTEXITCODE) {
+        write-host -fore red "Failed to generate include file."
+        break;
+    }
+
+    Write-Host -fore gray " - Fixing include file."
+    (gc $includeFile).replace('<Wix', '<Include') | Set-Content $includeFile
+    (gc $includeFile).replace('</Wix' ,'</Include') | Set-Content $includeFile
+    (gc $includeFile).replace('PSGetModuleInfo.xml" />' ,'PSGetModuleInfo.xml" Hidden="yes" />') | Set-Content $includeFile
```
markcowl: nice

setup/azurecmd.wxs
maddieclayton: Is it possible to set up PowerShell 5 to run with .Net < 4.5.2? Scott and I could not get PowerShell to run without 4.5.2 installed.

setup/generate.ps1
fearthecowboy: It does retain the upstream information, I only remove the `InstalledLocation` from the file (which WMF5.0 didn't support anyway, and the behavior is to assume the folder it's installed in.)

fearthecowboy: I did not look at how to alter the feed information -- I think it's a smidgen dangerous, IIRC, it records the 'name' of the feed, which would only be salient for the default feed AFAIK.

MiYanni: @fearthecowboy Can you sync with the Gallery team about this?

MiYanni:
@fearthecowboy See above comment about Gallery team.

setup/generate.ps1
fearthecowboy: @MiYanni  What specifically do you want me to 'sync with' them on.  The Gallery that's used is stored in   `RepositorySourceLocation` and `Repository` in the psgetmoduleinfo.xml. If the location and name don't match one the user has, it won't be able update. The only source you can bank on is the 'default' gallery.

fearthecowboy: Which one could overwrite, but given the fragility of the `psgetmoduleinfo.xml` file, I'd highly recommend not tweaking that as much as possible.

fearthecowboy: The only reason I can get away with it for InstalledLocation is that the original PSGet didn't store that, so they have a fall-back means to calculate the location based on where it was found

MiYanni:
@fearthecowboy After talking with @markcowl, it seems like we'd need a way to create the MSI using one PS Repo, but have those modules still marked for update from PSGallery. So, the modules are not marked as the same repo as the one they are generated from.

MiYanni:
@azuresdkci Retest this please

markcowl:
on demand run here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/powershell-demand/525/

signing run here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/ps-sign/211/

maddieclayton:
new sign job here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/ps-sign/
new on demand here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/powershell-demand/556/ (and 557)

maddieclayton:
sign job here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/ps-sign/238/

maddieclayton:
sign job with new msi here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/ps-sign/244/
posh pack job with new msi here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/posh-pack/48/console
on demand job here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/powershell-demand/586/ (and 587)

PR #5200, opened 2017-12-29
Update Cognitive Services Mgmt. PS with the latest Cognitive Services Mgmt. SDK

felixwa:
1. Apply Cognitive Services SDK 3.0.0.0
2. Update NuGet refs

## Description

---

### DIFF
--- a/src/ResourceManager/CognitiveServices/documentation/upcoming-breaking-changes.md
+++ b/src/ResourceManager/CognitiveServices/documentation/upcoming-breaking-changes.md
@@ -28,3 +28,5 @@
 # Upcoming Breaking Changes
 
 ## Release 1.0.x - May 2018
+
+* ProvisioningState property will be changed to a string in an upcoming release.


felixwa:
Hi reviewer, I guess that I need to bump the AssemblyVersion/AssemblyFileVersion in AssemblyInfo.cs but I'm not sure what the version it should be as I see your team controls that for every release. So let me know.

src/ResourceManager/CognitiveServices/Commands.Management.CognitiveServices/Models/PSCognitiveServicesAccount.cs
```diff
@@ -55,7 +55,7 @@ public PSCognitiveServicesAccount(CognitiveServicesModels.CognitiveServicesAccou
 
         public string Etag { get; private set; }
 
-        public ProvisioningState? ProvisioningState { get; private set; }
+        public string ProvisioningState { get; private set; }
```
felixwa: This is the only meaningful change introduced by new SDK (3.0.0.0) but I don't think it is breaking.

src/ResourceManager/CognitiveServices/Commands.Management.CognitiveServices/Commands.Management.CognitiveServices.csproj
```diff
@@ -48,9 +48,12 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Microsoft.Azure.Management.CognitiveServices, Version=2.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
-      <HintPath>..\..\..\packages\Microsoft.Azure.Management.CognitiveServices.2.0.0\lib\net452\Microsoft.Azure.Management.CognitiveServices.dll</HintPath>
+    <Reference Include="Microsoft.Azure.Management.CognitiveServices, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\..\..\packages\Microsoft.Azure.Management.CognitiveServices.3.0.0\lib\net452\Microsoft.Azure.Management.CognitiveServices.dll</HintPath>
     </Reference>
+    <Reference Include="Microsoft.Rest.ClientRuntime, Version=2.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL" />
```
markcowl: these should come from the common dependecies file

src/ResourceManager/CognitiveServices/Commands.Management.CognitiveServices/Models/PSCognitiveServicesAccount.cs
markcowl: @felixwa This is a breaking change, right.  As in, I could have had scripts with things like:

```PowerShell
$account.ProvisioningState == [ProvisioningState]::Running
```

and such a script will now be broken.  I think you need to adapt this type in PSCognitiveServicesAccount so that this is not broken.  It is OK to add an additional string field and deprecate this field, but you will need to do some translation between the string field from the wire and this enumeration until the next breaking change release (build)

tools/StaticAnalysis/Exceptions/BreakingChangeIssues.csv
```diff
@@ -781,3 +781,6 @@
...
 "Microsoft.Azure.Commands.Compute.dll","Microsoft.Azure.Commands.Compute.SetAzureVMCommand","Set-AzureRmVM","0","2080","The parameter 'Redeploy' in parameter set 'RedeployIdParameterSetName' for cmdlet 'Set-AzureRmVM' no longer has the attribute 'ValueFromPipelineByPropertyName'.","Add the attribute 'ValueFromPipelineByPropertyName' back to parameter 'Redeploy' in parameter set 'RedeployIdParameterSetName'."
+"d:\workspace\powershell\src\Package\Debug\ResourceManager\AzureResourceManager\AzureRM.CognitiveServices\Microsoft.Azure.Commands.Management.CognitiveServices.dll","Microsoft.Azure.Commands.Management.CognitiveServices.GetAzureCognitiveServicesAccountCommand","Get-AzureRmCognitiveServicesAccount","0","3000","The type of property 'ProvisioningState' of type 'Microsoft.Azure.Commands.Management.CognitiveServices.Models.PSCognitiveServicesAccount' has changed from 'System.Nullable`1[Microsoft.Azure.Management.CognitiveServices.Models.ProvisioningState]' to 'System.String'.","Change the type of property 'ProvisioningState' back to 'System.Nullable`1[Microsoft.Azure.Management.CognitiveServices.Models.ProvisioningState]'."
```
markcowl: not allowed

markcowl:
@felixwa Also, please update changelog.md with these changes

src/ResourceManager/CognitiveServices/Commands.Management.CognitiveServices/Models/PSCognitiveServicesAccount.cs
felixwa: @markcowl , the meaning of this PR is to introduce this change as we found we cannot enum all possible ProvisioningState so we changed the Swagger, SDK then this PS, change its type from an enum to a string.

I don't want to add back that enum in PS as it has gone in SDK also I cannot predicate possible values of that enum. If RP returns back a value not in the enum, PS throws an exception at client side.

Is there an option we wait "until the next breaking change release (build)"? If so we could wait but I really don't want to add that enum back in PS and only in PS.

tools/StaticAnalysis/Exceptions/BreakingChangeIssues.csv
markcowl: @felixwa breaking changes are not allowed until the breaking change release.  You will need to deprecate this field and adapt to the breaking change in the wire - likely by intrpoducing a new field to your output type, and providing the best available mapping in the existing enum property

PR #5264, opened 2018-01-11
Reserved instance auto register provider

corquiri:
## Description

---

This add new handler for Update-AzureRmReservation command such that if -appliedScopeType parameter is Single, the cmdlet registers the subscription with Microsoft.Compute. This step is necessary for Reservations api to work successfully.
Additionally, help document for Update-AzureRmReservation is updated.

### DIFF
--- a/src/ResourceManager/Reservations/ChangeLog.md
+++ b/src/ResourceManager/Reservations/ChangeLog.md
@@ -1,4 +1,5 @@
-﻿<!--
+
+<!--
     Please leave this section at the top of the change log.
 
     Changes for the current release should go under the section titled "Current Release", and should adhere to the following format:
@@ -34,3 +35,7 @@
     - cmdlet to retrieve list of reservation order ids that are applicable to subscription.
 * New Cmdlet Get-AzureRmReservationCatalog
     - cmdlet to retrieve available reservation catalog for the subscription.
+* Enable subscription Auto-Registration for the reservations provider
+
+## Version 0.1.0
+* Initial Release of Resrvations cmdlets


src/ResourceManager/Reservations/Commands.Reservations/Common/AzureReservationsAutoRegisterCmdletBase.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        protected override void BeginProcessing()
+        {
+            AzureSession.Instance.ClientFactory.RemoveHandler(typeof(AzureReservationsAutoRegisterDelegatingHandler));
+            IAzureContext context;
+            if (TryGetDefaultContext(out context)
+                && context.Account != null
+                && context.Subscription != null)
+            {
+                AzureSession.Instance.ClientFactory.AddHandler(new AzureReservationsAutoRegisterDelegatingHandler(
```
markcowl: This is unnecessary.  The existing auto-registration infrastructure auto-registers on any creation request that faile due to an unregistered subscription

markcowl:
@corquiri Auto-Registration is automatic for all  RPs in existing Azure PowerShell.  Why do you feel you need to ad this functionality?

src/ResourceManager/Reservations/Commands.Reservations/Common/AzureReservationsAutoRegisterCmdletBase.cs
corquiri: @markcowl the reason we are doing this is because we want to register the subscription with Microsoft.Compute - once this PATCH call comes to RI-RP (Reservations), we make a call to other service (CAS MCIO) which require the subscription to be registered with MS.Compute. So the initial call from Powershell => RI-RP can fail and the existing auto-registration infra might not be able to catch unregistered exception (because that exception is happening between RI-RP and CAS) so we need to manually do registration.

corquiri:
Comment is added explaining why we need extra auto-register logic.

src/ResourceManager/Reservations/Commands.Reservations/Common/AzureReservationsAutoRegisterCmdletBase.cs
markcowl: Do you have some evidence that this doesn't work in the existing infra?  I would prefer updating the existing auto-registration componenent rather than adding a new one.

markcowl:
@corquiri This requires some coordination with our team, closing for now, let's discuss next week and target this for the 2/9 release.

corquiri:
Is there any update on this PR? Please let me know if there was a discussion for this in Powershell team. Thanks.

corquiri:
Repopening this PR, this auto-register only applies to Reservations commandlets

src/ResourceManager/Reservations/Commands.Reservations/Common/AzureReservationsAutoRegisterCmdletBase.cs
markcowl: This is not limited to your cmdlets.  This is applying this to every client created from the client factory,  

IMO, this should not be done in a handler, but should simply be a call using the resource manager client before any cmdlets that need to be registered.

Howver, if you feel that you need this to be a DelegatingHandler for some reason, then you need to apply it to your client after creation has taken place.

src/ResourceManager/Reservations/Commands.Reservations/Common/AzureReservationsAutoRegisterDelegatingHandler.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.Commands.ResourceManager.Common.Properties;
+using Microsoft.Azure.Management.Internal.Resources;
+using Microsoft.Azure.Management.Internal.Resources.Models;
+using Microsoft.Rest.Azure;
+using Microsoft.WindowsAzure.Commands.Utilities.Common;
+
+namespace Microsoft.Azure.Commands.Reservations.Common
+{
+    class AzureReservationsAutoRegisterDelegatingHandler : DelegatingHandler, ICloneable
```
markcowl: why does this need to be in a delegating handler?

src/ResourceManager/Reservations/Commands.Reservations/Cmdlets/PatchReservation.cs
```diff
@@ -63,5 +73,68 @@ public override void ExecuteCmdlet()
...
+            {
+                IAzureContext context;
+                if (TryGetDefaultContext(out context)
+                    && context.Account != null
+                    && context.Subscription != null)
+                {
+                    var client = new ResourceManagementClient(
+                                context.Environment.GetEndpointAsUri(AzureEnvironment.Endpoint.ResourceManager),
+                                AzureSession.Instance.AuthenticationFactory.GetServiceClientCredentials(context, AzureEnvironment.Endpoint.ResourceManager));
+                    client.SubscriptionId = subscriptionId;
```
corquiri: Here I am setting client.SubscriptionId as one that is passed as argument to this cmdlet; It is not from default context. This is because we want to be able to pre-register for any subscription that is under the same tenant.. it does not have to be default subscription or billing subscription. To my understanding setting client.SubscriptionId here will cause client.Providers.Get to get the providers for that subscription. Please let me know if I am mistaken.

src/ResourceManager/Reservations/Commands.Reservations/help/Update-AzureRmReservation.md
```diff
@@ -76,6 +76,19 @@ Accept pipeline input: False
 Accept wildcard characters: False
 ```
 
+### -DefaultProfile
+The credentials, account, tenant, and subscription used for communication with Azure.```yaml
```
cormacpayne: @corquiri the build is failing because of the trailing ` ```yaml ` string, please move it to a new line (see other parameters as an example)

corquiri: Moved

src/ResourceManager/Reservations/Commands.Reservations/Cmdlets/PatchReservation.cs
```diff
@@ -6,6 +6,12 @@
 using System.Management.Automation;
 using Newtonsoft.Json;
 using Microsoft.Azure.Management.Reservations;
+using Microsoft.Azure.Commands.Common.Authentication.Abstractions;
```
markcowl: @corquiri please add a changelog entry in your changelog for this change.  Also, there should be an automated test for this scenario.  Please add a  test or file a test debt issue to add it next sprint.

markcowl:
on demand run here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/powershell-demand/428/

Issue #5053, opened 2017-11-30
Get-AzureStorageContainer returns HTTP 403 when using SAS-Token instead of StorageKey


karlti:
### PowerShell Version
5.1.15063.726

### Module Version
Azure.Storage, current version

### OS Version
10.0.15063.726

### Description
When calling `"Get-AzureStorageContainer"` with a `StorageContext` that has been created with a SAS Token, the call always fails with an Error: 

***"The remote server returned an error: (403) Forbidden. HTTP Status Code: 403 - HTTP Error Message: Server  failed to authenticate the request."***

when drilling into, it looks like the script queries the ACLs of the container and this cannot succeed with a SAS.

When I directly call into the "List Container" REST API, with a SAS-Token like this:

` Invoke-RestMethod -Method Get -Uri "https://<myAcc>.blob.core.windows.net/?sv=2017-04-17&ss=bfqt&srt=sco&sp=rwdlacup&se=2017-11-30T20:03:25Z&st=2017-11-127T12:03:25Z&spr=https&sig=tf1Fg0jGXzp7xSv4b%2BbzUsXXXXXXXXX%3D&comp=list"`

it succeeds.

Please either change the script to work with SAS Tokens or adjust the CmdLet Documentation.
Thank you.

with kind regards,
Karl Tietze

### Debug Output
*PS C:\windows\system32> Get-AzureStorageContainer -Context $context -Name test-cont-001
DEBUG: 15:03:38 - Init Operation Context for 'GetAzureStorageContainerCommand' with client request id  Azure-Storage-PowerShell-2eb24846-2cad-4a66-b90d-e4d100138d06. If you want to g
et more details, please add "-Debug" to your command.
DEBUG: 15:03:38 - GetAzureStorageContainerCommand begin processing with ParameterSet 'ContainerName'.
DEBUG: 15:03:38 - Use storage account '[SasToken]' from storage context.
DEBUG: 15:03:38 - Start 1th remote call, method: HEAD, destination: https://<myAccount>.blob.core.windows.net/test-cont-001?sv=2017-04-17&ss=bfqt&srt=sco&sp=rwdlacup&se=2017-12-10T23:01
:20Z&st=2017-11-30T15:01:20Z&spr=https&sig=KfdpXXXXXXXXXX2Bo0%3D&api-version=2017-04-17&restype=container.
[...]

PR #5420, opened 2018-02-01
Fix Get-AzureStorageContainer with Account SAS failure

blueww:
## Description

Get-AzureStorageContainer not work when use storage context created from SAS credential. https://github.com/Azure/azure-powershell/issues/5053

This is because we will get container ACL after list/get container finish, but get container ACL only work with Account Key credential. We have catch 404 before for permission issue, but currently server report 403 for permission issue. So the fix is also catch 403 when get container ACL failure. So we can just return the container object to customer when get container ACL failed with permission issue. We will also write a warning message when customer get container with SAS credential to remind them get container ACL will fail.

---

### DIFF
--- a/src/Storage/Commands.Storage/Blob/Cmdlet/GetAzureStorageContainer.cs
+++ b/src/Storage/Commands.Storage/Blob/Cmdlet/GetAzureStorageContainer.cs
@@ -206,11 +206,15 @@ namespace Microsoft.WindowsAzure.Commands.Storage.Blob.Cmdlet
                 return;
             }
 
-            // Write warning when user SAS credential since get container ACL will fail
-            AzureStorageContext storageContext = this.GetCmdletStorageContext();
-            if (storageContext != null && storageContext.StorageAccount != null && storageContext.StorageAccount.Credentials != null && storageContext.StorageAccount.Credentials.IsSAS)
+            // Only write warning for SAS when use cmdlet alias Get-AzureStorageContainerAcl, since the cmdlets alias specified get container ACL
+            if (this.MyInvocation.Line.ToLower().Contains("get-azurestoragecontaineracl"))
             {
-                WriteWarning("Get Container ACL will fail with SAS token credentials.");
+                // Write warning when user SAS credential since get container ACL will fail
+                AzureStorageContext storageContext = this.GetCmdletStorageContext();
+                if (storageContext != null && storageContext.StorageAccount != null && storageContext.StorageAccount.Credentials != null && storageContext.StorageAccount.Credentials.IsSAS)
+                {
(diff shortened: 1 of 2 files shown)


blueww:
On demand: https://azuresdkci.westus2.cloudapp.azure.com/job/powershell-demand/408/

src/Storage/Commands.Storage/Blob/Cmdlet/GetAzureStorageContainer.cs
```diff
@@ -233,11 +240,11 @@ internal void PackCloudBlobContainerWithAcl(IEnumerable<Tuple<CloudBlobContainer
             }
             catch (StorageException e)
             {
-                if (!e.IsNotFoundException())
+                if (!e.IsNotFoundException() && !e.IsFordiddenException())
```
cormacpayne: @blueww we should change how exceptions are being thrown here so that the properties of the exception are preserved. This means only catching the exceptions that throw a 403 or 404. We should also put the warning you're adding above in here rather than having it before the check. You can change the message slightly so it applies to both 403 and 404 scenarios. It should look something like the following:

```cs
catch (StorageException e) when (e.IsNotFoundException() || e.IsForbiddenException())
{
    WriteWarning("Denied permission to query the permission of the specified container.");
}
```

blueww: @cormacpayne 

If change as you required, will get following error, this is because query container ACL is executed in aync task. So we can't write warning in it. Any idea？
BTW，I have modify the catch exception as you required.

>>Get-AzureStorageContainer : The WriteObject and WriteError methods cannot be called from outside the overrides of the BeginProcessing, ProcessRecord, and EndProcessing methods, and they can only be called from within the same thread. Validate 
that the cmdlet makes these calls correctly, or contact Microsoft Customer Support Services.
At line:1 char:1
>>+ Get-AzureStorageContainer -Context $ctxsas
>>+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>>    + CategoryInfo          : CloseError: (:) [Get-AzureStorageContainer], PSInvalidOperationException
>>    + FullyQualifiedErrorId : PSInvalidOperationException,Microsoft.WindowsAzure.Commands.Storage.Blob.Cmdlet.GetAzureStorageContainerCommand

blueww:
@markcowl ， @cormacpayne 
I have discussed with Mark in a meeting this morning, and we agree to write warning only when run the cmdlet with alias "Get-AzureStorageContainerACL", since it specify to get container ACL. And we also agreed on write the warning before run get ACL task.

I have changed the code according. Please check.

cormacpayne:
On-demand: https://azuresdkci.westus2.cloudapp.azure.com/job/powershell-demand/439/

Issue #5192, opened 2017-12-27
Get-AzureRmRoleAssignment throws unhandled exception on Service Provider account


MiYanni:
### Cmdlet(s)
Get-AzureRmRoleAssignment
### PowerShell Version
PSVersion                      5.1.16299.98
### Module Version
5.1.1      AzureRM.Resources
### OS Version
BuildVersion                   10.0.16299.98
### Description
While logged into a Service Provider account, Get-AzureRmRoleAssignment (no params) throws an unhandled Microsoft.Rest.Azure.CloudException.
### Debug Output
DEBUG: ============================ HTTP RESPONSE ============================

Status Code:
Forbidden

Headers:
Pragma                        : no-cache
ocp-aad-diagnostics-server-name: OzArKxpPSemSWWoFLWW7BNCbp+jQjreuu63QDSj69JM=
request-id                    : 2942ed39-8c4c-4aff-8a0e-a0da9c3dbfdc
client-request-id             : b3453f77-27ea-44a0-b8a7-c300b13dc04d
x-ms-dirapi-data-contract-version: 1.6
ocp-aad-session-key           : JBc28mv92xPoNCqpvat40jTYbFLzcucNdsy6NrboPoadL9M-jJKeGqIO1b-SWNlG_FKvEVmDy5F1ZQaS90v7RU8105cmbWNyK3s1p9zkenmD46Boy39OVgEZYCK8avAG.qzQSHzbrH8FZFF7Ce2De25BXwObq2YYlS9GhMc8GyEs
X-Content-Type-Options        : nosniff
DataServiceVersion            : 3.0;
Strict-Transport-Security     : max-age=31536000; includeSubDomains
Access-Control-Allow-Origin   : *
Duration                      : 2474974
Cache-Control                 : no-cache
Server                        : Microsoft-IIS/8.5
X-AspNet-Version              : 4.0.30319
X-Powered-By                  : ASP.NET,ASP.NET
Date                          : Wed, 27 Dec 2017 18:24:27 GMT

Body:
{
  "odata.error": {
    "code": "Authorization_RequestDenied",
    "message": {
      "lang": "en",
      "value": "Insufficient privileges to complete the operation."
    }
  }
}

Get-AzureRmRoleAssignment : Exception of type 'Microsoft.Rest.Azure.CloudException' was thrown.
At line:1 char:1
+ Get-AzureRmRoleAssignment
+ ~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #5589, opened 2018-02-20
Fix for the issue https://github.com/Azure/azure-powershell/issues/5192 

darshanhs90:
Fix for issue https://github.com/Azure/azure-powershell/issues/5192

## Checklist


src/ResourceManager/Resources/Commands.Resources/Models.Authorization/AuthorizationClientExtensions.cs
```diff
@@ -117,7 +118,19 @@ private static IEnumerable<PSRoleAssignment> ToPSRoleAssignments(this IEnumerabl
 
             List<string> objectIds = new List<string>();
             objectIds.AddRange(assignments.Select(r => r.PrincipalId.ToString()));
-            List<PSADObject> adObjects = activeDirectoryClient.GetObjectsByObjectId(objectIds);
+            List<PSADObject> adObjects = null;
+            try
+            {
+                adObjects = activeDirectoryClient.GetObjectsByObjectId(objectIds);
+            }
+            catch (CloudException ce) {
```
cormacpayne: @darshanhs90 I think we should add a separate function that will check for the below condition:

```cs
public bool IsFooError(CloduException ce)
{
    if (ce.Response != null && ce.Response.StatusCode == HttpStatusCode.Unauthorized &&
        ce.Error != null && ce.Error.Code != null && string.Equals(ce.Error.Code, "Authorization_RequestDenied", StringComparison.OrdinalIgnoreCase))
    {
        return true;
    }
    
    return false
}
```

This allows us to handle this special case and return a custom exception, while allowing the other exceptions to be thrown as they usually would and preserve their error message and content:

```cs
catch (CloudException ex) when (IsFooError(ex))
{
    throw new InvalidOperationException(ProjectResources.InSufficientGraphPermission);
}
```

src/ResourceManager/Resources/Commands.Resources/Models.Authorization/AuthorizationClientExtensions.cs
```diff
@@ -117,7 +118,19 @@ private static IEnumerable<PSRoleAssignment> ToPSRoleAssignments(this IEnumerabl
...
+                if (ce.Response.StatusCode == HttpStatusCode.Unauthorized && ce.Error.Code.Equals("Authorization_RequestDenied",StringComparison.OrdinalIgnoreCase))
```
cormacpayne: @darshanhs90 nit: should we make the "Authorization_RequestDenied` string a constant?

darshanhs90: Done

src/ResourceManager/Resources/Commands.Resources/Models.Authorization/AuthorizationClientExtensions.cs
darshanhs90: Done

cormacpayne: @darshanhs90 this is almost correct -- we want to make sure that we're only catching this special type of exception, which is why we'd use the `when` check:

```cs
catch (CloudException ex) when (IsAuthorizationDeniedException(ex))
{
    throw new InvalidOperationException(ProjectResources.InsufficientGraphPermission);
}
```

cormacpayne:
On-demand: https://azuresdkci.westus2.cloudapp.azure.com/job/powershell-demand/504/

Issue #5564, opened 2018-02-16
Slow Resources tests in playback mode


MiYanni:
These are slow in playback:
- `Microsoft.Azure.Commands.Resources.Test.ScenarioTests.ManagedApplicationTests.TestManagedApplicationCRUD`
- `Microsoft.Azure.Commands.Resources.Test.ScenarioTests.ResourceTests.TestMoveAResourceTest`

![image](https://user-images.githubusercontent.com/17788297/36320841-7ebb0e50-12fc-11e8-9d18-e32661ad0200.png)

## Goals
- [ ] Figure out why they are slow
- [ ] Make them not slow

cormacpayne:
Looks like the issue is in the following while-loop:

https://github.com/Azure/azure-powershell/blob/2b6c7a526fe54caa4f021d493523b6419b925603/src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Components/LongRunningOperationHelper.cs#L87-L100

Specifically, the `Task.Delay` call here:

https://github.com/Azure/azure-powershell/blob/2b6c7a526fe54caa4f021d493523b6419b925603/src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Components/LongRunningOperationHelper.cs#L98

PR #5627, opened 2018-02-24
Fix resource tests duration

MiYanni:
## Description

This fixes https://github.com/Azure/azure-powershell/issues/5564

TestManagedApplicationCRUD and TestMoveAResourceTest took 5 mins and 2 mins in playback respectively. This fixes those by injecting Retry-After 0 into the response headers during record matching (playback). Any responses recorded with a Retry-After value are removed and replaced with Retry-After 0 during record matching.

This header is always injected in the response body because any response has the potential to be a 'waiting' response. So, we inject this when reading the records so every response will always only wait 0 seconds if that response is required to wait. If they are 'success' responses, they don't wait anyway, so this value causes no side effects.

## Checklist


maddieclayton:
on demand run here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/powershell-demand/511/

Issue #5496, opened 2018-02-12
Fix Output, Parameters, and Piping for KeyVault cmdlets


markcowl:
### Description
Work with KeyVault team to update KeyVault cmdlets to make the following changes.  In this release, we want to identify proposed cmdlet changes, present these to the KeyVautl team, and make the changes ourselves, or in coordination with them.

We want to normalize the cmdlets in the following ways.
- [ ] Correctly specify output type
- [ ] Enumerate collections
- [ ] Take piepline input as appropriate
  - [ ] Get-AzureRmKeyVault | Get-AzureKeyVaultSecret
  - [ ] Get-AzureKeyVaultSecret | Get-AzureKeyVaultSecret
- [ ] Accept pipeline input using object parameter set
- [ ] Accept pipeline input using resourceid parameter set (control plane only)
- [ ] Accept multiple instances as well as single instances for parameters that specify a resource.

Additionally we need to:
- [ ] Ensure all cmdlets are covered by tests
- [ ] Ensure all tests are passing with the new changes and skipped tests are fixed and rerecorded
- [ ] All documentation is updated and all examples/help messages are updated or fixed

maddieclayton:
Proposed changes here: https://github.com/maddieclayton/azure-powershell/wiki/KeyVault-Improvements

maddieclayton:
## Breaking Changes

### General
- **Make nullable parameters not nullable**
- **Figure out which cmdlets are asynchronous, add asjob, and update in the breaking change release**.
- **Remove ConfirmImpact (possibly add shouldcontinue in its place)**
- **Remove ValueFromPipelineByPropertyName where necessary to enable piping**

### Add-AzureKeyVaultCertificate
- **Change from asynchronous operation to synchronous, and change output to KeyVaultCertificate?**
- **Make policy mandatory as cmdlet fails without it.**

### Add-AzureKeyVaultCertificateContact
- **Change the output to just the contact rather than all contacts?**

### Add-AzureKeyVaultKey
- **Remove the Tags alias or ValueFromPipeline tag**
- **Add piping for Get-AzureRmResource | Add-AzureKeyVaultKey**

### Get-AzureKeyVaultCertificate
- **Remove ValueFromPipelineByPropertyName on Name to enable ResourceId parameter set**
- **Add piping for Get-AzureRmResource | Get-AzureKeyVaultCertificate**

### Get-AzureKeyVaultCertificateIssuer
[...]

Issue #2919, opened 2016-09-10
[2.1.0] Add Static Analysis rule for Verb-SingularNoun pattern in cmdlet names


alexandair:
I've mentioned this a couple of times to PG. Unfortunately, things are not getting better. We still have a bunch of cmdlets that don't follow a naming convention and use plurals:

``` powershell
   Source: AzureRM.ApiManagement

CommandType Name                                    Version Source               
----------- ----                                    ------- ------               
Cmdlet      Set-AzureRmApiManagementVirtualNetworks 2.1.0   AzureRM.ApiManagement
Cmdlet      Set-AzureRmApiManagementHostnames       2.1.0   AzureRM.ApiManagement


   Source: AzureRM.Backup

CommandType Name                              Version Source        
----------- ----                              ------- ------        
Cmdlet      Get-AzureRmBackupVaultCredentials 2.1.0   AzureRM.Backup
Cmdlet      Get-AzureRmBackupJobDetails       2.1.0   AzureRM.Backup


   Source: AzureRM.Batch

CommandType Name                               Version Source       
----------- ----                               ------- ------       
Cmdlet      Get-AzureRmBatchAccountKeys        2.1.0   AzureRM.Batch
Cmdlet      Get-AzureRmBatchSubscriptionQuotas 2.1.0   AzureRM.Batch
Cmdlet      Get-AzureBatchRemoteLoginSettings  2.1.0   AzureRM.Batch
Cmdlet      Get-AzureBatchPoolUsageMetrics     2.1.0   AzureRM.Batch


   Source: AzureRM.CognitiveServices

CommandType Name                                    Version Source                   
----------- ----                                    ------- ------                   
Cmdlet      Get-AzureRmCognitiveServicesAccountSkus 0.3.0   AzureRM.CognitiveServices


   Source: AzureRM.HDInsight

CommandType Name                                      Version Source           
----------- ----                                      ------- ------           
Cmdlet      Add-AzureRmHDInsightConfigValues          2.1.0   AzureRM.HDInsight
Cmdlet      Get-AzureRmHDInsightProperties            2.1.0   AzureRM.HDInsight



   Source: AzureRM.KeyVault

CommandType Name                                             Version Source          
----------- ----                                             ------- ------          
Cmdlet      New-AzureKeyVaultCertificateOrganizationDetails  2.1.0   AzureRM.KeyVault
[...]

PR #5654, opened 2018-03-02
KeyVault improvements

maddieclayton:
## Description
https://github.com/Azure/azure-powershell/issues/5496

## Checklist

### DIFF
--- a/src/ResourceManager/KeyVault/Commands.KeyVault/Commands/SetAzureKeyVaultCertificatePolicy.cs
+++ b/src/ResourceManager/KeyVault/Commands.KeyVault/Commands/SetAzureKeyVaultCertificatePolicy.cs
@@ -70,6 +70,7 @@ namespace Microsoft.Azure.Commands.KeyVault
         [Parameter(Mandatory = true,
             Position = 2,
             ValueFromPipeline = true,
+            ValueFromPipelineByPropertyName = true,
             ParameterSetName = ByValueParameterSet,
             HelpMessage = "Specifies the certificate policy.")]
         [ValidateNotNull]


maddieclayton:
on demand run here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/powershell-demand/544/

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/GetAzureKeyVaultCertificate.cs
```diff
@@ -88,56 +138,71 @@ public class GetAzureKeyVaultCertificate : KeyVaultCmdletBase
...
+                }
+                else
+                {
+                    PSDeletedKeyVaultCertificate deletedCert = PSDeletedKeyVaultCertificate.FromDeletedCertificateBundle(DataServiceClient.GetDeletedCertificate(VaultName, Name));
+                    WriteObject(deletedCert);
+                }
+            }
+            else
+            {
+                if (string.IsNullOrEmpty(Name))
```
dragav: It's the structure of the 'if' block that looks odd - would it be possible to refactor this with less nesting?

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/GetAzureKeyVaultCertificateOperation.cs
```diff
@@ -41,19 +50,39 @@ public class GetAzureKeyVaultCertificateOperation : KeyVaultCmdletBase
...
+            {
+                VaultName = InputObject.VaultName.ToString();
+                Name = InputObject.Name.ToString();
+            }
+
             var certificateOperation = this.DataServiceClient.GetCertificateOperation(VaultName, Name);
-            var kvCertificateOperation = KeyVaultCertificateOperation.FromCertificateOperation(certificateOperation);
+            var kvCertificateOperation = PSKeyVaultCertificateOperation.FromCertificateOperation(certificateOperation);
+            kvCertificateOperation.VaultName = VaultName;
+            kvCertificateOperation.Name = Name;
```
dragav: Are these assignments necessary here? Could they be moved into the FromCertOp method? (haven't checked if the vault name is available there)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/RemoveAzureKeyVaultAccessPolicy.cs
```diff
@@ -99,6 +191,10 @@ public class RemoveAzureKeyVaultAccessPolicy : KeyVaultManagementCmdletBase
             ParameterSetName = ByEmail,
             ValueFromPipelineByPropertyName = true,
             HelpMessage = "Specifies the email address of the user in Azure Active Directory for which to grant permissions.")]
+        [Parameter(Mandatory = true,
+            ParameterSetName = InputObjectByEmail,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Specifies the email address of the user in Azure Active Directory for which to grant permissions.")]
```
dragav: 'whose permissions should be deleted' (sorry, the help message was already incorrect, but since you're editing this :-))

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/RemoveAzureKeyVaultAccessPolicy.cs
```diff
@@ -109,25 +205,44 @@ public class RemoveAzureKeyVaultAccessPolicy : KeyVaultManagementCmdletBase
             ParameterSetName = ByObjectId,
             ValueFromPipelineByPropertyName = true,
             HelpMessage = "Specifies the ID of application that a user must.")]
+        [Parameter(Mandatory = false,
+            ParameterSetName = InputObjectByObjectId,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Specifies the ID of application that a user must.")]
```
dragav: 'whose permissions should be removed'. (Again, not your bug, but while you're in there..)

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/RemoveAzureKeyVaultAccessPolicy.cs
```diff
@@ -109,25 +205,44 @@ public class RemoveAzureKeyVaultAccessPolicy : KeyVaultManagementCmdletBase
...
             ParameterSetName = ForVault,
             ValueFromPipelineByPropertyName = true,
             HelpMessage = "If specified, disables the retrieval of secrets from this key vault by the Microsoft.Compute resource provider when referenced in resource creation.")]
-        public SwitchParameter EnabledForDeployment { get; set; }
+        [Parameter(Mandatory = false,
+            ParameterSetName = InputObjectForVault,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "If specified, disables the retrieval of secrets from this key vault by the Microsoft.Compute resource provider when referenced in resource creation.")]
+        [Alias("EnabledForDeployment")]
+        public SwitchParameter DisableDeployment { get; set; }
```
dragav: I have some concerns regarding this renaming (as well as the subsequent one, re: template deployment). These are actually vault properties, and the name was chosen specifically to match their corresponding vault properties. The renaming not only diverges from this implicit tie-in, but also reverses its semantics: the user must now negate this value to understand their effect.
Please consider reverting this specific change.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/RemoveAzureKeyVaultCertificate.cs
```diff
@@ -46,65 +55,83 @@ public class RemoveAzureKeyVaultCertificate : KeyVaultCmdletBase
...
+                   HelpMessage = "Certificate Object.")]
+        [ValidateNotNullOrEmpty]
+        public PSKeyVaultCertificateIdentityItem InputObject { get; set; }
 
         /// <summary>
         /// If present, operate on the deleted key entity.
         /// </summary>
-        [Parameter( Mandatory = false,
-                    HelpMessage = "Permanently remove the previously deleted certificate." )]
+        [Parameter(Mandatory = false,
```
dragav: (nitpick) I've seen that elsewhere in this changeset the InRemovedState parameter was marked as mandatory for the 'deleted' parameter set. Here it seems to be missing. Would you consider adding the 'deleted' parameter set to this cmdlet as well, and maintain consistency with other cmdlets?

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/RemoveAzureKeyVaultCertificateContact.cs
```diff
@@ -56,63 +48,67 @@ public class RemoveAzureKeyVaultCertificateContact : KeyVaultCmdletBase
...
-        [Parameter(HelpMessage = "This cmdlet does not return an object by default. If this switch is specified, it returns the contact object.")]
+        [Parameter(Mandatory = false,
+                   HelpMessage = "This cmdlet does not return an object by default. If this switch is specified, it returns the contact object.")]
         public SwitchParameter PassThru { get; set; }
 
-        protected override void ProcessRecord()
+        public override void ExecuteCmdlet()
         {
-            if (ShouldProcess(EmailAddress, Properties.Resources.RemoveCertificateContact))
+            foreach (string email in EmailAddress)
```
dragav: Same observation as in the matching AddCertContact cmdlet - consider storing the input email addresses in a hashset, making a single call to the service to retrieve the contacts, and then updating the contacts in a single pass.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/RestoreAzureKeyVaultKey.cs
```diff
@@ -24,42 +24,70 @@ namespace Microsoft.Azure.Commands.KeyVault
...
+
         /// <summary>
         /// The input file in which the backup blob is stored
         /// </summary>
         [Parameter(Mandatory = true,
                    Position = 1,
                    HelpMessage = "Input file. The input file containing the backed-up blob")]
         [ValidateNotNullOrEmpty]
-        public string InputFile { get; set; }
+        public string[] InputFile { get; set; }
```
dragav: Why an array? Is there a scenario that requires the batching of multiple Restore cmdlets in a single one? I get the convenience factor, but I'm concerned that error scenarios will lead to confusion: what was restored, which one failed and why, and do we continue on error or stop at the first one?

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/SetAzureKeyVaultAccessPolicy.cs
```diff
@@ -235,6 +367,22 @@ public class SetAzureKeyVaultAccessPolicy : KeyVaultManagementCmdletBase
...
+            HelpMessage = "Specifies certificate operation permissions to grant to a user or service principal.")]
+        [Parameter(Mandatory = false,
+            ParameterSetName = InputObjectByUserPrincipalName,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Specifies certificate operation permissions to grant to a user or service principal.")]
+        [Parameter(Mandatory = false,
+            ParameterSetName = InputObjectByEmailAddress,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Specifies certificate operation permissions to grant to a user or service principal.")]
         [ValidateSet("get", "list", "delete", "create", "import", "update", "managecontacts", "getissuers", "listissuers", "setissuers", "deleteissuers", "manageissuers", "recover", "purge", "all")]
```
dragav: As an aside, we wanted to deprecate 'all' as a permission for some time. Since this is a deprecating change, we should just remove it with this commit. If you'd rather not, that's perfectly fine - we'll follow up. Thanks.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/SetAzureKeyVaultCertificatePolicy.cs
```diff
@@ -64,11 +64,46 @@ public class SetAzureKeyVaultCertificatePolicy : KeyVaultCmdletBase
...
+        [Alias("CertificatePolicy")]
+        public PSKeyVaultCertificatePolicy InputObject { get; set; }
+
+        /// <summary>
+        /// RenewAtNumberOfDaysBeforeExpiry
+        /// </summary>
+        [Parameter(Mandatory = true,
+            ValueFromPipelineByPropertyName = true,
+            ParameterSetName = ExpandedRenewNumberParameterSet,
+            HelpMessage = "Specifies the number of days after which the automatic process for the certificate renewal begins.")]
```
dragav: "the number of days before expiration when automatic renewal should start".

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/AddAzureKeyVaultCertificateContact.cs
```diff
@@ -27,90 +28,128 @@ namespace Microsoft.Azure.Commands.KeyVault
...
         /// EmailAddress
         /// </summary>
         [Parameter(Mandatory = true,
-                   ParameterSetName = AddParameterSet,
                    Position = 1,
                    ValueFromPipelineByPropertyName = true,
                    HelpMessage = "Specifies the email address of the contact.")]
         [ValidateNotNullOrEmpty]
-        public string EmailAddress { get; set; }
+        public string[] EmailAddress { get; set; }
```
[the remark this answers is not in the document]
markcowl: @dragav For PowerShell, this is not a breaking change - this parameter will bind to a string or a list of strings.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/SetAzureKeyVaultAccessPolicy.cs
markcowl: This is not a breaking change, but we should deprecate anything we would like to remove at the next breaking change release

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/GetAzureKeyVaultCertificate.cs
```diff
@@ -28,15 +26,18 @@ namespace Microsoft.Azure.Commands.KeyVault
...
-        private const string ByVaultNameParameterSet = "ByVaultName";
-        private const string ByCertificateVersionsParameterSet = "ByCertificateVersions";
-        private const string ByDeletedCertificateParameterSet = "ByDeletedCertificates";
+        private const string ByVaultNameParameterSet = "ByName";
+        private const string ByCertificateNameandVersionParameterSet = "ByCertificateNameAndVersion";
+        private const string ByCertificateVersionsParameterSet = "ByCertificateAllVersions";
+
+        private const string InputObjectByVaultNameParameterSet = "ByNameInputObject";
+        private const string InputObjectByCertificateNameandVersionParameterSet = "ByCertificateNameAndVersionInputObject";
+        private const string InputObjectByCertificateVersionsParameterSet = "ByCertificateAllVersionsInputObject";
```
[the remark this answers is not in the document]
dragav: I wasn't suggesting any new scenario; it seemed odd to me to have an "cert all versions input object" parameter set - it implied that "cert all versions" can be _the_ input. Thanks for clarifying.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/GetAzureKeyVaultCertificate.cs
```diff
@@ -46,39 +47,88 @@ public class GetAzureKeyVaultCertificate : KeyVaultCmdletBase
...
+        [Parameter(Mandatory = true,
+                   ParameterSetName = InputObjectByVaultNameParameterSet,
+                   Position = 0,
+                   ValueFromPipeline = true,
+                   HelpMessage = "KeyVault object.")]
+        [Parameter(Mandatory = true,
+                   ParameterSetName = InputObjectByCertificateNameandVersionParameterSet,
+                   Position = 0,
+                   ValueFromPipeline = true,
+                   HelpMessage = "KeyVault object.")]
```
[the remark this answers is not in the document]
dragav: Yes, thank you for the clarification.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/SetAzureKeyVaultAccessPolicy.cs
dragav: All occurrences of "All" as a permission should be removed. Thanks.

maddieclayton: Added to list of breaking changes - will be included in my next PR.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ImportAzureKeyVaultCertificate.cs
```diff
@@ -105,8 +101,11 @@ public class ImportAzureKeyVaultCertificate : KeyVaultCmdletBase
         /// Certificate Collection
         /// </summary>
         [Parameter(Mandatory = true,
+                   Position = 2,
+                   ValueFromPipeline = true,
                    ParameterSetName = ImportWithPrivateKeyFromCollectionParameterSet,
                    HelpMessage = "Specifies the certificate collection to add to key vault.")]
+        [Alias("InputObject")]
```
cormacpayne: @maddieclayton any reason we don't want to call this parameter `InputObject` with an alias to `CertificateCollection` like we do for all other cmdlets?

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/SetAzureKeyVaultCertificateIssuer.cs
```diff
@@ -107,10 +107,11 @@ public class SetAzureKeyVaultCertificateIssuer : KeyVaultCmdletBase
         /// </summary>
         [Parameter(Mandatory = true,
                    ParameterSetName = ByValueParameterSet,
+                   ValueFromPipeline = true,
                    ValueFromPipelineByPropertyName = true,
                    HelpMessage = "Specifies the certificate issuer to set.")]
         [ValidateNotNullOrEmpty]
-        public KeyVaultCertificateIssuer Issuer { get; set; }
+        public PSKeyVaultCertificateIssuerIdentityItem Issuer { get; set; }
```
cormacpayne: @maddieclayton do we want to name this `InputObject` with an alias to `Issuer`?

src/ResourceManager/KeyVault/Commands.KeyVault/help/Set-AzureKeyVaultCertificateAttribute.md
```diff
@@ -1,4 +1,4 @@
-﻿---
```
cormacpayne: @maddieclayton it looks like the underlying call for this cmdlet is a [`PATCH`](https://github.com/Azure/azure-sdk-for-net/blob/psSdkJson6/src/SDKs/KeyVault/dataPlane/Microsoft.Azure.KeyVault/Generated/KeyVaultClient.cs#L9769), so we should look at renaming this cmdlet to `Update-AzureKeyVaultCertificateAttribute` with an alias to `Set-*`

src/ResourceManager/KeyVault/Commands.KeyVault/help/Set-AzureKeyVaultCertificatePolicy.md
```diff
@@ -1,4 +1,4 @@
-﻿---
```
cormacpayne: @maddieclayton same comment about this call being a [`PATCH`](https://github.com/Azure/azure-sdk-for-net/blob/psSdkJson6/src/SDKs/KeyVault/dataPlane/Microsoft.Azure.KeyVault/Generated/KeyVaultClient.cs#L9542), so we should change the verb to `Update`

src/ResourceManager/KeyVault/Commands.KeyVault/help/Set-AzureKeyVaultKeyAttribute.md
```diff
@@ -1,4 +1,4 @@
-﻿---
```
cormacpayne: @maddieclayton same comment about this call being a [`PATCH`](https://github.com/Azure/azure-sdk-for-net/blob/psSdkJson6/src/SDKs/KeyVault/dataPlane/Microsoft.Azure.KeyVault/Generated/KeyVaultClient.cs#L963), so we should change the verb to `Update`

src/ResourceManager/KeyVault/Commands.KeyVault/help/Set-AzureKeyVaultSecretAttribute.md
```diff
@@ -1,4 +1,4 @@
-﻿---
```
cormacpayne: @maddieclayton same comment about this call being a [`PATCH`](https://github.com/Azure/azure-sdk-for-net/blob/psSdkJson6/src/SDKs/KeyVault/dataPlane/Microsoft.Azure.KeyVault/Generated/KeyVaultClient.cs#L4792), so we should change the verb to `Update`

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/GetAzureKeyVaultCertificateIssuer.cs
```diff
@@ -42,17 +40,28 @@ public class GetAzureKeyVaultCertificateIssuer : KeyVaultCmdletBase
...
+            HelpMessage = "KeyVault object.")]
+        [ValidateNotNullOrEmpty]
+        public PSKeyVault InputObject { get; set; }
+
+        /// <summary>
+        /// Name.
+        /// </summary>
+        [Parameter(Mandatory = false,
             Position = 1,
             ValueFromPipelineByPropertyName = true,
```
markcowl: We should be removing these valuefrompipelinebypropertyname for the interactive parameter sets, right?  I assume this is happening in the breaking change update.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/GetAzureKeyVaultKey.cs
```diff
@@ -91,58 +123,83 @@ public class GetAzureKeyVaultKey : KeyVaultCmdletBase
         /// <summary>
         /// Key version.
         /// </summary>
-        [Parameter(Mandatory = false,
+        [Parameter(Mandatory = true,
```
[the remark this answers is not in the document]
markcowl: Same q.  Are we breaking any existing scenarios?

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/NewAzureKeyVaultCertificatePolicy.cs
```diff
@@ -27,130 +26,164 @@ namespace Microsoft.Azure.Commands.KeyVault
...
+
+        /// <summary>
+        /// DnsNames
+        /// </summary>
+        [Parameter(Mandatory = true,
+                   Position = 1,
+                   ParameterSetName = DNSNamesParameterSet,
+                   ValueFromPipelineByPropertyName = true,
+                   HelpMessage = "Specifies the DNS names in the certificate.")]
+        public List<string> DnsNames { get; set; }
```
markcowl: should rename as DnsName, with alias to old name DnsNames (and deprecate)

dragav: Why? This parameter does take a list..

cormacpayne: @dragav it comes from the [_Strongly Encouraged Developer Guidelines_](https://msdn.microsoft.com/en-us/library/dd878270(v=vs.85).aspx) for PowerShell cmdlets:

_Avoid using plural names for parameters whose value is a single element. This includes parameters that take arrays or lists because the user might supply an array or list with only one element._

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/GetAzureKeyVaultKey.cs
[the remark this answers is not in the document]
maddieclayton: If you look at the help file, I added an optional "Name" parameter to the first parameter set, which will take care of the case where -Version is not provided to this parameter set.  This leaves all scenarios working, and I believed that the parameter sets made more sense.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/GetAzureKeyVaultSecret.cs
```diff
@@ -89,58 +121,83 @@ public class GetAzureKeyVaultSecret : KeyVaultCmdletBase
         /// <summary>
         /// Secret version
         /// </summary>
-        [Parameter(Mandatory = false,
+        [Parameter(Mandatory = true,
```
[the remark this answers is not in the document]
maddieclayton: Same comment here. I added an optional "Name" parameter to the first parameter set, which will take care of the case where -Version is not provided to this parameter set. This leaves all scenarios working, and I believed that the parameter sets made more sense.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/SetAzureKeyVaultCertificatePolicy.cs
```diff
@@ -64,11 +64,46 @@ public class SetAzureKeyVaultCertificatePolicy : KeyVaultCmdletBase
         [Alias(Constants.CertificateName)]
         public string Name { get; set; }
 
+        /// <summary>
+        /// CertificatePolicy
+        /// </summary>
+        [Parameter(Mandatory = true,
+            Position = 2,
+            ValueFromPipelineByPropertyName = true,
```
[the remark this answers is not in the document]
maddieclayton: Good call, thanks for the catch.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/RemoveAzureKeyVault.cs
```diff
@@ -62,82 +90,112 @@ public class RemoveAzureKeyVault : KeyVaultManagementCmdletBase
...
-        /// </summary>
-        [Parameter(Mandatory = false,
-           HelpMessage = "Indicates that the cmdlet does not prompt you for confirmation. By default, this cmdlet prompts you to confirm that you want to delete the key vault.")]
-        public SwitchParameter Force { get; set; }
-
         /// <summary>
         /// If present, operate on the deleted vault entity.
         /// </summary>
-        [Parameter(Mandatory = false,
+        [Parameter(Mandatory = true,
```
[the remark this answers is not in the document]
maddieclayton: If they provided all the parameters except -InRemovedState, it would default to the first parameter set, because the first is the default and contains all the parameters in the second.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ImportAzureKeyVaultCertificate.cs
maddieclayton: Deleting alias as it doesn't really make sense in this context.

src/ResourceManager/KeyVault/Commands.KeyVault/help/Set-AzureKeyVaultCertificateAttribute.md
maddieclayton: Going to rename to Update-AzureKeyVaultSecret as I will combine this with Set-AzureKeyVaultSecret at the breaking change release and I don't want to aliases to get out of control.

dragav: Please don't do this. There is no separate command for adding or importing a secret - Set fits the intended profile here.

src/ResourceManager/KeyVault/Commands.KeyVault/help/Set-AzureKeyVaultSecretAttribute.md
maddieclayton: Going to rename to Update-AzureKeyVaultSecret as I will combine this with Set-AzureKeyVaultSecret at the breaking change release and I don't want to aliases to get out of control.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/NewAzureKeyVaultCertificatePolicy.cs
dragav: I disagree. We don't have a parameter called "PermissionToSecrets" - plural is appropriate here as it conveys the fact that the certificate accepts an enumeration of DNS names. Having a suggestively named parameter outweighs a convention based solely on the parameter's type.

src/ResourceManager/KeyVault/Commands.KeyVault/help/Set-AzureKeyVaultCertificatePolicy.md
dragav: Same comment as above - there is no Add-Policy, and we don't want one. A Set creates the policy if missing, and overwrites it if existing.

src/ResourceManager/KeyVault/Commands.KeyVault/help/Set-AzureKeyVaultCertificateAttribute.md
maddieclayton: I'm going to skip doing these aliases for now and we will discuss again for the breaking change PR - I think some of these aliases might make sense while others do not.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/GetAzureKeyVaultCertificateIssuer.cs
maddieclayton: Yes, this will be happening for the breaking change update.

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/SetAzureKeyVaultCertificateIssuer.cs
maddieclayton: Done

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/NewAzureKeyVaultCertificatePolicy.cs
cormacpayne: @dragav I agree that this convention is silly, but it is one that we enforce for new changes across _all_ modules and one that has been brought up by users numerous times (_e.g._, [this issue](https://github.com/Azure/azure-powershell/issues/2919)).

Going back to the guidelines mentioned above, as long as the user can provide a _single_ object to the list/array parameter, than the parameter noun should be singular. If, however, you can do extra validation to ensure that the user _always_ provides more than one object, then having a plural noun is allowed:

_Plural parameter names should be used only in those cases where the value of the parameter is always a multiple-element value. In these cases, the cmdlet should verify that multiple elements are supplied, and the cmdlet should display a warning to the user if multiple elements are not supplied._

maddieclayton:
on demand here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/powershell-demand/648/ (and 649)

Issue #5660, opened 2018-03-02
New-AzureRmVmss/Vm should support a custom image. 


sergey-shandar:
### Description

VM currently _partly_ support a custom image, see VHD.

URL to disk is a name (id?)

See https://review.docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/tutorial-use-custom-image-powershell?branch=pr-en-us-33477

### Script/Steps for Reproduction

```powershell

```

### Module Version

```powershell
Get-Module -Name AzureRM -ListAvailable
```

### Environment Data

```powershell
$PSVersionTable
```

### Debug Output

```

```

PR #5721, opened 2018-03-10
New-AzureRmVm/Vmss: Custom image

sergey-shandar:
## Description

New-AzureRmVm: Custom image.

Currently, only an image in the same resource group is supported. Id is not supported yet.

Issue #5660 

Depends on PR #5722 

## Checklist

- [ ] I have read the [_Submitting Changes_](https://github.com/Azure/azure-powershell/blob/preview/CONTRIBUTING.md#making-changes) section of [`CONTRIBUTING.md`](https://github.com/Azure/azure-powershell/blob/preview/CONTRIBUTING.md)

### DIFF
--- a/src/ResourceManager/Compute/ChangeLog.md
+++ b/src/ResourceManager/Compute/ChangeLog.md
@@ -18,6 +18,7 @@
         - Additional information about change #1
 -->
 ## Current Release
+* `New-AzureRmVM` and `New-AzureRmVMSS` support data disks.
 * `New-AzureRmVM` and `New-AzureRmVMSS` support custom image by name or by id.
 * Log analytic feature
     - Added `Export-AzureRmLogAnalyticRequestRateByInterval` cmdlet


src/ResourceManager/Compute/Commands.Compute/Strategies/ComputeRp/ImageEx.cs
```diff
@@ -25,14 +25,78 @@ namespace Microsoft.Azure.Commands.Compute.Strategies.ComputeRp
...
+                var resourceName = imageArray[8];
+
+                if (empty != string.Empty
+                    || subscriptions != SdkEngine.Subscriptions 
+                    || resourceGroups != ResourceType.ResourceGroups
+                    || providers != EntityConfigExtensions.Providers
+                    || providerNamespace != ComputeStrategy.Namespace
+                    || provider != "images")
+                {
+                    throw new ArgumentException("Invalid image resource id  '" + imageName + "'.");
```
markcowl: This should be in a resource string

src/ResourceManager/Compute/Commands.Compute/Strategies/ComputeRp/ImageEx.cs
```diff
@@ -25,14 +25,78 @@ namespace Microsoft.Azure.Commands.Compute.Strategies.ComputeRp
...
+                }
+
+                var compute = client.GetClient<ComputeManagementClient>();
+                if (compute.SubscriptionId != subscriptionId)
+                {
+                    throw new ArgumentException("The image subscription doesn't math the current subscription.");
```
markcowl: resource string, also 'math' -> 'match'

src/ResourceManager/Compute/Commands.Compute/Strategies/ComputeRp/ImageEx.cs
```diff
@@ -65,8 +129,21 @@ public static int[] UpdatePorts(this OperatingSystemTypes osType, int[] ports)
             }
             else
             {
-                // get image
-                return Images
+                var compute = client.GetClient<ComputeManagementClient>();
```
markcowl: We shouldn't duplicate this code -it feels like this should be in the same if case as if the image contains a '/'

src/ResourceManager/Compute/Commands.Compute/Strategies/ComputeRp/ImageEx.cs
```diff
@@ -25,14 +25,78 @@ namespace Microsoft.Azure.Commands.Compute.Strategies.ComputeRp
...
+            }
+
+            if (imageName.Contains("/"))
             {
+                var imageArray = imageName.Split('/');
+                if (imageArray.Length != 9)
+                {
+                    throw new ArgumentException("Invalid image resource id  '" + imageName + "'.");
+                }
+                // has to be ""
```
markcowl: I would move image resource id parsing into a helper method, or better yet, use this common code, and simply ensure that the properties are what you expect:  https://github.com/Azure/azure-powershell/blob/preview/src/ResourceManager/Common/Commands.ResourceManager.Common/Utilities/Models/ResourceIdentifier.cs

This will give you subscription, resourceGroupName, resourceName and resourceType (Microsoft.Comute/images)

src/ResourceManager/Compute/Commands.Compute/Strategies/ComputeRp/ImageEx.cs
markcowl: I would suggest first checking if the name contains a ':', then a single case for '/' or no '/'.

src/ResourceManager/Compute/Commands.Compute/Strategies/ComputeRp/ImageEx.cs
```diff
@@ -77,6 +154,14 @@ public static int[] UpdatePorts(this OperatingSystemTypes osType, int[] ports)
                                 : OperatingSystemTypes.Linux,
                             nameAndImage.Value)))
                     .FirstOrDefault();
+
+                if (result == null)
+                {
+                    // TODO: move it to resource.
+                    throw new ArgumentException("Can't find the image '" + imageName + "'");
```
markcowl: yes, please move to resource

src/ResourceManager/Common/Commands.Common.Strategies/ResourceId.cs
```diff
@@ -0,0 +1,79 @@
...
+        public const string Subscriptions = "subscriptions";
+        public const string ResourceGroups = "resourceGroups";
+        public const string Providers = "providers";
+
+        /// <summary>
+        /// Returns 'null' if the given id is not parsable.
+        /// </summary>
+        /// <param name="id"></param>
+        /// <returns></returns>
+        public static IResourceId TryParse(string id)
```
cormacpayne: @sergey-shandar any reason you implemented this class instead of using the existing [`ResourceIdentifier`](https://github.com/Azure/azure-powershell/blob/preview/src/ResourceManager/Common/Commands.ResourceManager.Common/Utilities/Models/ResourceIdentifier.cs) class?

src/ResourceManager/Compute/Commands.Compute/Strategies/ComputeRp/ImageEx.cs
```diff
@@ -81,10 +84,44 @@ private static int[] CreatePorts(this OperatingSystemTypes osType)
...
+                    localImage.StorageProfile.OsDisk.OsType,
+                    new ImageReference { Id = localImage.Id });
+            }
             else
             {
-                // get image
-                return Images
+                try
+                {
+                    var localImage = await compute.Images.GetAsync(resourceGroupName, imageName);
```
cormacpayne: @sergey-shandar does this address Mark's comments?

_We shouldn't duplicate this code -it feels like this should be in the same if case as if the image contains a '/'. I would suggest first checking if the name contains a ':', then a single case for '/' or no '/'._

From the comment, it seems like the above `else if` should just be an `else`.

src/ResourceManager/Common/Commands.Common.Strategies/ResourceId.cs
sergey-shandar: We also need this interface in template implementation. The ARM template implementation is  inside a Strategy library which doesn't depend on a library which implements `ResourceIdentifier` class.

src/ResourceManager/Compute/Commands.Compute/Strategies/ComputeRp/ImageEx.cs
sergey-shandar: The implementations are quite different and we don't want to invalid image ids to pass as image names.

cormacpayne:
On-demand: https://azuresdkci.westus2.cloudapp.azure.com/job/powershell-demand/631/

Issue #4807, opened 2017-10-18
Set-AzureStorageBlobContent is not working if FIPS is enabled. 


rgathub:
### Cmdlet(s)
Set-AzureStorageBlobContent 

### PowerShell Version
Instructions: to get PowerShell version, type `$PSVersionTable` and look for the value associated with `PSVersion`
5.1.15063.608   

### Module Version
3.4.1

### OS Version
Instructions: to get OS version, type `$PSversionTable` and look for value associated with `BuildVersion`
10.0.15063.608     

### Description
Unable to upload files to BLOB Storage if FIPS is enabled. This also prevents me to do ARM deployments with Custom Script Extensions as that relies on the cmdlet Set-AzureStorageBlobContent .

### Debug Output
Instructions: to get Debug Output, set `$DebugPreference="Continue"` and then execute the cmdlet or script causing the issue
DEBUG: 9:48:53 AM - NewAzureStorageContext begin processing with ParameterSet 'AccountNameAndKey'.
DEBUG: 9:48:53 AM - NewAzureStorageContext end processing.
DEBUG: 9:48:53 AM - Init Operation Context for 'SetAzureBlobContentCommand' with client request id  Azure-Storage-PowerShell-675d205d-ae74-4908-91b2-421df71b6c60. If you want to get more details, please add "-Debug" to your command.
DEBUG: 9:48:53 AM - SetAzureBlobContentCommand begin processing with ParameterSet 'SendManual'.
DEBUG: 9:48:53 AM - Use storage account 'strsspdiagnostics' from storage context.
Set-AzureStorageBlobContent : This implementation is not part of the Windows Platform FIPS validated cryptographic algorithms.
At line:9 char:1
+ Set-AzureStorageBlobContent -File $File -Context $StorageContext -Con ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : CloseError: (:) [Set-AzureStorageBlobContent], InvalidOperationException
    + FullyQualifiedErrorId : InvalidOperationException,Microsoft.WindowsAzure.Commands.Storage.Blob.SetAzureBlobContentCommand

[...]

PR #5746, opened 2018-03-15
[Storage] Fix MD5 not work on FIPS enabled machine issue

blueww:
## Description

Fix the issue  https://github.com/Azure/azure-powershell/issues/4807

The resolution:

- At the begin of the related cmdlets, try call the MD5 caculate function, if the MD5 function report FIPS exception, set DMlib to use native MD5; else still use the orignal MD5 algrithm.

With this change:

- we can get the issue resolved without adding new parameter, so script can be shared across machines without change
- PowerShell by default will still use the orignal MD5 algrithm, so aligned with .net SDK  (the MD5 check cost will be very low)
- It don’t need big cmdlet framework change.

## Checklist


blueww:
@praries880 

Currently, we don't have Unit test for E2E of Storage data plan cmdlets. But will cover it in another test framework. 
As I know, @MiYanni might will add Unit test for it in the future.  

So for this change, the test we can do is to check FIPSEnabled is disable on normal machine. But dur to the protect level, we can't access FIPSEnabled from test case. I can change FIPSEnabled to public to add a test. But it looks not proper to change an object from protected to public just for test.

So I think we can wait for @MiYanni to add the storage Unit test, then add test case for it.

MiYanni:
@blueww I'm not involved in this situation/issue. What's going on?

blueww:
@MiYanni  
Don't worry. It's just I am asked to add Unit test for some Storage data plan cmdlets change. But we don't have E2E Storage Dataplan test in PowerShell Unit test now.
As I remember, Mark send you a mail to add Storage Data plan test based on a script before. 
So this test should base on the test change that change.
I am not pushing you to add the test now. Just explain why we can't add test for this change now.

cormacpayne:
Sign job: https://azuresdkci.westus2.cloudapp.azure.com/job/ps-sign/265/

praries880:
@blueww   Thanks for the explanation. 
Since we cant create a test, would you mind testing the signed bits for your changes (found here https://azuresdkci.westus2.cloudapp.azure.com/job/ps-sign/265/artifact/src/Package/) and test the functionality on a machine where you can controll the FIPS setting?

src/Storage/Commands.Storage/Common/StorageCloudCmdletBase.cs
```diff
@@ -622,5 +622,32 @@ protected override void StopProcessing()
...
+                catch (System.Reflection.TargetInvocationException)
+                {
+                    fIPSEnabled = true;
+                    return true;
+                }
+                fIPSEnabled = false;
+                return false;
+            }
+        }
+        private static bool? fIPSEnabled = null;
```
markcowl: variable name should be fipsEnabled

src/Storage/Commands.Storage/Common/StorageCloudCmdletBase.cs
```diff
@@ -622,5 +622,32 @@ protected override void StopProcessing()
...
+                try
+                {
+                    System.Security.Cryptography.MD5.Create();
+                }
+                catch (System.Reflection.TargetInvocationException)
+                {
+                    fIPSEnabled = true;
+                    return true;
+                }
+                fIPSEnabled = false;
```
markcowl: vertical space

markcowl:
```c#
public static bool FipsEnabled {get;} = IsFIPSEnabled();

internal static bool ISFIPSEnabled()
{
                try
                {
                    System.Security.Cryptography.MD5.Create();
                    return false;
                }
              catch (System.Reflection.TargetInvocationException)
                {
                    return true;
                }
}
```

markcowl: I think we can take advantage of the fact that the static initializer will run once to simploify the code - so we cans et the value and then use it later

src/Storage/Commands.Storage/File/Cmdlet/SetAzureStorageFileContent.cs
```diff
@@ -80,6 +80,12 @@ public override void ExecuteCmdlet()
                 throw new FileNotFoundException(string.Format(CultureInfo.CurrentCulture, Resources.SourceFileNotFound, this.Source));
             }
 
+            // if FIPS policy is enabled, must use native MD5
+            if (FIPSEnabled)
+            {
+                CloudStorageAccount.UseV1MD5 = false;
```
markcowl: Is this the only cmdlet that would be impacted?  Wouldn't every blob-related cmdlet need this setting?

src/Storage/Commands.Storage/Common/StorageCloudCmdletBase.cs
blueww: fixed, use fipsEnabled

src/Storage/Commands.Storage/Common/StorageCloudCmdletBase.cs
```diff
@@ -622,5 +622,32 @@ protected override void StopProcessing()
...
         }
+
+        /// <summary>
+        /// true if FIPS policy is enabled on the current machine
+        /// </summary>
+        protected static bool FIPSEnabled
+        {
+            get
+            {
+                if (fIPSEnabled.HasValue)
```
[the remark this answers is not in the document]
blueww: fixed, use fipsEnabled

src/Storage/Commands.Storage/Common/StorageCloudCmdletBase.cs
blueww: Fixed as the code from you. Thanks!

src/Storage/Commands.Storage/File/Cmdlet/SetAzureStorageFileContent.cs
blueww: Only upload blob/file cmdlets need use MD5 function to calculate Property ContentMD5. Other cmdlets don't use this function. So only need change cmdlets Set-AzureStorageBlob(File)Content

blueww:
@markcowl 

I have fixed the comments from you, please help to check. Change log is also added.
The on-demand test can be found in: https://azuresdkci.westus2.cloudapp.azure.com/job/powershell-demand/650/

PR #5771, opened 2018-03-20
Add MSI eMSI support

markcowl:
## Checklist

### DIFF
--- a/src/ResourceManager/Profile/Commands.Profile/help/Connect-AzureRmAccount.md
+++ b/src/ResourceManager/Profile/Commands.Profile/help/Connect-AzureRmAccount.md
@@ -21,9 +21,10 @@ Connect-AzureRmAccount [-Environment <String>] [[-Credential] <PSCredential>] [-
 
 ### ServicePrincipalWithSubscriptionId
 ```
-Connect-AzureRmAccount [-Environment <String>] [-Credential] <PSCredential> [-ServicePrincipal] -TenantId <String>
- [-Subscription <String>] [-ContextName <String>] [-Force] [-Scope <ContextModificationScope>]
- [-DefaultProfile <IAzureContextContainer>] [-WhatIf] [-Confirm] [<CommonParameters>]
+Connect-AzureRmAccount [-Environment <String>] [-Credential] <PSCredential> [-ServicePrincipal]
+ -TenantId <String> [-Subscription <String>] [-ContextName <String>] [-Force]
+ [-Scope <ContextModificationScope>] [-DefaultProfile <IAzureContextContainer>] [-WhatIf] [-Confirm]
+ [<CommonParameters>]
 ```
 
 ### ServicePrincipalCertificateWithSubscriptionId
@@ -44,7 +45,7 @@ Connect-AzureRmAccount [-Environment <String>] [-TenantId <String>] -AccessToken
 
 ### ManagedServiceLogin
 ```
-Add-AzureRmAccount [-Environment <String>] [-TenantId <String>] [-AccountId <String>] [-ManagedService]
+Connect-AzureRmAccount [-Environment <String>] [-TenantId <String>] [-AccountId <String>] [-Identity]
  [-ManagedServicePort <Int32>] [-ManagedServiceHostName <String>] [-Subscription <String>]
  [-ContextName <String>] [-Force] [-Scope <ContextModificationScope>]
  [-DefaultProfile <IAzureContextContainer>] [-WhatIf] [-Confirm] [<CommonParameters>]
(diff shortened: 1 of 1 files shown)


src/Common/Commands.Common.Authentication/Authentication/ManagedServiceAccessToken.cs
```diff
@@ -93,22 +108,97 @@ public void AuthorizeRequest(Action<string, string> authTokenSetter)
 
         void GetOrRenewAuthentication()
         {
-            if (_expiration - DateTime.UtcNow < TimeSpan.FromMinutes(5))
+            if (_expiration - DateTime.UtcNow < TimeSpan.FromMinutes(4))
```
praries880: @markcowl A comment here about the significance of the 4 minute to expiration time check would be great... 

also what drove the reduction of the one minute here 5 to 4?

src/Common/Commands.Common.Authentication/Authentication/ManagedServiceTokenInfo.cs
```diff
@@ -42,9 +44,22 @@ public class ManagedServiceTokenInfo
...
         }
+
+        public bool ShouldCache()
+        {
+            return !IsExpired();
+        }
+
+        public bool IsExpired()
+        {
+            return DateTime.UtcNow - CreationDate > (TimeSpan.FromSeconds(ExpiresIn) - TimeSpan.FromMinutes(5));
```
praries880: Is the 5 minutes supposed to be a buffer time? So that we never end up with an expired token in our cache? Also is this related to the check for 4 mins in the code above?

If so, should we look at introducing a constant that is set to 5 and use <said constant> -1 in the code above?

src/Common/Commands.Common.Authentication/Factories/AuthenticationFactory.cs
```diff
@@ -26,7 +26,9 @@ namespace Microsoft.Azure.Commands.Common.Authentication.Factories
 {
     public class AuthenticationFactory : IAuthenticationFactory
     {
-        public const string CommonAdTenant = "Common", DefaultMSILoginUri = "http://localhost:50342/oauth2/token";
+        public const string CommonAdTenant = "Common", 
+            DefaultMSILoginUri = "http://169.254.169.254/metadata/identity/oauth2/token",
```
praries880: Why are we using this specific IP and is it never going to change or should we use a DNS name?

src/Common/Commands.Common.Authentication/Factories/AuthenticationFactory.cs
```diff
@@ -26,7 +26,9 @@ namespace Microsoft.Azure.Commands.Common.Authentication.Factories
 {
     public class AuthenticationFactory : IAuthenticationFactory
     {
-        public const string CommonAdTenant = "Common", DefaultMSILoginUri = "http://localhost:50342/oauth2/token";
+        public const string CommonAdTenant = "Common", 
+            DefaultMSILoginUri = "http://169.254.169.254/metadata/identity/oauth2/token", 
+            DefaultBackupMSILoginUri = "http://localhost:50342/oauth2/token";
```
praries880: Is this for testing purposes? How would we get back a valid token from the local host?

src/Common/Commands.Common.Authentication/Utilities/HttpRetryAlgorithm.cs
```diff
@@ -0,0 +1,101 @@
...
+            public ExponentialRetryAlgorithm(TimeSpan initialInterval, int multiplier, int maxTries)
+            {
+                _currentInterval = initialInterval;
+                _multiplier = multiplier;
+                _maxTries = maxTries;
+            }
+            public TimeSpan GetRetryInterval()
+            {
+                var result = _currentInterval;
+                _currentInterval = TimeSpan.FromTicks(_currentInterval.Ticks * _multiplier);
```
praries880: Can there be a case where we have multiple threads/processes retrying for the same endpoint at the same time? If so, we should look at adding a small amount of jitter to each retry... jitter helps relieve the case where multiple threads can end up retrying in sync.

src/Common/Commands.Common.Authentication/Factories/AuthenticationFactory.cs
markcowl: No, these are the real endpoints

src/Common/Commands.Common.Authentication/Factories/AuthenticationFactory.cs
markcowl: Because it is the defined ip of the metadata service on every VM host.

cormacpayne:
On-demand: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/powershell-demand/646/

markcowl:
third time? https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/powershell-demand/647/

PR #5807, opened 2018-03-25
[Compute] Update for Zone resilient

hyonholee:
## Checklist


src/ResourceManager/Compute/Commands.Compute/Generated/Disk/Config/NewAzureRmDiskConfigCommand.cs
```diff
@@ -38,7 +38,7 @@ public partial class NewAzureRmDiskConfigCommand : Microsoft.Azure.Commands.Reso
             Position = 0,
             ValueFromPipelineByPropertyName = true)]
         [Alias("AccountType")]
-        public StorageAccountTypes? SkuName { get; set; }
+        public string SkuName { get; set; }
```
cormacpayne: @hyonholee out of curiosity, why the switch from this enum to string across all cmdlets?

Based on the markdown files, it seems that the accepted values went from `StandardLRS` and `PremiumLRS` to `Standard_LRS` and `Premium_LRS`, respectively. This would be a breaking change for existing scripts, which is fine since this will go out as a preview, but it might be worth looking at accepting the old values as well and handling them as special cases until you can make the breaking change in May. Let me know what you think.

src/ResourceManager/Compute/Commands.Compute/Generated/Disk/DiskGrantAccessMethod.cs
```diff
@@ -155,28 +155,23 @@ public override void ExecuteCmdlet()
...
-        [AllowNull]
         public string DiskName { get; set; }
 
-
         [Parameter(
             ParameterSetName = "DefaultParameter",
             Position = 3,
             Mandatory = true)]
-        public AccessLevel Access { get; set; }
+        public string Access { get; set; }
```
cormacpayne: @hyonholee same comment here, why the switch from this enum to string?

src/ResourceManager/Compute/Commands.Compute/Generated/VirtualMachineScaleSet/Config/SetAzureRmVmssStorageProfileCommand.cs
```diff
@@ -81,7 +81,7 @@ public partial class SetAzureRmVmssStorageProfileCommand : Microsoft.Azure.Comma
             Mandatory = false,
             Position = 7,
             ValueFromPipelineByPropertyName = true)]
-        public DiskCreateOptionTypes? OsDiskCreateOption { get; set; }
+        public string OsDiskCreateOption { get; set; }
```
cormacpayne: @hyonholee same comment here, why the switch from enum to string?

src/ResourceManager/Compute/Commands.Compute/Generated/Disk/DiskGrantAccessMethod.cs
hyonholee: @cormacpayne  This is changed in Swagger spec and client library.  They need string for Access because they will keep adding more access type, like VM sizes.

src/ResourceManager/Compute/Commands.Compute/Generated/VirtualMachineScaleSet/Config/SetAzureRmVmssStorageProfileCommand.cs
hyonholee: @cormacpayne This is changed in Swagger spec and client library. They need string for create option because they will keep adding more create option type, like VM sizes.

src/ResourceManager/Compute/Commands.Compute/Generated/Disk/Config/NewAzureRmDiskConfigCommand.cs
hyonholee: @cormacpayne  Yes, there are breaking changes in the Swagger spec and client library.  That is why we want preview module.

PR #5827, opened 2018-03-27
Adding Guid check for Reservations cmdlet arguments

corquiri:
## Description

Adding Guid check for Reservations cmdlet arguments

## Checklist


src/ResourceManager/Reservations/ChangeLog.md
```diff
@@ -20,6 +20,10 @@
 -->
 ## Current Release
 
+## Version 0.1.2
```
cormacpayne: @corquiri you can remove the version header, the text content you're adding should go under the `## Current Release` header

src/ResourceManager/Reservations/Commands.Reservations/Common/AzureReservationsCmdletBase.cs
```diff
@@ -21,6 +22,15 @@ public abstract class AzureReservationsCmdletBase : AzureRMCmdlet
 
         private Dictionary<string, List<string>> _defaultRequestHeaders;
 
+        protected void ValidateGuidInput(string inputValue, string inputType)
```
cormacpayne: @corquiri is there any reason you can't just make the parameters of type `Guid` rather than of type `string`? If the server was never able to handle a non-Guid string before, then this won't be a breaking change. This will allow you to remove all of the validation you are trying to do over the strings.

maddieclayton:
@corquiri Ping - please address the comments above.

src/ResourceManager/Reservations/Commands.Reservations/Common/AzureReservationsCmdletBase.cs
corquiri: Do you mean if I change parameter to something like
         [Parameter(Mandatory = false)]
         [ValidateNotNull]
         public Guid SubscriptionId { get; set; }

Then the powershell console will just throw if the input is not a GUID? Where is the code that parses the input from console and populate the parameter of cmdlet classes? If I can verify that the parameter validation is already being done, I can follow your suggestion and change the parameters to Guid type.

cormacpayne: @corquiri the PowerShell engine (not controlled by us) automatically attempts to cast the provided object to whatever type the parameter is expecting, and if it's unable to do so, it will throw an error. In this case, if the user provides a string that is not in the form of a GUID, PowerShell will throw with an error saying that it is unable to convert the given string into a GUID.

corquiri:
Reverted Guid check function and just changed the type and references accordingly.
Question : in MergeReservation.cs command, there is this parameter.
[Parameter(ParameterSetName = Constants.ParameterSetNames.CommandParameterSet, 
            Mandatory = true)]
        [ValidateNotNull]
        [ValidateCount (2, 2)]
        public Guid[] ReservationId { get; set; }

When I call merge-azurermreservation without providing this paramter, PS prompts me to enter values for this parameter (array of Guid). However, even when I specify that the exact count of this array is 2, PS keeps asking for more values after I enter 2 values. Screenshot is attached.

capture

Is there any way to avoid this situation?

cormacpayne:
@corquiri the `ValidateCount` attribute does validation _after_ you provide a value for the given parameter, so PowerShell will not stop prompting you for values after you have provided two. To avoid this situation, once it prompts you for the third entry, you just press enter and it will stop prompting.

cormacpayne:
@corquiri ping

corquiri:
updated changelog and re-triggering CI build because previous was not found

corquiri:
travis-ci build is stuck
continuous-integration/travis-ci Expected — Waiting for status to be reported
anyone familiar with this issue?

cormacpayne:
@corquiri I found the build in Travis and kicked it off again, as well as pushed another commit to resolve the merge conflict, and GitHub was able to find the build

cormacpayne:
@corquiri the build is failing because the breaking change tool is flagging the change from `string` to `Guid` as a breaking change, which it isn't since previous calls using a non-Guid string would fail. To suppress these exceptions, please follow the steps [here](https://github.com/Azure/azure-powershell/blob/preview/documentation/Debugging-StaticAnalysis-Errors.md#breaking-changes).

corquiri:
followed the instruction and updated breaking changes file from web editor

Issue #5663, opened 2018-03-02
New-AzureRmVM/VMSS should use hex or Crockford's Base32 for FQDN


sergey-shandar:
### Description

https://en.wikipedia.org/wiki/Base32#Crockford's_Base32, it's compatible with domain name symbols. For 128 bit id we need 26 symbols instead of 32 (hex).

### Script/Steps for Reproduction

```powershell

```

### Module Version

```powershell
Get-Module -Name AzureRM -ListAvailable
```

### Environment Data

```powershell
$PSVersionTable
```

### Debug Output

```

```

Issue #5650, opened 2018-03-01
New-AzureRmVM should display information about the objects being created with -WhatIf


sergey-shandar:
### Description

New-AzureRmVM (probably other strategy cmdlets) should display information about the different resources being created with -WhatIf

### Script/Steps for Reproduction

```powershell

```

### Module Version

```powershell
Get-Module -Name AzureRM -ListAvailable
```

### Environment Data

```powershell
$PSVersionTable
```

### Debug Output

```

```

maddieclayton:
@sergey-shandar Is this completed?

sergey-shandar:
@maddieclayton yes

PR #5933, opened 2018-04-11
New-AzureRmVm/VMSS: verbose parameter output.

sergey-shandar:
## Description

#5650 

Known issues:
- [ ] It appears that the implementation also depends on the issue #5663. We need deterministic -parameter for a domain name label.
- [x] Complex object PS serialization.
   - `IEnumerable`
   - `null`
   - `.ToString()` for everything else

## Checklist


src/ResourceManager/Compute/Commands.Compute/Strategies/AsyncCmdletExtensions.cs
```diff
@@ -79,20 +81,39 @@ public static void WriteVerbose(this IAsyncCmdlet cmdlet, string message, params
...
+
+            public IEnumerable<KeyValuePair<string, object>> Parameters
+            {
+                get
+                {
+                    var psName = _Cmdlet.ParameterSetName;
+                    return typeof(T)
+                        .GetProperties()
+                        .Where(p => p
+                            .GetCustomAttributes(false)
```
praries880: we should check for null here, what if there is a random class property that does not have any custom attributes... will the code here throw a null ptr exception?

src/ResourceManager/Compute/Commands.Compute/Strategies/AsyncCmdletExtensions.cs
```diff
@@ -79,20 +81,39 @@ public static void WriteVerbose(this IAsyncCmdlet cmdlet, string message, params
...
+                            .OfType<ParameterAttribute>()
+                            .Any(a => a.ParameterSetName == psName))
```
praries880: what about parameters like : https://github.com/praries880/azure-powershell/blob/dependency_fix/src/ResourceManager/Compute/Commands.Compute/VirtualMachine/Operation/GetAzureVMCommand.cs#L64
which do not have the property specified? what happens in this case?

src/ResourceManager/Compute/Commands.Compute/Strategies/StrategyCmdlet.cs
```diff
@@ -38,5 +46,24 @@ static class StrategyCmdlet
...
+            if (s != null)
+            {
+                return "\"" + s + "\"";
+            }
+            var e = value as IEnumerable;
+            if (e != null)
+            {
+                return string.Join(",", e.Cast<object>().Select(ToPowerShellString));
+            }
+            return value.ToString();
```
praries880: so for this, if the type of the parameter (specifically for custom defined classes) does not implement a ToString() this would just print out the object type instead of a string representation of the data in the object? Do we want to do this?

praries880:
Want about adding unit tests for this scenario?

src/ResourceManager/Compute/Commands.Compute/Strategies/AsyncCmdletExtensions.cs
sergey-shandar: `Type.GetProperties()` and `PropertyInfo.GetCustomAttributes()` never returns `null`. 
- https://msdn.microsoft.com/en-us/library/kyaxdd3x(v=vs.110).aspx
- https://msdn.microsoft.com/en-us/library/kff8s254(v=vs.110).aspx

src/ResourceManager/Compute/Commands.Compute/Strategies/StrategyCmdlet.cs
sergey-shandar: Yes, if we don't know the type we use `object.ToString()` function. This's also default PowerShell behavior.

src/ResourceManager/Compute/Commands.Compute/Strategies/AsyncCmdletExtensions.cs
sergey-shandar: I've a fix. I'm going to push it soon.

src/ResourceManager/Compute/Commands.Compute/Strategies/AsyncCmdletExtensions.cs
```diff
@@ -79,20 +81,40 @@ public static void WriteVerbose(this IAsyncCmdlet cmdlet, string message, params
...
             public List<ITaskProgress> TaskProgressList { get; } 
                 = new List<ITaskProgress>();
 
-            public AsyncCmdlet(Cmdlet cmdlet)
+            public AsyncCmdlet(T cmdlet)
             {
                 _Cmdlet = cmdlet;
             }
 
+            public bool WhatIf
```
markcowl: @sergey-shandar As we have discussed, this is an anti-pattern, we should not print additional information based on our own setting of WhatIf - WhatIf should be handled automaticsally by the PowerShell runtime.  Additional information printed in WhatIf should use only ShouldProcess prompts.

Verbose output should simply be written to the verbose stream - we cna log such output to the Verbose or Debug streams, depending on the level of information (verbose versus debug/very verbose)

markcowl:
on demand run here:  https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/powershell-demand/764/

Issue #4715, opened 2017-10-04
Multiple logins will cause Select-AzureRmSubscription to select default subscription


tran-barry:
### Cmdlet(s)
- Import-AzureRmContext
- Select-AzureRmSubscription

### PowerShell Version
- 5.1.15063.296

### Module Version
- 4.4.0

### OS Version
- 10.0.15063.296

### Description
If you login multiple times within the same console session, every 2nd login will cause `Select-AzureRmSubscription` to select the default subscription associated with your login instead of the one you specify.

This is only an issue in 4.4.0. Downgrading to 4.3.1 resolved the issue.

Also, having an azure profile with multiple subscriptions defined would work around the issue.

#### First Invocation (correct):
```
PS C:\Users\barry.tran> Import-AzureRmContext $pathToProfile; Select-AzureRmSubscription -subscriptionName Sandbox
DEBUG: 12:51:11 PM - Autosave setting from startup session: 'Process'
DEBUG: 12:51:11 PM - No autosave setting detected in environment variable 'AzureRmContextAutoSave'. 
DEBUG: 12:51:11 PM - Using Autosave scope 'Process'


Account          : BarryT@iqmetrix.com
SubscriptionName : Operations
SubscriptionId   : ********-****-****-****-************
TenantId         : ********-****-****-****-************
Environment      : AzureCloud

DEBUG: 12:51:11 PM - ImportAzureRMContextCommand end processing.
DEBUG: 12:51:11 PM - ImportAzureRMContextCommand end processing.
DEBUG: 12:51:11 PM - SetAzureRMContextCommand begin processing with ParameterSet 'Subscription'.
DEBUG: 12:51:11 PM - using account id 'BarryT@iqmetrix.com'...
DEBUG: 12:51:11 PM - Autosave setting from startup session: 'Process'
DEBUG: 12:51:11 PM - No autosave setting detected in environment variable 'AzureRmContextAutoSave'. 
DEBUG: 12:51:11 PM - Using Autosave scope 'Process'
DEBUG: [Common.Authentication]: Authenticating using configuration values: Domain: '********-****-****-****-************', En
dpoint: 'https://login.microsoftonline.com/', ClientId: '********-****-****-****-************', ClientRedirect: 'urn:ietf:wg:
oauth:2.0:oob', ResourceClientUri: 'https://management.core.windows.net/', ValidateAuthrity: 'True'
[...]

PR #5958, opened 2018-04-16
Fix Import-AzureRmContext using -Path and -AzureContext Parameters (Issue 4715)

AMoghrabi:
## Description

* One of my commits is for [Issue 4715](https://github.com/Azure/azure-powershell/issues/4715)
    * When running `Import-AzureRmContext -Path {path}`, and then `Select-AzureRmSubscription -SubscriptionName {subscriptionName}`, and then importing your context file again using the first cmdlet, it will invalidate the default context by setting it to a new instantiated copy of `DefaultContext` . This makes `Select-AzureRmSubscription` fail to select the context because it has fields that are null

* Second commit is to fix the `-AzureContext` parameter, which is not in issue 4715 (I found it while investigating issue 4715). Currently, an exception is thrown when running the cmdlet example found on the [documentation page](https://docs.microsoft.com/en-us/powershell/module/azurerm.profile/import-azurermcontext?view=azurermps-5.7.0):

`Import-AzureRmContext -AzureContext (Connect-AzureRmAccount)`

Error:

```
Import-AzureRmContext : Collection was modified; enumeration operation may not execute.
At line:1 char:1
+ Import-AzureRmContext -AzureContext $bla
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : CloseError: (:) [Import-AzureRmContext], InvalidOperationException
    + FullyQualifiedErrorId : Microsoft.Azure.Commands.Profile.ImportAzureRMContextCommand
```

This is because when trying to copy the default context's extended properties, the objects passed in to [UpdateProperties](https://github.com/Azure/azure-powershell/blob/preview/src/Common/Commands.Common.Authentication.Abstractions/Extensions/ModelExtensions.cs#L124) happen to be referencing the same object, which throws an error while trying to modify it during iteration.

* I couldn't find any examples of tests for `Import-AzureRmContext`, and I'm not sure if I need to write one for this case. If tests are required, I'd appreciate some help from the team.

* Sorry if I missed or didn't include something. First time contributing to this repo and to a open source project in general. Also, let me know if you'd like to to squash my commits. I tried separating because they're two separate bugs, though it's a very small change.

## Checklist

- [N/A] For public API changes to cmdlets:
    - [N/A] the changes have gone through a [cmdlet design review](https://github.com/Azure/azure-powershell-cmdlet-review-pr)
    - [N/A] the cmdlet markdown files were [generated using the `platyPS` module](https://github.com/Azure/azure-powershell/blob/preview/documentation/development-docs/help-generation.md)


AMoghrabi:
Here's a (probably) stupid question: I'm not sure if I should be performing the same change in a different solution file... The solution I used is `Azure.PowerShell.Netcore` and it sounds like `ResourceManager.ForRefactoringOnly.sln` shouldn't be modified as the name suggests 😄 

Should I be doing something else for the `AzureRM` modules (non-.net core) -- or is there some magic behind the scenes that handles this for me?

cormacpayne:
@azuresdkci add to whitelist

src/ResourceManager/Common/Commands.Common.Authentication.ResourceManager/AzureRmProfile.cs
```diff
@@ -566,6 +566,7 @@ public bool TryRemoveEnvironment(string name, out IAzureEnvironment environment)
 
         public bool TryCopyProfile(AzureRmProfile other)
         {
+            this.DefaultContextKey = other.DefaultContextKey;
```
markcowl: I think this belongs after the contexts are copied from the remote profile

src/ResourceManager/Common/Commands.Common.Authentication.ResourceManager/AzureRmProfile.cs
```diff
@@ -577,7 +578,7 @@ public bool TryCopyProfile(AzureRmProfile other)
                 TrySetContext(context.Key, context.Value);
             }
 
-            if (other.DefaultContext != null)
+            if (other.DefaultContext != null && this.DefaultContext != other.DefaultContext)
```
markcowl: Will have to ensure that the default equality comparer does the right thing in this case (i.e. that it can handle differences in context names).  I wonder, if the DefaultContextKey has been set above, and the value is not null, whether this will evaluate to true.

src/ResourceManager/Common/Commands.Common.Authentication.ResourceManager/AzureRmProfile.cs
AMoghrabi: Hey Mark. I can move it to after the contexts are copied. I think I tested this to see if it would make a difference and I don't think it did. Anyways, I'll push a commit in a bit to fix this..

MiYanni:
@markcowl Ping (as it is now Post-Build).

MiYanni:
@markcowl Does @AMoghrabi need to make any changes or is this ready to go?

AMoghrabi:
It would be great if this can be in the next release as my team is waiting on this change so we can upgrade our CI agents.

AMoghrabi:
Good morning @markcowl and @MiYanni . Would you have an estimate timeline when this change will be reviewed and published? 

Thanks!

MiYanni:
@AMoghrabi I'll take this over since Mark is away on vacation.

AMoghrabi:
Thanks @MiYanni ! Please let me know if you need me to make any changes, I can start working on it after work.

src/ResourceManager/Common/Commands.Common.Authentication.ResourceManager/AzureRmProfile.cs
```diff
@@ -573,7 +573,8 @@ public bool TryCopyProfile(AzureRmProfile other)
                 TrySetContext(context.Key, context.Value);
             }
 
-            if (other.DefaultContext != null)
+            this.DefaultContextKey = other.DefaultContextKey;
```
MiYanni: If I'm reading the code properly, this assignment and changes to the if-statement are not necessary. The statement above this sets the Contexts. If `other.DefaultContextKey` exists in `other.Context`, then it will be assigned in [TrySetDefaultContext](https://github.com/AMoghrabi/azure-powershell/blob/dd953b32d04ff025ec69ff800f3e3506afe12bf8/src/ResourceManager/Common/Commands.Common.Authentication.ResourceManager/AzureRmProfile.cs#L468-L478).

AMoghrabi:
@MiYanni I think you're right, AzureRM 6.x has fixed it because all of the available contexts get loaded in. In the version this bug was reported in, only some of the contexts get loaded in (your default context, and whichever ones you select afterwards.)

I'll have a coworker verify if this works and I'll get back to you tomorrow.

However, the `-AzureContext` parameter isn't working and I've included a fix for that in this PR. I've tried it on `6.3.0`:

```
PS C:\WINDOWS\system32> Get-AzureRmContext -ListAvailable

PS C:\WINDOWS\system32> Import-AzureRmContext -AzureContext (Connect-AzureRmAccount)
Import-AzureRmContext : Collection was modified; enumeration operation may not execute.
At line:1 char:1
+ Import-AzureRmContext -AzureContext (Connect-AzureRmAccount)
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : CloseError: (:) [Import-AzureRmContext], InvalidOperationException
    + FullyQualifiedErrorId : Microsoft.Azure.Commands.Profile.ImportAzureRMContextCommand
```

We don't use this parameter so it doesn't affect my company. I haven't seen an issue reported for it either...

MiYanni:
@AMoghrabi I haven't tried doing that operation myself, but my hunch is that because you are in-lining the `Connect-AzureRmAccount` call, it modifies the context that is being passed in while `Import-AzureRmContext` is running. It is highly likely that error comes from one of the 2 `foreach`s above the code you changed; either the EnvironmentTable or Contexts are being modified while that foreach is running. Try this:

```powershell
PS > $ctx = Connect-AzureRmAccount
PS > Import-AzureRmContext -AzureContext $ctx
```

I noticed the line you posted is part of [this reference doc](https://docs.microsoft.com/en-us/powershell/module/azurerm.profile/import-azurermcontext?view=azurermps-6.3.0#examples). Logically, that scenario makes no sense to me. `Connect-AzureRmAccount` automatically 'imports' the context. After some internal discussion, this use case may be outdated. I'll create an issue for this, as fixing the issue isn't the concern. It is more likely that we'd want to remove that parameter set from `Import-AzureRmContext`.

AMoghrabi:
Hey @MiYanni 

I tried the code snippet you gave me, same issue: 

```
PS C:\WINDOWS\system32> $ctx = Connect-AzureRmAccount
PS C:\WINDOWS\system32> Import-AzureRmContext -AzureContext $ctx
Import-AzureRmContext : Collection was modified; enumeration operation may not execute.
At line:1 char:1
+ Import-AzureRmContext -AzureContext $ctx
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : CloseError: (:) [Import-AzureRmContext], InvalidOperationException
    + FullyQualifiedErrorId : Microsoft.Azure.Commands.Profile.ImportAzureRMContextCommand
```

It makes sense to pull the parameter out, The scenario didn't make sense to me either but I just assumed there was some sort of use case for it that I couldn't think of.

Seeing how I can't reproduce the original issue anymore, you may close the pull request. Thanks for your help.

MiYanni:
Thank you for the quick responses and detailed information! Closing the PR.

Issue #5902, opened 2018-04-09
Turn on Context Autosave By Default


markcowl:
### Desciprtion

- Turn on Context AutoSave by default
- Ensure graceful transition to non-autosaved context on permission failures

### Cost: 6

markcowl:
Fixed.

PR #6003, opened 2018-04-20
Enable context autosave by default

cormacpayne:
## Description

Fix for issue https://github.com/Azure/azure-powershell/issues/5902

- Enable context autosave by default
- When the user runs `Connect-AzureRmAccount` with the default (empty) context, populate the context list with _all_ subscriptions in the current tenant

## Checklist


src/Common/Commands.Common.Authentication/AzureSessionInitializer.cs
```diff
@@ -97,6 +97,12 @@ static ContextAutosaveSettings InitializeSessionSettings(IDataStore store, strin
                     result.Mode = settings.Mode;
                     result.ContextFile = settings.ContextFile ?? result.ContextFile;
                 }
+                else
+                {
+                    FileUtilities.EnsureDirectoryExists(profileDirectory);
```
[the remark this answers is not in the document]
cormacpayne: @markcowl updated this so it remains `Process` by default but switches to `CurrentUser` after we are able to write the settings file

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
```diff
@@ -261,6 +261,28 @@ public bool TryRemoveContext(IAzureContext context)
...
             _profile.DefaultContext.TokenCache = _cache;
+            var defaultContext = _profile.DefaultContext;
+            if (shouldPopulateContextList)
+            {
+                var subscriptions = ListSubscriptions(tenantId);
+                foreach (var subscription in subscriptions)
+                {
+                    var tempContext = new AzureContext(subscription, account, environment, newTenant);
+                    string tempName = null;
+                    if (!_profile.TryGetContextName(tempContext, out tempName))
```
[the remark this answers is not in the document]
cormacpayne: @markcowl updated the logic for this after discussion so that the name is now `{{subscriptionName}} - {{subscriptionId}}`

src/ResourceManager/Profile/Commands.Profile/Models/RMProfileClient.cs
```diff
@@ -227,6 +226,7 @@ public bool TryRemoveContext(IAzureContext context)
                 }
             }
 
+            bool shouldPopulateContextList = _profile.DefaultContext?.Account == null;
```
markcowl: Just in case, we should also ensure that _profile?.Contexts?.Count < 2 (actually, this will need to be two null checks, but you get the idea).

cormacpayne: @markcowl in this class, `_profile` is of the abstracted type `IProfileOperations`, which doesn't contain the `Contexts` property (the implementation `AzureRmProfile` does contain this property, but I doubt we want to use that in this class). Is there a separate check we can do here?

markcowl: makes sense. As long as we are not overwriting existing contexts, I think we are ok here

cormacpayne:
On-demand: https://azuresdkci.westus2.cloudapp.azure.com/job/powershell-demand/782/

Issue #5896, opened 2018-04-09
Fix PSCmdlet properties without setters in AsJob scenarios


markcowl:
### Description

Some properties (dynamic parameters, SessionState.Path) are not copied into the target cmdlet when creating a cmdlet copy when executed using AsJon.

- [ ] Identify properties with no setters in PSCmdlet that are used by our cmdlets (at least SessioNState.Path, and DynamicParameters
- [ ] For each property, create a common code mechanism for accessing the exposed functionality in ProcessRecord, and ensure that this functionality is transferred in when copying the cmdlet in AzureLongRunningJob
- [ ] Write unit tests for verifying access to these properties in jobs
- [ ] Find all usages of theese properties and transfer to the common code mechanism
- [ ] Verify that associated issues are properly addressed:
  - [ ] #5502 
  - [ ] #5548 

### Cost: 6

maddieclayton:
Fixed.

v-tisheg:
This is marked as fixed.  Does that mean that the Invoke-AzureRmVMRunCommand using the -AsJob parameter should now work in the production environment or just that the fix is now ready to be released as part of the Milestone 5/01 release?  

I just need to know whether I can tell my customer to give this a try.

maddieclayton:
Yes! The fix has been released, please let me know if your customers continue to see an issue (be sure to use the latest version of AzureRM).

v-tisheg:
My Customer just tried using the -ASJob on his command and he says:

Can you test on your end? The job seems stuck in a running state. I did a simple write-host $env:computername for my powershell script I passed in. 

I will try to get the exact syntax he used.

maddieclayton:
@v-tisheg Please open a new issue with all the relevant issue if you can get the exact syntax he used

v-tisheg:
Thank you, will do.

Timothy Shegrud

Support Engineer
Cloud & Infrastructure Solutions
Customer Services & Support | Microsoft Corporation
(425) 706-4408; * v-tisheg@microsoft.com<mailto:v-tisheg@microsoft.com> | Hours: Sun - Thur, 8:00am-5:00pm PST
Need help outside of my working hours for a technical issue?
*: AzBackup@microsoft.com<mailto:AzBackup@microsoft.com>

[...]

PR #6004, opened 2018-04-20
Fix Issue 5896

maddieclayton:
## Description
https://github.com/Azure/azure-powershell/issues/5896

## Checklist

### DIFF
--- a/src/ResourceManager/Scheduler/Commands.Scheduler/Utilities/ExtensionMethods.cs
+++ b/src/ResourceManager/Scheduler/Commands.Scheduler/Utilities/ExtensionMethods.cs
@@ -20,7 +20,6 @@ namespace Microsoft.Azure.Commands.Scheduler.Utilities
     using System.Linq;
     using System.Management.Automation;
     using Microsoft.Azure.Management.Scheduler.Models;
-    using WindowsAzure.Commands.Utilities.Common;
 
     public static class ExtensionMethods
     {


src/Common/Commands.Common/AzurePSCmdlet.cs
```diff
@@ -137,6 +137,39 @@ protected string PSVersion
 
         protected abstract string DataCollectionWarning { get; }
 
+        private SessionState _asJobSessionState;
+
+        public SessionState AsJobSessionState
+        {
+            get
+            {
+                if (_asJobSessionState == null)
```
markcowl: I don't think we can instantiate this lazily, because it might be first referenced inside a job - I think we need to instantiate it either during BeginProcessing. or in the cmdlet constructor

src/Common/Commands.Common/AzurePSCmdlet.cs
```diff
@@ -137,6 +137,39 @@ protected string PSVersion
...
+                }
+                return _asJobSessionState;
+            }
+            set
+            {
+                _asJobSessionState = value;
+            }
+        }
+
+        private RuntimeDefinedParameterDictionary _copiedDynamicParameters;
```
markcowl: How are we assignign a value to this?

src/Common/Commands.Common/AzurePSCmdlet.cs
maddieclayton: Done

src/Common/Commands.Common/AzurePSCmdlet.cs
maddieclayton: We assign it once the dynamic parameters are created in the cmdlet - there is no base property for DynamicParameters, and they can be changed after the AzurePSCmdlet is instantiated.

maddieclayton:
@azuresdkci Test this please

src/Common/Commands.Common/AzurePSCmdlet.cs
```diff
@@ -137,13 +137,45 @@ protected string PSVersion
...
+
+        public RuntimeDefinedParameterDictionary CopiedDynamicParameters
+        {
+            get
+            {
+                if (_copiedDynamicParameters == null)
+                {
+                    _copiedDynamicParameters = new RuntimeDefinedParameterDictionary();
+                    return _copiedDynamicParameters;
```
markcowl: this extra return is unnecessary

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/CmdletBase/ResourceWithParameterCmdletBase.cs
```diff
@@ -134,6 +134,7 @@ public object GetDynamicParameters()
                         MyInvocation.MyCommand.Parameters.Keys.ToArray());
                 }
             }
+            this.CopiedDynamicParameters = dynamicParameters;
```
markcowl: Hmm, this isn't the interface I was hoping for.  Is this the only place we can get DynamicParameters from?  

If we need to do this, I think a void method with a bit less esoteric name, like ```void RegisterDynamicParameters(RuntimeParamaterDictionary paremeters) might be a  more intuitive option for new usages.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/Policy/NewAzurePolicyAssignment.cs
```diff
@@ -247,6 +247,7 @@ object IDynamicParameters.GetDynamicParameters()
                     this.dynamicParameters.Add(param.Name, dp);
                 }
             }
+            this.CopiedDynamicParameters = dynamicParameters;
```
markcowl: extra vertical space after '}'

src/Common/Commands.Common/AzurePSCmdlet.cs
```diff
@@ -137,13 +137,45 @@ protected string PSVersion
 
         protected abstract string DataCollectionWarning { get; }
 
+        private SessionState _asJobSessionState;
+
+        public SessionState AsJobSessionState
```
markcowl: I wonder if we shouldn't Hide SessionState, and throw if it is used, thoughts?

markcowl: In fact, I'm wondering if my previous advice about not hiding was good.  If we can get the propertes using ```base.```, then hiding them would make the interface simpler. 

Otherwise, we should just rething the name of CopiedParameterDictionary (maybe AsJobParameterDictionary)

maddieclayton: I think it makes sense to hide it here - updated accordingly.

src/ResourceManager/Resources/Commands.ResourceManager/Cmdlets/Implementation/CmdletBase/ResourceWithParameterCmdletBase.cs
maddieclayton: Unfortunately I could not find any way to get the dynamic parameters outside of copying them over during runtime.  I created the void method and updated the relevant files.

markcowl:
on demand run here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/powershell-demand/766/

PR #6045, opened 2018-04-25
Make compliance changes to ManagedStorageAccount KeyVault cmdlets

maddieclayton:
## Checklist

### DIFF
--- a/src/ResourceManager/KeyVault/documentation/upcoming-breaking-changes.md
+++ b/src/ResourceManager/KeyVault/documentation/upcoming-breaking-changes.md
@@ -60,6 +60,8 @@ The following cmdlets were affected this release:
     - Get-AzureKeyVaultCertificateOperation
     - Get-AzureKeyVaultCertificatePolicy
     - Get-AzureKeyVaultKey
+    - Get-AzureKeyVaultManagedStorageAccount
+    - Get-AzureKeyVaultManagedStorageSasDefinition
     - Get-AzureKeyVaultSecret
     - Remove-AzureRmKeyVault
     - Remove-AzureRmKeyVaultAccessPolicy
@@ -68,6 +70,8 @@ The following cmdlets were affected this release:
     - Remove-AzureKeyVaultCertificateIssuer
     - Remove-AzureKeyVaultCertificateOperation
     - Remove-AzureKeyVaultKey
+    - Remove-AzureKeyVaultManagedStorageAccount
+    - Remove-AzureKeyVaultManagedStorageSasDefinition
     - Remove-AzureKeyVaultSecret
     - Restore-AzureKeyVaultKey
     - Restore-AzureKeyVaultSecret
@@ -76,6 +80,7 @@ The following cmdlets were affected this release:
     - Set-AzureKeyVaultCertificateIssuer
     - Set-AzureKeyVaultCertificatePolicy
     - Set-AzureKeyVaultKeyAttribute
+    - Set-AzureKeyVaultManagedStorageSasDefinition
     - Set-AzureKeyVaultSecret
     - Set-AzureKeyVaultSecretAttribute
     - Stop-AzureKeyVaultCertificateOperation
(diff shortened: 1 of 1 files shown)


maddieclayton:
on demand run here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/powershell-demand/770/

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/GetAzureKeyVaultManagedStorageAccount.cs
```diff
@@ -18,16 +18,19 @@
...
 using KeyVaultProperties = Microsoft.Azure.Commands.KeyVault.Properties;
 using Microsoft.Azure.Commands.KeyVault.Models.ManagedStorageAccounts;
+using Microsoft.Azure.Management.Internal.Resources.Utilities.Models;
 
 namespace Microsoft.Azure.Commands.KeyVault
 {
-    [Cmdlet( VerbsCommon.Get, CmdletNoun.AzureKeyVaultManagedStorageAccount)]
-    [OutputType( typeof( List<PSKeyVaultManagedStorageAccount> ), typeof( PSKeyVaultManagedStorageAccount ), typeof(List<PSDeletedKeyVaultManagedStorageAccountIdentityItem>), typeof(PSDeletedKeyVaultManagedStorageAccount) )]
+    [Cmdlet( VerbsCommon.Get, CmdletNoun.AzureKeyVaultManagedStorageAccount, DefaultParameterSetName = ByAccountNameParameterSet)]
+    [OutputType( typeof( List<PSKeyVaultManagedStorageAccountIdentityItem> ), typeof( PSKeyVaultManagedStorageAccount ), typeof(List<PSDeletedKeyVaultManagedStorageAccountIdentityItem>), typeof(PSDeletedKeyVaultManagedStorageAccount) )]
```
markcowl: we should try to have a single output type, if there is a common ancestor type, and it should be a single type, not an enumeration

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/GetAzureKeyVaultManagedStorageSasDefinition.cs
```diff
@@ -21,13 +21,14 @@
 
 namespace Microsoft.Azure.Commands.KeyVault
 {
-    [Cmdlet( VerbsCommon.Get, CmdletNoun.AzureKeyVaultManagedStorageSasDefinition)]
+    [Cmdlet( VerbsCommon.Get, CmdletNoun.AzureKeyVaultManagedStorageSasDefinition, DefaultParameterSetName = ByDefinitionNameParameterSet)]
```
markcowl: same comment on OutputType

src/ResourceManager/KeyVault/Commands.KeyVault/Commands/ManagedStorageAccounts/GetAzureKeyVaultManagedStorageAccount.cs
dragav: Re-reading my late night comment, I realize that is the very definition of a base type; I meant that deleted full-entity types are not direct descendants of full-entity types.

maddieclayton: @dragav It looks like all of the return types are descendants of KVEntityIdentityItem - changing the "OutputType" to be one concrete type will allow users to tab through all the properties of KVEntityIdentityItem when running a cmdlet like "select -Properties", while having multiple output types gives them the option to tab through all properties of all output types, which in this case could be confusing when they are retrieving a regular managedstorageaccount and get the properties for a deleted managedstorageaccount while tabbing.  Note: they can still retreive these properties by typing them out, but they will not be available through tabbing.

dragav:
If that means that, upon invoking GetDeletedMSA -Name <msaName>, the user can only tab through the properties of the EntityIdentityItem type, then I find that equally confusing/unhelpful.

maddieclayton:
@dragav If you'd rather it return all possible properties, I will do that - I'll update it in the next commit.

Issue #2342, opened 2016-05-26
Get-AzureRmTrafficManagerProfile With only -Name argument just returns, no error


timomta:
If you enter the Traffic Manager profile name without the resource group name, like this:
 Get-AzureRmTrafficManagerProfile -Name myprofilename
The cmdlet simply returns to the cmd prompt with no indication of success or failure.
It appears that Get-AzureRmTrafficManagerProfile requires both -Name and -ResourceGroupName to work properly, but it doesn't enforce that or give the user feedback.

AMoghrabi:
@cormacpayne Do you mind if I take a stab at this?

cormacpayne:
@AMoghrabi feel free to open a pull request whenever you'd like ☺️

PR #6097, opened 2018-05-03
Fix Get-AzureRmTrafficManagerProfile when calling with -Name Parameter Only (Issue #2342)

AMoghrabi:
## Description
This is a fix for issue [#2342](https://github.com/Azure/azure-powershell/issues/2342)

When calling `Get-AzureRmTrafficManagerProfile -Name $tmProfileName`, the cmdlet will not return anything because it executes an empty code path. When using the `-Name` parameter you should need to specify the `ResourceGroupName` parameter as well.

From my experience from using the Azure Powershell cmdlets, I found that there are two ways to solve this:

1. Use ParameterSets so that when calling with `-Name`, it will automatically prompt for `-ResourceGroupName` as well. This of course also allows the user to use `-ResourceGroupName` without specifying `-Name`.

2. Allow the user to pass `Name` only, fetch all profiles, and then search for the profile matching the name.

Option 2 is slow so I opted to do option 1. I have basically copied what [Get-AzureRmStorageAccount](https://github.com/Azure/azure-powershell/blob/preview/src/ResourceManager/Storage/Commands.Management.Storage/StorageAccount/GetAzureStorageAccount.cs#L28-L52) has set up.

I'm not really sure how to test this or if it does require a test case. If it does require a test case, I would appreciate some guidance. 

## Checklist

- [N/A] For public API changes to cmdlets:
    - [N/A] the changes have gone through a [cmdlet design review](https://github.com/Azure/azure-powershell-cmdlet-review-pr)
    - [N/A] the cmdlet markdown files were [generated using the `platyPS` module](https://github.com/Azure/azure-powershell/blob/preview/documentation/development-docs/help-generation.md)

### DIFF
--- a/src/ResourceManager/TrafficManager/ChangeLog.md
+++ b/src/ResourceManager/TrafficManager/ChangeLog.md
@@ -18,7 +18,7 @@
         - Additional information about change #1
 -->
 ## Current Release
-* Update the parameters so that -ResourceGroupName parameter is required when using -Name parameter.
+* Update the parameters for `Get-AzureRmTrafficManagerProfile` so that -ResourceGroupName parameter is required when using -Name parameter.
 
 ## Version 4.0.4
 * Set minimum dependency of module to PowerShell 5.0


src/ResourceManager/Storage/ChangeLog.md
```diff
@@ -18,6 +18,7 @@
         - Additional information about change #1
 -->
 ## Current Release
+* Update the parameters so that -ResourceGroupName parameter is required when using -Name parameter.
```
cormacpayne: @AMoghrabi would you mind moving this snippet from the `Storage` change log to the `TrafficManager` change log?

AMoghrabi:
Hey, thanks for the quick review @cormacpayne  :) 

I'll have those changes in once I'm home!

AMoghrabi:
Hey @cormacpayne , the changes you've requested are in now. 

To clean up the change log mistake I made in one of my commits, I decided to force push, only to realize that you made a commit earlier to suppress the false-positive. Sorry about that. I've made the same change you did and I'm hoping that will fix the issue. If it doesn't, I'd appreciate if you can make your change again 😄

PR #6140, opened 2018-05-07
Add KeyVault Provider

maddieclayton:
## Checklist


maddieclayton:
sign job here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/ps-sign/586/
on demand here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/powershell-demand/813/

maddieclayton:
sign job here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/ps-sign/601/

maddieclayton:
updated sign job here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/ps-sign/606/

src/ResourceManager/KeyVault/KeyVaultProvider/KeyVaultProvider.cs
```diff
@@ -0,0 +1,3365 @@
...
+                        var filter = namesFromPath[2].TrimEnd('*');
+                        var filteredPaths = new List<string>();
+                        var allStrings = vault.AccessPolicies;
+
+                        foreach (var stringItem in allStrings)
+                        {
+                            if (stringItem.ObjectId.StartsWith(filter))
+                            {
+                                int separatorIndex = path.LastIndexOf(pathSeparator);
+                                var trimmedPath = path.Substring(0, separatorIndex)
```
cormacpayne: @maddieclayton is this second check needed? If `resourceGroup` isn't `null`, then that means the vault already exists, so shouldn't the `GetVault(...)` call always return an object?

cormacpayne: @maddieclayton do we need a null-check on `dynamicParameters.Sku` before calling `ToString` on it?

cormacpayne: @maddieclayton nit: please put an empty line after closing curly braces (this applies elsewhere in this method)

cormacpayne: @maddieclayton nit: this can just be `if (vault.EnableSoftDelete)`

It looks like you do this in a few places throughout this class, so just make sure those are addressed as well

cormacpayne: @maddieclayton we can consolidate the below lines to be the following:

```cs
var secretExists = allSecrets.Any(s => string.Equals(s.Name, name, StringComparison.OrdinalIgnoreCase));
```

cormacpayne: @maddieclayton `certificatePolicy` is already guaranteed to be non-null here, so we don't need this check

cormacpayne: @maddieclayton nit: please add curly braces to this if-statement

cormacpayne: @maddieclayton from @sphibbs: " 🤘 🤘 "

cormacpayne: @maddieclayton why do we need to catch these exceptions and re-throw them as errors?

cormacpayne: @maddieclayton I think we should refactor this method to be the following:

```cs
private bool OnlyOneOf(params object[] list)
{
    return list.Where(o => o != null).Count() == 1;
}
```

cormacpayne: @maddieclayton what's the purpose of this if-statement if `result` get's overridden on the next line of code?

src/ResourceManager/KeyVault/KeyVaultProvider/KeyVaultProvider.cs
```diff
@@ -0,0 +1,3365 @@
...
+        #endregion DriveHelperMethods
+
+        #region Item Methods
+
+        /// The Windows PowerShell engine calls this method when the 
+        /// Get-Item cmdlet is run and the path to this provider is 
+        /// specified. This method retrieves an item using the specified path.
+        protected override void GetItem(string path)
+        {
+            isLoggedIn();
```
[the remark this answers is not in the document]
maddieclayton: note to self: throw for login error

src/ResourceManager/KeyVault/KeyVaultProvider/KeyVaultProvider.cs
maddieclayton: line 1340

maddieclayton: 1364 add ?

maddieclayton: 1504

maddieclayton: 1662

maddieclayton: note to self: do this everywhere.

maddieclayton: 2420

maddieclayton: 2530

Issue #6066, opened 2018-04-27
Websites:  Set-AzureRmWebApp & Set-AzureRmWebApp properties should be switchParameter type instead of bool


panchagnula:
https://docs.microsoft.com/en-us/powershell/module/azurerm.websites/set-azurermwebapp?view=azurermps-5.7.0&viewFallbackFrom=azurermps-5.0.0 
& https://docs.microsoft.com/en-us/powershell/module/azurerm.websites/set-azurermwebappslot?view=azurermps-5.7.0

Change the properties to be a switch parameter

panchagnula:
Closing this for now & will make the change if needed in the future.

PR #6168, opened 2018-05-09
Adding the ability to toggle ARRAffinity in PS

jocawtho:
## Description

This adds a flag to allow customers to turn off/on ARRAffinity for their webapps. I also added this to the test and help files.

## Checklist

### DIFF
--- a/src/ResourceManager/Websites/ChangeLog.md
+++ b/src/ResourceManager/Websites/ChangeLog.md
@@ -18,7 +18,7 @@
         - Additional information about change #1
 -->
 ## Current Release
-* Added -ClientAffinityEnabled properties for Set-AzureRmWebApp
+* Added -ClientAffinityEnabled property for Set-AzureRmWebApp
 
 ## Version 5.0.1
 * Set minimum dependency of module to PowerShell 5.0


MiYanni:
@jocawtho It looks like:
`TestSetWebApp: Cannot convert 'System.String' to the type 'System.Boolean' required by parameter 'ClientAffinityEnabled'.`

If you wrap the parameter in parenthesis, it should resolve just fine. you should have ran into this locally. Also, you need to re-record the test since you changed it.

jocawtho:
Thanks for this tips! Sorry, just learned how to run these locally. New to PS development. Fixes have been committed and pushed

src/ResourceManager/Websites/Commands.Websites/Cmdlets/WebApps/SetAzureWebApp.cs
```diff
@@ -151,7 +154,8 @@ public override void ExecuteCmdlet()
                             Location = location,
                             ServerFarmId = WebApp.ServerFarmId,
                             Identity = parameters.Contains("AssignIdentity") && AssignIdentity ? new ManagedServiceIdentity("SystemAssigned", null, null) : WebApp.Identity,
-                            HttpsOnly = parameters.Contains("HttpsOnly") ? HttpsOnly : WebApp.HttpsOnly
+                            HttpsOnly = parameters.Contains("HttpsOnly") ? HttpsOnly : WebApp.HttpsOnly,
+                            ClientAffinityEnabled = parameters.Contains("ClientAffinityEnabled") ? ClientAffinityEnabled : WebApp.ClientAffinityEnabled
```
cormacpayne: @jocawtho in the scenario test above, you get the value for `ClientAffinityEnabled` from a web app, but if you hadn't provided a value at all, this line would still get the value from the web app. Is there a scenario where the user would provide the value for `ClientAffinityEnabled` from anywhere but a web app? If not, then we should remove the parameter and just set the value to `WebApp.ClientAffinityEnabled` all the time

src/ResourceManager/Websites/ChangeLog.md
```diff
@@ -18,6 +18,7 @@
         - Additional information about change #1
 -->
 ## Current Release
+* Added -AssignIdentity & -Httpsonly properties for Set-AzureRmWebApp
```
cormacpayne: @jocawtho this change log entry needs to be updated -- it looks like it was copied from one of the previous entries

src/ResourceManager/Websites/Commands.Websites/Cmdlets/WebApps/SetAzureWebApp.cs
jocawtho: Customer's should be able to set this setting without getting the setting from a webApp, but I chose to get the bool value from a webApp property to ensure the test's accuracy by setting the opposite value no matter what the default may be.

panchagnula:
@jocawtho we have an item to update the set commadlet  for apps & slots to use the  'switch parameter' instead of the bool property. See #6066, we should all new parameters to follow this. So please go -ahead & update your code to use the switch parameter, if you cannot update the rest of the properties i will do so for the rest.

maddieclayton:
Closing as comments have not been addressed in over a week. Please reopen once you are ready to address comments.

PR #6244, opened 2018-05-17
Remove need for profile changes in storage OAuth support

markcowl:
## Checklist

- [x] I have read the [_Submitting Changes_](https://github.com/Azure/azure-powershell/blob/preview/CONTRIBUTING.md#submitting-changes) section of [`CONTRIBUTING.md`](https://github.com/Azure/azure-powershell/blob/preview/CONTRIBUTING.md)
- [x] The title of the PR is clear and informative
- [x] The appropriate [change log has been updated](https://github.com/Azure/azure-powershell/blob/preview/CONTRIBUTING.md#updating-the-change-log)
- [x] The PR does not introduce [breaking changes](https://github.com/Azure/azure-powershell/blob/preview/documentation/breaking-changes/breaking-changes-definition.md)
- [x] If applicable, the changes made in the PR have proper test coverage
- [ ] For public API changes to cmdlets:
    - [ ] the changes have gone through a [cmdlet design review](https://github.com/Azure/azure-powershell-cmdlet-review-pr)
    - [ ] the cmdlet markdown files were [generated using the `platyPS` module](https://github.com/Azure/azure-powershell/blob/preview/documentation/development-docs/help-generation.md)

PR #6210, opened 2018-05-14
[Storage] Support OAuth (moves PR #5960) to AzureRM.Storage.Location

praries880:
This commit moves PR #5960 from the branch Azure.Storage.Oauth to
Azure.Storage.Location which is the branch we will be releasing all the
preview releases off of.

PR #6247, opened 2018-05-17
Fix versioning for the Azure.Storage module, Revert The version bump to AzureRM.Storage module

praries880:
AzureRm.Storage module.
PR #6210 #6244


markcowl:
Closing, as it looks like there is a non-ASCII character in your psd1 change.  Updated PSD1 files in my PR instead.

Issue #4965, opened 2017-11-11
Update telemetry to capture missing data and implement needed changes


markcowl:
### Description

Insights telemetry is not capturing the following data, despite having fields:

- ExceptionType
- ExceptionStackTarce
- InnerExceptionType (shouls contain a list of inner exception types)

We also should add the following data
- InvocationName from MyInvocation (this will have the command actually used)
- ParameteSetName

Verify that the values are sent end-to-end, and are robust over failures (telemetry code should not throw)

### Cost: 4

praries880:
Talked to @twitchax, We don't want invocationname since users could be using a user defined aliasa and that could have uniquely identifying information in there.

[ExceptionType and stack trace](https://github.com/Azure/azure-powershell/blob/2af05ef7a6f271cc8e8fb416d97ea34f79bb8009/src/Common/Commands.Common/MetricHelper.cs#L227) already exist.

praries880:
Update, we are going to support "invocationname"..

PR #6274, opened 2018-05-22
Fix for #4965,

praries880:
Added InvocationName, InnerExceptionTypie and ParameterSetName to the telemetry emitted.

Issue #4965


src/Common/Commands.Common/MetricHelper.cs
```diff
@@ -266,6 +270,12 @@ private void PopulatePropertiesFromQos(AzurePSQoSEvent qos, IDictionary<string,
             eventProperties.Add("HashMacAddress", HashMacAddress);
             eventProperties.Add("PowerShellVersion", PSVersion);
             eventProperties.Add("Version", AzurePowerShell.AssemblyVersion);
+            eventProperties.Add("CommandParameterSetName", qos.ParameterSetName);
+            if (!string.IsNullOrWhiteSpace(qos.InvocationName))
```
cormacpayne: @praries880 why do we need this if-statement? I didn't think it was possible for `InvocationName` to be null or empty

src/ResourceManager/Common/Commands.ResourceManager.Common/AzureRMCmdlet.cs
```diff
@@ -214,6 +219,11 @@ protected override void InitializeQosEvent()
                         s => string.Format(CultureInfo.InvariantCulture, "-{0} ***", s)));
             }
 
+            if (!string.IsNullOrWhiteSpace(this.ParameterSetName))
```
cormacpayne: @praries880 since we don't need to do any additional checks (like we do with `this.MyInvocation` in the if-statements above), is there any reason we can't move the initialization of this property into the instantiation of the `AzurePSQoSEvent` object above?

```cs
_qosEvent = new AzurePSQoSEvent()
{
    CommandName = commandAlias,
    ...
    ParameterSetName = this.ParameterSetName
}
```

Also, I don't believe `ParameterSetName` can ever be null or empty (unless the cmdlet creator gives an empty set name, which doesn't happen in our cmdlets 😀 )

src/Common/Commands.Common/MetricHelper.cs
praries880: well, now I know!!!

src/ResourceManager/Common/Commands.ResourceManager.Common/AzureRMCmdlet.cs
praries880: didnt know that either ;)

src/ResourceManager/Common/Commands.ResourceManager.Common/AzureRMCmdlet.cs
```diff
@@ -204,7 +204,13 @@ protected override void InitializeQosEvent()
                 ClientRequestId = this._clientRequestId,
                 SessionId = _sessionId,
                 IsSuccess = true,
-            };
+                ParameterSetName = this.ParameterSetName
+        };
```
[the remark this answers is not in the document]
praries880: will fix it

src/Common/Commands.Common/MetricHelper.cs
```diff
@@ -226,6 +226,16 @@ private void LogExceptionEvent(AzurePSQoSEvent qos)
...
                 eventProperties.Add("StackTrace", qos.Exception.StackTrace);
                 eventProperties.Add("ExceptionType", qos.Exception.GetType().ToString());
+                Exception innerEx = qos.Exception.InnerException;
+                int exceptionCount = 0;
+                //keep goin till we get to the last inner exception
+                while (innerEx != null)
+                {
+                    //Increment the inner exception count so that we can tell which is the outermost
+                    //and which the innermost
+                    eventProperties.Add("InnerExceptionType-"+exceptionCount++, innerEx.GetType().ToString());
```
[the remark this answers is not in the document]
praries880: @cormacpayne I had initially missed this piece in the linked issue :"InnerExceptionType (shouls contain a list of inner exception types)"

Hence I am adding all the inner exceptions. I can create a comma separated list of exceptions and add it to the dictionary, but I thought it might be better to return each one as a separate item in the dictionary.

src/Common/Commands.Common/MetricHelper.cs
```diff
@@ -226,6 +226,16 @@ private void LogExceptionEvent(AzurePSQoSEvent qos)
                 eventProperties.Add("Message", "Message removed due to PII.");
                 eventProperties.Add("StackTrace", qos.Exception.StackTrace);
                 eventProperties.Add("ExceptionType", qos.Exception.GetType().ToString());
+                Exception innerEx = qos.Exception.InnerException;
```
twitchax: I forget, how do we feel about `var`?

src/ResourceManager/Common/Commands.ResourceManager.Common/AzureRMCmdlet.cs
```diff
@@ -204,8 +204,14 @@ protected override void InitializeQosEvent()
...
                 SessionId = _sessionId,
                 IsSuccess = true,
+                ParameterSetName = this.ParameterSetName
             };
 
+            if (this.MyInvocation != null && !string.IsNullOrWhiteSpace(this.MyInvocation.InvocationName))
+            {
+                _qosEvent.InvocationName = this.MyInvocation.InvocationName;
+            }
+
```
twitchax: Why not replace this whole thing (if-statement and all) with:

```
_qosEvent.InvocationName = this.MyInvocation?.InvocationName;
```

Or, can `this.MyInvocation.InvocationName` actually be whitespace?

src/Common/Commands.Common/MetricHelper.cs
praries880: not sure :)
Having said that... i would use a var when dealing with anonymous types.. in this case i odnt foresee it changeing from Exception, so i explicitly use the type.

praries880:
@cormacpayne  can you kindly sign off again, took care of a merge conflict.

praries880:
on demand : https://azuresdkci.westus2.cloudapp.azure.com/job/powershell-demand/897/

src/ResourceManager/Common/Commands.ResourceManager.Common/AzureRMCmdlet.cs
cormacpayne: @twitchax I don't believe that `this.MyInvocation.InvocationName` can be whitespace, but it never hurts to make the check 😀

Issue #6521, opened 2018-06-22
Get-AzureRmSubscription repeats subscription per tenantid when using AccessToken connection


tylerd:
### Description 

Executing `Get-AzureRmSubscription` results in repeat subscriptions per Tenant.

Interestingly, the Subscriptions listed are only under the Microsoft Tenant. There are other subscriptions under the other Tenants that are not listed (even though the TenantIds are listed)

### Script/Steps for Reproduction 

Generate Access Token to access Azure Management API.
Get all Subscriptions

```powershell
Import-Module AzureRM -Global
Import-Module Microsoft.ADAL.PowerShell -Global

$azEnv = Get-AzureRmEnvironment -Name "AzureCloud"
$clientId = [Microsoft.Azure.Commands.Common.Authentication.AdalConfiguration]::PowerShellClientId
$redirectUri = [Microsoft.Azure.Commands.Common.Authentication.AdalConfiguration]::PowerShellRedirectUri

$token = Get-ADALAccessToken -AuthorityName $MSFT -ClientId $clientId -ResourceId $azEnv.ActiveDirectoryServiceEndpointResourceId -RedirectUri $redirectUri

Connect-AzureRmAccount -AccessToken $token -AccountId $AccountId -TenantId $MSFT

# Done Login

Get-AzureRmSubscription | select Name,TenantId

```

### Module Version

```powershell
C:\Users\tylerd\DesktopGet-Module -Name AzureRM -ListAvailable


    Directory: C:\Users\tylerd\Documents\WindowsPowerShell\Modules


ModuleType Version    Name                                ExportedCommands
---------- -------    ----                                ----------------
Script     6.3.0      AzureRM
Script     6.2.0      AzureRM


    Directory: C:\Program Files\WindowsPowerShell\Modules


ModuleType Version    Name                                ExportedCommands
---------- -------    ----                                ----------------
Script     5.7.0      AzureRM
```

### Environment Data

```powershell
C:\Users\tylerd\Desktop$PSVersionTable

Name                           Value
----                           -----
PSVersion                      5.1.17134.112
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.17134.112
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

### Debug Output

Full Debug Output Attached:
[...]

Issue #6489, opened 2018-06-19
Change default context name to avoid key clashes


markcowl:
The default context key name uses just the subscription ID as a key - change this to use the subscription name and a hash of subscription Id, Account, and Tenant, so that it will be unique.

markcowl:
Fixed.

PR #6621, opened 2018-07-06
Fixing Issue #6521 - Duplicate subscription results per tenant when using Access Token

tylerd:
## Description

If the Get-RmSubscription command is executed or anything executes ListSubscriptions without including a Tenant, and the current context account is an Access Token account … then the TenantId associated with the current context is used. This prevents duplicate results when the same Access Token is reused for all of the tenants

## Checklist

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -161,6 +161,7 @@
 
 #### AzureRM.Profile
 * Ps1Xml attribute added to the basic output types
+* Fixed issue in Get-AzureRmSubscription where multiple sets of subscriptions returned per AAD Tenant
 
 #### AzureRM.Compute
 * IP Tag feature for VMSS


cormacpayne:
@tylerd Hey Tyler, what is the full set of commands that need to be run to reproduce the original error? This will help us in coming up with a test to cover this scenario

maddieclayton:
Closing as above question has not been answered. @tylerd Feel free to reopen this PR and answer Cormac's question if you are still running into this issue.

tylerd:
Issue #6521 has information on the issue.
I have not retested the root issue in the latest 6.5.0 update. Will re-test today.

markcowl:
@tylerd So, I don't think this is the root cause of the issue, although it will change this particular behavior in this narrow circumstance.  I believe the real issue is this: #6489 Which is due to be fixed in the next release.  We will verify this particular scenario when this is fixed.

Thank you for caring about this enough to make a contribution, and we will ensure that this particular scenario is fixed when we do the underlyign fix.

tylerd:
Debug output using AzureRM Module 6.5.0
[psoutput1.txt](https://github.com/Azure/azure-powershell/files/2236193/psoutput1.txt)

tylerd:
You can see in the Debug output that the code makes the same request multiple times, once for each Tenant that the current user is a part of … even though the Access Token has been generated from a single Tenant

markcowl:
@tylerd Got it - so how did you generate the token in this case?  The intended usage is that you provide the token for a single tenant and provide the tenant ID - in this case, if the tenantId were required, I think the problem would be resolved.  We want to discourage users from using the token associated with the Common tenant, for example.

tylerd:
@markcowl - As per the related issue discription, I generate the token using this code

```

$MSFT = "72f988bf-86f1-41af-91ab-2d7cd011db47"
$AccountId = "tylerd@microsoft.com"

Import-Module AzureRM
Import-Module Microsoft.ADAL.PowerShell

$azEnv = Get-AzureRmEnvironment -Name "AzureCloud"
$clientId = [Microsoft.Azure.Commands.Common.Authentication.AdalConfiguration]::PowerShellClientId
$redirectUri = [Microsoft.Azure.Commands.Common.Authentication.AdalConfiguration]::PowerShellRedirectUri

$token = Get-ADALAccessToken -AuthorityName $MSFT -ClientId $clientId -ResourceId $azEnv.ActiveDirectoryServiceEndpointResourceId -RedirectUri $redirectUri

Connect-AzureRmAccount -AccessToken $token -AccountId $AccountId -TenantId $MSFT
```

As you can see, I do specify the Tenant Id, even though it is not required. The specified Tenant Id is "saved" in the stored context somewhere, however on any subsequent cmdlets (Get-AzureRmSubscription) that TenantId scope is not respected.

Also note: If I call `Get-AzureRmSubscription -TenantId "<guid>"` I do not get repeated results. So it would be great if the TenantId used when calling `Connect-AzureRmAccount` was scoped for the entire session.

maddieclayton:
Closing as this issues will need to be updated in the azure-powershell-common repo rather than the azure-powershell repo.  Updated the issue as well.

Issue #1340, opened 2015-11-23
Set-AzureStorageBlobContent requires READ instead of only WRITE


GitHubBJW:
A few versions back (like 0.95) I was able to create a context using New-AzureStorageContext and a SAS token that used access policies.  I only needed to grant WRITE to the blob container access policy and Set-AzureStorageBlobContent worked fine.  Now it seems it requires READ and WRITE in order to write blob to a container!

$Context             = New-AzureStorageContext -StorageAccountName $AccountName -SasToken $SAS -Protocol https
# Used to work when WRITE was granted, but now requires READ & WRITE

Set-AzureStorageBlobContent -Container $Container -File $Folder\$File -Blob $File -Context $Context -Force

The impact of this, is that we used to have a secure way to accept uploads.  But with READ being enabled, this means those files are now available for anyone to download!

EmmaZhu:
@GitHubBJW 
The cmdlet needs to check destination's existence, which requires READ permission, to avoid overwrite anything without noticed. This is a bit inconvenient if destination doesn't exist or you always want to overwrite destination. We'll open a work item to find a way to suppress this checking.

dtrefilo-Quest:
The following snippet can be used to upload a file to blob without READ access.
`$context.StorageAccount.CreateCloudBlobClient().GetContainerReference($ContainerName).GetBlockBlobReference($item.Name).UploadFromFile($item.FullName, 'Open')`

cormacpayne:
@blueww Hey Wei, do you know if this issue is still active?

blueww:
@cormacpayne 
Yes, this issue is still active.
I will find some time try to fix it.

blueww:
This is already fixed in the latest Azure PSH build release 6.5.0. So close the issue.
https://github.com/Azure/azure-powershell/releases

PR #6626, opened 2018-07-09
[Storage] Fix issue that can't upload blob/file with write only SAS

blueww:
## Description

This is to fix the issue https://github.com/Azure/azure-powershell/issues/1340 
Modify by:
1. when use -Force, will use "TransferContext.ForceOverwrite" in DMlib transfer context, so DMlib won't check dest exist 
2. When Powershell upload blob/file cmdlet code try to read dest (to check exist, or print output), Don't report error when return 403 since it can be caused by write only sas.

## Checklist

### DIFF
--- a/src/Storage/Commands.Storage/File/Cmdlet/SetAzureStorageFileContent.cs
+++ b/src/Storage/Commands.Storage/File/Cmdlet/SetAzureStorageFileContent.cs
@@ -164,13 +164,24 @@ namespace Microsoft.WindowsAzure.Commands.Storage.File.Cmdlet
             catch (StorageException e)
             {
                 if (e.RequestInformation != null &&
-                    e.RequestInformation.HttpStatusCode == (int)HttpStatusCode.BadRequest &&
-                    e.RequestInformation.ExtendedErrorInformation == null)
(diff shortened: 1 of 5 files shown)


src/Storage/Commands.Storage/File/Cmdlet/SetAzureStorageFileContent.cs
```diff
@@ -163,14 +163,24 @@ public override void ExecuteCmdlet()
             }
             catch (StorageException e)
             {
-                if (e.RequestInformation != null &&
-                    e.RequestInformation.HttpStatusCode == (int)HttpStatusCode.BadRequest &&
-                    e.RequestInformation.ExtendedErrorInformation == null)
+                if (e.RequestInformation.HttpStatusCode == (int)HttpStatusCode.Forbidden)
```
cormacpayne: @blueww we should do the same null-check here for `e.RequestInformation`. Also, do we need to check the value of `e.RequestInformation.ExtendedErrorInformation`?

blueww: Add null-check for e.RequestInformation.
Since we don't print e.RequestInformation.ExtendedErrorInformation here, don't need to check the value of it.

PR #6642, opened 2018-07-11
Disable-AzureRmVmDiskEncryption Update

ejarvi:
## Description

Update Disable-AzureRmVmDiskEncryption to work for VMs encrypted without AAD parameters

## Checklist

- [ n/a] For public API changes to cmdlets:
    - [n/a] the changes have gone through a [cmdlet design review](https://github.com/Azure/azure-powershell-cmdlet-review-pr)
    - [n/a] the cmdlet markdown files were [generated using the `platyPS` module](https://github.com/Azure/azure-powershell/blob/preview/documentation/development-docs/help-generation.md)

### DIFF
--- a/src/ResourceManager/Compute/Commands.Compute/Extension/AzureDiskEncryption/DisableAzureDiskEncryption.cs
+++ b/src/ResourceManager/Compute/Commands.Compute/Extension/AzureDiskEncryption/DisableAzureDiskEncryption.cs
@@ -125,23 +125,12 @@ namespace Microsoft.Azure.Commands.Compute.Extension.AzureDiskEncryption
             return null;
         }
 
-        private bool IsDiskEncryptionSettingPresent(VirtualMachine vm)
+        private bool IsVmModelEncryptionSet(VirtualMachine vm)
         {
-            bool isPresent = false; 
-            if (vm != null && 
-                vm.InstanceView != null &&
-                vm.InstanceView.Disks != null &&
-                vm.InstanceView.Disks.Count > 0)
-            {
-                foreach ( DiskInstanceView d in vm.InstanceView.Disks )
-                {
-                    if (d.EncryptionSettings != null && d.EncryptionSettings.Count>0)
-                    {
-                        isPresent = true;
-                    }
-                }
-            }
-            return isPresent;
+            return (vm != null &&
+                vm.StorageProfile != null &&
+                vm.StorageProfile.OsDisk != null &&
+                vm.StorageProfile.OsDisk.EncryptionSettings != null);
         }
 
         private string GetVersionForDisable(VirtualMachine vm)
@@ -160,34 +149,33 @@ namespace Microsoft.Azure.Commands.Compute.Extension.AzureDiskEncryption
                 }
             }
 
-            // If we reach this point, there was no extension currently installed, so we begin a process of identifying
-            // which version of the extension had been installed previously. 
+            // If we reach this point, no extension is currently installed, even if the VM is encrypted.
+            // To disable encryption, we will select the extension version matching the encryption 
+            // settings present on the VM and then use that version to issue the disable operation. 
 
-            // This supports a corner case when the extension is removed from teh VM first without first disabling it, and 
-            // only later deciding to disable encryption on the VM by adding the extension back to disable it.  
-            // The disable operation must be done with a version that matches the version that was used at time of enable.
-            // So we will derive which strategy (single or dual) encrypted the VM and then return the corresponding version. 
-
-            // If per disk encryption settings are found, use the single pass version to disable:
-            if (IsDiskEncryptionSettingPresent(vm))
+            if (IsVmModelEncryptionSet(vm))
             {
-                switch (currentOSType)
+                // encryption settings present in VM model, use default dual pass version
+                if (currentOSType == OperatingSystemTypes.Linux)
                 {
-                    case OperatingSystemTypes.Windows:
-                        return AzureDiskEncryptionExtensionContext.ExtensionSinglePassVersion;
-                    case OperatingSystemTypes.Linux:
-                        return AzureDiskEncryptionExtensionContext.LinuxExtensionSinglePassVersion;
+                    return AzureDiskEncryptionExtensionContext.LinuxExtensionDefaultVersion;
                 }
-            }
-
-            // Otherwise, retain consistency and use the default dual pass version:
-            if (currentOSType == OperatingSystemTypes.Linux)
-            {
-                return AzureDiskEncryptionExtensionContext.LinuxExtensionDefaultVersion;
+                else
+                {
+                    return AzureDiskEncryptionExtensionContext.ExtensionDefaultVersion;
+                }
             }
             else
             {
-                return AzureDiskEncryptionExtensionContext.ExtensionDefaultVersion;
+                // encryption settings not present in the VM model, use single pass version
+                if (currentOSType == OperatingSystemTypes.Linux)
+                {
+                    return AzureDiskEncryptionExtensionContext.LinuxExtensionSinglePassVersion;
+                }
+                else
+                {
+                    return AzureDiskEncryptionExtensionContext.ExtensionSinglePassVersion;
+                }
             }
         }
 


src/ResourceManager/Compute/Commands.Compute.Test/ScenarioTests/VirtualMachineExtensionTests.cs
```diff
@@ -122,6 +122,13 @@ public void TestAzureDiskEncryptionExtensionSinglePass()
             ComputeTestController.NewInstance.RunPsTest("Test-AzureDiskEncryptionExtensionSinglePass");
         }
 
+        [Fact(Skip = "TODO: only works for live mode due to key vault dependency")]
```
cormacpayne: @ejarvi just for our understanding, is there a reason that we can't create a KeyVault client in the test controller that allows you to use the KeyVault cmdlets in playback/record mode?

ejarvi: Yes thanks for asking.  I would really like to get this enabled.  Once I have three different tests that I should be able to flip multiple tests to check-ins.  I spent time last night trying to get this to work locally yesterday but was unsuccessful.   After updating the package references to use the 4.2.1-alpha version of Key Vault so that it could find dependency references, I got hung up on an exception indicating the KeyVaultClient constructor was not being called in the correct format.  This was from within test controller code I'm not yet familiar with and apparently across a package boundary. If I recall correctly the exception surfaced from Microsoft.Azure.Common.Compute.Authentication.dll.   I haven't had time to drill deeper yet.

ejarvi: taking another stab at this now...

cormacpayne: @ejarvi okay thanks for the information. @maddieclayton and I can work with you in the next sprint to mitigate this issue, and we should be able to use the Service Fabric tests as an example since they are currently using KeyVault cmdlets in their tests

ejarvi:
@cormacpayne  Have you seen this result before in the CI build?  Trying to follow the pattern of other tests but haven't been able to resolve... 

ScenarioTests/VirtualMachineExtensionTests.cs(148,57): error CS1503: Argument 1: cannot convert from 'string' to 'Microsoft.Azure.ServiceManagemenet.Common.Models.XunitTracingInterceptor' [/home/travis/build/Azure/azure-powershell/src/ResourceManager/Compute/Commands.Compute.Test/Commands.Compute.Test.Netcore.csproj]

maddieclayton:
@ejarvi You will need to pull from upstream preview, then add the tracing interceptor to the run test call (you will see examples from the other tests when you pull from upstream).

ejarvi:
@maddieclayton - thank you, merged with upstream preview, updated to new syntax, and checks are passing now 

@cormacpayne - please let me know if anything else needed

src/ResourceManager/Compute/Commands.Compute/Extension/AzureDiskEncryption/DisableAzureDiskEncryption.cs
```diff
@@ -143,6 +143,34 @@ private VirtualMachineExtension GetVmExtensionParameters(VirtualMachine vmParame
...
+                    {
+                        installedTypeHandlerVersion = vme.TypeHandlerVersion;
+                        break;
+                    }
+                }
+            }
+
+            if (installedTypeHandlerVersion == null)
+            {
+                ThrowTerminatingError(
```
SudhakaraReddyEvuri: This adds a new failure condition where user won't be able to run 'Disable-AzureRmVmDiskEncryptionExtension' cmdlet on a VM that wasn't encrypted before . Or on a VM that was encrypted before but extension was removed (with encryption enabled).

src/ResourceManager/Compute/Commands.Compute/Extension/AzureDiskEncryption/DisableAzureDiskEncryption.cs
```diff
@@ -286,9 +314,9 @@ public override void ExecuteCmdlet()
                     // | Linux   | Not Specified | N/A                        |
                     // +---------+---------------+----------------------------+
 
-                    if (OperatingSystemTypes.Windows.Equals(currentOSType) &&
-                        !string.IsNullOrEmpty(VolumeType) &&
-                        VolumeType.Equals(AzureDiskEncryptionExtensionContext.VolumeTypeData, StringComparison.InvariantCultureIgnoreCase))
+                    if ((OperatingSystemTypes.Windows.Equals(currentOSType) && parameters.TypeHandlerVersion.Equals(AzureDiskEncryptionExtensionContext.ExtensionSinglePassVersion)) ||
```
SudhakaraReddyEvuri: This will cause UpdateVmEncryptionSettings to be called for Windows VMs + Two pass encryption + Data volume type case. Which shouldn't be.

ejarvi: It still skips UpdateVmEncryptionSettings for DualPass-Windows-DataVolume, but adds a skip for SinglePass-Windows and SinglePass-Linux.

src/ResourceManager/Compute/Commands.Compute/Extension/AzureDiskEncryption/DisableAzureDiskEncryption.cs
ejarvi: Thanks, good catch.  Didn't realize this was a supported use case, will revert this portion of the logic to previous behavior.

ejarvi: @SudhakaraReddyEvuri added support for this case, please review new changes.

ejarvi:
@SudhakaraReddyEvuri  - incorporated changes from your review 
@cormacpayne - assigning back to you, please reconsider for approval

src/ResourceManager/Compute/Commands.Compute/Extension/AzureDiskEncryption/DisableAzureDiskEncryption.cs
```diff
@@ -125,6 +125,72 @@ private Hashtable GetExtensionProtectedSettings()
...
+        }
+
+        private string GetVersionForDisable(VirtualMachine vm)
+        {
+            // if there is currently an extension installed, use that version for disable as well 
+            if (vm.Resources != null)
+            {
+                foreach (VirtualMachineExtension vme in vm.Resources)
+                {
+                    if (vme.Publisher.Equals(AzureDiskEncryptionExtensionContext.ExtensionDefaultPublisher) &&
```
SudhakaraReddyEvuri: Please remove the version check here. By doing just publisher check, you will use the existing version.

src/ResourceManager/Compute/Commands.Compute/Extension/AzureDiskEncryption/DisableAzureDiskEncryption.cs
```diff
@@ -125,6 +125,72 @@ private Hashtable GetExtensionProtectedSettings()
             return null;
         }
 
+        private bool IsDiskEncryptionSettingPresent(VirtualMachine vm)
```
SudhakaraReddyEvuri: I would suggest to just check encryption settings in VM model instead of looping through disks etc. In the caller, you can use two pass encryption if VM model has encryption settings otherwise use single pass extension

SudhakaraReddyEvuri: I don't have further comments. Once this is changed, consider approved from my side.

ejarvi: I'll replace the helper method with one that checks VM model, logic flow:   use existing extension version if it exists, if it doesn't exist check if vm model encryption settings exist, if so use dual pass, in all other cases use single pass.

Issue #6473, opened 2018-06-15
New-AzureRmVmss simple param set should fail if a LB with the user specified name exists and is not compatible with what the cmdlet expects


praries880:
This issue comes out of this [issue](https://github.com/Azure/azure-powershell/issues/6413).

The user created an LB which did not have the configuration the cmdlet expected. As a result the cmdlet tried to configure the LB which caused configuration to end up in  bad state and the VMSScreation failed as a result.

We should check if this issue can happen with other resources as well that are created by the cmdlet.

praries880:
PR https://github.com/Azure/azure-powershell/pull/6665 fixes part one... we will need another set of changes to validate the config of the existing LB to satisfy the second part of the issues requirements.

PR #6665, opened 2018-07-13
6473 fix

praries880:
Fix for issue #6473 

Add logic to validate configs of resources that already exist for the simple param set use cases.
As of now only LD logic implements the check and that too returns false if a LB with a user provided name exists.

If an LB with the default name chosen by the cmdlet exists the cmdlet does not fail.

In a future checkin we will implement an actual deep comparison of the existing LB's config with the target config.

### DIFF
--- a/src/ResourceManager/Compute/Commands.Compute/Manual/VirtualMachineScaleSetCreateOrUpdateMethod.cs
+++ b/src/ResourceManager/Compute/Commands.Compute/Manual/VirtualMachineScaleSetCreateOrUpdateMethod.cs
@@ -278,8 +278,9 @@ namespace Microsoft.Azure.Commands.Compute.Automation
 
             var parameters = new Parameters(this, client);
 
-            // If the user did not specify a load balancer name, As in we are using a
-            // just
+            // If the user did not specify a load balancer name, mark the LB setting to ignore
+            // preexisting check. The most common scenario is users will let the cmdlet create and name the LB for them with the default
+            // config. We do not want to block that scenario in case the cmdlet failed mid operation and tthe user kicks it off again.
             if (!loadBalancerNamePassedIn)
             {
                 LoadBalancerStrategy.IgnorePreExistingConfigCheck = true;


praries880:
@sergey-shandar Reason for `do not merge` : I am working on the test changes, I have manually validated the changes.

Will update the PR with the tests once i am done with those.

src/ResourceManager/Common/Commands.Common.Strategies/StateExtensions.cs
```diff
@@ -70,7 +70,19 @@ sealed class ContainsDispatchVisitor : IEntityConfigVisitor<IState, bool>
         {
             public bool Visit<TModel>(ResourceConfig<TModel> config, IState context)
                 where TModel : class
-                => context.Contains(config);
+            {
+                TModel currendEntityConfig = context.Get(config);
```
sergey-shandar: This's a generic algorithm. It supposes only check if the config exist. It shouldn't validate anything else. I would propose to create a separate algorithm with different name.

praries880: @sergey-shandar The implementation of the validation logic is generic as well, its the particular resource that can choose to override the validation logic, the default is true for all resources. As of now only the LB resource that implements the validation.

The purpose of the check is to see if the existing config is compatible with what we are targeting. The code path here is such that it have both... Infact, the existing logic checks if the curent config exists. This should be a great place to add the validate... it a logical extension of the check to see if a config already exists to add the logic to now see if it is compatible with the target config for the resource. Do let me know why you think the current changes are not generic...

cormacpayne:
@praries880 to get the build to pass, you'll need to update both the Compute and Profile change logs

maddieclayton:
@praries880 When you're ready to merge this, please remove the "Do Not Merge" Tag

praries880:
on demand : https://azuresdkci.westus2.cloudapp.azure.com/job/powershell-demand/1004/

src/ResourceManager/Compute/Commands.Compute/Manual/VirtualMachineScaleSetCreateOrUpdateMethod.cs
```diff
@@ -275,8 +278,20 @@ public string Location
 
             var parameters = new Parameters(this, client);
 
+            // If the user did not specify a load balancer name, As in we are using a
+            // just
+            if (!loadBalancerNamePassedIn)
```
cormacpayne: @praries880 nit: we can consolidate all of the changes made in this file into one line:

```cs
LoadBalancerStrategy.IgnorePreExistingConfigCheck = String.IsNullOrWhitespace(LoadBalancerName);
```

Also, would you mind updating the comment above this line since it appears to be incomplete

praries880: @sergey-shandar It behaves correctly, I have validated it multiple times on my desktop and while running the tests.

Issue #6673, opened 2018-07-16
Enabl AccelNet on capable VMs by default in PowerShell


praries880:
Changes similar to [the ones in CLI](https://github.com/Azure/azure-cli/blob/dev/src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_validators.py#L562)

Enable accelerated Networking on thw VMs nic if it meets the criteria. Check the changes linked to in CLI for the conditions.

Aggreed to changes : 
`
The strategy cmdlet (simple param sets) does not take in a configuration for a Network Interface, instead it creates one at runtime and currently the cmdlet has no flag to enable AccelNet.
 
As per my understanding, we are not adding a flag, the cmdlet will look at the passed in values for the SKU and image type to make the call. Also, the config will be set to false for vms created using custom images and disk files.
`

PR #6699, opened 2018-07-18
Update new-azurermvm simple param set to enable auto configuration of Accelerated Network for qualifying vms

praries880:
Fix for issue #6673

### DIFF
--- a/src/ResourceManager/Compute/Commands.Compute/Common/Utils.cs
+++ b/src/ResourceManager/Compute/Commands.Compute/Common/Utils.cs
@@ -1,4 +1,18 @@
-﻿using Microsoft.Azure.Commands.Compute.Strategies;
+// ----------------------------------------------------------------------------------
+//
+// Copyright Microsoft Corporation
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// ----------------------------------------------------------------------------------
+
+using Microsoft.Azure.Commands.Compute.Strategies;
 using Microsoft.Azure.Commands.Compute.Strategies.ComputeRp;
 using Microsoft.Azure.Management.Compute;
 using System;


praries880:
On Demand : https://azuresdkci.westus2.cloudapp.azure.com/job/powershell-demand/1005/

src/ResourceManager/Compute/Commands.Compute/Common/Utils.cs
```diff
@@ -0,0 +1,151 @@
...
+            public ImageTypeForAccelNet(string publisher, string offer, string skuPattern)
+            {
+                Publisher = publisher;
+                Offer = offer;
+                SkuPattern = skuPattern;
+            }
+
+            public bool Matches(string publisher, string offer, string sku)
+            {
+                return (Publisher.Equals(publisher) && (offer == null || offer.Equals(Offer)) && (sku == null || SkuPattern == null || (new Regex(SkuPattern)).IsMatch(sku)));
```
cormacpayne: @praries880 do we need to do a null-check on `publisher` and/or `Publisher` here? It seems we do so for the other parameters

src/ResourceManager/Compute/Commands.Compute/Common/Utils.cs
```diff
@@ -0,0 +1,151 @@
...
+
+            var sizeFound = ValidSizesForAccelNet.Where(x => x.ToLower().Equals(size.ToLower()));
+            if (sizeFound == null || sizeFound.Count() <= 0)
+            {
+                return false;
+            }
+
+            var fourCoreSize = FourCoreSizesForAccelNet.Where(x => x.ToLower().Equals(size.ToLower()));
+
+            if (fourCoreSize == null | fourCoreSize.Count() == 0)
```
cormacpayne: @praries880 should the operator here be `||` ?

src/ResourceManager/Compute/Commands.Compute/Common/Utils.cs
```diff
@@ -0,0 +1,151 @@
...
+            {
+                var locationToUse = location;
```
cormacpayne: @praries880 nit: we can consolidate the next few lines into one:

```cs
var locationToUse = String.IsNullOrWhitespace(location) ? defaultLocation : location;
```

src/ResourceManager/Compute/Commands.Compute/Common/Utils.cs
praries880: We don't need to, we validate that the publisher is not null at the top of the function.

src/ResourceManager/Compute/Commands.Compute/Common/Utils.cs
praries880: :) why.. yes.. yes it does

praries880:
@cormacpayne  took care of review comment

src/ResourceManager/Compute/Commands.Compute/Common/Utils.cs
sergey-shandar: @cormacpayne @praries880 we need Nullable Reference Types (from C# 8.0) https://blogs.msdn.microsoft.com/dotnet/2017/11/15/nullable-reference-types-in-csharp/

PR #6927, opened 2018-08-14
Enable/Disable-AzureRmAlias cmdlet

maddieclayton:
## Checklist


maddieclayton:
netcore sign job here: https://azuresdkci.westus2.cloudapp.azure.com/job/psnetsign/118/

maddieclayton:
netcore sign job here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/psnetsign/119/

src/ResourceManager/Profile/Commands.Profile/AzureRmAlias/DisableAzureRmAlias.cs
```diff
@@ -0,0 +1,188 @@
...
+            {
+                if (File.Exists(userprofile.ToString()))
+                {
+                    List<string> modulesToKeep = new List<String>();
+                    string filecontent = "";
+                    string originalText = File.ReadAllText(userprofile.ToString());
+                    if (originalText.Contains("#Begin Azure PowerShell alias import"))
+                    {
+                        // Add back profile code unrelated to Azure PowerShell aliases
+                        var splitOriginalText = originalText.Split(new string[] { "#Begin Azure PowerShell alias import", "#End Azure PowerShell alias import" }, StringSplitOptions.None);
```
markcowl: let's make these marker comments as constants, so we can reuse them, and would only need to change then in one place

src/ResourceManager/Profile/Commands.Profile/AzureRmAlias/DisableAzureRmAlias.cs
```diff
@@ -0,0 +1,188 @@
...
+                                }
+                                match = match.NextMatch();
+                            }
+                        }
+                    }
+
+                    // Add script to enable aliases to profile if there are any modules to keep
+                    if (modulesToKeep.Count > 0)
+                    {
+                        filecontent = "#Begin Azure PowerShell alias import" + Environment.NewLine + "`$error.clear()" + Environment.NewLine +
```
markcowl: It seems like we should factor out the manipulation of the file lists into a helper class, as a lot of this should be shared with the other cmdlet

src/ResourceManager/Profile/Commands.Profile/AzureRmAlias/DisableAzureRmAlias.cs
```diff
@@ -0,0 +1,188 @@
...
+                            "Import-Module Az.Profile -ErrorAction SilentlyContinue" + Environment.NewLine + "if (!`$error) { " + Environment.NewLine;
+
```
markcowl: Here, I think we are in trouble if $error is already non-null.  I Hink instead we can define a variable $loadError, and pass it into -ErrorVariable, and use our error variable to determine if there were any errors specifically caused by the cmdlet.

Alternately, you could srround the entire thing in a try/catch and use -ErrorAction Stop

src/ResourceManager/Profile/Commands.Profile/AzureRmAlias/EnableAzureRmAlias.cs
```diff
@@ -0,0 +1,201 @@
...
+
+        [Parameter(Mandatory = false, HelpMessage = "If specified, cmdlet will return all aliases enabled")]
+        public SwitchParameter PassThru { get; set; }
+
+        public override void ExecuteCmdlet()
+        {
+            if (ShouldProcess(Scope, "Add aliases"))
+            {
+                string jsonmapping = Mappings.jsonMappings;
+                Dictionary<string, object> caseSensitiveMapping = (Dictionary<string, object>)JsonConvert.DeserializeObject(jsonmapping, typeof(Dictionary<string, object>));
```
markcowl: Also, the code that reads from the mappings and produces a case-insensitve mapping should be common code.

src/ResourceManager/Profile/Commands.Profile/AzureRmAlias/EnableAzureRmAlias.cs
```diff
@@ -0,0 +1,201 @@
...
+                            WriteWarning("Module '" + module + "' is not a valid Az module.");
+                        }
+                    }
+                }
+
+                // Set path to user profile based on the Scope
+                object userprofile = "";
+                if (Scope != null && Scope.Equals("CurrentUser"))
+                {
+                    userprofile = SessionState.PSVariable.GetValue("env:USERPROFILE") + "\\Documents\\PowerShell\\profile.ps1";
```
markcowl: Use Path.Combine to deternmine paths

src/ResourceManager/Profile/Commands.Profile/AzureRmAlias/DisableAzureRmAlias.cs
```diff
@@ -0,0 +1,188 @@
...
+                        WriteWarning("Module '" + module + "' is not a valid Az module.");
+                    }
+                }
+            }
+
+            // Set path to user profile based on the Scope
+            object userprofile = "";
+            if (Scope != null && Scope.Equals("CurrentUser"))
+            {
+                userprofile = SessionState.PSVariable.GetValue("env:USERPROFILE") + "\\Documents\\PowerShell\\profile.ps1";
```
markcowl: Use path.combine to determien paths

src/ResourceManager/Profile/Commands.Profile/AzureRmAlias/EnableAzureRmAlias.cs
```diff
@@ -0,0 +1,201 @@
...
+
+        [Parameter(Mandatory = false, HelpMessage = "Indicates which modules to enable aliases for. If none are specified, default is all modules.")]
+        public string[] Module { get; set; }
+
+        [Parameter(Mandatory = false, HelpMessage = "If specified, cmdlet will return all aliases enabled")]
+        public SwitchParameter PassThru { get; set; }
+
+        public override void ExecuteCmdlet()
+        {
+            if (ShouldProcess(Scope, "Add aliases"))
```
markcowl: These should be in resources, as they are displayed to the customer

src/ResourceManager/Profile/Commands.Profile/AzureRmAlias/EnableAzureRmAlias.cs
```diff
@@ -0,0 +1,201 @@
...
+                }
+
+                // If no modules are specified, enable all aliases
+                if (Module == null)
+                {
+                    foreach (var key in mapping.Keys)
+                    {
+                        Dictionary<string, string> modulemapping = (Dictionary<string, string>)JsonConvert.DeserializeObject(mapping[key].ToString(), typeof(Dictionary<string, string>));
+
+                        foreach (var name in modulemapping.Keys)
```
markcowl: I think we wnat to filter the module list by installed modules - we could add this functionality after check-in, however.

src/ResourceManager/Profile/Commands.Profile/AzureRmAlias/EnableAzureRmAlias.cs
markcowl: It would be nicer if the ShouldProcess was once per module

src/ResourceManager/Profile/Commands.Profile/AzureRmAlias/DisableAzureRmAlias.cs
```diff
@@ -0,0 +1,188 @@
...
+
+        [Parameter(Mandatory = false, HelpMessage = "Indicates which modules to disable aliases for. If none are specified, default is all enabled modules.")]
+        public string[] Module { get; set; }
+
+        [Parameter(Mandatory = false, HelpMessage = "If specified, cmdlet will return all disabled aliases")]
+        public SwitchParameter PassThru { get; set; }
+
+        public override void ExecuteCmdlet()
+        {
+            string jsonmapping = Mappings.jsonMappings;
```
markcowl: I think if we factor this out into common mappings code, and factor the profile manipulation into a helper class, it will be easier to mock and unit test this cmdlet

maddieclayton:
Sign job for Netcore here: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/psnetsign/121/

PR #7108, opened 2018-08-31
Update Get-AzureRmStorageAccount.md to reflect that ResourceGroupName is required

blueelvis:
Tried with the latest Azure Powershell and `ResourceGroupName` is a required parameter. Otherwise, it will not work. The storage account actually exists as well.

```
PS C:\Users\Pranav> Get-AzureRmStorageAccount -name "pranavdevtest"

cmdlet Get-AzureRmStorageAccount at command pipeline position 1
Supply values for the following parameters:
(Type !? for Help.)
ResourceGroupName:
Get-AzureRmStorageAccount : Cannot validate argument on parameter 'ResourceGroupName'. The argument is null or empty. Provide an argument that is not null or empty, and then try the command
again.
At line:1 char:1
+ Get-AzureRmStorageAccount -name "pranavdevtest"
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidData: (:) [Get-AzureRmStorageAccount], ParameterBindingValidationException
    + FullyQualifiedErrorId : ParameterArgumentValidationError,Microsoft.Azure.Commands.Management.Storage.GetAzureStorageAccountCommand
```

Thanks,
Pranav

## Checklist

### DIFF
--- a/src/ResourceManager/Storage/Commands.Management.Storage/help/Get-AzureRmStorageAccount.md
+++ b/src/ResourceManager/Storage/Commands.Management.Storage/help/Get-AzureRmStorageAccount.md
@@ -91,7 +91,7 @@ Type: System.String
 Parameter Sets: ResourceGroupParameterSet
 Aliases:
 
-Required: False
+Required: True
 Position: 0
 Default value: None
 Accept pipeline input: True (ByPropertyName)


blueww:
@blueelvis 

The Resource Group name is only required, when you add storage account name, and in this case it's in parameter set "AccountNameParameterSet". And in "AccountNameParameterSet" in help , you can see the Resource Group name is already required.

The place you changed is in "ResourceGroupParameterSet ", which means don't give storage account name. In this case, when give Resouce Group name, will list account in this resource group name; when don't give resource group name, will list all account in the subscription.


So you change is not necessary, and should not be merge.



> -ResourceGroupName
> Specifies the name of the resource group that contains the Storage account to get.

> Type: System.String
> Parameter Sets: ResourceGroupParameterSet
> Aliases:
> 
> Required: False
> Position: 0
> Default value: None
> Accept pipeline input: True (ByPropertyName)
> Accept wildcard characters: False
> Type: System.String

> Parameter Sets: AccountNameParameterSet
> Aliases:
> 
> Required: True
> Position: 0
> Default value: None
> Accept pipeline input: True (ByPropertyName)
> Accept wildcard characters: False

src/ResourceManager/Storage/Commands.Management.Storage/help/Get-AzureRmStorageAccount.md
```diff
@@ -91,7 +91,7 @@ Type: System.String
 Parameter Sets: ResourceGroupParameterSet
 Aliases:
 
-Required: False
```
blueww: The place you changed is in "ResourceGroupParameterSet ", which means don't give storage account name. In this case, when give Resouce Group name, will list account in this resource group ; when don't give resource group name, will list all account in the subscription. So the Resouce Group name is not required, and the change is not right.

blueelvis:
@blueww - Thanks Wei! But, I think that it could be added that it is needed when you are fetching a single storage account. What do you think?

blueww:
This is right "it is needed when you are fetching a single storage account".
But this scenario refer to parameter set "AccountNameParameterSet", which already have resource group name as required in help.

blueelvis:
@blueww - My bad on the confusion. I meant on the docs over here - https://docs.microsoft.com/en-us/powershell/module/azurerm.storage/get-azurermstorageaccount?view=azurermps-6.8.1

PS - I did click on Edit button on top and landed here.

There doesn't seem to be any mention of `ResourceGroupName` as required when trying to fetch a single storage account.

blueww:
@blueelvis 
Thanks for point the doc out!
In the help *.md file you change, it already mention the parameter is required in "AccountNameParameterSet".
https://github.com/Azure/azure-powershell/blob/preview/src/ResourceManager/Storage/Commands.Management.Storage/help/Get-AzureRmStorageAccount.md

But in the online doc, it doesn't show parameter in parameter set, so the Resouce Group is not necessary, since it's not required when account name not supply.

I think a fix is change the description of the parameter,  to indicate it is required when the Strorage Account name is added.

blueelvis:
@blueww - Am I correct to assume that there is a custom render engine at work on that help page when viewed online on the docs website? Because not all of this information seems to be visible. If my assumption is correct, why not display the parameter sets as well? It will surely be helpful!

Would you agree with the following description for the `ResourceGroupName` parameter?
```
Specifies the name of the resource group that contains the Storage account to get. This parameter is **required** when querying for a single storage account.
```

Thanks,
-Pranav

blueww:
@markcowl , @cormacpayne 
Do you have any idea, why the help *.md file show parameter in parameter set, but the online help doc don't show parameter in parameter set. So some required parameter in some parameter are show as not require in online help doc (since it's not required in some paramterset). 
Could we show online help also with parameter set?


@blueelvis 
The help change looks good to me. @TaraMeyer  Do you have any comments for the description of resource group name?
```
Specifies the name of the resource group that contains the Storage account to get. This parameter is **required** when querying for a single storage account.
```

maddieclayton:
@blueelvis @blueww The update to docs to include the parameter set name in the syntax description is one of our top priorities - we will reference this issue when discussing with the docs team.  We also would like to get rid of the "Required Parameters" and the "Optional Parameters", because as you see, some parameters can be mandatory in one set and not in another.  I will close this PR for now, @blueww can you open up a new one if you plan to make the change to ResourceGroupName description?

blueww:
@maddieclayton ， @blueelvis 
If we can show the online help with parameter set, I don't think we need change the description for parameter, since the information already in parameterset. 
Let me know if you have different opinion.

PR #7299, opened 2018-09-19
Analysis Services: sovereign cloud support

pychuang:
## Description

This change enables sovereign cloud support for Analysis Services. It gets AAD and Analysis Services endpoints from AzureEnvironment.

## Checklist

### DIFF
--- a/src/ResourceManager/AnalysisServices/Commands.AnalysisServices.Dataplane/ChangeLog.md
+++ b/src/ResourceManager/AnalysisServices/Commands.AnalysisServices.Dataplane/ChangeLog.md
@@ -18,10 +18,10 @@
         - Additional information about change #1
 -->
 ## Current Release
+* Support sovereign clouds
 
 ## Version 0.5.4
 * Updated to the latest version of the Azure ClientRuntime.
-* Support sovereign clouds
 
 ## Version 0.5.3
 * Updated help files to include full parameter types and correct input/output types.


src/ResourceManager/AnalysisServices/Commands.AnalysisServices.Dataplane/ChangeLog.md
```diff
@@ -21,6 +21,7 @@
 
 ## Version 0.5.4
 * Updated to the latest version of the Azure ClientRuntime.
+* Support sovereign clouds
```
markcowl: This should go under the ```## Current Release``` heading

src/ResourceManager/AnalysisServices/Commands.AnalysisServices.Dataplane/Models/AsAzureClientSession.cs
```diff
@@ -136,14 +119,20 @@ public AsAzureProfile Login(AsAzureContext asAzureContext)
 
         public static string GetAuthorityUrlForEnvironment(AsAzureEnvironment environment)
         {
-            var environmentKey = AsAzureRolloutEnvironmentMapping.Keys.FirstOrDefault(s => environment.Name.Contains(s));
-            AsAzureAuthInfo authInfo = null;
-            if (string.IsNullOrEmpty(environmentKey) || !AsAzureRolloutEnvironmentMapping.TryGetValue(environmentKey, out authInfo))
+            var profile = AzureRmProfileProvider.Instance.GetProfile<AzureRmProfile>();
+            var availableEnvironments = profile.Environments;
```
markcowl: All you should need to do here is get DefaultContext.Environment - this will have the environment the user is currently logged in to.

adxsdkps:

src/ResourceManager/AnalysisServices/Commands.AnalysisServices.Dataplane/Models/AsAzureClientSession.cs
pychuang: Do you mean DefaultContext in AzureRMCmdlet?

Our data plane cmdlets inherit from AzurePSCmdlet and do not depend on Login-AzureRmAccount at all. We have Add-AzureAnalysisServicesAccount to login to a specific rollout (region), not environment (cloud).

If we instead depends on the state of what users do with Login-AzureRmAccount, it's a protocol change to our customers. It will be a really big change and probably needs PM involved.

Please correct me if I misunderstand something.

athipp: >             All you should need to do here is get DefaultContext.Environment - this will have the environment the user is currently logged in to.

This is breaking, because before this, customers who are using these dataplane cmdlets did not have to login to Azure RM. Now we are requiring them to login to RM to get  dataplane related data. @markcowl , is there a way to get this information without having to login. Because in our scenarios, the user who is logging to Azure ASAccount is a server admin (internal to ASAZure) and has nothing to do with the Azure Resource. The user will not have any access to subscription, so Add-AzureRM will fail. So this is actually more than just a breaking change, basically disabling the usecase. :)

maddieclayton:
@azuresdkci Add to whitelist

src/ResourceManager/AnalysisServices/Commands.AnalysisServices.Dataplane/Models/AsAzureClientSession.cs
markcowl: The issue is that you should not need to use Authentication.ResourceManager dll - I believe the only reason you are using it is to get at the environment, but you should be able to do this with:

```
AzureRmProfileProvider.Instance.Profile.DefaultContext.Environment
```

this returns the current environment withotu having to take any dependencies on AzureRMProfile at all, and with this, you should be able to remove this library dependency

athipp: @markcowl , If the user has not logged in to RM, what value will be returned by AzureRmProfileProvider.Instance.Profile.DefaultContext.Environment?

markcowl: Well, then, it seems like your current code won't work either, so I'm not sure what you are asking.  This is precisely the same as using AzureRMProfileProivder.Instance.GetProfile(), which is what you are currently doing, except that it doesn't require the additional assembly dependency.  So if this won't work, your proposed code also won't work

athipp: Yes that I totally agree. @pychuang do we need this PR at all then?

pychuang: Agree. This PR won't work either. We probably need to go back to the original approach.

I will create a PR with original approach, and we can continue the discussion then.

pychuang:
Will create another PR with original approach.

PR #7387, opened 2018-09-26
DeploymentManager: Expand GetRollout output to improve experience

deveshguha:
## Description

The premise of the change is to add a switch parameter to an existing cmdlet to improve user experience when getting information about a rollout. 

The Rollout object, which is the output of the GetRollout cmdlet, is an extremely complex object, with a hierarchy that is 4-5 levels deep. As the rollout progresses, more and more operations are completed and this Rollout object takes a tree form, which becomes very cumbersome. 
[E.g.: $rollout.Services[1].ServiceUnits[1].Steps[0].OperationInfo.Error]

The new switch parameter, -ShowDetails, is required to enhance the discoverability of information in this rollout object, and is used to display the expanded rollout object to the console.

## Checklist

### DIFF
--- a/src/ResourceManager/DeploymentManager/Commands.DeploymentManager/help/Get-AzureRmDeploymentManagerRollout.md
+++ b/src/ResourceManager/DeploymentManager/Commands.DeploymentManager/help/Get-AzureRmDeploymentManagerRollout.md
@@ -45,7 +45,7 @@ PS C:\> Get-AzureRmDeploymentManagerRollout -ResourceGroupName ContosoResourceGr
 
 This command gets a rollout named ContosoRollout in the ContosoResourceGroup. 
 
-### Example 2 Get and display the rollout hierarchically
+### Example 2 Get and display the rollout details
 ```powershell
 PS C:\> Get-AzureRmDeploymentManagerRollout -ResourceGroupName ContosoResourceGroup -Name ContosoRollout -Verbose
 ```


src/ResourceManager/DeploymentManager/Commands.DeploymentManager/Commands/RolloutCmdletBase.cs
```diff
@@ -0,0 +1,319 @@
...
+
+        private const int StepIndentFactor = 2;
+
+        private const int StepPropertiesIndentFactor = 3;
+
+        private const int ResourceOperationsIndentFactor = 4;
+
+        protected static void PrintVerbose(string message)
+        {
+            Console.WriteLine(message);
```
cormacpayne: @deveshguha you never want to use `Console.WriteLine(...)` in PowerShell -- PowerShell has separate streams that you can write to and handles their display differently.

For verbose text, you'll want to use the verbose stream, which you can write to with the method `WriteVerbose(...)`

src/ResourceManager/DeploymentManager/Commands.DeploymentManager/Commands/RolloutCmdletBase.cs
```diff
@@ -0,0 +1,319 @@
...
+        }
+
+        protected static void PrintError(string message)
+        {
+            ConsoleColor currentColor = Console.ForegroundColor;
+            Console.ForegroundColor = ConsoleColor.Red;
+            Console.WriteLine(message);
```
cormacpayne: @deveshguha same comment here -- you'll want to use `WriteError(...)` to write to the error stream, or just simply throw the error as an exception.

src/ResourceManager/DeploymentManager/Commands.DeploymentManager/Commands/GetRollout.cs
```diff
@@ -88,6 +94,12 @@ public override void ExecuteCmdlet()
             }
 
             var rollout = this.DeploymentManagerClient.GetRollout(this.ResourceGroupName, this.Name, this.RetryAttempt);
+
+            if (this.ShowDetails)
+            {
+                RolloutCmdletBase.PrintRollout(rollout);
```
cormacpayne: @deveshguha I think the best way to model this is to change the `Console.WriteLine(...)` calls to the appropriate stream writing calls, and then get rid of this new `-ShowDetails` parameter and _always_ make the calls to write to the verbose and error streams. This means that if the user ever wants the extra information that you're reporting back, they will always get it if they provide the `-Verbose` or `-Error` switch parameter that automatically is included with every cmdlet that we shipped (and is handled by the PowerShell engine). Let me know if you have any comments or concerns about this.

deveshguha:
@cormacpayne can we create a sign build and publish this package?

PR #7405, opened 2018-09-27
Update description for -SubscriptionId

cloudmelon:
I am an Azure book author, here is my propose update for this documentation.
-Subscription can be filled by subscription name or id , but while using subscription id , we should pass -SubscriptionId  as flag, tested with my past experience, and this also causes a misunderstanding from one of my readers.

## Checklist

### DIFF
--- a/src/ResourceManager/Profile/Commands.Profile/help/Set-AzureRmContext.md
+++ b/src/ResourceManager/Profile/Commands.Profile/help/Set-AzureRmContext.md
@@ -164,8 +164,7 @@ Accept wildcard characters: False
 ```
 
 ### -Subscription
-Subscription Name or Id 
-Whie using subscription Id as the parameter, this parameter is alias as -SubscriptionId 
+The name or id of the subscription that the context should be set to. This parameter has aliases to -SubscriptionName and -SubscriptionId, so, for clarity, either of these can be used instead of -Subscription when specifying name and id, respectively.
 
 ```yaml
 Type: System.String


cormacpayne:
@cloudmelon Hey Melony, thanks for opening this PR. I think this is a tricky issue that's worth discussing a little more. I agree that it can be confusing to have the `-Subscription` parameter accept _either_ a name or id, which is why we have two separate parameters for `-SubscriptionName` and `-SubscriptionId` in the [`Get-AzureRmSubscription`](https://github.com/Azure/azure-powershell/blob/preview/src/ResourceManager/Profile/Commands.Profile/help/Get-AzureRmSubscription.md) cmdlet. However, making a similar type of change for this cmdlet (along with [`Connect-AzureRmAccount`](https://github.com/Azure/azure-powershell/blob/preview/src/ResourceManager/Profile/Commands.Profile/help/Connect-AzureRmAccount.md), which follows the same pattern) would be considered a breaking change, so I don't think this is a feasible option _at the moment_.

I like the change that you are currently making to the documentation to let the user know there's an alias to `-SubscriptionId` that _should_ be used when the user is providing an id (and we can make a similar comment about the `-SubscriptionName` alias when the user provides a name), but I do want to make sure we handle this properly in the long term. Do you think the documentation updates that you're proposing now followed by the breaking change in the future to split the parameter into two will make for a better user experience?

Let me know what you think.

cloudmelon:
Hello @cormacpayne I really appreciate your attitude to open a discussion regarding this topic, it is a positive way to share our thoughts together :) Currently, as far as I know,  some readers do miss the fact of subscription can be Id or Name, in this case, breaking changes to divide the subscription to SubscriptionName and SubscriptionId will definitly to help them understand it better. 
However, I understand it may be redundant in that case. That was why I proposed to precise the parametter while passing subscription id by using SubscriptionId as alias. 
Let me know your thoughts. I would be very pleased to continue discussion with you :)

cormacpayne:
@cloudmelon Hey Melony, I made a change to the text in the file you're updating. Let me know what you think.

PR #7316, opened 2018-09-20
Update ClientRuntime and TestFramework

cormacpayne:
## Checklist

- [x] I have read the [_Submitting Changes_](https://github.com/Azure/azure-powershell/blob/preview/CONTRIBUTING.md#submitting-changes) section of [`CONTRIBUTING.md`](https://github.com/Azure/azure-powershell/blob/preview/CONTRIBUTING.md)
- [x] The title of the PR is clear and informative
- [x] The appropriate [change log has been updated](https://github.com/Azure/azure-powershell/blob/preview/CONTRIBUTING.md#updating-the-change-log)
- [x] The PR does not introduce [breaking changes](https://github.com/Azure/azure-powershell/blob/preview/documentation/breaking-changes/breaking-changes-definition.md)
- [ ] If applicable, the changes made in the PR have proper test coverage
- [ ] For public API changes to cmdlets:
    - [ ] the changes have gone through a [cmdlet design review](https://github.com/Azure/azure-powershell-cmdlet-review-pr)
    - [ ] the cmdlet markdown files were [generated using the `platyPS` module](https://github.com/Azure/azure-powershell/blob/preview/documentation/development-docs/help-generation.md)

PR #7429, opened 2018-09-29
[Compute] Upgrade compute client library version to 23.0.0.

hyonholee:
Shared Gallery cmdlets:
https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/135
https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/168

Add -Location parameter to Get-AzureRmVM: 
https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/172

Add DiffDiskSetting parameter to Set-AzureRmVmOSDisk and Set-AzureRmVmssStorageProfile:
https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/171

Add DiskIOPSReadWrite and DiskMBpsReadWrite parameters to New-AzureRmDiskConfig and New-AzureRmDiskUpdateConfig cmdlets
https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/173

Add UltraSSDEnabled parameter to New-AzureRmVMConfiig, New-AzureRmVmssConfig, Update-AuzreRmVM, and Update-AzureRmVmss.
https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/182

## Checklist


cormacpayne:
These changes depend on PR https://github.com/Azure/azure-powershell/pull/7316

cormacpayne:
@hyonholee would you mind resolving the merge conflicts? Also, I'm removing the `Do Not Merge` label so we know this should be merged when review is approved

src/ResourceManager/Compute/Commands.Compute/Generated/VirtualMachineScaleSet/Config/NewAzureRmVmssConfigCommand.cs
```diff
@@ -192,6 +192,11 @@ public partial class NewAzureRmVmssConfigCommand : Microsoft.Azure.Commands.Reso
             ValueFromPipelineByPropertyName = true)]
         public string[] IdentityId { get; set; }
 
+        [Parameter(
+           Mandatory = false,
+           ValueFromPipelineByPropertyName = true)]
+        public bool UltraSSDEnabled { get; set; }
```
cormacpayne: @hyonholee if that user provides no value for this parameter, what is the default value? It seems like it would be better to make this a switch parameter called `-EnableUltraSSD` that the user can see if the default behavior is disabled, but I'm not sure if that works with the logic below (instantiating the objects sent to the server, etc.).

This comment applies for all instances of this new parameter.

src/ResourceManager/Compute/Commands.Compute/help/New-AzureRmVmss.md
```diff
@@ -61,9 +61,9 @@ The command above creates the following with the name `$vmssName` :
 
 The default image chosen for the VMs in the VMSS is `2016-Datacenter Windows Server` and the SKU is `Standard_DS1_v2`
 
-
 ### Example 2: Create a VMSS using the **`DefaultParameterSet`**
-``` powershell
+```
+powershell
```
cormacpayne: @hyonholee this line should remain ` ```powershell `

tools/StaticAnalysis/Exceptions/AzureRM.Compute/BreakingChangeIssues.csv
```diff
@@ -99,4 +99,6 @@
...
+"Microsoft.Azure.Commands.Compute.dll","Microsoft.Azure.Commands.Compute.Extension.AzureDiskEncryption.RemoveAzureVmssDiskEncryptionCommand","Disable-AzureRmVmssDiskEncryption","0","3000","The type of property 'Identity' of type 'Microsoft.Azure.Commands.Compute.Automation.Models.PSVirtualMachineScaleSet' has changed from 'Microsoft.Azure.Management.Compute.Models.VirtualMachineScaleSetIdentity' to 'Microsoft.Azure.Commands.Compute.Automation.Models.PSVirtualMachineScaleSetIdentity'.","Change the type of property 'Identity' back to 'Microsoft.Azure.Management.Compute.Models.VirtualMachineScaleSetIdentity'."
+"Microsoft.Azure.Commands.Compute.dll","Microsoft.Azure.Commands.Compute.Extension.Diagnostics.AddAzureRmVmssDiagnosticsExtension","Add-AzureRmVmssDiagnosticsExtension","0","3000","The type of property 'UpgradePolicy' of type 'Microsoft.Azure.Commands.Compute.Automation.Models.PSVirtualMachineScaleSet' has changed from 'Microsoft.Azure.Management.Compute.Models.UpgradePolicy' to 'Microsoft.Azure.Commands.Compute.Automation.Models.PSUpgradePolicy'.","Change the type of property 'UpgradePolicy' back to 'Microsoft.Azure.Management.Compute.Models.UpgradePolicy'."
```
cormacpayne: @hyonholee I believe that the two exceptions flagged here are false-positives (couldn't find any removed properties), but I want to double-check that there are no properties in the mentioned types have been removed when going from `FooComputeType` to `PSFooComputeType`.

src/ResourceManager/Compute/Commands.Compute/Generated/VirtualMachineScaleSet/Config/NewAzureRmVmssConfigCommand.cs
praries880: @hyonholee
Is there a distinction between the case when UltraSSDEnabled -s false vs Absent?

The code [here](https://github.com/Azure/azure-powershell/pull/7429/files#diff-b6f0ad3c65642019eab97e326814dc46R430) treats the cases differently and suggests that there is a difference b/w the setting not being specified vs the setting it to false. What is the difference in terms of the vm behavior/config int he two cases?

hyonholee: @robhat1

singhkays: @praries880 If it’s a switch, then I don’t think we’ll be able to set it to false which is a legitimate scenario a customer might want to do with existing VM i.e. change it to false

praries880: @singhkays what happens if we just set it to false when the switch is not present? is that different form the setting being present on the vm but being set to false?
i.e :
```
                vVirtualMachineProfile.AdditionalCapabilities.UltraSSDEnabled = this.UltraSSDEnabled;
```
This setting will be false when the switch is false and true when the switch is present, but its always applied to the vm. Is that not acceptable?

hyonholee: This should be discussed before the Powershell design review was approved.  Anyway, I think the semantics for null value of AdditionalCapabilities and for false value of UItraSSDEnabled are different.   @robhat1 to confirm it.  (can we set it to false when the storage account type is not allowed for UltraSSD?)

singhkays: Talked to @rohbat1 offline. He'll respond here with the differences between false and null. Usually null in ARM APIs means nothing to change

robhat1: The property in CRP is a nullable Boolean. We don’t want it to be specified if user doesn’t want anything to do with UltraSSD. Sending a value of false by default would unnecessarily result in having CRP to persist the flag in its KVS.
Also, there is a difference between null and false.

This is the behavior that we want:
1.	Send a value true if UltraSSD has to be enabled.
2.	Send a value false if UltraSSD has to be disabled.
3.	Send null if user doesn’t have anything to do with UltraSSD or does not want any change to existing persisted value in CRP. If the persisted value in CRP is true and null is sent, CRP simply ignores it and assumes user does not want to change anything.

praries880: @singhkays @robhat1 Thanks for the explanation...
So If im reading this right... the UltraSSD setting is for more than the current vm/vmss it is being applied to. The text above seems to suggest that the setting (if specified) is toggled for the user rather than just the vm or vmss being created here. 

Is my understanding correct? If it is then this seems like it should be an entirely different cmdlet itself that the user uses to toggle the UltraSSD setting n shouldn't be bundled with the new-AzVM(ss)config command.

praries880: @hyonholee @robhat1 
I just synced with @singhkays offline.
For the  New-AzureRmVMConfiig, New-AzureRmVmssConfig c,dlets, i propose adding a switcvh called "EanbleUltraSSD", which if present sets the config to true. If the switch is not present, we do not  send in the configuration. Given that we are creating a new VM/Vmss, the effect is that ultraSSD is not enabled.

For Update-AuzreRmVM, and Update-AzureRmVmss i am fine with keeping it as a boolean value since here the three different states do make a difference.

Let me know if this sounds resonable.

praries880:
@hyonholee kindly take care of the merge conflicts

hyonholee:
@cormacpayne @praries880 Hi, I resolved merge conflicts, and also updated EnableUltraSSD to switch parameters as suggested.  Please review it.  Thanks.

src/ResourceManager/Compute/Commands.Compute/VirtualMachine/Config/NewAzureVMConfigCommand.cs
```diff
@@ -133,6 +138,11 @@ public override void ExecuteCmdlet()
                 vm.HardwareProfile.VmSize = this.VMSize;
             }
 
+            if (this.EnableUltraSSD.IsPresent)
+            {
+                vm.AdditionalCapabilities = new AdditionalCapabilities(true);
```
praries880: @hyonholee 
Don't we need to actually assign the ultrassd capability like we do [here](https://github.com/Azure/azure-powershell/pull/7429/files#diff-9c9de3599461d0d75d69ca1fd1421fa5R178)?

hyonholee: @praries880 new AdditionalCapabilities(true) creates the object with UltraSSDEnabled = true.  The links is for Update, where it creates the object and then sets the Boolean value as the same as the passed value.

praries880: @hyonholee  thanks for clarifying

hyonholee:
Any update?  Can this PR be merged in this release?  Thanks.

praries880:
@hyonholee It was too late on Friday to merge this change into the release branch and get it out. We decided to merge it in preview once the release branch got merged back into it. Kindly take care of the merge conflicts.

praries880:
@azuresdkci test this please

PR #7492, opened 2018-10-08
Adding the output model change in Get-AzureRmResource

bganapa:
## Description
Just an addition in the migration guide docs


bganapa:
@cormacpayne Thanks for the edits

PR #7555, opened 2018-10-14
Made Name parameter as optional.

tanveery:
The Name parameter was specified as required which is not true. For this command, the Name parameter is optional.

## Checklist

### DIFF
--- a/src/ResourceManager/Storage/Commands.Management.Storage/help/Get-AzureRmStorageAccount.md
+++ b/src/ResourceManager/Storage/Commands.Management.Storage/help/Get-AzureRmStorageAccount.md
@@ -76,7 +76,7 @@ Type: System.String
 Parameter Sets: AccountNameParameterSet
 Aliases: StorageAccountName, AccountName
 
-Required: True
+Required: False
 Position: 1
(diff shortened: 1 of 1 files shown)


cormacpayne:
@tanveery Hey Tanveer, there are two parameter sets for the `Get-AzureRmStorageAccount` cmdlet:

(1) The default parameter set where `-ResourceGroupName` is optional -- this allows the user to provide either no parameters (returns all storage accounts in the current context) or a value for `-ResourceGroupName` (returns all storage accounts in the provided resource group)

(2) The other parameter set where `-ResourceGroupName` and `-Name` are required -- this allows the user to provide a value for both of these parameters and get back a specific storage account matching the criteria provided

In the case for parameter set (2), if the user provides a value for the `-Name` parameter, but not for the `-ResourceGroupName` parameter, PowerShell will prompt the user to provide a value since it has determined the user is in set (2) and needs a value for all required parameters.

The markdown file you are editing specifies that the `-Name` parameter is available in only one parameter set and that it is required, which are both true. As a result, the requested change should not be made.

Please let me know if you have any questions 😁

PR #7576, opened 2018-10-15
Pull in latest changes in preview into the UltraSSD branch

praries880:
These changes get in the latest version of the Client runtime and test framework into the branch


cormacpayne:
@praries880 would you mind pulling in the latest from `preview` into your branch?

praries880:
@cormacpayne Thanks for the fix.. I pulled it in.

Issue #7479, opened 2018-10-05
Get-RoleDefinition cmdlet throws null reference when there is no context


markcowl:
```powershell
PS C:\Users\markcowl> Get-AzRoleDefinition
Get-AzRoleDefinition : Object reference not set to an instance of an object.
At line:1 char:1
+ Get-AzRoleDefinition
+ ~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : CloseError: (:) [Get-AzRoleDefinition], NullReferenceException
    + FullyQualifiedErrorId : Microsoft.Azure.Commands.Resources.GetAzureRoleDefinitionCommand

PS C:\Users\markcowl> Resolve-AzError


   HistoryId: 1


Message        : Object reference not set to an instance of an object.
StackTrace     :    at Microsoft.Azure.Commands.Resources.GetAzureRoleDefinitionCommand.ExecuteCmdlet()
                    at Microsoft.WindowsAzure.Commands.Utilities.Common.AzurePSCmdlet.ProcessRecord()
Exception      : System.NullReferenceException
InvocationInfo : {Get-AzRoleDefinition}
Line           : Get-AzRoleDefinition
Position       : At line:1 char:1
                 + Get-AzRoleDefinition
                 + ~~~~~~~~~~~~~~~~~~~~
HistoryId      : 1


```

praries880:
Pr merged : https://github.com/Azure/azure-powershell/pull/7592

hausec:
Still getting this error on azure-cli 2.5.1

There is a subscription for the tenant.

PR #7592, opened 2018-10-16
Fix for issue #7479

praries880:


src/ResourceManager/Resources/Commands.Resources/RoleDefinitions/GetAzureRoleDefinitionCommand.cs
```diff
@@ -54,15 +54,25 @@ public override void ExecuteCmdlet()
                 Scope = Scope,
                 ResourceIdentifier = new ResourceIdentifier
                 {
-                    Subscription = DefaultProfile.DefaultContext.Subscription.Id.ToString()
+                    Subscription = (DefaultProfile.DefaultContext.Subscription.Id == null)?null:DefaultProfile.DefaultContext.Subscription.Id.ToString()
```
MiYanni:
```suggestion
                    Subscription = DefaultProfile.DefaultContext.Subscription.Id?.ToString()
```
Just use the [null conditional operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-conditional-operators) instead.

Issue #7530, opened 2018-10-12
Add message to deprecation warnings with instuctions on how to turn off warnings


markcowl:
### Description
- Add additional details to the warning messages for deprecation
  * To stop display of deprecation warnings, set the environment variable <variable name> to "true", using Set-Item Env:\<variablename> true.  Add this line to your user profile to alaways disable these warnings.
  * Consider adding a cmdlet to enable and disable deprecation warnings:
  - Enable-AzDeprecationWarnings / Disable-AzDeprecationWarnings [-Scope Session | CurrentUser | LocalMachine] 

### Cost: 2

praries880:
### Prs
* common repo : https://github.com/Azure/azure-powershell-common/pull/87
* Ps repo : https://github.com/Azure/azure-powershell/pull/7685

maddieclayton:
https://github.com/Azure/azure-powershell-common/pull/115
https://github.com/Azure/azure-powershell/pull/8202

PR #7685, opened 2018-10-26
Add help file to be linked from the breaking change warnings 

praries880:
Part 1 of the fix for : https://github.com/Azure/azure-powershell/issues/7530
This part is about adding the markup file that will be linked from the actual text added to the warnings a user sees.

The aka link that will point to this is : https://aka.ms/AzPsSdkBreakingChangeAttributeInformation 

Second part will be the actual code change in the common repo.

### DIFF
--- a/documentation/breaking-changes/breaking-change-messages-help.md
+++ b/documentation/breaking-changes/breaking-change-messages-help.md
@@ -1,14 +1,21 @@
 # Breaking Change Messages Help
 
-## What is this warning all about?
+## What is this warning and what is it about?
 
-Breaking change warnings are a means for the cmdlet authors to communicate with the end users any upcoming breaking changes. Most of these changes will be taking effect in the next breaking change release.
+Breaking change warnings are a means for the cmdlet authors to communicate with the end users any upcoming breaking changes in the cmdlet. Most of these changes will be taking effect in the next breaking change release.
 
 ## How do I get rid of the warnings?
 
-As of now the only way a user can supress the warnings is to set an environment variable, [details here](https://github.com/Azure/azure-powershell/blob/preview/documentation/breaking-changes/breaking-changes-attribute-help.md#supress-the-breaking-change-messages-at-runtime).
+As of now the only way a user can supress the warnings is to set an environment variable `SuppressAzurePowerShellBreakingChangeWarnings` to `true`, [more details here](https://github.com/Azure/azure-powershell/blob/preview/documentation/breaking-changes/breaking-changes-attribute-help.md#supress-the-breaking-change-messages-at-runtime).
 
-## More information
+## Further reading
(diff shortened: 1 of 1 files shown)


praries880:
PRin the common repo : https://github.com/Azure/azure-powershell-common/pull/87

documentation/breaking-changes/breaking-change-messages-help.md
```diff
@@ -0,0 +1,24 @@
+# Breaking Change Messages Help
+
+## What is this warning all about?
+
+Breaking change warnings are a means for the cmdlet authors to communicate with the end users any upcoming breaking changes. Most of these changes will be taking effect in the next breaking change release.
```
markcowl: This should be written from the perspecitve of a customer, who doesn't know or care anything about our internal org structure or processes.  They need to know:
- Why they are getting a breaking change warning [because a service change that causes a breaking change will be coming in a major release]
- What they can do about it - the warning itself should contain advice about how to make the change, and a migration guide describing how scripts need to change in detail will be published when the breakign change is released (pointer to migration guides)
- When the breaking change will happen - there is one breaking change release in late Spring and one in Fall.  A breakign change warnign indicates that a change will happen at the next breaking change release.  Also point them at the milestone schedule.

documentation/breaking-changes/breaking-change-messages-help.md
```diff
@@ -0,0 +1,24 @@
...
+
+## How do I get rid of the warnings?
+
+As of now the only way a user can supress the warnings is to set an environment variable, [details here](https://github.com/Azure/azure-powershell/blob/preview/documentation/breaking-changes/breaking-changes-attribute-help.md#supress-the-breaking-change-messages-at-runtime).
```
markcowl: We should explicitly state here how to turn off the warnings.  A specific mention of the environment variable, and how to disable and enable warnings in powershell script shoudl be included.

documentation/breaking-changes/breaking-change-messages-help.md
```diff
@@ -0,0 +1,24 @@
...
+
+## More information
+
+### What are the different types of breaking changes possible in the cmdlet?
+
+[Here](https://github.com/Azure/azure-powershell/blob/preview/documentation/breaking-changes/breaking-changes-definition.md
+) is a good resource on the various types of breaking changes authors can call out in a cmdlet.
+
+### How are these warnings added to the cmdlets
```
markcowl: I would remove this section on how the warnings are added

maddieclayton:
@praries880 Looks like something weird happened with your commits - can you fix?

praries880:
deprecated in favor of : https://github.com/Azure/azure-powershell/pull/7812

PR #7914, opened 2018-11-27
2: Removed StackAdmin folder

MiYanni:
Removes StackAdmin folder.


deathly809:
Could you give some info on this?  Will this eventually make it to master/preview branch?

MiYanni:
@deathly809 For Stack, we are opting to have it be separate since most of the code/cmdlets aren't following the same processes as the rest of the service teams. Our initial thought is there would be a permanent `stack` branch. Your code should need nothing from this repo directly, so that's why we are handling it as a separate endeavor. You use our [common repo](https://github.com/Azure/azure-powershell-common) packages for your process, so nothing here in our module repo should apply.

Currently, you have separate copies of our modules here too, since they are locked to certain APIs. We will be doing an Azure Profile integration for the modules in 2019. So, eventually, you'll just consume a Profile of those cmdlets, and not need your own separate copies of the source code either.

bganapa:
@MiYanni we have a release scheduled for Dec 1st week. Can this wait till then?

an open PR for this release  https://github.com/Azure/azure-powershell/pull/7924

MiYanni:
@bganapa This work is scheduled post Az 1.0 release. So, past December 18th. After then, we are migrating from having the primary branch of this repo be `preview` to instead be `dev`. The branch that this PR is targeting, `az-cleanup`, will be merged into `dev` in late December.

PR #7915, opened 2018-11-27
3: Move Storage to ResourceManager/Storage

MiYanni:
This moves the Storage projects to the ResourceManager/Storage folder as to match how multiple projects are handled for every other module in the repo. This keeps all Storage-related projects in the same folder.


blueww:
@MiYanni 

I am not sure why we need to do this. But this will take a lot of conflict when merge preview back to Storage branch. 
So you mind to also apply it to the storage branch "AzureRM.Storage.Location"

MiYanni:
@blueww Yes, there will be a high number of merge issues for every PR. We already have a plan to handle this from our team in stages. There will be a stage where we update the service branches with our overall changes. So, for now, do not worry about this.

Overall, the effort from the `az-cleanup` branch is to remove all `AzureRM` relevant structuring, simplify the codebase, and make Az the only development experience for the repository.

blueww:
@MiYanni 

Thanks for the information!
Please help to migrate this change from "preview" to branch "AzureRM.Storage.Location" after this PR is merged.

PR #7929, opened 2018-11-28
4: Remove net472 testing build process

MiYanni:
This is so that the per-module solution won't be overly complicated to build/run the tests. We are removing the TestPrerequisites concept and will have a different solution for testing/creating net472 builds/dlls for testing properly. However, I have left the net472 block/logic in the .targets file for testing later when we do get net472 builds working again.


cormacpayne:
@MiYanni would you mind taking a look at the merge conflict?

PR #7991, opened 2018-12-03
Remove Resources dependency on common Graph.RBAC library and use SDK instead

cormacpayne:
## Description

Remove the Resources project dependency on the common Graph.RBAC library and depend on the `Microsoft.Azure.Graph.RBAC` SDK instead.

## Checklist


cormacpayne:
CC: @grlin

src/ResourceManager/Resources/Commands.Resources/ActiveDirectory/Models/ActiveDirectoryClient.cs
```diff
@@ -0,0 +1,818 @@
...
+                    servicePrincipals.Add(servicePrincipal.ToPSADServicePrincipal());
+                }
+            }
+            else
+            {
+                Rest.Azure.OData.ODataQuery<ServicePrincipal> odataQuery = null;
+                if (!string.IsNullOrEmpty(options.SearchString) && options.SearchString.EndsWith("*"))
+                {
+                    options.SearchString = options.SearchString.TrimEnd('*');
+                    odataQuery = new Rest.Azure.OData.ODataQuery<ServicePrincipal>(s => s.DisplayName.StartsWith(options.SearchString));
```
markcowl: Does DisplayName always have a value?  If not, we should guard nulls

src/ResourceManager/Resources/Commands.Resources/ActiveDirectory/Models/ActiveDirectoryClient.cs
```diff
@@ -0,0 +1,818 @@
...
+                    return result.Select(u => u.ToPSADUser());
+                }
+            }
+            else
+            {
+                Rest.Azure.OData.ODataQuery<User> odataQuery = null;
+                if (!string.IsNullOrEmpty(options.SearchString) && options.SearchString.EndsWith("*"))
+                {
+                    options.SearchString = options.SearchString.TrimEnd('*');
+                    odataQuery = new Rest.Azure.OData.ODataQuery<User>(u => u.DisplayName.StartsWith(options.SearchString));
```
markcowl: same comment.  ALso, it would be nice if we only hand to do this logic one time, rather than once per entity

src/ResourceManager/Resources/Commands.Resources/ActiveDirectory/Cmdlets/ActiveDirectoryBaseCmdlet.cs
```diff
@@ -0,0 +1,111 @@
...
+            else
+            {
+                throw exception;
+            }
+        }
+
+
+        private Exception ParseResponse(GraphErrorException graphEx)
+        {
+            int exceptionMessageIndex = graphEx.Response.Content.IndexOf("\"value\":", StringComparison.CurrentCultureIgnoreCase);
```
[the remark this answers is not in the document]
cormacpayne: I'm not sure, but I added a null-check just to be safe.

src/ResourceManager/Resources/Commands.Resources/ActiveDirectory/Models/ActiveDirectoryClient.cs
cormacpayne: I think it always has a value, but I added a null-check just in case.

src/ResourceManager/Resources/Commands.Resources/ActiveDirectory/Models/ActiveDirectoryClient.cs
cormacpayne: Same comment. I'm not sure if we can make this logic generic since the underlying types (`User`, `ServicePrincipal`, etc.) all derive from the same base type, but that base type doesn't have the `DisplayName` property that we're filtering over.

src/ResourceManager/Resources/Commands.Resources.Test/Commands.Resources.Test.Netcore.csproj
```diff
@@ -24,6 +24,7 @@
   </PropertyGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' != 'net472'">
+    <PackageReference Include="Microsoft.Azure.Graph.RBAC" Version="3.5.0-preview" />
```
markcowl: It looks l9ike this isn't published.  Can we publish it?  Also, does this have the correct newtonsoft.json dependency in netstandard2.0?

src/ResourceManager/Resources/Commands.Resources/ActiveDirectory/Models/ParameterSet.cs
```diff
@@ -0,0 +1,199 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// ----------------------------------------------------------------------------------
+
+namespace Microsoft.Azure.Commands.ActiveDirectory
+{
+    public static class ParameterSet
+    {
+        public const string Mail = "MailParameterSet";
+
```
markcowl: rmove spaces between constant definition statements

src/ResourceManager/Resources/Commands.Resources/Properties/Resources.resx
```diff
@@ -420,4 +420,28 @@
   <data name="DefaultEndDateUsed" xml:space="preserve">
     <value>No value specified for -EndDate parameter; setting the value to one year after start date.</value>
   </data>
+  <data name="GroupWithDisplayNameDoesntExist" xml:space="preserve">
+    <value>Group with display name '{0}' does not exist.</value>
+  </data>
+  <data name="MultipleApplicationsWithDisplayNameFound" xml:space="preserve">
+    <value>More than one application found with display name '{0}'. Please use the Get-AzureRmADApplication cmdlet to get the object id of the desired application.</value>
```
markcowl: Get-AzAdApplication: Please go through these to remove references to 'AzureRm' in cmdlets

tools/StaticAnalysis/Exceptions/Az.Resources/BreakingChangeIssues.csv
```diff
@@ -1 +1,73 @@
 "AssemblyFileName","ClassName","Target","Severity","ProblemId","Description","Remediation"
```
markcowl: Note: is there a reason we can't publish the nuget package above?

src/ResourceManager/Resources/Commands.Resources/ActiveDirectory/Cmdlets/ActiveDirectoryBaseCmdlet.cs
```diff
@@ -0,0 +1,115 @@
...
+using System;
+using System.Management.Automation;
+using System.Net;
+using ProjectResources = Microsoft.Azure.Commands.Resources.Properties.Resources;
+
+namespace Microsoft.Azure.Commands.ActiveDirectory
+{
+    public abstract class ActiveDirectoryBaseCmdlet : AzureRMCmdlet
+    {
+        private ActiveDirectoryClient activeDirectoryClient;
```
markcowl: use '_activeDirectoryClient' - private fields shoudl start with underscore

src/ResourceManager/Resources/Commands.Resources/ActiveDirectory/Models/ActiveDirectoryClient.cs
```diff
@@ -0,0 +1,818 @@
...
+
+        /// <summary>
+        /// Creates new ActiveDirectoryClient using WindowsAzureSubscription.
+        /// </summary>
+        /// <param name="context"></param>
+        public ActiveDirectoryClient(IAzureContext context)
+        {
+            GraphClient = AzureSession.Instance.ClientFactory.CreateArmClient<GraphRbacManagementClient>(
+                context, AzureEnvironment.Endpoint.Graph);
+
```
markcowl: remove extra vertical space

src/ResourceManager/Resources/Commands.Resources/ActiveDirectory/Models/ActiveDirectoryClient.cs
```diff
@@ -0,0 +1,818 @@
...
+            GraphClient.TenantID = context.Tenant.Id.ToString();
+        }
+
+        public PSADObject GetADObject(ADObjectFilterOptions options)
+        {
+            PSADObject result = null;
+
```
markcowl: remove extra vertical space - only after ending brackets '}' is it appropriate inside a method

src/ResourceManager/Resources/Commands.Resources/ActiveDirectory/Models/ActiveDirectoryClient.cs
```diff
@@ -0,0 +1,818 @@
...
+                }
+
+                return new GenericPageEnumerable<ADGroup>(
+                    delegate ()
+                    {
+                        return GraphClient.Groups.List(odataQuery);
+                    }, GraphClient.Groups.ListNext, first, skip).Select(g => g.ToPSADGroup());
+            }
+
+            return new List<PSADGroup>();
```
markcowl: So, this is a  bit difficult to follow.  I would suggest moving this return value into the upper if, and making this one

```c#
if(!string.IsNullOrEmpty(options.Id)) {
}
else if (options.Mail != null) {
}
else {
}
``` 

statement

src/ResourceManager/Resources/Commands.Resources.Test/Commands.Resources.Test.Netcore.csproj
cormacpayne: Package has been published and verified that it has proper `Newtonsoft.Json` dependency in .NET Standard

tools/StaticAnalysis/Exceptions/Az.Resources/BreakingChangeIssues.csv
cormacpayne: The package has been published and is being referenced instead of the local `.nupkg` file.

Issue #7374, opened 2018-09-25
Rename Az module assemblies


MiYanni:
Let's do either:
```
Microsoft.Azure.PowerShell.Module.*.dll
```
or:
```
Microsoft.Azure.PowerShell.Cmdlet.*.dll
```

MiYanni:
Discussed with Mark. Going with:
```
Microsoft.Azure.PowerShell.Cmdlets.*.dll
```

MiYanni:
After a design review meeting, here is what we've decided. The plurality is based on:
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-namespaces
Basically, namespaces should be plural when appropriate. So, we are using plural names here.

### Cmdlets/Modules
```
Microsoft.Azure.PowerShell.Cmdlets.*.dll
```

### Clients (Management/Dataplane) [in Common]
```
Microsoft.Azure.PowerShell.Clients.*.dll
```

### Framework (no designator) [in Common]
```
Microsoft.Azure.PowerShell.*.dll
```

MiYanni:
Completed as part of Az 1.0.

PR #8034, opened 2018-12-06
11: Update module & Common DLL names

MiYanni:
This changes the names of the DLLs to coincide with https://github.com/Azure/azure-powershell/issues/7374 
The Common changes are here: https://github.com/Azure/azure-powershell-common/pull/101


src/ResourceManager/Profile/Commands.Profile/Az.Accounts.psd1
```diff
@@ -57,24 +57,24 @@ DotNetFrameworkVersion = '4.7.2'
 RequiredAssemblies = '.\Microsoft.Azure.PowerShell.Authentication.Abstractions.dll', 
     '.\Microsoft.Azure.PowerShell.Authentication.dll', 
     '.\Microsoft.Azure.PowerShell.Authentication.ResourceManager.dll', 
-    '.\Microsoft.Azure.PowerShell.Authorization.dll', 
-    '.\Microsoft.Azure.PowerShell.Compute.dll', 
-    '.\Microsoft.Azure.PowerShell.Graph.Rbac.dll', 
-    '.\Microsoft.Azure.PowerShell.Network.dll', 
-    '.\Microsoft.Azure.PowerShell.ResourceManager.dll', 
+    '.\Microsoft.Azure.PowerShell.Clients.Authorization.dll',
```
markcowl: Just a quick question: are we going to change the profile directory as part of this, or is that too big a change?

tools/Common.Netcore.Dependencies.targets
```diff
@@ -66,7 +66,7 @@
     <PackageReference Include="PowerShellStandard.Library" Version="5.1.0-preview-06" PrivateAssets="All" />
   </ItemGroup>
   <PropertyGroup>
-    <StorageToolsPath>$(NugetPackageRoot)microsoft.azure.powershell.storage\1.0.231-preview\tools\</StorageToolsPath>
+    <StorageToolsPath>$(NugetPackageRoot)microsoft.azure.powershell.storage\1.0.233-preview\tools\</StorageToolsPath>
```
markcowl: Per previous note, this is OK for now, but will need to update to the latest published before we release

src/ResourceManager/Profile/Commands.Profile/Az.Accounts.psd1
MiYanni: No directory changes/moving files happened in az-cleanup. It would have caused a lot of unnecessary merge conflicts and had no benefits for Az 1.0. I'll be cleaning/updating the repo in the next few weeks to reduce the directory structure and rename things. As a note, a lot of our processes are tied to names and directory structure. So, this would have been fairly massive to change for the entire repo right now. After Az 1.0, it will be easier to do.

PR #8108, opened 2018-12-12
 Corrects error when full path does not exist in AzError.ps1 and AzureRMError.ps1

ManagedUx:
Although the error is trapped by the try/catch block
the error is added to $Error. If a script or runbook
checks the count or value of $Error this can cause
an issue.

## Description

We recently update the azure modules on our hybrid worker.  The file StartupScripts\AzureRmError.ps1 (duplicated in StartupScripts\AzError.ps1) throws and exception if the full path (directory) does not already exist for any reason. This error is trapped by the try/catch but still manifests in the $Error collection. If the runbook, like ours, is checking the $Error collection as part of it's process verification this causes an issue. This behavior was not observed in prior versions.  This PR corrects the issue by first checking for and creating the full path prior to the New-Item commandlet attempting to create the file.

## Checklist

- [NA] If applicable, the changes made in the PR have proper test coverage
- [NA] For public API changes to cmdlets:
    - [NA] the changes have gone through a [cmdlet design review](https://github.com/Azure/azure-powershell-cmdlet-review-pr) (Microsoft internal partners only)
    - [NA] the cmdlet markdown files were [generated using the `platyPS` module](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/help-generation.md)

### DIFF
--- a/src/ResourceManager/Profile/Commands.Profile/StartupScripts/AzError.ps1
+++ b/src/ResourceManager/Profile/Commands.Profile/StartupScripts/AzError.ps1
@@ -16,10 +16,9 @@
         }
 
         $installationChecksDirectory = Split-Path -Path $pathToInstallationChecks -Parent -ErrorAction Stop
-        if (-not (Test-Path $installationChecksDirectory -ErrorAction Ignore)) {
-            New-Item -Path $installationChecksDirectory -ErrorAction Stop -ItemType Directory
+        if (Test-Path $installationChecksDirectory -ErrorAction Ignore) {
+            New-Item -Path $pathToInstallationChecks -ErrorAction Stop -ItemType File -Value ($installationchecks | ConvertTo-Json -ErrorAction Stop)
         }
-        New-Item -Path $pathToInstallationChecks -ErrorAction Stop -ItemType File -Value ($installationchecks | ConvertTo-Json -ErrorAction Stop)
     }
     catch
     { 
--- a/src/ResourceManager/Profile/Commands.Profile/StartupScripts/AzureRmError.ps1
+++ b/src/ResourceManager/Profile/Commands.Profile/StartupScripts/AzureRmError.ps1
@@ -16,10 +16,9 @@ function Write-InstallationCheckToFile
         }
         
         $installationChecksDirectory = Split-Path -Path $pathToInstallationChecks -Parent -ErrorAction Stop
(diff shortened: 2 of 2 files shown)


adxsdkps:

markcowl:
@azuresdkci add to whitelist

markcowl:
@azuresdkci test this please

markcowl:
@ManagedUx Thanks for the contribution, we will take  alook.

Note that in Automation (or any unattended scripting scenario), you should set the SkipAzInstallationChecks environment variable to true in any case, as these add no value in that scenario:

```powershell
Set-Item Env:\SkipAzInstallationChecks "true"
```

ManagedUx:
@markcowl Good to know (and I see that in the code).  However, given this breaks an already working AzureRm environment / scripts simply by doing "Update-Module AzureRm" I figured it was worth submitting.  Thanks again.

src/ResourceManager/Profile/Commands.Profile/StartupScripts/AzError.ps1
```diff
@@ -15,6 +15,10 @@
             Remove-Item -Path $pathToInstallationChecks -ErrorAction Stop
         }
 
+        $installationChecksDirectory = Split-Path -Path $pathToInstallationChecks -Parent -ErrorAction Stop
+        if (-not (Test-Path $installationChecksDirectory -ErrorAction Ignore)) {
```
markcowl: I think we should only write the file if the directory exists - we essentiallyshouldn't write errors in headless scenarios, I think, i.e.:

```powershell
$installationChecksDirectory = Split-Path -Path $pathToInstallationChecks -Parent -ErrorAction Stop
if (Test-Path -Path $installationChecksDirectory -ErrorAction Ignore) {
    New-Item -Path $pathToInstallationChecks -ErrorAction Stop -ItemType File -Value ($installationchecks | ConvertTo-Json -ErrorAction Stop)
}
```

markcowl:
@ManagedUX - this looks great.  Because users can turn off these checks, I think this works better if we only write the file when the path to the parent directory already exists.  We can make this change if you don't have time to update.

markcowl:
@ManagedUx  Made this change in Az. Will make sure this is added in AzureRM at the first bug fix release.

markcowl:
Closign for now, as change is in Az.  Will mnove change to AzureRM in next AzureRM release. Issue tracking in #8151

PR #8178, opened 2018-12-19
Removes duplicate MSSharedLibKey.snk files

MiYanni:
This PR removes MSSharedLibKey.snk from the individual module folders. All modules now use the **shared** MSSharedLibKey.snk in `src`.

PR also:
- Fixes Advisor to use new `props` pattern for the csproj files
- Removes `AllowUnsafeBlocks` from projects that were not using `unsafe` blocks
  - KeyVault is the only module that actually uses `unsafe` code blocks

### DIFF
--- a/tools/GenerateHelp.ps1
+++ b/tools/GenerateHelp.ps1
@@ -11,6 +11,7 @@ Param(
     [string]$FilteredModules
 )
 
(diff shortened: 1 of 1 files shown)


MiYanni:
sign job: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/psnetsign/392/

PR #8198, opened 2018-12-22
[ESRP AzureRm] removed redundant script

vladimir-shcherbakov:
* tabs replaced with spaces
* ps script to remove codesign client artifacts removed and replaced with a direct powershell call from the build script.


vladimir-shcherbakov:
signing: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/ps-sign-esrp/77/

PR #8207, opened 2018-12-28
2 of 3: Rename project and module folders

MiYanni:
This is the second PR of three. This includes:
- Renaming project folders to be consistent with module/service names
  - This primarily removed `Commands.` from the start of these folders
  - Some folders were simply inconsistent with their project name. These have been fixed.
- Renaming module folders to be consistent with module names
  - Includes:
    - `AzureBatch` -> `Batch`
    - `DataFactoryV2` -> `DataFactory`
      - Note: The `DataFactoryV2` project still remains in a `DataFactoryV2` project folder. However, the overall module folder matches the module name now.
    - `Insights` -> `Monitor`
    - `Profile` -> `Accounts`
- Moved a tool project in `Batch` to the tools folder
- Moved the `ResourceManager` project in `Resource` up a directory
  - This removed the `Cmdlets` folder that the project was nested in. This matches all other project depths now.


markcowl:
@MiYanni  This PR is too large to review.  We can either split this up into multiple PRs, or split it into multiple chunks

PR #8208, opened 2018-12-28
3 of 3: Remove ResourceManager folder

MiYanni:
This is the third PR of three. This includes:
- Moves all module folders up into `src`. Thus, removing the `ResourceManager` folder.
- Fixes references in scripts/docs to `src/ResourceManager`


MiYanni:
@cormacpayne Updated the `net472` script. I removed the stuff about Storage folder since it no longer is a separate path depth. For Version Controller, a majority of the paths come from `VersionFileHelper.cs`. The actual project folder is determined as a parent of the module's folder, so that specific property didn't need to be changed. But, a couple others did. Please look over the commit.

MiYanni:
@azuresdkci Retest this please

Issue #8124, opened 2018-12-13
Update AzureRM.Resources.ps1 and AzureRM.Storage.ps1 to use Az naming


MiYanni:
Our test scripts we share out of `ScenarioTests.ResourceManager` use the `AzureRM` naming convention. We should update these to have aliases to an `Az` version of those functions. This allows new tests being created to only have to use the `Az` nomenclature.

maddieclayton:
Done. I have an additional PR open to rename all of the tests, but this is unrelated

PR #8263, opened 2019-01-09
Fix prefix in test files

maddieclayton:
## Description
Fix issue: https://github.com/Azure/azure-powershell/issues/8124
Common pr: https://github.com/Azure/azure-powershell-common/pull/120

## Checklist


markcowl:
@maddieclayton Why do we need to do this?  We should certainly enable folks to use either prefix, but I don't know that we shoudl change this wholesale

maddieclayton:
@markcowl I thought it would be good for all of the tests to have the Az prefix so that partners coming in would see that they should use the Az prefix across the board for tests.  I also added some tests for Enable/Disable-AzureRmAlias, so we don't need to use aliases to test those cmdlets anymore.

src/LogicApp/LogicApp.Test/ScenarioTests/WorkflowController.cs
```diff
@@ -90,7 +90,7 @@ public void RunPowerShellTest(XunitTracingInterceptor logger, params string[] sc
                     "ScenarioTests\\Common.ps1",
                     "ScenarioTests\\" + callingClassName + ".ps1",
                     _helper.RMProfileModule,
-                    _helper.GetRMModulePath(@"AzureRM.LogicApp.psd1"),
+                    _helper.GetRMModulePath(@"Az.LogicApp.psd1"),
```
MiYanni: I don't think we support this yet. AFAIK, it still needs to be `"AzureRM.LogicApp.psd1"` for it to resolve correctly in `GetRMModulePath`.

MiYanni: Discussed offline. Should work with `"Az.LogicApp.psd1"`.

PR #8275, opened 2019-01-10
Update Get-AzureRmResource.md

cynthn:
None of the parameters should be marked as "Required = True" as you can use the cmdlet without parameters (as shown in Example 1 in the published documentation).

## Checklist

### DIFF
--- a/src/ResourceManager/Resources/Commands.Resources/help/Get-AzureRmResource.md
+++ b/src/ResourceManager/Resources/Commands.Resources/help/Get-AzureRmResource.md
@@ -259,7 +259,7 @@ Type: System.String
 Parameter Sets: ByResourceId
 Aliases: Id
 
-Required: True
+Required: False
 Position: Named
 Default value: None
 Accept pipeline input: True (ByPropertyName)
@@ -290,7 +290,7 @@ Type: System.Collections.Hashtable
 Parameter Sets: ByTagObjectParameterSet
 Aliases:
 
-Required: True
+Required: False
 Position: Named
 Default value: None
 Accept pipeline input: False


adxsdkps:

cormacpayne:
@cynthn Hey Cynthia, it's true that these parameters aren't _always_ required, but they are required within the parameter sets they appear in. For example, if you look at the section for the `-ResourceId` parameter, it has the following property:

```
Parameter Sets: ByResourceId
```

This defines the parameter set(s) that `-ResourceId` appears in, and all of the subsequent properties in that section apply to that specific parameter set(s).

I don't believe this change should be made as it's the pattern that the PowerShell Core team enforces and would also require us to change the documentation for most cmdlets that also have a similar issue. Please let me know if you have any questions or concerns.

cynthn:
But, when it gets rendered, that piece is lost? 

This is the rendered topic: https://docs.microsoft.com/en-us/powershell/module/azurerm.resources/get-azurermresource?view=azurermps-6.13.0 and it doesn't show anything about being specific to a parameter set:

## Required Parameters
**-ResourceId**
Specifies the fully qualified resource ID, as in the following example /subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/providers/Microsoft.Compute/virtualMachines
Type: | String
-- | --
Aliases: | Id
Position: | Named
Default value: | None
Accept pipeline input: | True (ByPropertyName)
Accept wildcard characters: | False

I understand that this would be a larger change, but it would be good to look at addressing this across the content. It is very confusing for customers who only see that a parameter is required and nothing about which parameter set requires it.

Thanks :)

PR #8294, opened 2019-01-11
Fix help for New-AzureRmPolicyDefinition cmdlet

mentat9:
Default value of -Mode parameter was wrong


cormacpayne:
@mentat9 Hey Chris, it still looks like there are some merge conflicts. We recently removed the `ResourceManager` folder that was found in `src` and moved all of the service folders down a level (`src/ResourceManager/Resources` --> `src/Resources`), so you will want to make sure that the changes you're making in this PR are to the new path files.

mentat9:
Since the preview branch is dead, I've moved these changes over to master in a new PR: https://github.com/Azure/azure-powershell/pull/8331

PR #8387, opened 2019-01-23
[GuestConfiguration] Supports Audit policies, fixed online doc link for help files

viveklingaiah:
## Checklist

### DIFF
--- a/tools/StaticAnalysis/Exceptions/Az.GuestConfiguration/MissingAssemblies.csv
+++ b/tools/StaticAnalysis/Exceptions/Az.GuestConfiguration/MissingAssemblies.csv
@@ -1,0 +1,3 @@
+"Directory","Assembly Name","Assembly Version","Referencing Assembly","Severity","ProblemId","Description","Remediation"
+"c:\workspace\powershell\artifacts\Debug\Az.GuestConfiguration","Microsoft.Win32.Registry","4.1.1.0","System.Management.Automation","0","3000","Missing assembly Microsoft.Win32.Registry referenced from System.Management.Automation","Ensure that the assembly is included in the Wix file or directory"
+"c:\workspace\powershell\artifacts\Debug\Az.GuestConfiguration","Microsoft.PowerShell.CoreCLR.Eventing","6.1.0.0","System.Management.Automation","0","3000","Missing assembly Microsoft.PowerShell.CoreCLR.Eventing referenced from System.Management.Automation","Ensure that the assembly is included in the Wix file or directory"


vivlingaiah:
@MiYanni I'm facing two issues below:
1) I'm still not able to load help content when I do get-help <cmdlet name>
2) when i just copy the GuestConfiguration folde, zip it and unzip it on a different box and try to import module, it throws errors.

MiYanni:
@vivlingaiah 
1. You have to generate the -Help.xml files. When you run the build.proj, include the property, `/p:RunStaticAnalysis=true`.
2. Run your fork and branch through this Jenkins project: https://azuresdkci.westus2.cloudapp.azure.com/view/PowerShell/job/psnetsign-esrp/build?delay=0sec
We publish the module with a completely different set of DLLs than what is used to build the module. This is because we load shared DLLs into the Accounts module, and have all modules take a dependency on Accounts. This way, every module doesn't need to have copies of the same DLLs.
After that build finishes, share the Az.GuestConfiguration.nupkg with them. Have them put it into a folder on their machine (like `C:\Modules\`) and run these commands to register the folder as a module repo and then install the module:
```powershell
Register-PSRepository -Name GuestConfigurationPreview -SourceLocation {{path to the folder where the nupkg is}} -PackageManagementProvider NuGet -InstallationPolicy Trusted
Install-Module -Name Az.GuestConfiguration -Repository GuestConfigurationPreview
```
Additionally, are they running Windows PowerShell or PowerShell Core? They'll need .NET Framework 4.7.2 installed if they are using Windows PowerShell. PowerShell Core does not have this dependency.

vivlingaiah:
@MiYanni , 
- I have fixed tests in commit 2, because build was failing.
- I also did include dll-help.xml file anywhere in the commit. Is that the right thing to do ?

vivlingaiah:
I put the nupkg files together under a dir and imported the module successfully. I'm still getting below the error message:

image

vivlingaiah:
also help is missing with nupkg. I'm able to access local content in local dev env

vivlingaiah:
build is also failing due to below error. KustoClusterNotExist wasn't added as part of PR

Properties\Resources.resx : warning MSB3568: Duplicate resource name "KustoClusterNotExist" is not allowed, ignored. [c:\workspace\powershell\src\Kusto\Kusto\Kusto.csproj] [c:\workspace\powershell\build.proj]
EXEC : warning : The names of some imported commands from the module 'HelpGeneration' include unapproved verbs that might make them less discoverable. To find the commands with unapproved verbs, run the Import-Module command again with the Verbose parameter. For a list of approved verbs, type Get-Verb. [c:\workspace\powershell\build.proj]

src/GuestConfiguration/GuestConfiguration/Az.GuestConfiguration.psd1
```diff
@@ -107,7 +107,7 @@ PrivateData = @{
         ReleaseNotes = 'Initial Release'
 
         # Prerelease string of this module
-        Prerelease = 'preview'
+        #Prerelease = 'preview'
```
MiYanni: This should still be a preview release. Please revert. We agreed that it would not GA until if it determined to be a separate module or not.

MiYanni: I was incorrect in this statement.

vivlingaiah: As per our conversation with Scott, there won't be a "*-preview" in the version in the end. but it should be less than 1.0.0. We've already released 0.9.0-preview. So. it should be greater that without "preview" suffix. Once we determine a location and release it from there, it should be released as 1.0.0. This module is not preview for all practical purposes. It is like a standalone stable/GA module.

Issue #8235, opened 2019-01-04
 Get-AzureRmRoleAssignment (AzureRM.Resources 6.0.2) doesn't respect -ErrorAction 


acmcg:
### Description

Get-AzureRmRoleAssignment (AzureRM.Resources 6.0.2) doesn't respect -ErrorAction 

### Script/Steps for Reproduction

Note: Commands are being run against AzureStack 

Run the commandlet against a non existent policy assignment with ErrorAction set to SilentlyContinue
get-AzureRmPolicyAssignment -Name "BogusPolicyAssignment" -Scope "/subscriptions/$SubscriptionId/resourceGroups/$resourceGroup" -ErrorAction SilentlyContinue

### Module Version

```
Get-Module -Name AzureRM.Resources  -ListAvailable

ModuleType Version    Name                                ExportedCommands
---------- -------    ----                                ----------------
Script     6.0.2      AzureRM.Resources                   {Get-AzureRmProviderOperation, Remove-AzureRmRoleAssignment, Get-AzureRmRoleAssignment, New-AzureRmRoleAssignment...}
```

### Environment Data

```
Name                           Value
----                           -----
PSVersion                      5.1.17134.407
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.17134.407
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

### Debug Output

```
DEBUG: 5:01:43 PM - GetAzurePolicyAssignmentCmdlet begin processing with ParameterSet 'GetPolicyAssignmentName'.
DEBUG: 5:01:43 PM - using account id 'andrew.mcgregor@hpeasicsingapore.onmicrosoft.com'...
DEBUG: [Common.Authentication]: Authenticating using Account: 'andrew.mcgregor@hpeasicsingapore.onmicrosoft.com', environment: 'AzureStackUser', tenant: '62574f56-4554-49e3-9194-34570282a2e3'
DEBUG: Microsoft.IdentityModel.Clients.ActiveDirectory Information: 2 :
DEBUG: 01/04/2019 06:01:43:  - TokenCache: Serializing token cache with 11 items.

DEBUG: Microsoft.IdentityModel.Clients.ActiveDirectory Information: 2 :
[...]

PR #8399, opened 2019-01-24
Fix ErrorAction issue

maddieclayton:
## Description
https://github.com/Azure/azure-powershell/issues/8235

## Checklist

### DIFF
--- a/src/Resources/ResourceManager/Implementation/CmdletBase/ResourceManagerCmdletBase.cs
+++ b/src/Resources/ResourceManager/Implementation/CmdletBase/ResourceManagerCmdletBase.cs
@@ -423,8 +423,7 @@ namespace Microsoft.Azure.Commands.ResourceManager.Cmdlets.Implementation
                 var errorResponseException = capturedException.SourceException as ErrorResponseMessageException;
                 if (errorResponseException != null)
                 {
-                    this.WriteError(errorResponseException.ToErrorRecord());
-                    return;
+                    throw errorResponseException;
                 }
 
                 var aggregateException = capturedException.SourceException as AggregateException;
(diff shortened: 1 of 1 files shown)


src/Resources/ResourceManager/Implementation/CmdletBase/ResourceManagerCmdletBase.cs
```diff
@@ -423,7 +423,8 @@ private void HandleException(ExceptionDispatchInfo capturedException)
                 var errorResponseException = capturedException.SourceException as ErrorResponseMessageException;
                 if (errorResponseException != null)
                 {
-                    this.ThrowTerminatingError(errorResponseException.ToErrorRecord());
+                    this.WriteError(errorResponseException.ToErrorRecord());
```
markcowl: So, there are cases in which we will want to throw a terminating exception (such as PipelineStoppedException)

If this is extending AzurePSCmdlet and usign the default ProcessRecord implementation, we can make use of it's default error handling, and simply throw these exceptions (it will write them to the error stream): https://github.com/Azure/azure-powershell-common/blob/master/src/Common/AzurePSCmdlet.cs#L735.

Otherwise, we want to throw any exception that is a terminating exception and write others to the error stream. We can use the inherited IsTerminatingException method to do this.

maddieclayton:
@markcowl Looks like I can throw and the ErrorAction preference is working corrects (this base class inherits from AzurePSCmdlet and does not override ProcessRecord). Do we still need unit tests specifically for Resources now that we are using the common implementation?

markcowl:
> @markcowl Looks like I can throw and the ErrorAction preference is working corrects (this base class inherits from AzurePSCmdlet and does not override ProcessRecord). Do we still need unit tests specifically for Resources now that we are using the common implementation?

I believe we already have tests for the default implementation of this, so this is fine

PR #8452, opened 2019-01-30
[LogicApp] Update LA SDK version

refortie:
## Description
Updating the referenced SDK to 4.1.0 and updating the necessary files based on changes in the SDK

## Checklist

### DIFF
--- a/src/LogicApp/LogicApp.Test/ScenarioTests/WorkflowTests.ps1
+++ b/src/LogicApp/LogicApp.Test/ScenarioTests/WorkflowTests.ps1
@@ -315,6 +315,16 @@ function Test-ValidateLogicApp
 	$definition = [IO.File]::ReadAllText($definitionFilePath)
 	Test-AzLogicApp -ResourceGroupName $resourceGroup.ResourceGroupName -Name $workflowName -Location $locationName -Definition $definition -ParameterFilePath $parameterFilePath
 
+	# Test 3: Failure for an invalid definition.
+	try
+	{
+		Test-AzLogicApp -ResourceGroupName $resourceGroupName -Name $workflowName -Location $locationName -Definition '{}'
+	}
+	catch
+	{
+		Assert-AreEqual $_.Exception.Message "The request content is not valid and could not be deserialized: 'Required property '`$schema' not found in JSON. Path 'properties.definition', line 4, position 20.'."
+	}
+
 	Remove-AzLogicApp -ResourceGroupName $resourceGroupName -Name $workflowName -Force
 }
 
--- a/src/LogicApp/LogicApp.Test/ScenarioTests/IntegrationAccountGeneratedIcnTests.ps1
+++ b/src/LogicApp/LogicApp.Test/ScenarioTests/IntegrationAccountGeneratedIcnTests.ps1
@@ -13,6 +13,7 @@
 # ----------------------------------------------------------------------------------
 
 # Constants used in the tests.
+# To see up new resources for the tests, see "how-to-create-a-session.md"
 $SESSION_RG = "flowrg"
 $SESSION_IA = "PS-Tests-Sessions"
 $AGREEMENT_X12_Name = "PS-X12-Agreement"
--- a/src/LogicApp/LogicApp.Test/ScenarioTests/IntegrationAccountReceivedIcnTests.ps1
+++ b/src/LogicApp/LogicApp.Test/ScenarioTests/IntegrationAccountReceivedIcnTests.ps1
@@ -13,6 +13,7 @@
 # ----------------------------------------------------------------------------------
 
 # Constants used in the tests.
+# To see up new resources for the tests, see "how-to-create-a-session.md"
 $SESSION_RG = "flowrg"
 $SESSION_IA = "PS-Tests-Sessions"
 $AGREEMENT_X12_Name = "PS-X12-Agreement"
--- a/src/LogicApp/LogicApp.Test/ScenarioTests/how-to-create-a-session.md
+++ b/src/LogicApp/LogicApp.Test/ScenarioTests/how-to-create-a-session.md
@@ -1,0 +1,110 @@
+The session integration account artifact is unique amongst the rest of the artifacts. It is a byproduct of connecting 2 partners with an agreement and sending a message from partner 1 to partner 2. The session is required for the RecievedIcn and GeneratedIcn tests.
+
+## Regenerate Sessions Using Existing Logic App
+There should be a Logic App already in the location specified by the test constants. If it is still there, all you need to do is trigger the run and it should create everything that is needed. If it isn't there and the sessions don't exist, you'll need to create the Logic App yourself.
+
+## How to Create a new Logic App to generate Sessions
+
+1. Create an Integration Account
+2. Create 2 partners. Use the ZZ qualifier and give each partner a different value. You can use "ZZ" for one and "AA" for the other as an example
+3. Create an X12 agreement
+4. Create an Edifact agreement
+5. Create a Logic App in the same sub, resource group, and location to keep things tidy
+6. Add a Request trigger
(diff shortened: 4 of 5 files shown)


maddieclayton:
@refortie You'll need to add the updated recorded files because of the API version bump.

refortie:
@maddieclayton Yeah. I'm getting an `One or more errors occurred. (AADSTS50012: Invalid client secret is provided.` when trying to re-record them. I've never seen this before and was attempting to find out if this local to my machine only or not

refortie:
@maddieclayton I got this issue when using the old recorded session when trying to do new recording

`Message: System.MissingMethodException : Method not found: 'Void Newtonsoft.Json.JsonSerializerSettings.set_TypeNameAssemblyFormat(System.Runtime.Serialization.Formatters.FormatterAssemblyStyle)'.`

Is this a known thing or something specific to my machine? I've been recording tests for the last couple of months and never had this come up

maddieclayton:
@refortie If you pulled from master recently, this is a known issue.  The fix is in this PR: https://github.com/Azure/azure-powershell/pull/8448 which has been approved and will be merged as soon as it passes CI.  In the meantime (which should be maximum of one hour) you can use this work around:

```
Single test
dotnet test {path_to_solution}\{solution_name}.sln --filter DisplayName={test_namespace}.{test_class}.{test_name} --framework netcoreapp2.0

All tests in a class
dotnet test {path_to_solution}\{solution_name}.sln --filter ClassName={test_namespace}.{test_class} --framework netcoreapp2.0

All check-in tests in a solution
dotnet test {path_to_solution}\{solution_name}.sln --filter AcceptanceType=CheckIn&RunType!=DesktopOnly --framework netcoreapp2.0

All tests in a solution
dotnet test {path_to_solution}\{solution_name}.sln --framework netcoreapp2.0
```
You will need to set the connection string by hand, so it might be worth waiting the one hour for the PR to get merged, then pulling from upstream.

maddieclayton:
Update: Fix has been merged, please pull using "git pull upstream master" and you should no longer see this issue.

maddieclayton:
@refortie Can you explain why you need to many changes to the tests? If there are no breaking changes in the SDK, why are entire sections of the tests rewritten (and new harded coded values added, which makes the test very hard to rerecord)?

refortie:
For the chunks that I had to modify:
- New-AzResource doesn't appear to be a command we can call anymore, so any tests that used that needed to be modified. I would have preferred to remove it in a different PR, but it didn't seem worth the effort to modify the tests just to strip them out once this is merged. As a parallel, all the tests that used New-AzResource were to either provision an App Service plan, which isn't how we do Logic Apps anymore and haven't for a couple years I think. The other was to create a data plane artifact which was causing us to break fundamentals because we weren't documenting it, every time we recorded tests. No customer was ever supposed to manually create it, it is a byproduct of other processes, and I think was only exposed, so that we could run these tests and access will be blocked at some point.
- It seems like when whoever changed over the file paths to use Join-Path that they missed a number of them (or didn't realize that you needed to do both pieces of the folder structure)
- I also had to change a couple of spots where I think you had made resources (like a keyvault), I don't have access to your resources and so I couldn't record new versions without providing my own. Also, those resources probably should live in our sub anyway.

src/LogicApp/LogicApp.Test/ScenarioTests/WorkflowTests.ps1
```diff
@@ -215,9 +195,10 @@ function Test-CreateAndGetLogicAppUsingDefinitionWithActions
 	Assert-True { $workflow5.Length -ge 1 }
 
 	# Test 6: Get workflow with just name parameter
-	$workflow6 = Get-AzLogicApp -Name $workflowName
-	Assert-NotNull $workflow6
-	Assert-True { $workflow6.Length -ge 1 }
+	# refortie (1/31/19): TODO Test is disabled for now, we only return the first page on get logic app by subscription
+	# $workflow6 = Get-AzLogicApp -Name $workflowName
```
maddieclayton: What happened here? We should be testing the list by subscription call

src/LogicApp/LogicApp.Test/ScenarioTests/WorkflowTests.ps1
```diff
@@ -302,8 +280,8 @@ function Test-UpdateLogicApp
 	try
 	{
 		$workflowName = "82D2D842-C312-445C-8A4D-E3EE9542436D"
-		$definitionFilePath = Join-Path $TestOutputRoot "\Resources\TestSimpleWorkflowTriggerDefinition.json"
-		Set-AzLogicApp -ResourceGroupName $resourceGroupName -Name $workflowName -AppServicePlan $planName -DefinitionFilePath $definitionFilePath -Force
```
maddieclayton: What happens if the user sends in the AppServicePlan parameter? It is still an option in the cmdlet.

src/LogicApp/LogicApp.Test/ScenarioTests/WorkflowTests.ps1
```diff
@@ -338,18 +314,8 @@ function Test-ValidateLogicApp
 	# Test 2: Using definition object and parameter file.
 	$definition = [IO.File]::ReadAllText($definitionFilePath)
 	Test-AzLogicApp -ResourceGroupName $resourceGroup.ResourceGroupName -Name $workflowName -Location $locationName -Definition $definition -ParameterFilePath $parameterFilePath
-    
-	# Test 3: Failure for an invalid definition.
-	try
-	{
-		Test-AzLogicApp -ResourceGroupName $resourceGroupName -Name $workflowName -Location $locationName -Definition '{}'
```
maddieclayton: Why are you removing this test?

maddieclayton:
@refortie The comments above make sense.  Only extra comment that I have is that you should add comments above the hardcoded value stating how to properly set up your subscription to rerecord these tests.

src/LogicApp/LogicApp.Test/ScenarioTests/WorkflowTests.ps1
refortie: So I ran into a problem here where the underlying PS cmdlet only returns the first page of results and in the test sub I am using, we are way beyond that, the `Assert-NotNull $workflow6` fails (I'm guessing most customers don't have thousands of LA in their sub). I was trying to keep the changes small and wanted to change that in a subsequent PR.

src/LogicApp/LogicApp.Test/ScenarioTests/WorkflowTests.ps1
refortie: So I asked another developer who's been here longer. At one point we allowed App Service plans as an option, but we have long since turned off that functionality (as of like 2 years ago). We still allow a customer to specify one and we record it, but from what I can find out, the API doesn't _do_ anything with that value

src/LogicApp/LogicApp.Test/ScenarioTests/WorkflowTests.ps1
refortie: Based on the error message I was getting back, I thought it was a test case that would need to be handled in the parameters itself (like fixing the input of the cmdlet to accept an actual class instead of `Object`). Looking deeper, that doesn't seem to be the case. It's just a bad error message.

I've added the test back in with the new error message

refortie:
For the tests that take special work to set up the state properly: would you like those instructions inline/in file or in an external file (like perhaps in the documentation folder)? There are quite a few steps to recreate it and it might be a long in file comment

maddieclayton:
@refortie An external file sounds perfect - please put this new file in the ScenarioTests folder and link to it inline in the ps1 file.

src/LogicApp/LogicApp.Test/ScenarioTests/WorkflowTests.ps1
refortie: I do. I've got in my notes as something that needs to be fixed

src/LogicApp/LogicApp.Test/ScenarioTests/WorkflowTests.ps1
refortie: Will do

refortie:
Added the readme file

Issue #7759, opened 2018-11-06
Get-AzIntegrationAccountMap returns only maps in XSLT format


ajhall:
### Description

`Get-AzIntegrationAccountMap` only supports XSLT format and does not provide a way to work with other formats. Maps in Liquid format, for example, can be added via the portal or the REST API but are inaccessible via Azure PowerShell.

When listing maps, the list is filtered on the following line to show only XSLT:

https://github.com/Azure/azure-powershell/blob/b61106c491fc624776f3be3f933879c8cf7ffb36/src/ResourceManager/LogicApp/Commands.LogicApp/Utilities/IntegrationAccountMapOperations.cs#L105

The `-MapName` parameter can be used to retrieve a specific map, but retrieving a Liquid format map results in an "Unable to deserialize the response" error.

I would understand the limitation if the content of the map was included in the response and needed to be deserialized into a PowerShell object, but the response actually just includes a `contentLink` object with a URI inside to get the map content. The only significant difference in the API response compared to an XSLT map is the `mapType` property.

### Script/Steps for Reproduction

First, create an integration account in the Azure Portal and add a map in Liquid format. (This is unsupported in Azure PowerShell). Optionally, add another map in XSLT format to illustrate that the list of maps is filtered to XSLT.

Run `Get-AzIntegrationAccountMap` without specifying `-MapName`. Only XSLT maps are returned.
```powershell
Get-AzIntegrationAccountMap `
 -ResourceGroupName MyIA-rg `
 -Name MyIA
```

Run `Get-AzIntegrationAccountMap` with a `-MapName` specified to specifically request the Liquid map. The cmdlet will return "Unable to deserialize the response."

```powershell
Get-AzIntegrationAccountMap `
 -ResourceGroupName MyIA-rg `
 -Name MyIA `
 -MapName MyLiquidMap
```

### Module Version

```powershell
Get-Module -ListAvailable

ModuleType Version    Name                                PSEdition ExportedCommands
---------- -------    ----                                --------- ----------------
Script     0.4.0      Az.Aks                              Core,Desk {Get-AzAks, New-AzAks, Remove-AzAks, Import-AzAksCredential...}
Script     0.4.0      Az.AnalysisServices                 Core,Desk {Resume-AzAnalysisServicesServer, Suspend-AzAnalysisServicesServer, Get-AzAnalysisServicesServer, Remove-AzAnalysisServicesServer...}
[...]

Issue #6261, opened 2018-05-21
Unable to deserialize the response


ToMakeSense:
I have encountered this issue on VSTS for Azure resources deployment with azure PowerShell recently, the script doesn’t work for one resource group, but it works for others, strange.  It seemed the same issue as #5835 , I  tried the solution given in my local machine, it's not work for me.
```powershell
PS C:\WINDOWS\system32> Get-AzureRmIntegrationAccount -Name XXXXX -ResourceGroupName XXXXX
Get-AzureRmIntegrationAccount : Unable to deserialize the response.
At line:1 char:1
+ Get-AzureRmIntegrationAccount -Name nok-ipaas-shared-we-iac-01 -Resou ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : CloseError: (:) [Get-AzureRmIntegrationAccount], SerializationException
    + FullyQualifiedErrorId : Microsoft.Azure.Commands.LogicApp.Cmdlets.GetAzureIntegrationAccountCommand
```

maddieclayton:
@ToMakeSense Can you give us the current modules you are running by giving us the output of "Get-Module -ListAvailable"?  It looks like some people were not seeing this issue anymore in AzureRM 6.0.1, so if you have not updated to 6.0.1 can you update and let me know if you still see this issue?

ToMakeSense:
I had updated to the latest version
```powershell
PS C:\WINDOWS\system32> Get-Module -name AzureRM -ListAvailable


    Directory: C:\Program Files\WindowsPowerShell\Modules


ModuleType Version    Name                                ExportedCommands                                                                                 
---------- -------    ----                                ----------------                                                                                 
Script     6.0.1      AzureRM                                                                                                                              
Script     5.6.0      AzureRM                                                                                                                              
Script     5.1.1      AzureRM      
```

maddieclayton:
@ToMakeSense Thanks for the confirmation.  Can you print the output of running ```Get-AzureRmIntegrationAccount``` as well as the output of Resolve-AzureRmError (run after you get this deserialization error).  This will help us narrow down where the error is coming from. Thanks!

ToMakeSense:
```powershell
PS C:\WINDOWS\system32> Get-AzureRmIntegrationAccount
[...]

PR #8482, opened 2019-02-06
[LogicApp] Expose new enum types

refortie:
## Description

Explicitly added in the enum types that were implemented in the last SDK update. Specifically 3 new map types and 1 new integration account sku were added.

* Missing Map Types: #7759
* Missing Basic Sku: #6261

## Checklist


vladimir-shcherbakov:
@refortie For public API changes to cmdlets the changes should go through a [cmdlet design review](https://github.com/Azure/azure-powershell-cmdlet-review-pr)

src/LogicApp/LogicApp/Cmdlets/IntegrationAccount/NewAzureIntegrationAccountCommand.cs
```diff
@@ -51,6 +51,7 @@ public class NewAzureIntegrationAccountCommand : LogicAppBaseCmdlet
 
         [Parameter(Mandatory = false, HelpMessage = "The integration account SKU.",
             ValueFromPipelineByPropertyName = true)]
+        [ValidateSet("Free", "Basic", "Standard", IgnoreCase = true)]
```
maddieclayton: If a user provides a string outside these three values will the cmdlet always error? If it previously would not always error, this is a breaking change.  However, if the cmdlet would ALWAYS error when there was a SKU provided outside these three string values, this is fine.

src/LogicApp/LogicApp/Cmdlets/IntegrationAccount/UpdateAzureIntegrationAccountCommand.cs
```diff
@@ -51,6 +51,7 @@ public class UpdateAzureIntegrationAccountCommand : LogicAppBaseCmdlet
 
         [Parameter(Mandatory = false, HelpMessage = "The integration account SKU.",
             ValueFromPipelineByPropertyName = true)]
+        [ValidateSet("Free", "Basic", "Standard", IgnoreCase = true)]
```
maddieclayton: same comment here about errors/breaking changes

src/LogicApp/LogicApp/Cmdlets/IntegrationAccount/UpdateAzureIntegrationAccountMapCommand.cs
```diff
@@ -135,16 +112,13 @@ public override void ExecuteCmdlet()
...
-            {
-                integrationAccountMapCopy.ContentType = this.contentType;
-            }
-
             if (!string.IsNullOrEmpty(this.MapType))
             {
                 integrationAccountMapCopy.MapType = this.MapType;
             }
 
+            integrationAccountMapCopy.ContentType = this.MapType.Equals("liquid", StringComparison.CurrentCultureIgnoreCase) ? "text/plain" : "application/xml";
```
maddieclayton: Is the actual ContentType parameter used anymore? I don't see any usage.

maddieclayton:
@refortie No need to fill out a cmdlet review as all that is being changed is validate sets and internal logic.

src/LogicApp/LogicApp/Cmdlets/IntegrationAccount/NewAzureIntegrationAccountCommand.cs
refortie: It would always error even without this ValidateSet, this is just documenting the existing limitations. I would assume that adding in the ValdiateSet would change the error message returned. Before the API would return an error stating that the sku isn't allowed and now the PS cmdlet will. Just wanted to confirm that is alright?

src/LogicApp/LogicApp/Cmdlets/IntegrationAccount/UpdateAzureIntegrationAccountCommand.cs
refortie: My response is the same as the other comment :)

src/LogicApp/LogicApp/Cmdlets/IntegrationAccount/UpdateAzureIntegrationAccountMapCommand.cs
refortie: It isn't. I debated whether accepting the customer's input would override what I can infer. I know that if the customer provides a liquid file it is 1 content type, otherwise it is the other. Should I use the customer input instead and override if only is isn't provided?

src/LogicApp/LogicApp/Cmdlets/IntegrationAccount/NewAzureIntegrationAccountCommand.cs
maddieclayton: Yup! As long as you aren't erroring in cases where no error was previously thrown before, it is fine for a different error to be thrown (and it is nice not to have to make the call to the server and that the valid values are documented).

src/LogicApp/LogicApp/Cmdlets/IntegrationAccount/UpdateAzureIntegrationAccountMapCommand.cs
maddieclayton: It should be fine to no longer use the user's input.  What you should do is deprecate the ContentType parameter because you are no longer using it (https://github.com/Azure/azure-powershell/blob/master/documentation/breaking-changes/breaking-changes-attribute-help.md#a-parameter-is-being-deprecated).

refortie: I've marked both New and Set to be deprecated so that they match (instead of just Set)

PR #8532, opened 2019-02-13
Get Server Variables Cmdlets Powershell Changes

nirkum18:
## Description

Application Gateway supports certain server variables that can be used for Header CRUD feature. There are lists of suggested Request and Response Headers. These commandlets will allow the user to fetch the supported server variables' list and the suggested request and response headers' lists.

## Checklist


nirkum18:
Design Review: [Azure/azure-powershell-cmdlet-review-pr#224](https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/224)

src/Network/Network.Test/Network.Test.csproj
```diff
@@ -14,7 +14,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.Graph.RBAC" Version="3.4.0-preview" />
-    <PackageReference Include="Microsoft.Azure.Management.Network" Version="19.8.1-preview" />
+    <PackageReference Include="Microsoft.Azure.Management.Network" Version="20.0.0-preview" />
```
number213: You need to commit local nuget package (the one in tools\LocalFeed)

nirkum18: @number213 Nuget packages have been added.

cormacpayne:
@number213 @EvgenyAgafonchikov should we open a separate PR to update the version of the Network SDK being used in PowerShell? It looks like all build failures in both Jenkins and Travis are resulting from the new version.

number213:
@cormacpayne that's right, we've just release new SDK and I've opened PR to update it (https://github.com/Azure/azure-powershell/pull/8545). After it gets merged, I'll rebase this PR.

number213:
Rebased after SDK update and removed unneeded models (the server returns array of strings), @cormacpayne please take a look.

number213:
@cormacpayne gentle ping

src/Network/Network.Test/ScenarioTests/ApplicationGatewayTests.ps1
```diff
@@ -1554,3 +1554,27 @@ function Test-ApplicationGatewayCRUDSubItems2
 		Clean-ResourceGroup $rgname
 	}
 }
+
+function Test-AvailableServerVariables
+{
+	$result = Get-AzApplicationGatewayAvailableServerVariables
```
cormacpayne: @nirkum18 for the tests you're adding, we need proper test setup so that other users will be able to record this test (including the Network test recording job that we use whenever an SDK is released with a new API version) -- running just these cmdlets will fail if the user does not have the appropriate setup in their subscription, so you should do the necessary resource construction at the beginning of the test and deconstruction at the end of the test. Please see the other tests in this file for an example of what this should look like.

src/Network/Network/ApplicationGateway/GetAzureApplicationGatewayAvailableRequestHeaders.cs
```diff
@@ -0,0 +1,39 @@
...
+using Microsoft.Azure.Commands.Network.Models;
+using Microsoft.Azure.Management.Network;
+using Microsoft.WindowsAzure.Commands.Common;
+using System.IO;
+using System.Management.Automation;
+using System.Reflection;
+
+namespace Microsoft.Azure.Commands.Network
+{
+    [Cmdlet("Get", ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "ApplicationGatewayAvailableRequestHeaders"), OutputType(typeof(string))]
```
cormacpayne: @nirkum18 we cannot return a `string` object from cmdlets as this will affect piping scenarios. We should create new types for each property that we can return instead that contains properties that the user would find useful.

tools/StaticAnalysis/Exceptions/Az.Network/SignatureIssues.csv
```diff
@@ -293,3 +293,6 @@
...
 "Microsoft.Azure.PowerShell.Cmdlets.Network.dll","Microsoft.Azure.Commands.Network.InvokeNetworkWatcherNetworkConfigurationDiagnosticCommand","Invoke-AzNetworkWatcherNetworkConfigurationDiagnostic","1","8100","Invoke-AzNetworkWatcherNetworkConfigurationDiagnostic Does not support ShouldProcess but the cmdlet verb Invoke indicates that it should.","Determine if the cmdlet should implement ShouldProcess and if so determine if it should implement Force / ShouldContinue"
+"Microsoft.Azure.PowerShell.Cmdlets.Network.dll","Microsoft.Azure.Commands.Network.GetAzureApplicationGatewayAvailableRequestHeaders","Get-AzApplicationGatewayAvailableRequestHeaders","1","8400","Get-AzApplicationGatewayAvailableRequestHeaders uses the noun 'AzApplicationGatewayAvailableRequestHeaders', which does not follow the enforced naming convention of using a singular noun for a cmdlet name.","Consider using a singular noun for the cmdlet name."
```
cormacpayne: @nirkum18 these exceptions shouldn't be suppressed, you should have the noun name for the cmdlets be singular to follow the singularity rule in PowerShell: https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/azure-powershell-design-guidelines.md#noun-singularity

src/Network/Network.Test/ScenarioTests/ApplicationGatewayTests.ps1
number213: @cormacpayne these tests don't need any setup, all they do is query backend for a list of predefined values

src/Network/Network/ApplicationGateway/GetAzureApplicationGatewayAvailableRequestHeaders.cs
number213: @cormacpayne the server returns an array of strings for all of those new cmdlets (e.g. `[ "A-IM", "Accept", "Accept-Charset"]`), how should we proceed here? I don't think there's a need for custom types here

tools/StaticAnalysis/Exceptions/Az.Network/SignatureIssues.csv
number213: @cormacpayne new cmdlets are similar to `Get-AzApplicationGatewayAvailableSslOptions` and `Get-AzApplicationGatewayAvailableWafRuleSets`. They always return array of predefined values, so I think it falls under

> Plural parameter names should be used only in those cases where the value of the parameter is always a multiple-element value.

nirkum18: @cormacpayne as @number213 mentioned above, these commandlets return an array of strings, so it should fall under plural naming.

number213:
@cormacpayne addressed your feedback, please take a look

src/Network/Network/ApplicationGateway/GetAzureApplicationGatewayAvailableRequestHeaders.cs
cormacpayne: @nirkum18 @number213 the reason for advising against `string` output types is because it can cause issues with parameter binding when used in piping scenarios -- it is much easier in PowerShell to use concrete objects and pass those around then using strings. As mentioned in the design review, the ideal pattern would be to bundle these cmdlets into one, introducing a new concrete type that has the result of these cmdlets as properties of the type, and users would only have to call one cmdlet to get an object back that has all of these values. It's mentioned that these cmdlets are following the pattern of existing cmdlets, but those cmdlets return concrete types from their API calls, whereas these calls all return a list of strings, so it would be beneficial to bundle them together for a better user experience.

tools/StaticAnalysis/Exceptions/Az.Network/SignatureIssues.csv
cormacpayne: @nirkum18 @number213 I either misspoke or PowerShell recently updated their patterns 😁 It looks like they advise following the singular noun rule for cmdlets in _all_ cases:

From the [_Strongly Encouraged Development Guidelines_](https://docs.microsoft.com/en-us/powershell/developer/cmdlet/strongly-encouraged-development-guidelines#use-a-specific-noun-for-a-cmdlet-name-sd01):

_To enhance the user experience, the noun that you choose for a cmdlet name should be singular. For example, use the name Get-Process instead of Get-Processes. It is best to follow this rule for all cmdlet names, even when a cmdlet is likely to act upon more than one item._

src/Network/Network/ApplicationGateway/GetAzureApplicationGatewayAvailableRequestHeaders.cs
number213: Sounds reasonable to me. @nirkum18, this is your feature, so please consider Cormac's feedback

tools/StaticAnalysis/Exceptions/Az.Network/SignatureIssues.csv
nirkum18: @cormacpayne I have updated the PR with singular naming convention.

number213:
@nirkum18 you don't need to add Microsoft.Azure.Management.Network.20.0.0-preview.nupkg anymore

nirkum18:
> @nirkum18 you don't need to add Microsoft.Azure.Management.Network.20.0.0-preview.nupkg anymore

@number213 i have removed the files now

tools/StaticAnalysis/Exceptions/Az.Network/SignatureIssues.csv
cormacpayne: @nirkum18 you can revert the changes made in this file now that this won't be flagged

src/Network/Network/ApplicationGateway/GetAzureApplicationGatewayAvailableRequestHeaders.cs
cormacpayne: @nirkum18 any update on this?

nirkum18: @cormacpayne The same cmdlet behavior is implemented in the service side, swagger and powershell review doc. So we would like to keep the cmdlets consistent at all places.

tools/StaticAnalysis/Exceptions/Az.Network/SignatureIssues.csv
nirkum18: @cormacpayne removed these changes

cormacpayne:
Closing this PR in favor of https://github.com/Azure/azure-powershell/pull/8705

PR #8534, opened 2019-02-13
[Compute] update for Update-AzImage and Get-AzVmssVM

hyonholee:
1. Add Tag and ResourceId parameters to Update-AzImage cmdlet.

https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/231

2. Get-AzVmssVM without instance ID and with InstanceView can list VMSS VMs with instance view.

## Checklist


hyonholee:
When I see the build failure, the static analysis is complaining that some parameter sets are removed from a bunch of cmdlets.
However, I did not remove any parameter set.
I just change starting value of positional parameters from 1 to 0.
Could you investigate it?

maddieclayton:
The current version of our breaking change tool will show the “The parameter set 'FriendMethod' for cmdlet 'Set-AzVmssVM' has been removed" whenever a syntax change is made to a parameter set so that the tool cannot match up the newly changed parameter sets to the older parameter sets (I think in this case enough of the positions were changed that it no longer could match the original parameter set).  Feel free to suppress these messages, and I will do a close check to make sure there are no actual breaking changes.

src/Compute/Compute.Test/ScenarioTests/ImageTests.ps1
```diff
@@ -129,11 +129,10 @@ function Test-Image
         Assert-AreEqual 1 $images.Count;
 
         # Update Image Tag
-        $image = Get-AzImage -ResourceGroupName $rgname -ImageName $imageName;
-        $image.Tags =  New-Object "System.Collections.Generic.Dictionary``2[System.String,System.String]";
-        $image.Tags.Add("test1", "testval3");
-        $image.Tags.Add("test2", "testval4");
-        Update-AzImage -ResourceGroupName $rgname -ImageName $imageName -Image $image;
+        #Update-AzImage -ResourceGroupName $rgname -ImageName $imageName -Tag @{test1 = "testval3"; test2 = "testval4"};
```
maddieclayton: Why are all these lines commented out?

src/Compute/Compute/Generated/Image/ImageCreateOrUpdateMethod.cs
```diff
@@ -82,7 +83,7 @@ public override void ExecuteCmdlet()
         public SwitchParameter AsJob { get; set; }
     }
 
-    [Cmdlet(VerbsData.Update, ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "Image", DefaultParameterSetName = "DefaultParameter", SupportsShouldProcess = true)]
+    [Cmdlet(VerbsData.Update, ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "Image", DefaultParameterSetName = "ObjectParameter", SupportsShouldProcess = true)]
```
maddieclayton: DefaultParameter should always be the default parameter set - it doesn't make sense to prompt for an object instead of a string if no parameters are provided.

src/Compute/Compute/Generated/Image/ImageCreateOrUpdateMethod.cs
```diff
@@ -108,27 +125,46 @@ public override void ExecuteCmdlet()
...
         [Parameter(
             ParameterSetName = "DefaultParameter",
-            Position = 2,
+            Position = 1,
             Mandatory = true,
             ValueFromPipelineByPropertyName = true)]
         [Alias("Name")]
         public string ImageName { get; set; }
 
+        [CmdletParameterBreakingChange("Image", "Image parameter will become optional when ResourceGroupName and ImageName parameters are given")]
```
maddieclayton: I believe the breaking change is that Image parameter will be removed for the ResourceGroupName/ImageName parameter set in an upcoming breaking change release.

src/Compute/Compute/Generated/Image/ImageCreateOrUpdateMethod.cs
hyonholee: @maddieclayton  Since I changed Image from mandatory to optional for DefaultParameter set, if I set "DefaultParameter" set to default, then Update-AzImage with -Image object throw an error because it also looks for resource group name and image name.  So, I changed default parameter set to "ObjectParameter" set until we can remove Image parameter from "DefaultParameter" set in the upcoming breaking change release.

maddieclayton: This is fine for now then

PR #8537, opened 2019-02-14
Initial PR for Blueprints PowerShell Module

filizt:
## Description

This is the first PR for the Blueprints PowerShell module. Blueprints is a Azure service that is in public preview. We're looking to get first line of reviews for our PowerShell module implementations.

Blueprints powershell cmdlets design review: https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/179

## Checklist

### DIFF
--- a/src/Blueprint/Blueprint/help/Get-AzBlueprint.md
+++ b/src/Blueprint/Blueprint/help/Get-AzBlueprint.md
@@ -72,7 +72,7 @@ Id                   : /subscriptions/00000000-1111-0000-1111-000000000000/provi
 DefinitionLocationId : 00000000-1111-0000-1111-000000000000
 Versions             : {1.0}
 Description          : Powershell test blueprint
-LastModified         : 2019-02-01
+TimeCreated          : 2019-02-01
 TargetScope          : Subscription
 Parameters           : {storageData_storageAccountType, storageData_location, allowedlocations_listOfAllowedLocations}
 ResourceGroups       : ResourceGroup
@@ -89,7 +89,7 @@ Name                 : PS-MG-BlueprintDefinition
 Id                   : /providers/Microsoft.Management/managementGroups/myManagementGroupId/providers/Microsoft.Blueprint/blueprints/PS-MG-BlueprintDefinition
 DefinitionLocationId : myManagementGroupId
 Versions             : {1.0, 2.0, 3.0, 4.0}
-LastModified         : 2019-03-04
+TimeCreated          : 2019-03-04
 TargetScope          : Subscription
 Parameters           : {enforcetaganditsvalue_tagName, enforcetaganditsvalue_tagValue, [Usergrouporapplicationname]:Contributor_RoleAssignmentName,
                        [Usergrouporapplicationname]:Owner_RoleAssignmentName}
@@ -107,7 +107,7 @@ Id                   : /subscriptions/00000000-1111-0000-1111-000000000000/provi
 DefinitionLocationId : 00000000-1111-0000-1111-000000000000
 Versions             : {1.0}
 Description          : Powershell test blueprint
-LastModified         : 2019-02-01
+TimeCreated          : 2019-02-01
 TargetScope          : Subscription
 Parameters           : {storageData_storageAccountType, storageData_location, allowedlocations_listOfAllowedLocations}
 ResourceGroups       : ResourceGroup
--- a/src/Blueprint/Blueprint/Blueprint.format.ps1xml
+++ b/src/Blueprint/Blueprint/Blueprint.format.ps1xml
@@ -43,8 +43,8 @@
                 </ItemSelectionCondition>
               </ListItem>
               <ListItem>
-                <Label>LastModified</Label>
-                <ScriptBlock>$_.Status.LastModified.ToString("yyyy-MM-dd")</ScriptBlock>
+                <Label>TimeCreated</Label>
+                <ScriptBlock>$_.Status.TimeCreated.ToString("yyyy-MM-dd")</ScriptBlock>
               </ListItem>
               <ListItem>
                 <Label>TargetScope</Label>
@@ -97,8 +97,8 @@
                 </ItemSelectionCondition>
               </ListItem>
               <ListItem>
-                <Label>LastModified</Label>
-                <ScriptBlock>$_.Status.LastModified.ToString("yyyy-MM-dd")</ScriptBlock>
+                <Label>TimeCreated</Label>
+                <ScriptBlock>$_.Status.TimeCreated.ToString("yyyy-MM-dd")</ScriptBlock>
               </ListItem>
               <ListItem>
                 <Label>TargetScope</Label>
--- a/src/Blueprint/Blueprint/Common/BlueprintClient.cs
+++ b/src/Blueprint/Blueprint/Common/BlueprintClient.cs
@@ -91,7 +91,7 @@ namespace Microsoft.Azure.Commands.Blueprint.Common
             {
                 if (latest == null)
                     latest = blueprint;
-                else if (CompareDates(blueprint.Status.LastModified, latest.Status.LastModified) > 0)
+                else if (CompareDates(blueprint.Status.TimeCreated, latest.Status.TimeCreated) > 0)
                     latest = blueprint;
             }
 


src/Blueprint/Blueprint/ChangeLog.md
```diff
@@ -0,0 +1,21 @@
+<!--
+    Please leave this section at the top of the change log.
```
majastrz: >Please [](start = 4, length = 6)

Should we fill this out? What do other modules do with these files?

src/Blueprint/Blueprint/Blueprint.csproj
```diff
@@ -0,0 +1,34 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
majastrz: >Project [](start = 1, length = 7)

After we migrate to CBT in our Blueprint repo, we'll be able to use thin projects like this :)

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
```diff
@@ -0,0 +1,382 @@
...
+        /// The blueprint client.
+        /// </summary>
+        private IBlueprintClient blueprintClient;
+
+        public IBlueprintClient BlueprintClient
+        {
+            get
+            {
+                return blueprintClient = blueprintClient ?? new BlueprintClient(DefaultProfile.DefaultContext);
+            }
```
majastrz: nit - we should probably keep the getter and setter using the same style if they're both simple return statements. For this one, it'd be ok to make the get a lambda. Easier to read that way.

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
```diff
@@ -0,0 +1,382 @@
...
+            set => blueprintClient = value;
+        }
+
+        /// <summary>
+        /// The blueprint client with delegating handler. The delegating handler is needed to get blueprint versions info.
```
majastrz: >The delegating handler is needed to get blueprint versions info. [](start = 58, length = 64)

We should probably file a task to model $expand in the swagger or via some SDK customization so it's easy for users of the SDK to do it.

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
```diff
@@ -0,0 +1,382 @@
...
+            {
+                foreach (var e in aggEx.Flatten().InnerExceptions)
+                {
+                    WriteExceptionError(e);
+                }
+
+                return;
+            }
+
+            throw ex;
```
majastrz: >ex [](start = 18, length = 2)

Should this write the exception as well? If not, we should probably have a comment explaining the reasons why not.

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
```diff
@@ -0,0 +1,382 @@
...
+        #endregion Cmdlet Overrides
+
+        #region Protected Methods
+        /// <summary>
+        /// Check if an assignment with the same name exists
+        /// </summary>
+        /// <param name="scope"> Subscription scope</param>
+        /// <param name="name"> Assignment name</param>
+        /// <param name="invertCheck"> Inverts the check logic from check for exists to check for not-exits </param>
+        protected void CheckIfAssignmentExist(string scope, string name, bool invertCheck)
```
majastrz: >Exist [](start = 40, length = 5)

nit - Exists

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
```diff
@@ -0,0 +1,382 @@
...
+            try
+            {
+                existingAssignment = BlueprintClient.GetBlueprintAssignment(scope, name);
+            }
+            catch (Exception ex)
+            {
+                if (ex is CloudException cex && cex.Response.StatusCode != System.Net.HttpStatusCode.NotFound)
+                {
+                    throw new CloudException(cex.Message);
+                }
```
majastrz: This eats the exception if it's not a CloudException at which point the check is really inconclusive but following code will assume it doesn't exist. I'd probably also rename the method to ThrowIfAssignmentExists to make it clearer to the caller what will happen.

Also is CloudException the right one to throw. My expectation would be to have CloudException be fully populated with Http status, code, message, etc.

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
```diff
@@ -0,0 +1,382 @@
...
+            }
+
+            if (invertCheck)
+            {
+                if (existingAssignment == null)
+                {
+                    throw new Exception(string.Format(Resources.AssignmentNotExist, name, scope));
```
majastrz: >string.Format( [](start = 40, length = 14)

All string.Format calls should specify a culture, so any numbers get formatted correctly.

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
```diff
@@ -0,0 +1,382 @@
...
+
+            try
+            {
+                existingAssignment = BlueprintClient.GetBlueprintAssignment(scope, name);
+            }
+            catch (Exception ex)
+            {
+                if (ex is CloudException cex && cex.Response.StatusCode != System.Net.HttpStatusCode.NotFound)
+                {
+                    throw new CloudException(cex.Message);
```
majastrz: >cex.Message [](start = 45, length = 11)

This comes back out as a localized message, but where do we configure the language we want on the client?

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
```diff
@@ -0,0 +1,382 @@
...
+        /// </summary>
+        /// <returns></returns>
+        protected ServicePrincipal GetBlueprintSpn()
+        {
+            var odataQuery = new Rest.Azure.OData.ODataQuery<ServicePrincipal>(s => s.ServicePrincipalNames.Contains(BlueprintConstants.AzureBlueprintAppId));
+            var servicePrincipal = GraphRbacManagementClient.ServicePrincipals.List(odataQuery.ToString())
+                .FirstOrDefault();
+
+            return servicePrincipal;
+        }
```
majastrz: This only really applies to cmdlets that create or upgrade an assignment. Should we move that over to some base class somewhere?

majastrz: Also the method below

---
In reply to: [257093525](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257093525) [](ancestors = 257093525)

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
```diff
@@ -0,0 +1,382 @@
...
+                var managementGroupAncestorsObjects = resultJObjects["managementGroupAncestors"].Children().ToList();
+
+                foreach (var mgObject in managementGroupAncestorsObjects)
+                {
+                    managementGroupAncestors.Add(mgObject.ToString());
+                }
+
+                if (managementGroupAncestors.Count > 0)
+                {
+                    // Remove the last item which is the tenantId.
```
majastrz: >tenantId [](start = 57, length = 8)

It's not. The last item is the root MG whose ID is the same guid as the tenant ID.

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
```diff
@@ -0,0 +1,382 @@
...
+            }
+            catch (Exception ex)
+            {
+                if (ex is CloudException cex && cex.Response.StatusCode != System.Net.HttpStatusCode.NotFound)
+                {
+                    throw new CloudException(cex.Message);
+                }
+            }
+
+            if (invertCheck)
```
majastrz: >invertCheck [](start = 16, length = 11)

nit - the bool flag could get missed easily. I'd consider to split this into 2 methods.

src/Blueprint/Blueprint/Cmdlets/GetAzureRMBlueprint.cs
```diff
@@ -0,0 +1,119 @@
...
+        [Parameter(ParameterSetName = ParameterSetNames.ByManagementGroupNameAndLatestPublished, Position = 2, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = ParameterHelpMessages.LatestPublishedFlag)]
+        public SwitchParameter LatestPublished { get; set; }
+
+        #endregion Parameters
+
+        #region Cmdlet Overrides
+        public override void ExecuteCmdlet()
+        {
+
+            var scope = getCurrentScope();
```
majastrz: >g [](start = 24, length = 1)

nit - c# naming convention is to capitalize method names.

src/Blueprint/Blueprint/Cmdlets/GetAzureRMBlueprintAssignment.cs
```diff
@@ -0,0 +1,50 @@
...
+        #region Cmdlet Overrides
+        public override void ExecuteCmdlet()
+        {
+            var subscription = SubscriptionId ?? DefaultContext.Subscription.Id;
+
+            try
+            {
+                switch (ParameterSetName) {
+                    case ParameterSetNames.BlueprintAssignmentsBySubscription:
+                        foreach (var assignment in BlueprintClient.ListBlueprintAssignments(string.Format(BlueprintConstants.SubscriptionScope, subscription)))
```
majastrz: >string.Format(BlueprintConstants.SubscriptionScope, subscription) [](start = 92, length = 65)

This sort of thing repeats a lot, so it probably should be a utility function somewhere.

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
majastrz: Ones that deal with localized strings should probably use either CurrentCulture or CurrentUICulture (not sure which). Anything that is not localized text (like scope strings, for example) should use InvariantCulture

---
In reply to: [257093111](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257093111) [](ancestors = 257093111)

majastrz: Also we probably should turn on some static analysis to catch stuff like this for us.

---
In reply to: [257118910](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257118910) [](ancestors = 257118910,257093111)

src/Blueprint/Blueprint/Cmdlets/NewAzureRMBlueprintAssignment.cs
```diff
@@ -0,0 +1,90 @@
...
+using ParameterHelpMessages = Microsoft.Azure.Commands.Blueprint.Common.BlueprintConstants.ParameterHelpMessages;
+using System.Text.RegularExpressions;
+using Microsoft.Azure.Commands.Blueprint.Common;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.Commands.Blueprint.Cmdlets
+{
+    [Cmdlet(VerbsCommon.New, ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "BlueprintAssignment", SupportsShouldProcess = true)]
```
majastrz: >ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "BlueprintAssignment" [](start = 29, length = 77)

I'm wondering if we should had a Nouns class to avoid weird inconsistencies in spelling or capitalization (didn't find any, though).

src/Blueprint/Blueprint/Cmdlets/NewAzureRMBlueprintAssignment.cs
```diff
@@ -0,0 +1,90 @@
...
+    public class NewAzureRmBlueprintAssignment : BlueprintCmdletBase
+    {
+        #region Parameters
+        [Parameter(ParameterSetName = ParameterSetNames.CreateBlueprintAssignment, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = ParameterHelpMessages.BlueprintAssignmentName)]
+        [ValidatePattern("^[0-9a-zA-Z_-]*$", Options = RegexOptions.Compiled | RegexOptions.CultureInvariant)]
```
majastrz: >] [](start = 39, length = 1)

We can probably add a length validator as well or add a quantifier to the regex.

src/Blueprint/Blueprint/Cmdlets/NewAzureRMBlueprintAssignment.cs
```diff
@@ -0,0 +1,90 @@
...
+        [ValidateNotNull]
+        public PSBlueprintBase Blueprint { get; set; }
+
+        [Parameter(ParameterSetName = ParameterSetNames.CreateBlueprintAssignment, Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = ParameterHelpMessages.SubscriptionIdToAssign)]
+        [ValidateNotNullOrEmpty]
+        public string[] SubscriptionId { get; set; }
+
+        [Parameter(ParameterSetName = ParameterSetNames.CreateBlueprintAssignment, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = ParameterHelpMessages.Location)]
+        [ValidateNotNullOrEmpty]
+        [LocationCompleter("Microsoft.Batch/operations")]
```
majastrz: >Microsoft.Batch/operations [](start = 28, length = 26)

Is that intended to be that resource type? It's not one of ours.

src/Blueprint/Blueprint/Models/PSLockMode.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Commands.Blueprint.Models
+{
+    public enum PSLockMode
+    {
+        Unknown = 0,
+        None = 1,
+        AllResources = 2
+    }
```
majastrz: These are from the old API version. We should use the new ones.

src/Blueprint/Blueprint/Models/PSAssignmentLockSettings.cs
```diff
@@ -0,0 +1,27 @@
+﻿// ----------------------------------------------------------------------------------
+//
+// Copyright Microsoft Corporation
+// Licensed under the Apache License, Version 2.0 (the "License");
```
majastrz: >Apache [](start = 22, length = 6)

Is this the right license and do we need all files to have it? (I hope not.)

src/Blueprint/Blueprint/Models/PSBlueprint.cs
```diff
@@ -0,0 +1,120 @@
...
+                TargetScope = PSBlueprintTargetScope.Unknown,
+                Parameters = new Dictionary<string, PSParameterDefinition>(),
+                ResourceGroups = new Dictionary<string, PSResourceGroupDefinition>(),
+                Versions = model.Versions,
+                ParametersDisplayList = new List<string>(),
+                ResourceGroupDisplayList = new List<string>(),
+                VersionList = new List<string>()
+            };
+
+            if (DateTime.TryParse(model.Status.TimeCreated, out DateTime timeCreated))
```
majastrz: >TimeCreated [](start = 47, length = 11)

Is this coming from the SDK? Should we fix it so the SDK generates date/time values and not strings?

majastrz: >TryParse [](start = 25, length = 8)

If it fails to parse, we silently skip assinging the value. That probably means TimeCreated set to Datetime.MinValue. Do we want that?

majastrz: Same for all the TryParse calls here.

---
In reply to: [257121303](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257121303) [](ancestors = 257121303)

src/Blueprint/Blueprint/Models/PSManagedServiceIdentity.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Commands.Blueprint.Models
+{
+    public class PSManagedServiceIdentity
+    {
+        public string Type { get; set; }
+        public string PrincipalId { get; set; }
+        public string TenantId { get; set; }
+    }
```
majastrz: This needs to handle user-assigned identities

src/Blueprint/Blueprint/Models/PSBlueprintTargetScope.cs
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+// ----------------------------------------------------------------------------------
+
+namespace Microsoft.Azure.Commands.Blueprint.Models
+{
+    public enum PSBlueprintTargetScope
+    {
+        Unknown = 0,
+        Subscription = 1,
+        ManagementGroup = 2,
```
majastrz: >ManagementGroup = 2, [](start = 8, length = 20)

This isn't allowed... yet.

src/Blueprint/Blueprint/Models/PSBlueprintAssignment.cs
```diff
@@ -0,0 +1,99 @@
...
+                Id = assignment.Id,
+                Type = assignment.Type,
+                Location = assignment.Location,
+                Scope = scope,
+                Identity = new PSManagedServiceIdentity
+                {
+                    PrincipalId = assignment.Identity.PrincipalId,
+                    TenantId = assignment.Identity.TenantId,
+                    Type = assignment.Type
+                },
```
majastrz: We need to update this for user-assigned also

src/Blueprint/Blueprint/Models/PSParameterValueBase.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Commands.Blueprint.Models
+{
+    public class PSParameterValueBase
```
majastrz: >PSParameterValueBase [](start = 17, length = 20)

Do we have display name also?

src/Blueprint/Blueprint/Models/PSPublishedBlueprint.cs
```diff
@@ -0,0 +1,105 @@
...
+        public List<string> ParametersDisplayList { get; set; }
+        public List<string> ResourceGroupDisplayList { get; set; }
+
+        /// <summary>
+        /// Create a PSPublishedBlueprint object from a PublishedBlueprint.
+        /// </summary>
+        /// <param name="model"></param>
+        /// <param name="managementGroupName">Name of the management group the blueprint belongs to.</param>
+        /// <returns>A new PSPublishedBlueprint object</returns>
+        internal static PSPublishedBlueprint FromPublishedBlueprintModel(PublishedBlueprint model, string scope)
```
majastrz: >FromPublishedBlueprintModel [](start = 45, length = 27)

Same comments as in the PSBlueprint conversion logic.

src/Blueprint/Blueprint/Properties/Resources.resx
```diff
@@ -0,0 +1,147 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="AssignmentExists" xml:space="preserve">
+    <value>An assignment with name '{0}' in Subscription '{1}' already exists. Please use Set-AzBlueprintAssignment to update an existing assignment.</value>
```
majastrz: >Subscription [](start = 44, length = 12)

nit - should be lower case to match the other convention (like "artifact" or "assignment")

src/Blueprint/Blueprint/Properties/Resources.resx
```diff
@@ -0,0 +1,147 @@
...
+  </data>
+  <data name="AssignmentNotExist" xml:space="preserve">
+    <value>An assignment with name '{0}' in Subscription '{1}' does not exists. Please use New-AzBlueprintAssignment to create a new assignment.</value>
```
majastrz: >exists [](start = 72, length = 6)

typo - exist

src/Blueprint/Blueprint/Properties/Resources.resx
majastrz: below as well

---
In reply to: [257122665](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257122665) [](ancestors = 257122665)

src/Blueprint/Blueprint/Properties/Resources.resx
```diff
@@ -0,0 +1,147 @@
...
+    <value>Creating blueprint assignment '{0}'</value>
+  </data>
+  <data name="DeleteAssignmentShouldProcessString" xml:space="preserve">
+    <value>Deleting blueprint assignment '{0}' in Subscription '{1}'</value>
+  </data>
+  <data name="ManagementGroupNotFound" xml:space="preserve">
+    <value>We can't find any Management Groups that you have access to.</value>
+  </data>
+  <data name="NoBlueprintAssignmentsNotFound" xml:space="preserve">
+    <value>Could not find any Blueprints assignments within '{0}'. Please check the Subscription Id and try again.</value>
```
majastrz: >Blueprints [](start = 30, length = 10)

blueprint

src/Blueprint/Blueprint/Properties/Resources.resx
```diff
@@ -0,0 +1,147 @@
...
+  </data>
+  <data name="NoBlueprintsFound" xml:space="preserve">
+    <value>Could not find any Blueprints.</value>
```
majastrz: >Blueprints [](start = 30, length = 10)

lower case

src/Blueprint/Blueprint/Properties/Resources.resx
```diff
@@ -0,0 +1,147 @@
...
+  </data>
+  <data name="PublishedBlueprintNotFound" xml:space="preserve">
+    <value>Published Blueprint '{0}' could not be found in management group '{1}'.</value>
```
majastrz: >Blueprint [](start = 21, length = 9)

lower case

src/Blueprint/Blueprint/Properties/Resources.resx
```diff
@@ -0,0 +1,147 @@
...
+  <data name="ManagementGroupNotFound" xml:space="preserve">
+    <value>We can't find any Management Groups that you have access to.</value>
+  </data>
+  <data name="NoBlueprintAssignmentsNotFound" xml:space="preserve">
+    <value>Could not find any Blueprints assignments within '{0}'. Please check the Subscription Id and try again.</value>
+  </data>
+  <data name="NoBlueprintsFound" xml:space="preserve">
+    <value>Could not find any Blueprints.</value>
+  </data>
+  <data name="PublishedBlueprintNotFound" xml:space="preserve">
```
majastrz: >PublishedBlueprintNotFound [](start = 14, length = 26)

Do we need a separate message if they're looking for a specific version?

src/Blueprint/Blueprint.Test/ScenarioTests/BlueprintAssignmentTests.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public BlueprintAssignmentTests(Xunit.Abstractions.ITestOutputHelper output)
+        {
+            _logger = new ServiceManagement.Common.Models.XunitTracingInterceptor(output);
+            ServiceManagement.Common.Models.XunitTracingInterceptor.AddToContext(_logger);
+            TestExecutionHelpers.SetUpSessionAndProfile();
+        }
+
+        [Fact]
+        [Trait(Category.AcceptanceType, Category.CheckIn)]
```
majastrz: Ooh... what are these?

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
filizt: Created a task. [https://msazure.visualstudio.com/One/_workitems/edit/3926599](https://msazure.visualstudio.com/One/_workitems/edit/3926599) 


---
In reply to: [257091752](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257091752) [](ancestors = 257091752)

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
filizt: Just to check, this is in the cmdletbase class. Do you think it'd be better if we had a base class for create/upgrade operations?

---
In reply to: [257093525](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257093525) [](ancestors = 257093525)

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
filizt: Thanks for catching that. I'll update the code.

---
In reply to: [257093763](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257093763) [](ancestors = 257093763)

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
filizt: Sounds good. 

---
In reply to: [257093950](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257093950) [](ancestors = 257093950)

src/Blueprint/Blueprint/Cmdlets/NewAzureRMBlueprintAssignment.cs
filizt: I can add length validator no problem. But I checked this on the Portal and it didn't look like we had a length requirement for assignment name(I was able to enter arbitrarily long assignment name there). Do we have a check for that during deployment?

---
In reply to: [257119662](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257119662) [](ancestors = 257119662)

src/Blueprint/Blueprint/Cmdlets/GetAzureRMBlueprint.cs
filizt: Done

---
In reply to: [257116891](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257116891) [](ancestors = 257116891)

src/Blueprint/Blueprint/Cmdlets/NewAzureRMBlueprintAssignment.cs
filizt: This is a recommended argument completer if you have a "location" parameter in the cmdlets. [https://github.com/Azure/azure-powershell/blob/89eea9c118fa45f05f04e5dd09dc285f4a70a22e/documentation/development-docs/azure-powershell-developer-guide.md#location-completer](https://github.com/Azure/azure-powershell/blob/89eea9c118fa45f05f04e5dd09dc285f4a70a22e/documentation/development-docs/azure-powershell-developer-guide.md#location-completer) 

---
In reply to: [257119814](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257119814) [](ancestors = 257119814)

src/Blueprint/Blueprint/Models/PSLockMode.cs
filizt: Updated (They were updated at some point, must have discarded them somehow and didn't notice. Anyway, thanks for pointing out).

---
In reply to: [257120202](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257120202) [](ancestors = 257120202)

src/Blueprint/Blueprint/Models/PSManagedServiceIdentity.cs
filizt: Adding.

---
In reply to: [257122019](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257122019) [](ancestors = 257122019)

src/Blueprint/Blueprint/Cmdlets/NewAzureRMBlueprintAssignment.cs
majastrz: The completer probably queries /providers/{namespace} in ARM that looks at the resource type to figure out allowed location. Can you try it with blueprint assignment resource type first. If it doesn't work right (our manifest says "" for location), we should switch to something from Microsoft.Resources namespace instead. That's guaranteed to be updated when new regions show up.

---
In reply to: [257331542](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257331542) [](ancestors = 257331542,257119814)

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
majastrz: I was thinking about moving that into an Assignment specific base class somewhere. Any assignment related cmdlet would need this logic. (Unless I'm wrong and there are other usages.)

---
In reply to: [257326054](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257326054) [](ancestors = 257326054,257093525)

src/Blueprint/Blueprint/Cmdlets/NewAzureRMBlueprintAssignment.cs
majastrz: UI might not check - I'd check if backend does some validation.

---
In reply to: [257330223](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257330223) [](ancestors = 257330223,257119662)

src/Blueprint/Blueprint/Cmdlets/GetAzureRMBlueprint.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            try
```
majastrz: >try [](start = 12, length = 3)

Btw, I really like how clean and simple the cmdlet code is.

src/Blueprint/Blueprint/Cmdlets/RemoveAzureRMBlueprintAssignment.cs
```diff
@@ -0,0 +1,71 @@
...
+                {
+                    case ParameterSetNames.DeleteBlueprintAssignmentByName:
+                        if (ShouldProcess(Name, string.Format(Resources.DeleteAssignmentShouldProcessString, Name, SubscriptionId)))
+                        {
+                            BlueprintClient.DeleteBlueprintAssignment(SubscriptionId, Name);
+
+                            if (PassThru.IsPresent)
+                            {
+                                WriteObject(true);
+                            }
```
majastrz: I think we should return the PSObject representation of what was deleted or nothing if it didn't exist. The following would still be possible in PowerShell with that change because PS has a concept of truthiness similar to JS:
$deletionResult = Remove-AzBlueprintAssignment -dsfjkhdskfjhdskjfhk
if($deletionResult) {} else {}

src/Blueprint/Blueprint/Cmdlets/RemoveAzureRMBlueprintAssignment.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        [Parameter(ParameterSetName = ParameterSetNames.DeleteBlueprintAssignmentByName, Position = 1, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = ParameterHelpMessages.BlueprintAssignmentName)]
+        [ValidateNotNull]
+        public string Name { get; set; }
+
+        [Parameter(ParameterSetName = ParameterSetNames.DeleteBlueprintAssignmentByObject, Position = 0, Mandatory = true, ValueFromPipeline = true, HelpMessage = ParameterHelpMessages.BlueprintAssignmentObject)]
+        public PSBlueprintAssignment BlueprintAssignmentObject { get; set; }
+
+        [Parameter(Mandatory = false)]
+        public SwitchParameter PassThru { get; set; }
```
majastrz: >PassThru [](start = 31, length = 8)

What does that do?

src/Blueprint/Blueprint/Common/Utils.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.Azure.Commands.Common.Authentication;
+using AzureSession = Microsoft.Azure.Common.Authentication.AzureSession;
+
+namespace Microsoft.Azure.PowerShell.Cmdlets.Blueprint.Common
+{
+    class Utils
```
majastrz: >class [](start = 4, length = 5)

nit - missing access modifier

src/Blueprint/Blueprint/Common/Utils.cs
```diff
@@ -0,0 +1,38 @@
...
+
+            if (tokens != null && tokens.Length == 2)
+            {
+                locationId = tokens[0] == "subscriptions" ? tokens[1] : null;      
+            }
+
+            if (tokens != null && tokens.Length == 4)
+            {
+                locationId = tokens[2] == "managementGroups" ? tokens[3] : null;
+            }
```
majastrz: Azure resource IDs are not case-sensitive. We also have some similar parsing logic in the backend that could potentially be reused (as long as it doesn't depend on an internal library).

src/Blueprint/Blueprint/Common/BlueprintConstants.cs
```diff
@@ -0,0 +1,57 @@
...
+            public const string BlueprintAssignmentsBySubscription = "BlueprintAssignmentsBySubscription";
+            public const string BlueprintAssignmentByName = "BlueprintAssignmentByName";
+            public const string CreateBlueprintAssignment = "BlueprintAssignment";
+            public const string DeleteBlueprintAssignmentByName = "DeleteBlueprintAssignmentByName";
+            public const string DeleteBlueprintAssignmentByObject = "DeleteBlueprintAssignmentByObject";
+        }
+
+        public static class ParameterHelpMessages
+        {
+            public const string SubscriptionId = "Subscription Id.";
```
majastrz: >SubscriptionId [](start = 32, length = 14)

I see them being used in the cmdlet attributes. How are they different from the .md files in the /help folder?

src/Blueprint/Blueprint/Cmdlets/NewAzureRMBlueprintAssignment.cs
```diff
@@ -0,0 +1,90 @@
...
+                    var subscriptionsList = SubscriptionId ?? new[] { DefaultContext.Subscription.Id };
+
+                    foreach (var subscription in subscriptionsList)
+                    {
+                        var scope = string.Format(BlueprintConstants.SubscriptionScope, subscription);
+                        CheckIfAssignmentExist(scope, Name, false);
+                        // First Register Blueprint RP and grant owner permission to BP service principal
+                        RegisterBlueprintRp(subscription);
+                        var servicePrincipal = GetBlueprintSpn();
+                        AssignOwnerPermission(subscription, servicePrincipal);
```
majastrz: >AssignOwnerPermission [](start = 24, length = 21)

We should probably discuss with the rest of the team, but I'm wondering if we should have a switch to make this conditional.

src/Blueprint/Blueprint/Models/PSManagedServiceIdentity.cs
filizt: Actually adding this as one of the tasks for next phase.

---
In reply to: [257333104](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257333104) [](ancestors = 257333104,257122019)

src/Blueprint/Blueprint/Cmdlets/RemoveAzureRMBlueprintAssignment.cs
filizt: This is to provide output to the user when no output would normally be provided. It's a requirement for Remove-* cmdlets. [https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/patterns-for-standard-cmdlets.md#remove--cmdlet](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/patterns-for-standard-cmdlets.md#remove--cmdlet)

---
In reply to: [257385808](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257385808) [](ancestors = 257385808)

src/Blueprint/Blueprint/Common/BlueprintConstants.cs
filizt: The tool that produces the .md files look at the cmdlet attribute "HelpMessage" for the explanations of the parameters.

---
In reply to: [257388050](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257388050) [](ancestors = 257388050)

src/Blueprint/Blueprint.Test/ScenarioTests/BlueprintAssignmentTests.cs
filizt: This is to mark the test to be run during CI in playback mode. [https://github.com/Azure/azure-powershell/blob/89eea9c118fa45f05f04e5dd09dc285f4a70a22e/documentation/development-docs/azure-powershell-developer-guide.md](https://github.com/Azure/azure-powershell/blob/89eea9c118fa45f05f04e5dd09dc285f4a70a22e/documentation/development-docs/azure-powershell-developer-guide.md)

---
In reply to: [257123201](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257123201) [](ancestors = 257123201)

src/Blueprint/Blueprint/Properties/Resources.resx
filizt: These are the artifacts from the time I was trying to produce proper error messaging. Then decided to relay the error our API sent. So, removing them from the .resx file.

---
In reply to: [257123122](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257123122) [](ancestors = 257123122)

src/Blueprint/Blueprint/Models/PSBlueprintTargetScope.cs
filizt: Removing.. till then :)

---
In reply to: [257122123](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257122123) [](ancestors = 257122123)

src/Blueprint/Blueprint/ChangeLog.md
filizt: Added brief change log text

---
In reply to: [257091072](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257091072) [](ancestors = 257091072)

src/Blueprint/Blueprint/Cmdlets/NewAzureRMBlueprintAssignment.cs
filizt: I see. I updated it to use  Microsoft.Resources namespace with resourceGroups resource type since I saw this being used in other cmdlets as well.

---
In reply to: [257382175](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257382175) [](ancestors = 257382175,257331542,257119814)

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
filizt: We talked about this offline. PowerShell doesn't support localization.

---
In reply to: [257093171](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257093171) [](ancestors = 257093171)

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
filizt: Refactored the function and fixed the throw statement.

---
In reply to: [257093029](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257093029) [](ancestors = 257093029)

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
filizt: I'll follow up on static analysis offline. Resolving rest of the comments as by design. 

---
In reply to: [257119216](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257119216) [](ancestors = 257119216,257118910,257093111)

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
filizt: Sounds good. I'll work on that in the next phase of the cmdlets.

---
In reply to: [257382430](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257382430) [](ancestors = 257382430,257326054,257093525)

src/Blueprint/Blueprint/Cmdlets/NewAzureRMBlueprintAssignment.cs
filizt: Sounds good. I'll add that in the next phase of the cmdlets.

---
In reply to: [257119476](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257119476) [](ancestors = 257119476)

src/Blueprint/Blueprint/Cmdlets/NewAzureRMBlueprintAssignment.cs
filizt: I couldn't find any length validation on the backend either. 

---
In reply to: [257382671](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257382671) [](ancestors = 257382671,257330223,257119662)

src/Blueprint/Blueprint/Cmdlets/NewAzureRMBlueprintAssignment.cs
filizt: I'll bring this up with the team. Adding it as an item for the next phase of cmdlets.

---
In reply to: [257388642](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257388642) [](ancestors = 257388642)

src/Blueprint/Blueprint/Models/PSAssignmentLockSettings.cs
filizt: Looks like it. I see this in every other project as well. 

---
In reply to: [257120446](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257120446) [](ancestors = 257120446)

src/Blueprint/Blueprint/Models/PSBlueprint.cs
filizt: Created a task for this 3942542

---
In reply to: [257121161](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257121161) [](ancestors = 257121161)

src/Blueprint/Blueprint/Cmdlets/RemoveAzureRMBlueprintAssignment.cs
filizt: With passThru we can still write the above if statement like below. Using other cmdlets, and looking at their implementation it seemed like this is a common way of implementing the delete functionality. But, it also makes sense to differentiate the result of the delete operation between if the object to be deleted existed or not. I'll ask for recommendation on this during the PR process.
 $deletionResult = Remove-AzBlueprintAssignment -dsfjkhdskfjhdskjfhk -PassThru
if($deletionResult) {} else {}\

---
In reply to: [257385758](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257385758) [](ancestors = 257385758)

src/Blueprint/Blueprint/Common/Utils.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            if (string.IsNullOrEmpty(scope))
+                return null;
+
+            string[] tokens = scope.Split(new[] {'/'}, StringSplitOptions.RemoveEmptyEntries);
+            string locationId = null;
+
+            if (tokens != null && tokens.Length == 2)
+            {
+                locationId = string.Equals(tokens[0], "subscriptions", StringComparison.InvariantCultureIgnoreCase) ? tokens[1] : null;
```
majastrz: >InvariantCultureIgnoreCase [](start = 88, length = 26)

We're doing OrdinalIgnoreCase in the backend. Same comment for the comparison below as well.

src/Blueprint/Blueprint/Cmdlets/RemoveAzureRMBlueprintAssignment.cs
majastrz: Makes sense. My point is that we can just return the PS object (if -PassThru switch is set) representing the resource that was deleted if the actual deletion was done. It gives more info that just true/false.

---
In reply to: [257734556](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257734556) [](ancestors = 257734556,257385758)

src/Blueprint/Blueprint/Models/PSBlueprint.cs
```diff
@@ -0,0 +1,132 @@
...
+                psBlueprint.Status.LastModified = null;
+            }
+
+            if (Enum.TryParse(model.TargetScope, true, out PSBlueprintTargetScope targetScope))
+            {
+                psBlueprint.TargetScope = targetScope;
+            }
+            else
+            {
+                psBlueprint.TargetScope = PSBlueprintTargetScope.Unknown;
```
majastrz: >Unknown [](start = 65, length = 7)

This is fine for now. But in which scenario will any of the else branches in this function actually be taken?

src/Blueprint/Blueprint/Cmdlets/NewAzureRMBlueprintAssignment.cs
majastrz: Then never mind :)

---
In reply to: [257712941](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257712941) [](ancestors = 257712941,257382671,257330223,257119662)

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
```diff
@@ -0,0 +1,360 @@
...
+            try
+            {
+                assignment = BlueprintClient.GetBlueprintAssignment(scope, name);
+            }
+            catch (Exception ex)
+            {
+                if (ex is CloudException cex && cex.Response.StatusCode != System.Net.HttpStatusCode.NotFound)
+                {
+                    // if exception is for a reason other than .NotFound, pass it to the caller.
+                    throw ex;
```
majastrz: >throw ex; [](start = 20, length = 9)

If you throw an exception that was already thrown somewhere else, it wipes out the original stack trace. (We might need it for debugging sometime.) You should use "throw;" instead.

src/Blueprint/Blueprint/Models/PSParameterValueBase.cs
filizt: Couldn't find display name in SDK models. Looked at ParameterValue and ParameterValueBase.

---
In reply to: [257122404](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257122404) [](ancestors = 257122404)

src/Blueprint/Blueprint/Cmdlets/RemoveAzureRMBlueprintAssignment.cs
filizt: Agreed.

---
In reply to: [258255855](https://github.com/Azure/azure-powershell/pull/8537#discussion_r258255855) [](ancestors = 258255855,257734556,257385758)

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
filizt: It does write the exception but I made a change to call the base class explicitly. 

---
In reply to: [257092353](https://github.com/Azure/azure-powershell/pull/8537#discussion_r257092353) [](ancestors = 257092353)

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
filizt: Updated the style to match.

src/Blueprint/Blueprint/Models/PSBlueprint.cs
filizt: Resolving for now. I'll follow up on this offline.

---
In reply to: [258257485](https://github.com/Azure/azure-powershell/pull/8537#discussion_r258257485) [](ancestors = 258257485)

src/Blueprint/Blueprint/Models/PSParameterValueBase.cs
majastrz: Maybe they're missing from swagger, but yeah we can deal with that later.

---
In reply to: [258966822](https://github.com/Azure/azure-powershell/pull/8537#discussion_r258966822) [](ancestors = 258966822,257122404)

maddieclayton:
@filizt Please fix help issues: https://azuresdkci.westus2.cloudapp.azure.com/job/powershell/10936/artifact/artifacts/ValidateHelpIssues.csv/*view*/

It also looks like your tests are failing here: https://travis-ci.org/Azure/azure-powershell/builds/498428680?utm_source=github_status&utm_medium=notification

src/Blueprint/Blueprint/Cmdlets/NewAzureRMBlueprintAssignment.cs
```diff
@@ -0,0 +1,130 @@
...
+        [ValidateNotNull]
+        public PSBlueprintBase Blueprint { get; set; }
+
+        [Parameter(ParameterSetName = ParameterSetNames.CreateBlueprintAssignment, Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = ParameterHelpMessages.SubscriptionIdToAssign)]
+        [ValidateNotNullOrEmpty]
+        public string[] SubscriptionId { get; set; }
+
+        [Parameter(ParameterSetName = ParameterSetNames.CreateBlueprintAssignment, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = ParameterHelpMessages.Location)]
+        [ValidateNotNullOrEmpty]
+        [LocationCompleter("Microsoft.Resources/resourceGroups")]
```
cormacpayne: @filizt usually the `LocationCompleter` determines the valid locations for the resource being created -- is a Blueprint assignment a tracked resource? If so, we should make sure the argument for this completer accurately reflects the correct resource type.

src/Blueprint/Blueprint/Cmdlets/RemoveAzureRMBlueprintAssignment.cs
```diff
@@ -0,0 +1,74 @@
...
+        [Parameter(ParameterSetName = ParameterSetNames.DeleteBlueprintAssignmentByName, Position = 0, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = ParameterHelpMessages.AssignmentSubscriptionId)]
+        [ValidateNotNullOrEmpty]
+        public string SubscriptionId { get; set; }
+
+        [Parameter(ParameterSetName = ParameterSetNames.DeleteBlueprintAssignmentByName, Position = 1, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = ParameterHelpMessages.BlueprintAssignmentName)]
+        [ValidateNotNull]
+        public string Name { get; set; }
+
+        [Parameter(ParameterSetName = ParameterSetNames.DeleteBlueprintAssignmentByObject, Position = 1, Mandatory = true, ValueFromPipeline = true, HelpMessage = ParameterHelpMessages.BlueprintAssignmentObject)]
+        public PSBlueprintAssignment Assignment { get; set; }
```
cormacpayne: @filizt this should be renamed to `-InputObject` (you can add an alias to `-Assignment`)

src/Blueprint/Blueprint/Cmdlets/SetAzureRMBlueprintAssignment.cs
```diff
@@ -0,0 +1,124 @@
...
+                            ResourceGroup);
+
+                        foreach (var subscription in subscriptionsList)
+                        {
+                            var scope = Utils.GetScopeForSubscription(subscription);
+                            ThrowIfAssignmentNotExist(scope, Name);
+                            // Register Blueprint RP
+                            RegisterBlueprintRp(subscription);
+
+                            WriteObject(BlueprintClient.CreateOrUpdateBlueprintAssignment(scope, Name, assignment));
```
cormacpayne: @filizt is the underlying  call for `CreateOrUpdateBlueprintAssignment` a `PUT` or a `PATCH`? If it's a `PATCH`, then we should rename the cmdlet to be `Update-*` instead of `Set-*`

src/Blueprint/Blueprint/Cmdlets/SetAzureRMBlueprintAssignment.cs
```diff
@@ -0,0 +1,124 @@
...
+    {
+        #region Parameters
+        [Parameter(ParameterSetName = ParameterSetNames.CreateBlueprintAssignment, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = ParameterHelpMessages.BlueprintAssignmentName)]
+        [ValidatePattern("^[0-9a-zA-Z_-]*$", Options = RegexOptions.Compiled | RegexOptions.CultureInvariant)]
+        [ValidateNotNullOrEmpty]
+        public string Name { get; set; }
+
+        [Parameter(ParameterSetName = ParameterSetNames.CreateBlueprintAssignment, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = ParameterHelpMessages.BlueprintObject)]
+        [ValidateNotNull]
+        public PSBlueprintBase Blueprint { get; set; }
```
cormacpayne: @filizt any reason that we can't pass a Blueprint assignment object to this cmdlet as a parameter like we do for `Remove-AzBlueprintAssignment`?

src/Blueprint/Blueprint/Common/BlueprintConstants.cs
```diff
@@ -0,0 +1,53 @@
+﻿
+namespace Microsoft.Azure.Commands.Blueprint.Common
+{
+    public static class BlueprintConstants
+    {
+        public const string AzureBlueprintAppId = "f71766dc-90d9-4b7d-bd9d-4499c4331c3f";
+        public const string OwnerRoleDefinitionId = "/providers/Microsoft.Authorization/roleDefinitions/8e3af657-a8ff-443c-a75c-2fe8c4bcb635";
```
cormacpayne: @filizt what are these constants used for? Should they be used by the public?

src/Blueprint/Blueprint/help/Get-AzBlueprint.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+## DESCRIPTION
+Get one or more blueprint definitions. Blueprint definitions exist at the management group or subscription scope.
+
+## EXAMPLES
+
+### Example 1
+```powershell
+PS> Get-AzBlueprint
```
cormacpayne: @filizt for the examples you provide in the cmdlet help files, would you mind including the resulting output of each command so users have an idea of what is returned?

src/Blueprint/Blueprint/help/Get-AzBlueprint.md
```diff
@@ -0,0 +1,214 @@
...
+### CommonParameters
+This cmdlet supports the common parameters: -Debug, -ErrorAction, -ErrorVariable, -InformationAction, -InformationVariable, -OutVariable, -OutBuffer, -PipelineVariable, -Verbose, -WarningAction, and -WarningVariable. For more information, see about_CommonParameters (http://go.microsoft.com/fwlink/?LinkID=113216).
+
+## INPUTS
+
+### System.String
+
+## OUTPUTS
+
+### System.Object
```
cormacpayne: @filizt you'll need to regenerate the help for these cmdlets again once you've finalized all your changes since some of this information will be outdated

src/Blueprint/Blueprint/Az.Blueprint.psd1
```diff
@@ -0,0 +1,140 @@
...
+DotNetFrameworkVersion = '4.7.2'
+
+# Minimum version of the common language runtime (CLR) required by this module. This prerequisite is valid for the PowerShell Desktop edition only.
+# CLRVersion = ''
+
+# Processor architecture (None, X86, Amd64) required by this module
+# ProcessorArchitecture = ''
+
+# Modules that must be imported into the global environment prior to importing this module
+RequiredModules = @(@{ModuleName = 'Az.Accounts'; ModuleVersion = '1.0.0'; })
```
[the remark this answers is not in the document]
filizt: Will do.

---
In reply to: [261381261](https://github.com/Azure/azure-powershell/pull/8537#discussion_r261381261) [](ancestors = 261381261)

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
```diff
@@ -0,0 +1,387 @@
+﻿using Microsoft.Azure.Commands.Blueprint.Common;
```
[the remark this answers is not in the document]
filizt: Done.

---
In reply to: [261382467](https://github.com/Azure/azure-powershell/pull/8537#discussion_r261382467) [](ancestors = 261382467)

src/Blueprint/Blueprint/Cmdlets/SetAzureRMBlueprintAssignment.cs
filizt: It's a PUT call

---
In reply to: [261403792](https://github.com/Azure/azure-powershell/pull/8537#discussion_r261403792) [](ancestors = 261403792)

src/Blueprint/Blueprint/Common/BlueprintConstants.cs
filizt: The AzureBlueprintAppId is used  to query Blueprint service's first party app object id and OwnerRoleDefinition is used as the definition id to assign owner role to blueprint first party app. They can both be public.

---
In reply to: [261404672](https://github.com/Azure/azure-powershell/pull/8537#discussion_r261404672) [](ancestors = 261404672)

src/Blueprint/Blueprint/help/Get-AzBlueprint.md
filizt: Ok - keeping this feedback active to track at the very end.

---
In reply to: [261405912](https://github.com/Azure/azure-powershell/pull/8537#discussion_r261405912) [](ancestors = 261405912)

src/Blueprint/Blueprint/Blueprint.csproj
```diff
@@ -0,0 +1,30 @@
...
+  <PropertyGroup>
+    <PsModuleName>Blueprint</PsModuleName>
+    <RootNamespace>Microsoft.Azure.PowerShell.Cmdlets.Blueprint</RootNamespace>
+  </PropertyGroup>
+
+  <Import Project="$(MSBuildThisFileDirectory)..\..\Az.props" />
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.Management.Blueprint" Version="0.10.0-preview" />
+    <PackageReference Include="System.Security.Permissions" Version="4.5.0" />
```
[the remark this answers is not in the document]
filizt: Artifact from the previous developer of the cmdlets. Doesn't seem to be necessary for anything in the project. Removing.

---
In reply to: [261381889](https://github.com/Azure/azure-powershell/pull/8537#discussion_r261381889) [](ancestors = 261381889)

src/Blueprint/Blueprint/Cmdlets/NewAzureRMBlueprintAssignment.cs
```diff
@@ -0,0 +1,130 @@
...
+        public string Location { get; set; }
+
+        [Parameter(ParameterSetName = ParameterSetNames.CreateBlueprintAssignment, Mandatory = false, ValueFromPipelineByPropertyName = true)]
+        [ValidateNotNullOrEmpty]
+        public Hashtable ResourceGroup { get; set; }
```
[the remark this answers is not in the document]
filizt: This ResourceGroup parameter is for the blueprint assignment. It's a key value pairing where key is the resourcegroup name user defined in the blueprint and the value being resourcegroup type we define. User can also provide multiple resource groups for the blueprint assignment. If the parameter name "ResourceGroup" can be confusing for the user, we can rename it.

---
In reply to: [261388522](https://github.com/Azure/azure-powershell/pull/8537#discussion_r261388522) [](ancestors = 261388522)

src/Blueprint/Blueprint/Cmdlets/SetAzureRMBlueprintAssignment.cs
filizt: We can. Would it be ok to move this to next cmdlet development iteration?

---
In reply to: [261404022](https://github.com/Azure/azure-powershell/pull/8537#discussion_r261404022) [](ancestors = 261404022)

src/Blueprint/Blueprint.Test/ScenarioTests/BlueprintAssignmentTests.ps1
```diff
@@ -0,0 +1,141 @@
...
+	Assert-NotNull $assignments[0].BlueprintId
+	Assert-NotNull $assignments[0].Scope
+	Assert-NotNull $assignments[0].Location
+}
+
+function Test-NewBlueprintAssignment
+{
+	$mgId = "AzBlueprintAssignTest"
+	$blueprintName = "Filiz-Ps-Test1"
+	$subscriptionId = "28cbf98f-381d-4425-9ac4-cf342dab9753"
```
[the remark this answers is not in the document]
filizt: Adding this as a note for the next iteration.

---
In reply to: [261380808](https://github.com/Azure/azure-powershell/pull/8537#discussion_r261380808) [](ancestors = 261380808)

src/Blueprint/Blueprint/Cmdlets/BlueprintCmdletBase.cs
```diff
@@ -0,0 +1,387 @@
...
+
+        protected void ThrowIfAssignmentExits(string scope, string name)
+        {
+            PSBlueprintAssignment assignment = null;
+
+            try
+            {
+                assignment = BlueprintClient.GetBlueprintAssignment(scope, name);
+            }
+            catch (Exception ex)
```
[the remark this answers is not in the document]
filizt: Neat! I'll move to using "when" in the exception catching logic in the next iteration. 

---
In reply to: [261384307](https://github.com/Azure/azure-powershell/pull/8537#discussion_r261384307) [](ancestors = 261384307)

src/Blueprint/Blueprint/Cmdlets/NewAzureRMBlueprintAssignment.cs
```diff
@@ -0,0 +1,130 @@
...
+    [Cmdlet(VerbsCommon.New, ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "BlueprintAssignment", SupportsShouldProcess = true, DefaultParameterSetName = ParameterSetNames.CreateBlueprintAssignment), OutputType(typeof(PSBlueprintAssignment))]
+    public class NewAzureRmBlueprintAssignment : BlueprintCmdletBase
+    {
+        #region Parameters
+        [Parameter(ParameterSetName = ParameterSetNames.CreateBlueprintAssignment, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = ParameterHelpMessages.BlueprintAssignmentName)]
+        [ValidatePattern("^[0-9a-zA-Z_-]*$", Options = RegexOptions.Compiled | RegexOptions.CultureInvariant)]
+        [ValidateNotNullOrEmpty]
+        public string Name { get; set; }
+
+        [Parameter(ParameterSetName = ParameterSetNames.CreateBlueprintAssignment, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = ParameterHelpMessages.BlueprintObject)]
```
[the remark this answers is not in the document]
filizt: Adding this as an item for the next iteration.

---
In reply to: [261389042](https://github.com/Azure/azure-powershell/pull/8537#discussion_r261389042) [](ancestors = 261389042)

src/Blueprint/Blueprint/Cmdlets/NewAzureRMBlueprintAssignment.cs
filizt: We chatted about this offline. Keeping it as is for now.

---
In reply to: [261402747](https://github.com/Azure/azure-powershell/pull/8537#discussion_r261402747) [](ancestors = 261402747)

src/Blueprint/Blueprint/Cmdlets/RemoveAzureRMBlueprintAssignment.cs
filizt: Renamed.

---
In reply to: [261403238](https://github.com/Azure/azure-powershell/pull/8537#discussion_r261403238) [](ancestors = 261403238)

cormacpayne:
Sign job test: https://azuresdkci.westus2.cloudapp.azure.com/job/psnetsign-esrp/202/

cormacpayne:
Sign job: https://azuresdkci.westus2.cloudapp.azure.com/job/psnetsign-esrp/213/

cormacpayne:
Sign job: https://azuresdkci.westus2.cloudapp.azure.com/job/psnetsign-esrp/217/

cormacpayne:
Sign job: https://azuresdkci.westus2.cloudapp.azure.com/job/psnetsign-esrp/221/

Issue #8494, opened 2019-02-07
Set-AzureKeyVaultSecret -tag doesn't accept numbers in tags


amcgregor1:
### Description

Unable to use a date stamp for tagging KeyVault Secrets.
Important: I am running this against AzureStack so need to use the specific modules listed below

### Script/Steps for Reproduction

```powershell
PS C:\Users\mcgregan\hpeazservices\AzSIaC\pipeline> $Secret = ConvertTo-SecureString -String 'Password' -AsPlainText -Force
PS C:\Users\mcgregan\hpeazservices\AzSIaC\pipeline> $Tags = @{ 'RandomTag' = 'Seems:OKwith-these_Chars'}
PS C:\Users\mcgregan\hpeazservices\AzSIaC\pipeline>
PS C:\Users\mcgregan\hpeazservices\AzSIaC\pipeline> Set-AzureKeyVaultSecret -VaultName 'acm-dev-equinix-kvt03' -Name 'ITSecret' -SecretValue $Secret -tag $tags


Vault Name   : acm-dev-equinix-kvt03
Name         : ITSecret
Version      : 285039cec5dd4bc0a7ffa6661669c3c1
Id           : https://acm-dev-equinix-kvt03.vault.equinix.vigilant.it:443/secrets/ITSecret/285039cec5dd4bc0a7ffa6661669c3c1
Enabled      : True
Expires      :
Not Before   :
Created      : 7/02/2019 9:10:18 AM
Updated      : 7/02/2019 9:10:18 AM
Content Type :
Tags         : Name       Value
               RandomTag  Seems:OKwith-these_Chars

PS C:\Users\mcgregan\hpeazservices\AzSIaC\pipeline> $Tags = @{ 'NotOKwithThese' = (get-date -UFormat %Y%m%d-%H:%M:%S)}
PS C:\Users\mcgregan\hpeazservices\AzSIaC\pipeline> Set-AzureKeyVaultSecret -VaultName 'acm-dev-equinix-kvt03' -Name 'ITSecret' -SecretValue $Secret -tag $tags
Set-AzureKeyVaultSecret : Tag has invalid value
At line:1 char:1
+ Set-AzureKeyVaultSecret -VaultName 'acm-dev-equinix-kvt03' -Name 'ITS ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #8553, opened 2019-02-16
Fixes Set-AzKeyVaultSecret to allow objects as Tags (Issue #8494)

lwajswaj:
Fixes issue #8494 by forcing all values in Tags hashtable to be strings.

## Checklist

### DIFF
--- a/src/KeyVault/KeyVault/Commands/SetAzureKeyVaultSecret.cs
+++ b/src/KeyVault/KeyVault/Commands/SetAzureKeyVaultSecret.cs
@@ -1,4 +1,4 @@
-﻿// ----------------------------------------------------------------------------------
+// ----------------------------------------------------------------------------------
 //
 // Copyright Microsoft Corporation
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


lwajswaj:
Attempted to use methods exposed at "TagsHelper" but they didn't adjust properly as they are receiving Dictionary<string, string> to get a HashTable OR a hashtable to return a Dictionary<string,string>. Made more sense to just create a new extension method.

maddieclayton:
@lwajswaj I think what makes the most sense here is to convert the Hashtable to a Dictionary<string, string> using https://github.com/Azure/azure-powershell-common/blob/84adc2e8e833a22caa8ea714cd5c367e63da227c/src/ResourceManager/Version2016_09_01/Tags/TagsConversionHelper.cs#L47, then convert that dictionary back to a Hashtable using https://github.com/Azure/azure-powershell-common/blob/84adc2e8e833a22caa8ea714cd5c367e63da227c/src/ResourceManager/Version2016_09_01/Tags/TagsConversionHelper.cs#L82.  This should do exactly what your code is doing, and will guarantee that KeyVault works the exact same as other cmdlets in our repo that convert between Hashtable and Dictionary<string, string>.  I believe that this cmdlet fails where others do not because most services accept the Dictionary<string, string> rather than a Hashtable, so we are doing the CreateTagDictionary call which casts all tag keys and values to strings for all other cmdlet calls.

lwajswaj:
@maddieclayton, the problem with using that approach is the actual exception is being thrown by these lines:

```c#
if (tag.Value != null && !(tag.Value is string))
                    throw new ArgumentException("Tag has invalid value");
```

In TagHelper.ConvertToDictionary. Should we remove that check and replace the following line with a ToString() instead of casting it?

```c#
string value = (tag.Value == null) ? string.Empty : tag.Value.ToString();
```

thoughts? That's why I chose create that extra extension, it seemed cleaner

lwajswaj:
so, ConvertToDictionary would look like this:

```c#
public static Dictionary<string, string> ConvertToDictionary(this Hashtable tags)
        {
            Dictionary<string, string> tagsDictionary = new Dictionary<string, string>();
            foreach (DictionaryEntry tag in tags)
            {
                string key = tag.Key as string;
                if (string.IsNullOrWhiteSpace(key))
                    throw new ArgumentException("Invalid tag name");

                string value = (tag.Value == null) ? string.Empty : tag.Value.ToString();
                tagsDictionary[key] = value;
            }

            return tagsDictionary;
        }
```

lwajswaj:
Hey @maddieclayton, sorry for the coming and going.... I think updating "ConvertToDictionary" with the logic I just pushed (it's basically based on your recommendation) it's better as I see it's referenced 16 different times and each one of them is a possible issue if the passed hashtable doesn't contain all strings. In this way, we prevent other issues to occur.

maddieclayton:
In my opinion, rather than fixing the KeyVault specific implementation, we should change the cmdlet to use the general implementation: https://github.com/Azure/azure-powershell-common/blob/84adc2e8e833a22caa8ea714cd5c367e63da227c/src/ResourceManager/Version2016_09_01/Tags/TagsConversionHelper.cs. I believe this already does what you are fixing in the KeyVault implementation.

If this is too much work, I can take over the PR - just let me know

lwajswaj:
yeap, you are right. For this cmdlet to use general implementation, it's an easy fix. Just pushed the updates to do so

maddieclayton:
Just to make this a little more simple I edited the change down to just the error handling of the Tag hashtable - should be good to go once CI passes!

maddieclayton:
@lwajswaj Thanks for the contribution!!

PR #8598, opened 2019-02-22
[Storage] Move DMlib reference from from Az.Accounts to Az.Storage

blueww:
## Description

Move DMlib reference from from Az.Accounts to Az.Storage, since it only used by storage module.
If don't move it, upgrade DMlib with XSCL upgrade in storage module will fail:

- If we don’t include DMlib in “RequiredAssemblies” of Storage.Management\Az.Storage.psd1, run DMlib related cmdlets in Powershell will fail as new version DMlib not found.
- If we include DMlib in “RequiredAssemblies” of Storage.Management\Az.Storage.psd1, Unit test will fail: 2 version DMlib in module az.Accounts and Az.Storage, cause import Az.Storage fail as a different version of DMlib already imported with Az.Accounts, so storage cmdlets can’t be found in Unit test.

## Checklist


markcowl:
@blueww So, we cannot do this - older versions of Az.Storage may be used with the current Az.Accounts, and those modules would fail to load in this case.

For now, we will have to update DMLib in Az.Accounts, and it will have to remain backward compatible.

blueww:
@mark
I see, will try to do it in Build release which allows breaking.
Close this for now.

PR #8616, opened 2019-02-25
added permissions to cmdlet

MashaMSFT:
## Description

added permissions for this cmdlet

## Checklist

- [x ] I have read the [_Submitting Changes_](https://github.com/Azure/azure-powershell/blob/master/CONTRIBUTING.md#submitting-changes) section of [`CONTRIBUTING.md`](https://github.com/Azure/azure-powershell/blob/master/CONTRIBUTING.md)
- [ x] The title of the PR is clear and informative
- [x ] The PR does not introduce [breaking changes](https://github.com/Azure/azure-powershell/blob/master/documentation/breaking-changes/breaking-changes-definition.md)
- [x ] If applicable, the changes made in the PR have proper test coverage

### DIFF
--- a/src/ResourceManager/Sql/Commands.Sql/help/Switch-AzureRmSqlDatabaseFailoverGroup.md
+++ b/src/ResourceManager/Sql/Commands.Sql/help/Switch-AzureRmSqlDatabaseFailoverGroup.md
@@ -24,7 +24,7 @@ The Failover Group's secondary server must be used to execute this command.
 
 ## Permissions
 
-Failover is an action on the target-side failover group entity, so to failover you need RBAC write access to the failover group resource path on the new primary server (ex: `/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Sql/servers/{serverName}/failoverGroups/{failoverGroupName}`)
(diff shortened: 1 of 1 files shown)


adxsdkps:

cormacpayne:
@MashaMSFT Hey Masha, two comments:

(1) Back in December, we switched from shipping the `AzureRM` module every two weeks to the new `Az` module, which includes the help content that appears on https://docs.microsoft.com. As a result, we kept the legacy code for the `AzureRM` module in the `preview` branch and moved all active development of the `Az` module (and its docs) into the `master` branch. As a result, you'll want to close this PR and create one that targets the `master` branch and the `Switch-AzSqlDatabaseFailoverGroup.md` file.

(2) We use the [`platyPS`](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/help-generation.md) PowerShell module to generate the markdown help files, which creates the files with specific headers and expects those headers when converting the markdown files into a "PowerShell readable form" (`dll-Help.xml`). My concern about your change is that the next time a user attempts to regenerate the help for `Switch-AzSqlDatabaseFailoverGroup`, it will undo the change that you've made (since it expects the content in a specific form), or it will throw an error because the format is different than expected. I would suggest that we move the content you're adding to be under the `## DESCRIPTION` or `## NOTES` headers as its own line, which will guarantee it's not removed when regenerating the file.

I'm going to close this PR for now since it should be re-opened against the `master` branch, but let me know if you have any questions!

PR #8646, opened 2019-02-28
Az.StorageSync - Handle AzureUSGovernment in Cloudendpoint provisioning for ensuring Service Principals

ankushbindlish2:
## Checklist


src/StorageSync/StorageSync/Common/StorageSyncClientWrapper.cs
```diff
@@ -146,6 +153,24 @@ public string AfsAgentInstallerPath
...
+        /// Gets the current application identifier.
+        /// </summary>
+        /// <value>The current application identifier.</value>
+        public Guid CurrentApplicationId
+        {
+            get
+            {
+                switch (AzureRmProfileProvider.Instance?.Profile?.DefaultContext?.Environment?.Name)
+                {
+                    case EnvironmentName.AzureUSGovernment:
```
markcowl: Ideally, this should be discoverable via API.  Failing that, this should be an extended property of the built-in environment, meaning that you can just get the value from the environment in the same way.

For a preview this is OK, but please file an issue to either add a discovery API, or add this value to the built-in environment extensions

src/StorageSync/StorageSync/Common/StorageSyncClientWrapper.cs
```diff
@@ -41,9 +42,15 @@ namespace Microsoft.Azure.Commands.StorageSync.Common
     public class StorageSyncClientWrapper : IStorageSyncClientWrapper
     {
         /// <summary>
-        /// The kailani application identifier
+        /// The azure cloud kailani application identifier
         /// </summary>
-        public static Guid KailaniAppId = new Guid(StorageSyncResources.KailaniApplicationId);
+        private static Guid AzureCloudKailaniAppId = new Guid(StorageSyncResources.AzureCloudKailaniAppId);
```
markcowl: this is OK, but these don't need to be resource strings, since they are not user-facing (the reason we make things resource strings is so they can be translated, if necessary)

src/StorageSync/StorageSync/Common/StorageSyncClientWrapper.cs
ankushbindlish2: https://github.com/Azure/azure-powershell/issues/8652

Issue #6549, opened 2018-06-27
Add wildcard support and ensure top level resources can be filtered by Name only


kilasuit:
### Description

I should be able to return VM's without needing to supply a Resource Group name directly from calling  get-AzureRMVM without having to pipe this to Where-Object 

### Script/Steps for Reproduction

```powershell
$VMName = 'MyVM'
Get-AzureRMVM -Name  $VMName 

```

forces a request for the resource group name - this should not be a requirement when Name is provided and should be 2 seperate Parameter Sets, one that allows you to get the VM by name and one that allows you to get all VM's in a Resource Group by Resource Group Name

This is likely not just limited to VM's but all resource types
### Module Version

AzureRM.Compute 5.1.1
AzureRM 6.3.0
### Environment Data

``` powershell
$psversiontable

Major  Minor  Build  Revision
-----  -----  -----  --------
5      1      16299  431     
```

cormacpayne:
@kilasuit Hey Ryan, thanks for the feedback -- I think this is something that should be done for all `Get-*` cmdlets to allow for another type of filter 😀

@praries880 would you mind syncing up with @hyonholee about what would need to be done to enable this scenario in the generated cmdlet code for Compute?

kilasuit:
Thanks @cormacpayne  - @hyonholee & @praries880 - Any updates on this at all?

markcowl:
@kilasuit Yes, we are looking at a common set of functionality across cmdlets to allow wildcards and completion over the components of the resource ID.  This would include allowing you to filter the list of VMs by name (or partial name) for example, and potentially auto-complete VM names by partial name.

We plan to prototype this functionality post-ignite, and have this released later this year.

markcowl:
@maddieclayton Post Resource completer - Get cmdlets for top level resources that:
(1) Allow providing a resource name or wildcard to list all VMs with that name

cormacpayne:
@kilasuit the has been added to `Get-AzVM`, but we are looking for more areas to add this functionality as well

maddieclayton:
This seems to be multiple parts:
- [ ] Ensure that all top level resources can list over subscription as well as resourcegroupname
- [ ] Add wildcard support for all Get "Name" parameters to allow listing with filters

We need to decide whether this will just be in top level resources, and whether we only filter by "Name", or if we would support filtering by parent resource name as well (this could increase the time that it takes to run the cmdlet by a fair amount).

maddieclayton:
[...]

PR #8711, opened 2019-03-06
Add wildcard support to KeyVault cmdlets

maddieclayton:
## Description
https://github.com/Azure/azure-powershell/issues/6549

## Checklist

### DIFF
--- a/src/KeyVault/KeyVault/Models/KeyVaultManagementCmdletBase.cs
+++ b/src/KeyVault/KeyVault/Models/KeyVaultManagementCmdletBase.cs
@@ -126,32 +126,9 @@ namespace Microsoft.Azure.Commands.KeyVault
             return listResult;
         }
 
-        protected PSKeyVault FilterByTag(PSKeyVault keyvault, Hashtable tag)
+        protected PSKeyVault FilterByTag(PSKeyVault keyVault, Hashtable tag)
         {
-            var tagValuePair = new PSTagValuePair();
-            if (tag != null && tag.Count > 0)
-            {
-                tagValuePair = TagsConversionHelper.Create(tag);
-                if (tagValuePair == null)
-                {
-                    throw new ArgumentException(PSKeyVaultProperties.Resources.InvalidTagFormat);
-                }
-            }
-
-            List<PSKeyVault> listResult = new List<PSKeyVault>();
-            listResult.Add(keyvault);
-
-            if (!string.IsNullOrEmpty(tagValuePair.Name))
-            {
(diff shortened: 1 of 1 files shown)


src/KeyVault/KeyVault/Models/KeyVaultCmdletBase.cs
```diff
@@ -68,5 +70,38 @@ protected void GetAndWriteObjects<TObject>(KeyVaultObjectFilterOptions options,
                 WriteObject(pageResults, true);
             } while (!string.IsNullOrEmpty(options.NextLink));
         }
+        
+        public List<T> KVSubResourceWildcardFilter<T>(string name, IEnumerable<T> resources)
```
cormacpayne: @maddieclayton is there any reason that we need to implement this method and can't just use the following:

```cs
TopLevelWildcardFilter(null, name, resources)
```

src/KeyVault/KeyVault/Models/KeyVaultManagementCmdletBase.cs
```diff
@@ -112,9 +112,53 @@ protected List<PSKeyVaultIdentityItem> ListVaults(string resourceGroupName, Hash
...
+
+            if (!string.IsNullOrEmpty(tagValuePair.Value))
+            {
+                listResult = listResult.Where(r => r.Tags?.Values != null && r.Tags.ConvertToDictionary().Values.Any(v => string.Equals(v, tagValuePair.Value, StringComparison.OrdinalIgnoreCase))).ToList();
+            }
+
+            return listResult;
+        }
+
+        protected PSKeyVault FilterByTag(PSKeyVault keyvault, Hashtable tag)
```
cormacpayne: @maddieclayton is this method just checking to see if the provided `PSKeyVault` object has the given tag?

maddieclayton: Yes - I couldn't think of a good way to have parameter set that accepted either a name or a resourcegroup (but not both) and tags do to filtering, so I added in this extra logic in case you give both, as I would expect the cmdlet to return nothing if the tags of the specific KV don't match.

cormacpayne: @cormacpayne to prevent duplicating code, we should have this function be implemented as follows:

```cs
protected PSKeyVault FilterByTag(PSKeyVault keyVault, Hashtable tag)
{
    return FilterByTag(new List<PSKeyVaultIdentityItem> { keyVault }, tag).FirstOrDefault();
}
```

PR #8815, opened 2019-03-21
EventGrid: PowerShell for Api-Version 2019-02-01-preview

ahamad-MS:
## Description

Adding PowerShell support for the new API Version 2019-02-01-preview in Event Grid. This includes adding new cmdlets for create/delete domaintopics, and updating current cmdlets to support servicebusqueue as destination, pagination and search filtering for List operation and disallow All for IncludedEventTypes

The design review of this work is covered here: https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/257

## Checklist

    - [X] the changes have gone through a [cmdlet design review](https://github.com/Azure/azure-powershell-cmdlet-review-pr) (Microsoft internal partners only)
    - Link to the design review: https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/257

### DIFF
--- a/src/EventGrid/EventGrid/ChangeLog.md
+++ b/src/EventGrid/EventGrid/ChangeLog.md
@@ -19,7 +19,6 @@
 -->
 ## Upcoming Release
 
-## Version 1.1.1
 * Updated to use the 2019-02-01-preview API version.
 * New commands:
     - New-AzureRmEventGridDomainTopic:
--- a/src/EventGrid/EventGrid/Domain/GetAzureEventGridDomain.cs
+++ b/src/EventGrid/EventGrid/Domain/GetAzureEventGridDomain.cs
@@ -107,7 +107,7 @@ namespace Microsoft.Azure.Commands.EventGrid
             HelpMessage = EventGridConstants.TopHelp,
             ParameterSetName = ResourceIdEventSubscriptionParameterSet)]
         [ValidateRange(1, 100)]
-        public int? Top { get; set; }
+        public int Top { get; set; }
 
         [Parameter(
             Mandatory = false,
@@ -123,6 +123,12 @@ namespace Microsoft.Azure.Commands.EventGrid
             string domainName = string.Empty;
             IEnumerable<Domain> domainsList;
             string newNextLink = null;
+            int? providedTop = null;
+
+            if (MyInvocation.BoundParameters.ContainsKey(nameof(this.Top)))
+            {
+                providedTop = this.Top;
+            }
 
             if (!string.IsNullOrEmpty(this.ResourceId))
             {
@@ -168,14 +174,14 @@ namespace Microsoft.Azure.Commands.EventGrid
             else if (!string.IsNullOrEmpty(resourceGroupName) && string.IsNullOrEmpty(domainName))
             {
                 // List all Event Grid domains in the given resource group
-                (domainsList, newNextLink) = this.Client.ListDomainsByResourceGroup(resourceGroupName, this.ODataQuery, this.Top);
+                (domainsList, newNextLink) = this.Client.ListDomainsByResourceGroup(resourceGroupName, this.ODataQuery, providedTop);
                 PSDomainListPagedInstance pSDomainListPagedInstance = new PSDomainListPagedInstance(domainsList, newNextLink);
                 this.WriteObject(pSDomainListPagedInstance, true);
             }
             else if (string.IsNullOrEmpty(resourceGroupName) && string.IsNullOrEmpty(domainName))
             {
                 // List all Event Grid domains in the given subscription
-                (domainsList, newNextLink) = this.Client.ListDomainsBySubscription(this.ODataQuery, this.Top);
+                (domainsList, newNextLink) = this.Client.ListDomainsBySubscription(this.ODataQuery, providedTop);
                 PSDomainListPagedInstance pSDomainListPagedInstance = new PSDomainListPagedInstance(domainsList, newNextLink);
                 this.WriteObject(pSDomainListPagedInstance, true);
             }
--- a/src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomainTopic.cs
+++ b/src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomainTopic.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Azure.Commands.EventGrid
     {
         [Parameter(
             Mandatory = true,
-            ValueFromPipelineByPropertyName = true,
+            ValueFromPipelineByPropertyName = false,
             Position = 0,
             HelpMessage = EventGridConstants.ResourceGroupNameHelp,
             ParameterSetName = DomainTopicNameParameterSet)]
--- a/src/EventGrid/EventGrid/help/Get-AzEventGridDomain.md
+++ b/src/EventGrid/EventGrid/help/Get-AzEventGridDomain.md
@@ -195,10 +195,10 @@ Accept wildcard characters: False
 ```
 
 ### -Top
-The maximum number of resources to be obtained. Valid value is between 1 and 100. If top value is specified and more results are still available, the result will contain a link to the next page to be queried in NextLink. If the Top value is null or not specified, the full list of resources will be returned at once.
+The maximum number of resources to be obtained. Valid value is between 1 and 100. If top value is specified and more results are still available, the result will contain a link to the next page to be queried in NextLink. If the Top value is not specified, the full list of resources will be returned at once.
 
 ```yaml
-Type: System.Nullable`1[System.Int32]
+Type: System.Int32
 Parameter Sets: ResourceGroupNameParameterSet, DomainNameParameterSet, ResourceIdEventSubscriptionParameterSet
 Aliases:
 
--- a/src/EventGrid/EventGrid/Domain/GetAzureEventGridDomainTopic.cs
+++ b/src/EventGrid/EventGrid/Domain/GetAzureEventGridDomainTopic.cs
@@ -101,7 +101,7 @@ namespace Microsoft.Azure.Commands.EventGrid
             HelpMessage = EventGridConstants.ODataQueryHelp,
             ParameterSetName = ResourceIdDomainTopicParameterSet)]
         [ValidateRange(1, 100)]
-        public int? Top { get; set; }
+        public int Top { get; set; }
 
         [Parameter(
             Mandatory = false,
@@ -119,6 +119,12 @@ namespace Microsoft.Azure.Commands.EventGrid
             string newNextLink = null;
             IEnumerable<DomainTopic> domainTopicsList;
             List <PSDomainTopicListInstance> psDomainTopicsList = new List<PSDomainTopicListInstance>();
+            int? providedTop = null;
+
+            if (MyInvocation.BoundParameters.ContainsKey(nameof(this.Top)))
+            {
+                providedTop = this.Top;
+            }
 
             if (!string.IsNullOrEmpty(this.ResourceId))
             {
@@ -152,7 +158,7 @@ namespace Microsoft.Azure.Commands.EventGrid
                 else
                 {
                     // List all Event Grid domain topics in the given resource group/domain
-                    (domainTopicsList, newNextLink) = this.Client.ListDomainTopicsByDomain(resourceGroupName, domainName, this.ODataQuery, this.Top);
+                    (domainTopicsList, newNextLink) = this.Client.ListDomainTopicsByDomain(resourceGroupName, domainName, this.ODataQuery, providedTop);
                     PSDomainTopicListPagedInstance pSDomainTopicListPagedInstance = new PSDomainTopicListPagedInstance(domainTopicsList, newNextLink);
                     this.WriteObject(pSDomainTopicListPagedInstance, true);
                 }
--- a/src/EventGrid/EventGrid/EventSubscription/GetAzureEventGridSubscription.cs
+++ b/src/EventGrid/EventGrid/EventSubscription/GetAzureEventGridSubscription.cs
@@ -250,6 +250,12 @@ namespace Microsoft.Azure.Commands.EventGrid
             string scope;
             bool includeFullEndpointUrl = this.IncludeFullEndpointUrl.IsPresent;
             string newNextLink = null;
+            int? providedTop = null;
+
+            if (MyInvocation.BoundParameters.ContainsKey(nameof(this.Top)))
+            {
+                providedTop = this.Top;
+            }
 
             if (!string.IsNullOrEmpty(this.EventSubscriptionName))
             {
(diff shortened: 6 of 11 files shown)


src/EventGrid/EventGrid.Test/ScenarioTests/Common.ps1
```diff
@@ -55,7 +55,7 @@ Get Hybrid Connection ResourceId
 #>
 function Get-HybridConnectionResourceId
 {
-    return "/subscriptions/$subscriptionId/resourceGroups/<ResourceGroupName>/providers/Microsoft.Relay/namespaces/<NameSpace>/hybridConnections/<HybridConnectionName>"
+    return "/subscriptions/5b4b650e-28b9-4790-b3ab-ddbd88d727c4/resourceGroups/DevExpRg/providers/Microsoft.Relay/namespaces/DevExpRelayNamespace/hybridConnections/hydbridconnectiondestination"
```
cormacpayne: @ahamad-MS this is fine since we're in a separate preview branch, but we should make sure that resource ids aren't hard-coded in order to allow any user to come in and record your tests

src/EventGrid/EventGrid/ChangeLog.md
```diff
@@ -19,6 +19,21 @@
...
+* New commands:
+    - New-AzureRmEventGridDomainTopic:
+        - Creates a new Azure Event Grid Domain Topic.
+    - Remove-AzureRmEventGridDomainTopic:
+        - Remove an existing Azure Event Grid Domain Topic.
+
+* Update the following cmdlets to support new scenarios in 2019-02-01-preview API version
+    - New-AzureRmEventGridSubscription/Update-AzureRmEventGridSubscription:
+        - Add new enum for servicebusqueue as destination.
+        - Disllow usage of "All" in IncludesEventTypes option and replace it with $null
```
kalyanaj: Nit: IncludesEventTypes  should be -IncludedEventType (without the plural) per https://docs.microsoft.com/en-us/powershell/module/azurerm.eventgrid/new-azurermeventgridsubscription?view=azurermps-6.13.0.

src/EventGrid/EventGrid/ChangeLog.md
```diff
@@ -19,6 +19,21 @@
 -->
 ## Upcoming Release
 
+## Version 1.1.1
+* Updated to use the 2019-02-01-preview API version.
+* New commands:
+    - New-AzureRmEventGridDomainTopic:
+        - Creates a new Azure Event Grid Domain Topic.
+    - Remove-AzureRmEventGridDomainTopic:
+        - Remove an existing Azure Event Grid Domain Topic.
```
kalyanaj: Nit: Remove should be "Removes" to be consistent with "Creates" in the previous line above.

src/EventGrid/EventGrid/ChangeLog.md
kalyanaj: Nit: Typo in Disallow.

src/EventGrid/EventGrid/ChangeLog.md
```diff
@@ -19,6 +19,21 @@
 -->
 ## Upcoming Release
 
+## Version 1.1.1
```
cormacpayne: @ahamad-MS you can remove this header and keep all of the text under the `## Upcoming Release` header -- we'll do the versioning for this branch in a separate PR

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomain.cs
```diff
@@ -70,10 +72,57 @@ public class GetAzureEventGridDomain : AzureEventGridCmdletBase
...
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = EventGridConstants.TopHelp,
+            ParameterSetName = ResourceGroupNameParameterSet)]
+        [Parameter(
+            Mandatory = false,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = EventGridConstants.TopHelp,
+            ParameterSetName = ResourceIdEventSubscriptionParameterSet)]
+        [ValidateRange(1, 100)]
+        public int? Top { get; set; }
```
cormacpayne: @ahamad-MS nit: should this be nullable? If the user provides a value, it should be an integer between 1 and 100.

(this comment applies to all instances of this in the PR)

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomain.cs
```diff
@@ -89,8 +138,31 @@ public override void ExecuteCmdlet()
             {
                 resourceGroupName = this.ResourceGroupName;
             }
+            else if (!string.IsNullOrEmpty(this.NextLink))
```
cormacpayne: @ahamad-MS is this if-statement necessary? It seems that you should be able to replace the usage of `nextLink` with `NextLink` in the code below.

(this comment applies to all instances of this in the PR)

src/EventGrid/EventGrid/Domain/NewAzureEventGridDomainTopic.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        [Parameter(
+            Mandatory = true,
+            ValueFromPipelineByPropertyName = true,
+            Position = 1,
+            HelpMessage = EventGridConstants.DomainNameHelp,
+            ParameterSetName = DomainTopicNameParameterSet)]
+        [ValidateNotNullOrEmpty]
+        [Alias("Domain")]
+        public string DomainName { get; set; }
```
cormacpayne: @ahamad-MS in the cases where a user should provide the name of a resource that exists, you should add the [`ResourceNameCompleter`](https://github.com/cormacpayne/azure-powershell/blob/docs-update/documentation/development-docs/design-guidelines/parameter-best-practices.md#resource-name-completer) to allow users to tab through all of the values

(this comment applies to all instances of this in the PR)

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomainTopic.cs
```diff
@@ -0,0 +1,115 @@
...
+        ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "EventGridDomainTopic",
+        DefaultParameterSetName = DomainTopicNameParameterSet,
+        SupportsShouldProcess = true),
+    OutputType(typeof(bool))]
+
+    public class RemoveAzureEventGridDomainTopic : AzureEventGridCmdletBase
+    {
+        [Parameter(
+            Mandatory = true,
+            ValueFromPipelineByPropertyName = true,
```
cormacpayne: @ahamad-MS in cmdlets where you have a `-ResourceId` parameter, the `-ResourceId` parameter should be the only parameter with `ValueFromPipelineByPropertyName = true`

src/EventGrid/EventGrid/EventSubscription/NewAzureEventGridSubscription.cs
```diff
@@ -783,6 +783,13 @@ public override void ExecuteCmdlet()
                     WriteWarning(EventGridConstants.EventSubscriptionHandshakeValidationMessage);
                 }
 
+                if (this.IncludedEventType != null && this.IncludedEventType.Length == 1 && string.Equals(this.IncludedEventType[0], "All", StringComparison.OrdinalIgnoreCase))
```
cormacpayne: @ahamad-MS out of curiosity, is there a closed set of values that the user can provide to the `-IncludedEventType` parameter? If so, we should use a generic [`PSArgumentCompleter`](https://github.com/cormacpayne/azure-powershell/blob/docs-update/documentation/development-docs/design-guidelines/parameter-best-practices.md#generic-argument-completer) to specify these values and allow users to tab over them.

src/EventGrid/EventGrid/help/Get-AzEventGridDomain.md
```diff
@@ -14,63 +14,89 @@ Gets the details of an Event Grid domain, or gets a list of all Event Grid domai
...
 ```
 
-List all the Event Grid domains in resource group \`MyResourceGroupName\`.
+List all the Event Grid domains in resource group \`MyResourceGroupName\` without pagination.
 
 ### Example 4
+```powershell
+$odataFilter = "Name ne 'ABCD'"
+PS C:\> $result = Get-AzEventGridDomain -ResourceGroup MyResourceGroupName -Top 10 -ODataQuery $odataFilter
+PS C:\> Get-AzEventGridDomain $result.NextLink
```
cormacpayne: @ahamad-MS for the examples you're adding, would you mind also including the corresponding output that the user would see if they ran the corresponding command?

src/EventGrid/EventGrid/ChangeLog.md
ahamad-MS: Fix. thx

src/EventGrid/EventGrid/ChangeLog.md
ahamad-MS: Fixed. thx

src/EventGrid/EventGrid/ChangeLog.md
ahamad-MS: fixed. thx

src/EventGrid/EventGrid/ChangeLog.md
ahamad-MS: Thx. fixed.

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomain.cs
ahamad-MS: Yes since we need to support no pagination scenario too (which is what previous customers are used to). In this context, if Top is null or not specified, then we will get all the results without pagination (old behavior which users are used to). 

If Top is specified, then the pagination behavior will kick in. 

If user specified Top as 0 or -negative  it will be an invalid value for pagination scenario and not a trigger for not supporting pagination.

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomain.cs
ahamad-MS: OK. Fixed.

src/EventGrid/EventGrid/Domain/NewAzureEventGridDomainTopic.cs
ahamad-MS: Adding this to all parameters (new and existing) where applicable.

src/EventGrid/EventGrid/EventSubscription/NewAzureEventGridSubscription.cs
ahamad-MS: There is no fixed list here. We have different lists for different topic types and even these lists are changing continuously. So we cannot use PSArgumentCompleter here efficiently.

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomainTopic.cs
ahamad-MS: Fixed

src/EventGrid/EventGrid/help/Get-AzEventGridDomain.md
ahamad-MS: Is there specific pattern or format for adding the output here. I checked multiple other cmdlets (for storage) and I don't seem to see such a pattern applied. I would like to address this comment for all the examples in the help to have consistent behavior but this will require separate changelist with additional info. Please feel free to open an issue and assign it to me and I will address it in separate PR if you don't mind.

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomain.cs
ahamad-MS: Fixed.

src/EventGrid/EventGrid/ChangeLog.md
cormacpayne: @ahamad-MS it doesn't look like this comment was resolved

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomain.cs
cormacpayne: @ahamad-MS if the concern is about determining if the user provided a value to the `-Top` parameter, then you can do this two ways:

(1) The recommended way is to use `MyInvocation.BoundParameters.ContainsKey[nameof(Top)]` -- this is built into every cmdlet and will tell you whether or not the user provided a value for `-Top`. This will work if the parameter is nullable or not.

(2) If the parameter isn't nullable, then you can always check to see if the value is 0 -- even though the default value is outside of the range, PowerShell only does validation on the values provided by the user, so if you find the `-Top` is set to 0 in the cmdlet, then you know that the user didn't provide a value for it.

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomainTopic.cs
cormacpayne: @ahamad-MS it looks like this comment wasn't resolved

src/EventGrid/EventGrid/help/Get-AzEventGridDomain.md
cormacpayne: @ahamad-MS in general, output from cmdlets should follow the below pattern:

- If there is a single object returned from a cmdlet, it should be in a list view
- If there is more than one object returned from a cmdlet, it should be in a table view

Information about how to properly format these outputs can be found [here](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/generate-format.ps1xml-file.md).

Examples should contain the corresponding output that users would see if they were to run the command in the example -- this allows users to see properties (with sample values) are returned with the object. You can check out the [KeyVault help folder](https://github.com/Azure/azure-powershell/tree/master/src/KeyVault/KeyVault/help) for examples of existing cmdlets with outputs displayed in their examples.

src/EventGrid/EventGrid/ChangeLog.md
ahamad-MS: hhhmm.. Weird. re-fixing this.

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomainTopic.cs
ahamad-MS: fixed.

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomain.cs
ahamad-MS: OK, we are taking option 1 of using MyInvocation.BoundParameters.ContainsKey(nameof(Top)) and updating code and help text accordingly

src/EventGrid/EventGrid/help/Get-AzEventGridDomain.md
ahamad-MS: Thx for the info. Per our offline discussion, we will address this comment in separate checkin.

Issue #2626, opened 2016-07-18
Set-AzureRmVMCustomScriptExtension is not supporting PSVirtualMachine parameter


iricigor:
At the moment, there is only `-VMName` as parameter of type [string]. I cannot pass VM configuration object to script (type `[Microsoft.Azure.Commands.Compute.Models.PSVirtualMachine]` or similar). This is making command unusable for VMs that are being provisioned. One needs to provision non-customized (or unsecured machine) and wait for for its availability.
This functionality was present in old Azure PowerShell, i.e. command `Set-AzureVMCustomScriptExtension` was working just fine. The same is with other verbs (Get- , Remove-) in new Azure PowerShell.

maddieclayton:
@danielsollondon Can you take a look at this issue?

PR #8816, opened 2019-03-21
Add new parameter sets to Set-AzVMCustomScriptExtension with piping for resource id, resource and top level resource objects

niander:
## Description

https://github.com/Azure/azure-powershell/issues/2626

Also includes,
* Small bug fix for `VirtualMachineCustomScriptExtensionContext` not populating the `VMName` field
* Two new helper functions for testing: `Assert-Match` and `Assert-NotMatch` with development-doc updates

## Checklist

### DIFF
--- a/src/Compute/Compute/help/Set-AzVMCustomScriptExtension.md
+++ b/src/Compute/Compute/help/Set-AzVMCustomScriptExtension.md
@@ -15,7 +15,7 @@ Adds a custom script extension to a virtual machine.
 
 ### ByNameWithContainerAndFileNamesParameterSet (Default)
 ```
-Set-AzVMCustomScriptExtension [-ResourceGroupName] <String> -VMName <String> [[-Name] <String>]
+Set-AzVMCustomScriptExtension [-ResourceGroupName] <String> [-VMName] <String> [-Name <String>]
  -ContainerName <String> -FileName <String[]> [-StorageAccountName <String>] [-StorageEndpointSuffix <String>]
  [-StorageAccountKey <String>] [-Run <String>] [-Argument <String>] [-SecureExecution]
  [-TypeHandlerVersion <String>] [-Location <String>] [-DisableAutoUpgradeMinorVersion] [-ForceRerun <String>]
@@ -24,7 +24,7 @@ Set-AzVMCustomScriptExtension [-ResourceGroupName] <String> -VMName <String> [[-
 
 ### ByNameWithFileUriParameterSet
 ```
-Set-AzVMCustomScriptExtension [-ResourceGroupName] <String> -VMName <String> [[-Name] <String>]
+Set-AzVMCustomScriptExtension [-ResourceGroupName] <String> [-VMName] <String> [-Name <String>]
  [-FileUri <String[]>] [-Run <String>] [-Argument <String>] [-SecureExecution] [-TypeHandlerVersion <String>]
  [-Location <String>] [-DisableAutoUpgradeMinorVersion] [-ForceRerun <String>]
  [-DefaultProfile <IAzureContextContainer>] [-WhatIf] [-Confirm] [<CommonParameters>]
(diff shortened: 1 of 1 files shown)


adxsdkps:

maddieclayton:
@azuresdkci Add to whitelist

src/Compute/Compute/help/Set-AzVMCustomScriptExtension.md
```diff
@@ -13,23 +13,73 @@ Adds a custom script extension to a virtual machine.
 
 ## SYNTAX
 
-### SetCustomScriptExtensionByContainerAndFileNames (Default)
+### ByNameWithContainerAndFileNamesParameterSet (Default)
```
maddieclayton: Can you regenerate this to pick up the new changes?

src/Compute/Compute/Extension/CustomScript/SetAzureVMCustomScriptExtensionCommand.cs
```diff
@@ -123,7 +311,38 @@ public override void ExecuteCmdlet()
...
+                    {
+                        this.ResourceGroupName = this.InputObject.ResourceGroupName;
+                        this.VMName = this.InputObject.VMName;
+                        this.Name = this.InputObject.Name;
+
+                        this.Location = string.IsNullOrEmpty(this.Location) ? this.InputObject.Location : this.Location;
+                        this.TypeHandlerVersion = string.IsNullOrEmpty(this.TypeHandlerVersion) ? this.InputObject.TypeHandlerVersion : this.TypeHandlerVersion;
+                        this.FileUri = (this.FileUri == null || !this.FileUri.Any()) ? this.InputObject.Uri : this.FileUri;
+
+                        string cmdToExe = this.InputObject.CommandToExecute;
```
maddieclayton: Is there any error checking we should add around these five lines? If not and we are always guaranteed to have the same format you can leave it as is.

niander: What kind of format are you talking about?

maddieclayton: Can you always split InputObject.CommandToExecute into Run and Argument, or could the formatting be different? Or would that fail to create in the first place? If it's not possible to have that incorrect formatting then there's no need to do error checking

niander: I changed the code a little bit. But it should always have `-file` since the custom script is a powershell script.

src/Compute/Compute/help/Set-AzVMCustomScriptExtension.md
```diff
@@ -295,16 +459,31 @@ This cmdlet adds the custom script extension for the virtual machine that this p
 
 ```yaml
 Type: System.String
-Parameter Sets: (All)
+Parameter Sets: ByNameWithContainerAndFileNamesParameterSet, ByNameWithFileUriParameterSet
 Aliases: ResourceName
 
 Required: True
-Position: 1
+Position: Named
```
maddieclayton: Hmm looked like this wasn't updated for some reason. Can you rebuild and regenerate again to fix this?

src/Compute/Compute/help/Set-AzVMCustomScriptExtension.md
```diff
@@ -172,11 +273,23 @@ Specifies the name of the custom script extension.
 
 ```yaml
 Type: System.String
-Parameter Sets: (All)
+Parameter Sets: ByNameWithContainerAndFileNamesParameterSet, ByNameWithFileUriParameterSet
 Aliases: ExtensionName
 
 Required: False
-Position: Named
+Position: 1
```
maddieclayton: Same here

PR #8838, opened 2019-03-25
Az.batch preview

v-akhana:
We connect to Azure from internal network through authentication proxy server. In case of using PowerShell version 5.1, proxy settings can be done as below.

-------------------
$proxyUrl = 'proxy URL'
$proxyId = 'proxy user ID'
$proxyPwd = 'proxy password'

$password = ConvertTo-SecureString $proxyPwd -AsPlainText -Force
$cred = New-Object System.Management.Automation.PSCredential $proxyId, $password
$proxy = New-Object System.Net.WebProxy $proxyUrl
$proxy.Credentials = $cred
[System.Net.WebRequest]::DefaultWebProxy = $proxy
------------------

In terms of PowerShell 6.x, I would like to use proxy as well, but cannot use the command let above with proxy error.
 I would like to ask you how to connect with “Connect-AzAccount” using proxy server and PowerShell 6.x.


adxsdkps:

cormacpayne:
@v-akhana please file an issue [here](https://github.com/Azure/azure-powershell/issues) and fill out the corresponding issue template.

PR #8853, opened 2019-03-26
SQR (Scheduled Query Rule) PS cmdlets and tests implementation

shpimpal:
## Checklist

### DIFF
--- a/src/Monitor/Monitor/Az.Monitor.psd1
+++ b/src/Monitor/Monitor/Az.Monitor.psd1
@@ -139,19 +139,7 @@ PrivateData = @{
       - Remove-AzMetricAlertRuleV2
       - Get-AzMetricAlertRuleV2
       - Add-AzMetricAlertRuleV2
-		* Updated Monitor SDK to version 0.22.0-preview
-		* New cmdlets for Log Search Alerts (Scheduled Query Rules)
-      - New-AzScheduledQueryRuleSource
-      - New-AzScheduledQueryRuleSchedule
-      - New-AzScheduledQueryRuleTriggerCondition
-      - New-AzScheduledQueryRuleLogMetricTrigger
-      - New-AzScheduledQueryRuleAznsActionGroup
-      - New-AzScheduledQueryRuleAlertingAction
-      - New-AzScheduledQueryRule
-      - Get-AzScheduledQueryRule
-      - Set-AzScheduledQueryRule
-      - Update-AzScheduledQueryRule
-      - Remove-AzScheduledQueryRule'
+  * Updated Monitor SDK to version 0.22.0-preview'
 
         # Prerelease string of this module
         # Prerelease = ''


adxsdkps:

cormacpayne:
Corresponding design review: https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/229

cormacpayne:
@shpimpal as mentioned in the Teams message, our static analysis tool caught a few exceptions:

(1) There is a property in the header of each cmdlet help markdown file you added that needs to be added. The values full values can be found [here](https://dev.azure.com/azure-sdk/_apis/resources/Containers/1770393?itemPath=artifacts%2FStaticAnalysisResults%2FValidateHelpIssues.csv)

(2) There are a few cases where you need to implement ShouldProcess and update plural nouns to be singular in parameters. The full exceptions can be found [here](https://dev.azure.com/azure-sdk/_apis/resources/Containers/1770393?itemPath=artifacts%2FStaticAnalysisResults%2FSignatureIssues.csv)

shpimpal:
Thanks @cormacpayne ! 
 (1) Do these links have to be added manually? I did not add anything extra (except for examples) to the md files than what Update-MarkdownHelpModule cmdlet created

(2) All but one New-* cmdlets create in-memory objects that are consumed by a New-* cmdlet that creates the actual resource in the backend. I will implement ShouldProcess for that cmdlet. But what about other cmdlets that just create objects in the memory? I believe they should not implement ShouldProcess.

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleAlertingActionCommand.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        [Parameter(Mandatory = false, HelpMessage = "The AzNS action details")]
+        [ValidateNotNullOrEmpty]
+        public PSScheduledQueryRuleAznsAction AznsAction { get; set; }
+
+        [Parameter(Mandatory = true, HelpMessage = "The severity for this alert")]
+        public string Severity { get; set; }
+
+        [Parameter(Mandatory = false, HelpMessage = "The duration in minutes for which alert should be throttled")]
+        public int? ThrottlingInMin { get; set; }
```
cormacpayne: @shpimpal should this be nullable? If the user wants this value to be `null`, then they should ignore providing a value for this parameter and it will be `null` by default

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleAznsActionGroupCommand.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// Create a ScheduledQueryRule Source object
+    /// </summary>
+    [Cmdlet(VerbsCommon.New, ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "ScheduledQueryRuleAznsActionGroup"), OutputType(typeof(PSScheduledQueryRuleAznsAction))]
+    public class NewScheduledQueryRuleAznsActionGroupCommand : MonitorCmdletBase
+    {
+
+        #region Cmdlet parameters
+
+        [Parameter(Mandatory = false, HelpMessage = "The list of action groups to send notification to")]
+        public List<string> ActionGroup { get; set; }
```
cormacpayne: @shpimpal this should be a `string[]` rather than `List<string>` (there's no difference for how users provide this data, but the documentation is represented better as an array)

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,128 @@
...
+        [Parameter(Mandatory = true, HelpMessage = "The scheduled query rule Alerting Action")]
+        [ValidateNotNullOrEmpty]
+        public PSScheduledQueryRuleAlertingAction Action { get; set; }
+
+        //
+        // Summary:
+        //     Region where alert is to be created
+        [Parameter(Mandatory = true, HelpMessage = "The location for this alert")]
+        [ValidateNotNullOrEmpty]
+        public string Location { get; set; }
```
cormacpayne: @shpimpal this should have the [`LocationCompleter`](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/design-guidelines/parameter-best-practices.md#location-completer) attribute to allow users to tab through the valid locations

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,128 @@
...
+        //     Alert description
+        [Parameter(Mandatory = false, HelpMessage = "The description for this alert")]
+        public string Description { get; set; }
+
+        //
+        // Summary:
+        //     Alert name
+        [Parameter(Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The alert name")]
+        [ResourceNameCompleter("Microsoft.insights/scheduledqueryrules", nameof(ResourceGroupName))]
+        public string RuleName { get; set; }
```
cormacpayne: @shpimpal same comment as before: this should be renamed to `-Name` and you can add an alias to `-RuleName`

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,128 @@
...
+        [Parameter(Mandatory = false, HelpMessage = "Resource tags")]
+        public IDictionary<string, string> Tag;
+
+        //
+        // Summary:
+        //     Alert status - enabled or not
+        [Parameter(Mandatory = true, HelpMessage = "The azure alert state - valid values - true, false")]
+        [ValidateSet("true", "false")]
+        [PSArgumentCompleter("true", "false")]
+        public string Enabled { get; set; }
```
cormacpayne: @shpimpal this shouldn't be a `string` -- you should use a `SwitchParameter` to represent this parameter. What is the default value for this resource? If it's enabled, then we should change this to a parameter called `Disable`, and vice versa if the default behavior is disabled.

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleLogMetricTriggerCommand.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    [Cmdlet(VerbsCommon.New, ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "ScheduledQueryRuleLogMetricTrigger"), OutputType(typeof(PSScheduledQueryRuleLogMetricTrigger))]
+    public class NewScheduledQueryRuleLogMetricTriggerCommand : MonitorCmdletBase
+    {
+
+        #region Cmdlet parameters
+
+        [Parameter(Mandatory = true, HelpMessage = "The metric threshold operator : GreaterThan, LessThan, Equal")]
+        [ValidateNotNullOrEmpty]
+        [ValidateSet("GreaterThan", "LessThan", "Equal")]
```
cormacpayne: @shpimpal this should be `PSArgumentCompleter` to allow for the same tab functionality that is allowed by `ValidateSet`, but also allows users to provide values outside this set (to immediately support new values as they're unveiled by the API)

This comment can be applied to all instances of `ValidateSet` being used

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleAlertingActionCommand.cs
cormacpayne: @shpimpal I would rename this to `ThrottlingInMinutes` to be consistent with other parameters introduced in this parameter

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleSourceCommand.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+
+        #region Cmdlet parameters
+
+        [Parameter(Mandatory = true, HelpMessage = "The alert query")]
+        [ValidateNotNullOrEmpty]
+        public string Query { get; set; }
+
+        [Parameter(Mandatory = false, HelpMessage = "The list of authorized resources for this alert")]
+        public List<string> AuthorizedResource { get; set; }
```
cormacpayne: @shpimpal same comment about changing this to `string[]`

src/Monitor/Monitor/ScheduledQueryRules/RemoveScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,108 @@
...
+        public PSScheduledQueryRuleResource InputObject { get; set; }
+
+        [Parameter(ParameterSetName = ByResourceId, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The resource Id")]
+        [ValidateNotNullOrEmpty]
+        public string ResourceId { get; set; }
+
+        /// <summary>
+        /// Alert name
+        /// </summary>
+        [Parameter(ParameterSetName = ByRuleName, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The alert name")]
```
cormacpayne: @shpimpal the `ValueFromPipelineByPropertyName` property should only be used for the `-ResourceId` parameter

src/Monitor/Monitor/ScheduledQueryRules/RemoveScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,108 @@
...
+        [ResourceNameCompleter("Microsoft.insights/scheduledqueryrules", nameof(ResourceGroupName))]
+        [ValidateNotNullOrEmpty]
+        public string RuleName { get; set; }
```
cormacpayne: @shpimpal same comment about changing this to `-Name` and possibly adding an alias to `-RuleName`

src/Monitor/Monitor/ScheduledQueryRules/SetScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,229 @@
...
+        public PSScheduledQueryRuleResource InputObject { get; set; }
+
+        [Parameter(ParameterSetName = ByResourceId, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The resource Id")]
+        [ValidateNotNullOrEmpty]
+        public string ResourceId { get; set; }
+
+        //
+        // Summary:
+        //     Gets or sets source (Query, DataSourceId, etc.) for rule.
+        [Parameter(ParameterSetName = ByRuleName, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The scheduled query rule source")]
```
cormacpayne: @shpimpal same comment about removing `ValueFromPipelineByPropertyName` from all parameters except `-ResourceId`

src/Monitor/Monitor/ScheduledQueryRules/SetScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,229 @@
...
+        public PSScheduledQueryRuleAlertingAction Action { get; set; }
+
+        //
+        // Summary:
+        //     Region where alert is to be created
+        [Parameter(ParameterSetName = ByRuleName, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The location for this alert")]
+        [Parameter(ParameterSetName = ByInputObject, Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The location for this alert")]
+        [Parameter(ParameterSetName = ByResourceId, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The location for this alert")]
+        [ValidateNotNullOrEmpty]
+        public string Location { get; set; }
```
cormacpayne: @shpimpal same comment about adding the `LocationCompleter`

src/Monitor/Monitor/ScheduledQueryRules/SetScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,229 @@
...
+
+        //
+        // Summary:
+        //     Alert status - enabled or not, supported values - "true", "false"
+        [Parameter(ParameterSetName = ByRuleName, Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The azure alert state - valid values - true, false")]
+        [Parameter(ParameterSetName = ByInputObject, Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The azure alert state - valid values - true, false")]
+        [Parameter(ParameterSetName = ByResourceId, Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = "The azure alert state - valid values - true, false")]
+        [ValidateSet("true", "false")]
+        [PSArgumentCompleter("true", "false")]
+        public string Enabled { get; set; }
```
cormacpayne: @shpimpal same comment about changing to this a `SwitchParameter`

src/Monitor/Monitor/ScheduledQueryRules/SetScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,229 @@
...
+        public SwitchParameter AsJob { get; set; }
+        #endregion
+
+        protected override void ProcessRecordInternal()
+        {
+            ResourceIdentifier resourceIdentifier = null;
+            ScheduledQueryRuleResource requestBody = null;
+
+            // ByInputObject parameter set
+            if (this.IsParameterBound(c => c.InputObject) || this.InputObject != null)
```
cormacpayne: @shpimpal you can remove the check for `this.InputObject != null` since we have the `ValidateNotNull` attribute on the `-InputObject` parameter

src/Monitor/Monitor/ScheduledQueryRules/SetScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,229 @@
...
+                resourceIdentifier = new ResourceIdentifier(this.InputObject.Id);
+                this.RuleName = resourceIdentifier.ResourceName;
+                this.ResourceGroupName = resourceIdentifier.ResourceGroupName;
+
+                requestBody = UpdateScheduledQueryRuleResource(this.InputObject);
+            }
+            else
+            {
+                // ByRuleName and ByResourceId parameter sets
+                if (this.IsParameterBound(c => c.ResourceId) || !string.IsNullOrWhiteSpace(this.ResourceId))
```
cormacpayne: @shpimpal you can remove the `!string.IsNullOrWhiteSpace(this.ResourceId)` check since there is a `ValidateNotNullOrEmpty` attribute on the `-ResourceId` parameter

src/Monitor/Monitor/ScheduledQueryRules/UpdateScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,138 @@
...
+using Microsoft.WindowsAzure.Commands.Utilities.Common;
+using Microsoft.Azure.Management.Monitor.Management.Models;
+using Microsoft.WindowsAzure.Commands.Common.CustomAttributes;
+
+namespace Microsoft.Azure.Commands.Insights.ScheduledQueryRules
+{
+    /// <summary>
+    /// Updates a ScheduledQueryRule object
+    /// </summary>
+    [Cmdlet(VerbsData.Update, ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "ScheduledQueryRule",
```
cormacpayne: @shpimpal is there a reason we have both `Set-*` and `Update-*` cmdlets? If this API supports an `Update` call with an underlying `PATCH` operation, then we should just have a `New-*` cmdlet with an `Update-*` cmdlet. Otherwise, it should be a `New-*` cmdlet with a `Set-*` cmdlet.

src/Monitor/Monitor/ScheduledQueryRules/UpdateScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        [Parameter(ParameterSetName = ByResourceId, Mandatory = true, ValueFromPipelineByPropertyName = true,
+            HelpMessage = "The resource Id")]
+        [ValidateNotNullOrEmpty]
+        public string ResourceId { get; set; }
+
+        /// <summary>
+        /// Alert name
+        /// </summary>
+        [Parameter(ParameterSetName = ByRuleName, Mandatory = true, ValueFromPipelineByPropertyName = true,
```
cormacpayne: @shpimpal same comment about removing `ValueFromPipelineByPropertyName` from all other parameters that aren't `-ResourceId`

src/Monitor/Monitor/ScheduledQueryRules/UpdateScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,138 @@
...
+            HelpMessage = "The alert name")]
+        [ResourceNameCompleter("Microsoft.insights/scheduledqueryrules", nameof(ResourceGroupName))]
+        [ValidateNotNullOrEmpty]
+        public string RuleName { get; set; }
```
cormacpayne: @shpimpal same comment about renaming this to `-Name` and possibly adding an alias to `-RuleName`

src/Monitor/Monitor/ScheduledQueryRules/UpdateScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,138 @@
...
+        /// </summary>
+        [Parameter(ParameterSetName = ByRuleName, Mandatory = true,
+            HelpMessage = "The azure alert state - valid values - true, false")]
+        [Parameter(ParameterSetName = ByInputObject, Mandatory = true,
+            HelpMessage = "The azure alert state - valid values - true, false")]
+        [Parameter(ParameterSetName = ByResourceId, Mandatory = true,
+            HelpMessage = "The azure alert state - valid values - true, false")]
+        [ValidateSet("true", "false")]
+        [PSArgumentCompleter("true", "false")]
+        public string Enabled { get; set; }
```
cormacpayne: @shpimpal same comment about changing this to `SwitchParameter`

src/Monitor/Monitor/ScheduledQueryRules/UpdateScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        #endregion
+
+        protected override void ProcessRecordInternal()
+        {
+            ResourceIdentifier resourceIdentifier = null;
+            ScheduledQueryRuleResource resource = null;
+
+            // ByInputObject parameter set
+            if (this.IsParameterBound(c => c.InputObject) || this.InputObject != null)
```
cormacpayne: @shpimpal same comment in your cmdlets about removing additional null checks for parameters that have a validation attribute

src/Monitor/Monitor/help/Get-AzScheduledQueryRule.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+## DESCRIPTION
+Gets Scheduled Query Resources
+
+## EXAMPLES
+
+### Example 1
+```powershell
+Parameter Set: BySubscriptionOrResourceGroup
```
cormacpayne: @shpimpal you can remove these `Parameter Set` comments made in this example and have multiple examples that display each of the different parameter sets

src/Accounts/Accounts/AzureRmAlias/Mappings.json
```diff
@@ -1248,7 +1248,18 @@
     "Set-AzActionGroup": "Set-AzureRmActionGroup",
     "Get-AzActionGroup": "Get-AzureRmActionGroup",
     "Remove-AzActionGroup": "Remove-AzureRmActionGroup",
-    "New-AzMetricFilter": "New-AzureRmMetricFilter"
+    "New-AzMetricFilter": "New-AzureRmMetricFilter",
+    "New-AzScheduledQueryRuleSource": "New-AzureRmScheduledQueryRuleSource",
```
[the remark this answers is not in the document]
shpimpal: ok, will remove these.

src/Monitor/Monitor.Test/ScenarioTests/ScheduledQueryRulesTests.ps1
```diff
@@ -0,0 +1,247 @@
...
+$emailSubject = "SQR Log alert trigger notification"
+$customWebhookPayload = "{}"
+$thresholdOperator = "GreaterThan"
+$threshold = 5
+$metricTriggerType = "Total"
+$metricTriggerColumn = "timestamp"
+$frequencyInMin = 5
+$timeWindowInMin = 5
+$query = "traces | summarize AggregatedValue = count() by bin(timestamp, 5m)"
+$authorizedResources = "/subscriptions/ad825170-845c-47db-8f00-11978947b089/resourceGroups/PSCmdletRG/providers/microsoft.insights/components/PSCmdletsAI"
```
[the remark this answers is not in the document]
shpimpal: are they supposed to be left blank or some dummy value? I have seen other implementations hardcoding these the same way.

src/Monitor/Monitor.Test/ScenarioTests/ScheduledQueryRulesTests.ps1
```diff
@@ -0,0 +1,247 @@
...
+
+}
+
+<#
+.SYNOPSIS
+
+#>
+function Test-NewGetUpdateSetRemoveScheduledQueryRule
+{
+	Write-Output "Starting Test-NewGetUpdateSetRemoveScheduledQueryRule"
```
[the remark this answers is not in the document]
shpimpal: sure, will use Write-Debug instead of Write-Object and Write-Verbose.

src/Monitor/Monitor/ChangeLog.md
```diff
@@ -19,6 +19,21 @@
 -->
 ## Upcoming Release
 
+## Version 1.0.2
```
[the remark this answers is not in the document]
shpimpal: Okay.

src/Monitor/Monitor/OutputClasses/PSScheduledQueryRuleLogMetricTrigger.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
```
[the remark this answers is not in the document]
shpimpal: Sure.

src/Monitor/Monitor/ScheduledQueryRules/GetScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,93 @@
...
+        [Parameter(Mandatory = false, ParameterSetName = BySubscriptionOrResourceGroup, HelpMessage = "The resource group name")]
+        [Parameter(Mandatory = true, ParameterSetName = ByRuleName, HelpMessage = "The resource group name")]
+        [ResourceGroupCompleter]
+        [ValidateNotNullOrEmpty]
+        public string ResourceGroupName { get; set; }
+
+        [Parameter(Mandatory = true, ParameterSetName = ByRuleName, HelpMessage = "The alert rule name")]
+        [ValidateNotNullOrEmpty]
+        [ResourceNameCompleter("Microsoft.insights/scheduledqueryrules", nameof(ResourceGroupName))]
+        public string RuleName { get; set; }
```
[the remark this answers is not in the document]
shpimpal: Okay.

src/Monitor/Monitor/ScheduledQueryRules/GetScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        [Parameter(Mandatory = true, ParameterSetName = ByResourceId, HelpMessage = "The resource Id")]
```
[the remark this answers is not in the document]
shpimpal: isn't the cmdlet that accepts piped output supposed to have these? Get-* cmdlet does not accept the piped object, but its output is piped to other cmdlets.

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleAlertingActionCommand.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+
+        #region Cmdlet parameters
+
+        [Parameter(Mandatory = false, HelpMessage = "The AzNS action details")]
+        [ValidateNotNullOrEmpty]
+        public PSScheduledQueryRuleAznsAction AznsAction { get; set; }
+
+        [Parameter(Mandatory = true, HelpMessage = "The severity for this alert")]
+        public string Severity { get; set; }
```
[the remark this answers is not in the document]
shpimpal: Yup, will add them.

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleAlertingActionCommand.cs
shpimpal: Removed nullable.

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleCommand.cs
shpimpal: will add it.

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleCommand.cs
shpimpal: will take care.

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleLogMetricTriggerCommand.cs
shpimpal: will add `PSArgumentCompleter`. But we do not support values other than the ones in the set.

src/Monitor/Monitor/ScheduledQueryRules/RemoveScheduledQueryRuleCommand.cs
shpimpal: Do the Name and ResourceGroup parameters not need to have these attributes ?

src/Monitor/Monitor/ScheduledQueryRules/RemoveScheduledQueryRuleCommand.cs
shpimpal: will take care.

src/Monitor/Monitor/ScheduledQueryRules/SetScheduledQueryRuleCommand.cs
shpimpal: sure.

src/Monitor/Monitor/ScheduledQueryRules/SetScheduledQueryRuleCommand.cs
shpimpal: Had to add the check as the parameter was not being binded when invoked from mock tests.

src/Monitor/Monitor/ScheduledQueryRules/UpdateScheduledQueryRuleCommand.cs
shpimpal: The API supports patch operation but only one field can be updated with the patch call, to update that parameter added `Update-*` cmdlet. So as per documentation, `New-*` cmdlet will create a new resource and `Set-*` can be used to update any of the fields of existing resource.

src/Monitor/Monitor/ScheduledQueryRules/UpdateScheduledQueryRuleCommand.cs
shpimpal: Sure.

src/Monitor/Monitor/ScheduledQueryRules/SetScheduledQueryRuleCommand.cs
shpimpal: Had to add the check as the parameter was not being binded when invoked from mock tests.

src/Monitor/Monitor/ScheduledQueryRules/UpdateScheduledQueryRuleCommand.cs
shpimpal: Had to add the check as the parameter was not being binded when invoked from mock tests.

src/Monitor/Monitor/help/Get-AzScheduledQueryRule.md
shpimpal: Sorry, I didn't understand. There are examples for each parameter set.

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleCommand.cs
shpimpal: It's directly passed as a string to the parent resource class. It is a mandatory parameter so there is no default value.

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleAlertingActionCommand.cs
shpimpal: Other parameters meaning ? The name was kept the same as what is in the API swagger.

src/Monitor/Monitor.Test/ScenarioTests/ScheduledQueryRulesTests.ps1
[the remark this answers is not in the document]
cormacpayne: @shpimpal it looks like you're still calling `Write-Object` in a few places in your tests

src/Monitor/Monitor/ScheduledQueryRules/GetScheduledQueryRuleCommand.cs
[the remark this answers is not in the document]
cormacpayne: @shpimpal you can find our full guide on piping scenarios [here](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/design-guidelines/piping-best-practices.md), but in general there are two scenarios we want to support:

(1) Piping by object
    - The object returned from the `Get-*` cmdlet (and any other cmdlet that returns the same object type) can be piped to `Set-*`/`Update-*` and `Remove-*`

(2) Piping by resource id
    - An object with a `ResourceId` property (such as that returned from `Get-AzResource`) can be piped to `Get-*`, `Set-*`/``Update-*` and `Remove-*`

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleAlertingActionCommand.cs
cormacpayne: @jdhuntington sorry, I meant PR, not parameter. There are other parameters in this PR that have "InMinutes" suffix, but this one is "InMin". PowerShell doesn't need to map 1-to-1 when it comes to the REST API spec properties, so we should be able to do `ThrottlingInMinutes` here.

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleSourceCommand.cs
cormacpayne: @shpimpal ping on this comment

src/Monitor/Monitor/ScheduledQueryRules/RemoveScheduledQueryRuleCommand.cs
cormacpayne: @shpimpal no, `-Name` and `-ResourceGroupName` won't have this attribute. You can take a look at my previous comment above with information about our piping scenarios, but `-ResourceId` should be the only parameter with the `ValueFromPipelineByPropertyName` attribute, and `-InputObject` should be the only parameter with the `ValueFromPipeline` attribute

src/Monitor/Monitor/ScheduledQueryRules/SetScheduledQueryRuleCommand.cs
cormacpayne: @shpimpal same comment here

src/Monitor/Monitor/ScheduledQueryRules/SetScheduledQueryRuleCommand.cs
cormacpayne: @shpimpal same comment here about changing this to a `bool` and removing the attributes

src/Monitor/Monitor/ScheduledQueryRules/UpdateScheduledQueryRuleCommand.cs
cormacpayne: @shpimpal would you mind explaining the `PATCH` API operation? You're saying that you can only update one property at a time in your `Update` SDK call?

src/Monitor/Monitor/ScheduledQueryRules/UpdateScheduledQueryRuleCommand.cs
cormacpayne: @shpimpal same comment here

src/Monitor/Monitor/ScheduledQueryRules/UpdateScheduledQueryRuleCommand.cs
cormacpayne: @shpimpal same comment here about changing this to a `bool` and removing the attributes

src/Monitor/Monitor/help/Get-AzScheduledQueryRule.md
cormacpayne: @shpimpal rather than putting an example for each parameter set all into one example, you'll create separate examples in separate headers for each parameter set. It would look like the following:

```
## EXAMPLES

### Example 1 - List by resource group

```powershell
PS C:\> Get-AzScheduledQueryRule -ResourceGroupName "Rac46PostSwapRG"
    ```

### Example 2 - Get by name

```powershell
PS C:\> Get-AzScheduledQueryRule -ResourceGroupName "Rac46PostSwapRG" -Name "logalertfoo"
    ```

### Example 3 - Get by resource id

```powershell
PS C:\> Get-AzScheduledQueryRule -ResourceId "/subscriptions/b67f7fec-69fc-4974-9099-a26bd6ffeda3/resourceGroups/Rac46PostSwapRG/providers/microsoft.insights/scheduledQueryRules/logalertfoo"
    ```
```

_Note_: the set of three backticks (` ``` `) at the end of every code block have an indentation to them for this comment -- there should be no indentation on the line with those backticks.

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleCommand.cs
shpimpal: I tried changing it to bool, the cmdlet accepts only 0 or 1 (or any number) as a value. If I provide true or false, it takes it as a string and fails as the argument is of type bool.

src/Monitor/Monitor.Test/ScenarioTests/ScheduledQueryRulesTests.ps1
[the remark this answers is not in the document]
shpimpal: Do you mean `Write-Output` ? Should everything be `Write-Debug` ?

src/Monitor/Monitor/ScheduledQueryRules/UpdateScheduledQueryRuleCommand.cs
shpimpal: Yes. I meant that `PATCH` call to our SQR API accepts only one field of the resource that can be updated which is "Enabled". We cannot update other fields using `PATCH` call, we will have to use `Set-*` cmdlet to update any field of an existing resource which uses `PUT` semantics.

maddieclayton:
@shpimpal Ping on above comments.

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleAlertingActionCommand.cs
shpimpal: done

src/Monitor/Monitor/ScheduledQueryRules/RemoveScheduledQueryRuleCommand.cs
shpimpal: Got it, updated the PR.

src/Monitor/Monitor/ScheduledQueryRules/SetScheduledQueryRuleCommand.cs
shpimpal: Updated the PR.

src/Monitor/Monitor/ScheduledQueryRules/UpdateScheduledQueryRuleCommand.cs
shpimpal: Updated the PR.

src/Monitor/Monitor/help/Get-AzScheduledQueryRule.md
shpimpal: Done.

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleAlertingActionCommand.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+
+        #region Cmdlet parameters
+
+        [Parameter(Mandatory = false, HelpMessage = "The AzNS action details")]
+        [ValidateNotNullOrEmpty]
+        public PSScheduledQueryRuleAznsAction AznsAction { get; set; }
+
+        [Parameter(Mandatory = true, HelpMessage = "The severity for this alert")]
+        [ValidateSet("0", "1", "2", "3", "4")]
```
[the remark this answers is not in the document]
shpimpal: Updated the PR.

shpimpal:
> @shpimpal Ping on above comments.
Hi @maddieclayton, I pushed new changes and answered the comments above.

src/Monitor/Monitor.Test/ScenarioTests/ScheduledQueryRulesTests.ps1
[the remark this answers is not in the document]
cormacpayne: @shpimpal yes, sorry -- in your tests, there should only be calls to `Write-Debug`, not `Write-Output`, `Write-Object` or `Write-Verbose`

src/Monitor/Monitor/ScheduledQueryRules/SetScheduledQueryRuleCommand.cs
```diff
@@ -0,0 +1,229 @@
...
+        [Parameter(ParameterSetName = ByResourceId, Mandatory = false, HelpMessage = "Resource tags")]
+        public IDictionary<string, string> Tag { get; set; }
+
+        //
+        // Summary:
+        //     Alert status - enabled or not, supported values - "true", "false"
+        [Parameter(ParameterSetName = ByRuleName, Mandatory = false, HelpMessage = "The azure alert state - valid values - true, false")]
+        [Parameter(ParameterSetName = ByInputObject, Mandatory = false, HelpMessage = "The azure alert state - valid values - true, false")]
+        [Parameter(ParameterSetName = ByResourceId, Mandatory = false, HelpMessage = "The azure alert state - valid values - true, false")]
+        [PSArgumentCompleter("true", "false")]
```
cormacpayne: @shpimpal you can remove this `PSArgumentCompleter` attribute

src/Monitor/Monitor/ScheduledQueryRules/UpdateScheduledQueryRuleCommand.cs
cormacpayne: @shpimpal in my opinion, since the only thing that you can change with this cmdlet is if the rule is enabled or not, it would be best to split it into two cmdlets: `Enable-AzScheduledQueryRule` and `Disable-AzScheduledQueryRule`

With this, users won't get confused by the fact there are two cmdlets with similar "updating a rule" experience, and then you have two cmdlets that allow the user to enable and disable a rule.

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleTriggerConditionCommand.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// </summary>
+    [Cmdlet(VerbsCommon.New, ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "ScheduledQueryRuleTriggerCondition"), OutputType(typeof(PSScheduledQueryRuleTriggerCondition))]
+    public class NewScheduledQueryRuleTriggerConditionCommand : MonitorCmdletBase
+    {
+
+        #region Cmdlet parameters
+
+        [Parameter(Mandatory = true, HelpMessage = "The threshold operator : GreaterThan, LessThan or Equal")]
+        [ValidateNotNullOrEmpty]
+        [ValidateSet("GreaterThan", "LessThan", "Equal")]
```
cormacpayne: @shpimpal this should be a `ValidateSet`

src/Monitor/Monitor/help/Get-AzScheduledQueryRule.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+## DESCRIPTION
+Gets Scheduled Query Resources
+
+## EXAMPLES
+
+### Example 1 - List by subscription or resource group
+```powershell
+PS C:\> Get-AzScheduledQueryRule -ResourceGroupName "Rac46PostSwapRG"
```
cormacpayne: @shpimpal for your examples, you should have the corresponding output from the cmdlet being executed in the code block so users know what the command will return

cormacpayne:
@shpimpal the following tests are failing:

- `TestPipingRemoveSetUpdateScheduledQueryRule`
- `TestNewGetUpdateSetRemoveScheduledQueryRule`

From the errors, it looks like the recordings may be out of date, so my suggestion would be to re-record these tests and push the new `.json` files to this PR and see if the failures are resolved.

src/Monitor/Monitor/ScheduledQueryRules/NewScheduledQueryRuleSourceCommand.cs
shpimpal: addressed

src/Monitor/Monitor.Test/ScenarioTests/ScheduledQueryRulesTests.ps1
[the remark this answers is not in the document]
shpimpal: Done

cormacpayne:
Spoke offline with @shpimpal -- merge conflicts will be resolved and new recordings for failing test(s) will be pushed as well.

tools/StaticAnalysis/Exceptions/Az.Monitor/SignatureIssues.csv
```diff
@@ -1,3 +1,9 @@
-"AssemblyFileName","ClassName","Target","Severity","ProblemId","Description","Remediation"
-"Microsoft.Azure.PowerShell.Cmdlets.Monitor.dll","Microsoft.Azure.Commands.Insights.Alerts.NewAzureRmMetricAlertRuleV2CriteriaCommand","New-AzMetricAlertRuleV2Criteria","1","8100","New-AzMetricAlertRuleV2Criteria Does not support ShouldProcess but the cmdlet verb New indicates that it should.","Determine if the cmdlet should implement ShouldProcess and if so determine if it should implement Force / ShouldContinue"
-"Microsoft.Azure.PowerShell.Cmdlets.Monitor.dll","Microsoft.Azure.Commands.Insights.Alerts.NewAzureRmMetricAlertRuleV2DimensionSelectionCommand","New-AzMetricAlertRuleV2DimensionSelection","1","8100","New-AzMetricAlertRuleV2DimensionSelection Does not support ShouldProcess but the cmdlet verb New indicates that it should.","Determine if the cmdlet should implement ShouldProcess and if so determine if it should implement Force / ShouldContinue"
+AssemblyFileName,ClassName,Target,Severity,ProblemId,Description,Remediation
```
cormacpayne: @shpimpal it looks like StaticAnalysis is failing because each column entry in this file isn't wrapped in quotes (see the deleted lines for how this should look) -- this happens when the contents are copied over from Excel and not done using a text editor, such as VS Code

src/Monitor/Monitor/help/Get-AzScheduledQueryRule.md
shpimpal: Done

tools/StaticAnalysis/Exceptions/Az.Monitor/SignatureIssues.csv
shpimpal: Corrected

src/Monitor/Monitor/Az.Monitor.psd1
```diff
@@ -133,7 +139,19 @@ PrivateData = @{
       - Remove-AzMetricAlertRuleV2
       - Get-AzMetricAlertRuleV2
       - Add-AzMetricAlertRuleV2
-  * Updated Monitor SDK to version 0.22.0-preview'
+		* Updated Monitor SDK to version 0.22.0-preview
```
cormacpayne: @shpimpal this change to the release notes can be reverted since we will update this section during our release

PR #8861, opened 2019-03-26
Update WAF Policy cmdlets to support newest API features

digavino:
Update WAF Policy cmdlets to support newest API features

## Checklist


adxsdkps:

vladimir-shcherbakov:
@azuresdkci Add to whitelist

digavino:
Hi, I added the following design review - https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/268

src/FrontDoor/FrontDoor/Cmdlets/SetAzureRmFrontDoorFireWallPolicy.cs
```diff
@@ -87,7 +84,23 @@ public class SetAzureRmFrontDoorFireWallPolicy : AzureFrontDoorCmdletBase
...
+        /// Redirect URL used for redirect actions
+        /// </summary>
+        [Parameter(Mandatory = false, HelpMessage = "Redirect URL")]
+        public string RedirectUrl { get; set; }
 
+        /// <summary>
+        /// Custom block response code used for block actions
+        /// </summary>
+        [Parameter(Mandatory = false, HelpMessage = "Custom Response Status Code")]
+        public ushort? CustomBlockResponseStatusCode { get; set; }
```
markcowl: Same comment as above - this should be an int with a ValidateRange attribute

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorManagedRuleObject.cs
```diff
@@ -31,15 +26,15 @@ namespace Microsoft.Azure.Commands.FrontDoor.Cmdlets
     public class NewAzureRmFrontDoorManagedRuleObject : AzureFrontDoorCmdletBase
     {
         /// <summary>
-        /// Describes priority of the rule
+        /// Type of the ruleset (e.g.: DefaultRuleSet)
         /// </summary>
-        [Parameter(Mandatory = true, HelpMessage = "Describes priority of the rule")]
-        public int Priority { get; set; }
+        [Parameter(Mandatory = true, HelpMessage = "Type of the ruleset")]
+        public string Type { get; set; }
```
markcowl: Is there a know set of types?  In this case, this should have a completer

src/FrontDoor/FrontDoor/help/Get-AzFrontDoorFireWallPolicy.md
```diff
@@ -22,55 +22,61 @@ The **Get-AzFrontDoorFireWallPolicy** cmdletGet gets WAF policy in a resource gr
...
-Etag               :
-ProvisioningState  : Succeeded
-Tags               :
-Id                 : /subscriptions/{subid}/resourcegroups/{resourceGroupName}/providers/Micr
-                     osoft.Network/frontdoorwebapplicationfirewallpolicies/{Name}
-Name               : {Name}
-Type               :
+PS C:\> Get-AzFrontDoorFireWallPolicy -Name $policyName -ResourceGroupName $resourceGroupName
+
+PolicyMode                    : Prevention
```
markcowl: Default display should be table format, see: https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/generate-format.ps1xml-file.md

src/FrontDoor/FrontDoor/help/Get-AzFrontDoorFireWallPolicy.md
```diff
@@ -22,55 +22,61 @@ The **Get-AzFrontDoorFireWallPolicy** cmdletGet gets WAF policy in a resource gr
...
-CustomRules        : {Rule1}
-ManagedRules       : {Microsoft.Azure.Commands.FrontDoor.Models.PSAzureManagedRule}
-Etag               :
-ProvisioningState  : Succeeded
-Tags               :
-Id                 : /subscriptions/{subid}/resourcegroups/{resourceGroupName}/providers/Micr
-                     osoft.Network/frontdoorwebapplicationfirewallpolicies/{Name2}
-Name               : {Name2}
-Type               :
+PS C:\> Get-AzFrontDoorFireWallPolicy -ResourceGroupName $resourceGroupName
```
markcowl: same comment - default format shoudl be table view

src/FrontDoor/FrontDoor/help/New-AzFrontDoorFireWallPolicy.md
```diff
@@ -25,27 +26,57 @@ The **New-AzFrontDoorFireWallPolicy** cmdlet creates a new Azure WAF policy in t
...
-Etag               :
-ProvisioningState  : Succeeded
-Tags               :
-Id                 : /subscriptions/{subid}/resourcegroups/{resourceGroupName}/providers/Micr
-                     osoft.Network/frontdoorwebapplicationfirewallpolicies/{Name}
-Name               : {Name}
-Type               :
+PS C:\> New-AzFrontDoorFireWallPolicy -Name $policyName -ResourceGroupName $resourceGroupName -Customrule $customRule1,$customRule2 -ManagedRule $managedRule1 -EnabledState Enabled -Mode Prevention -RedirectUrl "https://www.bing.com/" -CustomBlockResponseStatusCode 405 -CustomBlockResponseBody "<html><head><title>You are blocked!</title></head><body></body></html>"
+
+PolicyMode                    : Prevention
```
markcowl: Default format should be table

src/FrontDoor/FrontDoor/help/Set-AzFrontDoorFireWallPolicy.md
```diff
@@ -8,102 +8,141 @@ schema: 2.0.0
...
-CustomRules        : {Rule1}
-ManagedRules       : {Microsoft.Azure.Commands.FrontDoor.Models.PSAzureManagedRule}
-Etag               :
-ProvisioningState  : Succeeded
-Tags               :
-Id                 : /subscriptions/{subid}/resourcegroups/{resourceGroupName}/providers/Micr
-                     osoft.Network/frontdoorwebapplicationfirewallpolicies/{Name}
-Name               : {Name}
-Type               :
+PS C:\> Set-AzFrontDoorFireWallPolicy -Name $policyName -ResourceGroupName $resourceGroupName -CustomBlockResponseStatusCode 403
```
markcowl: default format == table

src/FrontDoor/FrontDoor/help/Set-AzFrontDoorFireWallPolicy.md
```diff
@@ -8,102 +8,141 @@ schema: 2.0.0
...
-CustomRules        : {Rule1}
-ManagedRules       : {Microsoft.Azure.Commands.FrontDoor.Models.PSAzureManagedRule}
-Etag               :
-ProvisioningState  : Succeeded
-Tags               :
-Id                 : /subscriptions/{subid}/resourcegroups/{resourceGroupName}/providers/Micr
-                     osoft.Network/frontdoorwebapplicationfirewallpolicies/{Name}
-Name               : {Name}
-Type               :
+PS C:\> Set-AzFrontDoorFireWallPolicy -Name $policyName -ResourceGroupName $resourceGroupName -Mode Detection
```
markcowl: default format == table

src/FrontDoor/FrontDoor/help/Set-AzFrontDoorFireWallPolicy.md
```diff
@@ -8,102 +8,141 @@ schema: 2.0.0
...
-CustomRules        : {Rule1}
-ManagedRules       : {Microsoft.Azure.Commands.FrontDoor.Models.PSAzureManagedRule}
-Etag               :
-ProvisioningState  : Succeeded
-Tags               :
-Id                 : /subscriptions/{subid}/resourcegroups/{resourceGroupName}/providers/Micr
-                     osoft.Network/frontdoorwebapplicationfirewallpolicies/{Name}
-Name               : {Name}
-Type               :
+PS C:\> Set-AzFrontDoorFireWallPolicy -Name $policyName -ResourceGroupName $resourceGroupName -Mode Detection -EnabledState Disabled
```
markcowl: same comment

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorFireWallPolicy.cs
```diff
@@ -72,7 +73,23 @@ public class NewAzureRmFrontDoorFireWallPolicy : AzureFrontDoorCmdletBase
...
+        /// Redirect URL used for redirect actions
+        /// </summary>
+        [Parameter(Mandatory = false, HelpMessage = "Redirect URL")]
+        public string RedirectUrl { get; set; }
+
+        /// <summary>
+        /// Custom block response code used for block actions
+        /// </summary>
+        [Parameter(Mandatory = false, HelpMessage = "Custom Response Status Code")]
+        public ushort? CustomBlockResponseStatusCode { get; set; }
```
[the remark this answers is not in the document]
digavino: Will be fixed in next iteration.

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorManagedRuleOverrideObject.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// Defines the New-AzFrontDoorManagedRuleOverrideObject cmdlet.
+    /// </summary>
+    [Cmdlet("New", ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "FrontDoorManagedRuleOverrideObject"), OutputType(typeof(PSAzureManagedRuleOverride))]
+    public class NewAzureRmFrontDoorManagedRuleOverrideObject : AzureFrontDoorCmdletBase
+    {
+        /// <summary>
+        /// Rule ID
+        /// </summary>
+        [Parameter(Mandatory = true, HelpMessage = "Rule ID")]
+        public string RuleId { get; set; }
```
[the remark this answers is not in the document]
digavino: It's an ID from a long list of options that we generate (and we may add things in the future so we don't want it to be fixed).

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorManagedRuleOverrideObject.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// Override Action
+        /// </summary>
+        [Parameter(Mandatory = false, HelpMessage = "Override Action")]
+        public PSAction? Action { get; set; }
+
+        /// <summary>
+        /// Enabled State
+        /// </summary>
+        [Parameter(Mandatory = false, HelpMessage = "Enabled State")]
+        public PSEnabledState? EnabledState { get; set; }
```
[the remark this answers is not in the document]
digavino: Fixed in the next iteration.

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorManagedRuleOverrideObject.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// Rule ID
+        /// </summary>
+        [Parameter(Mandatory = true, HelpMessage = "Rule ID")]
+        public string RuleId { get; set; }
+
+        /// <summary>
+        /// Override Action
+        /// </summary>
+        [Parameter(Mandatory = false, HelpMessage = "Override Action")]
+        public PSAction? Action { get; set; }
```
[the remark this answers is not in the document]
digavino: I could do a normal enum I guess since it's not a mandatory parameter (then I can check if it was passed in).

src/FrontDoor/FrontDoor/Cmdlets/SetAzureRmFrontDoorFireWallPolicy.cs
digavino: Fixed in next iteration.

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorManagedRuleObject.cs
digavino: There is but we might add some later and we would not like to change the cmdlets whenever we add another.

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorFireWallPolicy.cs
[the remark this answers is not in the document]
markcowl: No, this would be a breakign chane later.  Please fix now.

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorManagedRuleOverrideObject.cs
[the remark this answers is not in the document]
markcowl: Correct, that is the suggested pattern - nullable value types are difficult to reflect over in PowerShell and are not needed (use the IsParameterBound extension to easily check for the parameter's presence or absence)

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorManagedRuleOverrideObject.cs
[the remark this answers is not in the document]
markcowl: What is the advantage in requiring the user to provide this? Please make this optional and generate a value if the user doesn't provide one.

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorManagedRuleOverrideObject.cs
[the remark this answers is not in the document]
markcowl: Again, this would be a  breaking change in the next iteration, please fix.

src/FrontDoor/FrontDoor/Cmdlets/SetAzureRmFrontDoorFireWallPolicy.cs
markcowl: Again, this is a breaking change in the next iteration.  Please fix.

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorManagedRuleOverrideObject.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// Rule ID
+        /// </summary>
+        [Parameter(Mandatory = true, HelpMessage = "Rule ID")]
+        public string RuleId { get; set; }
+
+        /// <summary>
+        /// Override Action
+        /// </summary>
+        [Parameter(Mandatory = false, HelpMessage = "Override Action")]
+        public PSAction Action { get; set; }
```
markcowl: Use a string with a completer, this will future-proof the API from future breaking changes and still allow the user to select from appropriate values.

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorFireWallPolicy.cs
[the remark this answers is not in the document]
digavino: By iteration, I meant in this code review... I already fixed this (made an int like you suggested).

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorManagedRuleOverrideObject.cs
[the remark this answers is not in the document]
digavino: I already changed this in my latest commit here.

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorManagedRuleOverrideObject.cs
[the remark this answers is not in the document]
digavino: I cannot generate a value if the user doesn't provide this... There is a huge list of IDs (that may change later) and a random string would not be correct here. Also, the IDs depend on the rule group (so an ID valid for a group is not going to be valid for another). The user needs to provide this.

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorManagedRuleOverrideObject.cs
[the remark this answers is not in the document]
digavino: By iteration, I meant in this code review... I already fixed this (made an int like you suggested).

src/FrontDoor/FrontDoor/Cmdlets/SetAzureRmFrontDoorFireWallPolicy.cs
digavino: By iteration, I meant in this code review... I already fixed this (made an int like you suggested).

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorManagedRuleOverrideObject.cs
digavino: Why would you prefer a string other than an Enum ? If I add anything to the Enum, then the next version will pick it up. If I was deleting anything from the Enum, that would be a breaking change regardless.

src/FrontDoor/FrontDoor/help/Get-AzFrontDoorFireWallPolicy.md
digavino: Changed help to show a table.

src/FrontDoor/FrontDoor/Cmdlets/SetAzureRmFrontDoorFireWallPolicy.cs
digavino: Also, just to be clear about something... I would have rather leave the ushort? because that allowed to set it back to $null while doing an integer with a range does not allow that. It's not a big deal since the customer could always type the default status code (but it would have been nice to have a nullable int with a range). So, there is a valid reason to sometimes use nullable parameters in arguments (when you want the user to be able to unset an optional property).

digavino: Reopening this conversation since I think this should really be an optional null-able int anyway. I don't still understand why you are against null-able parameters when it seems to be the only way to unset an optional property.

markcowl: If you are not replacing all values with the values the user provides, then the verb for this cmdlet shoudl be Update and not Set.  Set indicates that the values you are providing are replacing the value on the service.  For Patch-like semantics, you should use Update.

If you are replacing, then absence of the bound parameter indicates that the value is cleared.

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorManagedRuleOverrideObject.cs
markcowl: This takes some of the urgency from releasing and installing a new version every time the enum values are updated - users would be able to use new values without you havign to release a new version fo the cmdlets (or them havign to install a new version).

digavino: Not really, because I still need to create an object of type PSAzureManagedRuleOverride (which is then used in other objects). Then the model gets converted (and eventually a string is actually sent).

(line 348 of ModelExtensions.cs)
Action = ruleOverride.Action.HasValue ? ruleOverride.Action.Value.ToString() : null,

So, I understand what you are proposing but it would requiring touching pretty much every file and I still believe the user would have to install a new version if we ever add a value (since the API version is still hardcoded). If I want this to be a string, then the model has to be a string, then the model extension has to change, then I also have to change the parameters in all other cmdlets that have been using Enums (to be consistent). At the end, if we ever add a value, the user might have to install a new package version anyway (because of the hard coded version).

Also, this particular Enum (PSAction) is likely not going to change for a while (if ever) and any change would still require a new swagger with a new API version.

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorManagedRuleObject.cs
markcowl: You need to add a completer for the known set of types - this is essentially the same as an expandable enum.  See: https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/azure-powershell-developer-guide.md#generic-argument-completer

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorManagedRuleOverrideObject.cs
markcowl: Is this modeled as a string in yoru API, or as an enum?  If this is modeled as a string in your SDK, then you should change this to string and update the types to match.

If it is modeled as an enum in the SDK, then thsi is OK for now, but in the next swagger iteration, this should be modeled as an expandable enum (modelAsString: true)

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorCustomRuleObject.cs
```diff
@@ -76,12 +76,6 @@ public class NewAzureRmFrontDoorCustomRuleObject : AzureFrontDoorCmdletBase
         [Parameter(Mandatory = false, HelpMessage = "Rate limit thresold")]
         public int? RateLimitThreshold { get; set; }
 
-        /// <summary>
```
markcowl: Note: please also add a change description to the changelog - this is the trigger for updating the version and releasing the module

src/FrontDoor/FrontDoor/Cmdlets/NewAzureRmFrontDoorManagedRuleOverrideObject.cs
digavino: It's an Enum in the model right now.

PR #8874, opened 2019-03-27
[Storage] Support Blob Service Properties in SRP

blueww:
## Description

This PR is to support manage Storage Blob Service Properties in SRP mode.
It target for 4/6 release.

The design already pass review in https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/263 

## Checklist

### DIFF
--- a/src/Storage/Storage.Management/Blob/DisableAzureStorageBlobDeleteRetentionPolicy.cs
+++ b/src/Storage/Storage.Management/Blob/DisableAzureStorageBlobDeleteRetentionPolicy.cs
@@ -60,7 +60,7 @@ namespace Microsoft.Azure.Commands.Management.Storage
             HelpMessage = "Storage Account Name.",
            ParameterSetName = AccountNameParameterSet)]
         [ResourceNameCompleter("Microsoft.Storage/storageAccounts", nameof(ResourceGroupName))]
-        [Alias(AccountNameAlias)]
+        [Alias(AccountNameAlias, NameAlias)]
         [ValidateNotNullOrEmpty]
         public string StorageAccountName { get; set; }
 
--- a/src/Storage/Storage.Management/Blob/EnableAzureStorageBlobDeleteRetentionPolicy.cs
+++ b/src/Storage/Storage.Management/Blob/EnableAzureStorageBlobDeleteRetentionPolicy.cs
@@ -60,7 +60,7 @@ namespace Microsoft.Azure.Commands.Management.Storage
             HelpMessage = "Storage Account Name.",
            ParameterSetName = AccountNameParameterSet)]
         [ResourceNameCompleter("Microsoft.Storage/storageAccounts", nameof(ResourceGroupName))]
-        [Alias(AccountNameAlias)]
+        [Alias(AccountNameAlias, NameAlias)]
         [ValidateNotNullOrEmpty]
         public string StorageAccountName { get; set; }
 
--- a/src/Storage/Storage.Management/Blob/GetAzureStorageBlobServiceProperties.cs
+++ b/src/Storage/Storage.Management/Blob/GetAzureStorageBlobServiceProperties.cs
@@ -60,7 +60,7 @@ namespace Microsoft.Azure.Commands.Management.Storage
             HelpMessage = "Storage Account Name.",
            ParameterSetName = AccountNameParameterSet)]
         [ResourceNameCompleter("Microsoft.Storage/storageAccounts", nameof(ResourceGroupName))]
-        [Alias(AccountNameAlias)]
+        [Alias(AccountNameAlias, NameAlias)]
         [ValidateNotNullOrEmpty]
         public string StorageAccountName { get; set; }
 
--- a/src/Storage/Storage.Management/Blob/UpdateAzureStorageBlobServiceProperties.cs
+++ b/src/Storage/Storage.Management/Blob/UpdateAzureStorageBlobServiceProperties.cs
@@ -59,7 +59,7 @@ namespace Microsoft.Azure.Commands.Management.Storage
             HelpMessage = "Storage Account Name.",
            ParameterSetName = AccountNameParameterSet)]
         [ResourceNameCompleter("Microsoft.Storage/storageAccounts", nameof(ResourceGroupName))]
-        [Alias(AccountNameAlias)]
+        [Alias(AccountNameAlias, NameAlias)]
         [ValidateNotNullOrEmpty]
         public string StorageAccountName { get; set; }
 
--- a/src/Storage/Storage.Management/Blob/StorageBlobBaseCmdlet.cs
+++ b/src/Storage/Storage.Management/Blob/StorageBlobBaseCmdlet.cs
@@ -29,6 +29,7 @@ namespace Microsoft.Azure.Commands.Management.Storage
         private StorageManagementClientWrapper storageClientWrapper;
 
         protected const string AccountNameAlias = "AccountName";
+        protected const string NameAlias = "Name";
 
         protected const string StorageContainerNounStr = "StorageContainer";
         protected const string StorageContainerImmutabilityPolicyNounStr = StorageContainerNounStr + "ImmutabilityPolicy";
--- a/src/Storage/Storage.Management/help/Disable-AzStorageBlobDeleteRetentionPolicy.md
+++ b/src/Storage/Storage.Management/help/Disable-AzStorageBlobDeleteRetentionPolicy.md
@@ -129,7 +129,7 @@ Storage Account Name.
 ```yaml
 Type: System.String
 Parameter Sets: AccountName
-Aliases: AccountName
+Aliases: AccountName, Name
 
 Required: True
 Position: 1
--- a/src/Storage/Storage.Management/help/Enable-AzStorageBlobDeleteRetentionPolicy.md
+++ b/src/Storage/Storage.Management/help/Enable-AzStorageBlobDeleteRetentionPolicy.md
@@ -145,7 +145,7 @@ Storage Account Name.
 ```yaml
 Type: System.String
 Parameter Sets: AccountName
-Aliases: AccountName
+Aliases: AccountName, Name
 
 Required: True
 Position: 1
--- a/src/Storage/Storage.Management/help/Get-AzStorageBlobServiceProperty.md
+++ b/src/Storage/Storage.Management/help/Get-AzStorageBlobServiceProperty.md
@@ -114,7 +114,7 @@ Storage Account Name.
 ```yaml
 Type: System.String
 Parameter Sets: AccountName
-Aliases: AccountName
+Aliases: AccountName, Name
 
 Required: True
 Position: 1
--- a/src/Storage/Storage.Management/help/Update-AzStorageBlobServiceProperty.md
+++ b/src/Storage/Storage.Management/help/Update-AzStorageBlobServiceProperty.md
@@ -130,7 +130,7 @@ Storage Account Name.
 ```yaml
 Type: System.String
 Parameter Sets: AccountName
-Aliases: AccountName
+Aliases: AccountName, Name
 
 Required: True
 Position: 1


blueww:
@markcowl , @maddieclayton , @cormacpayne 
could you share the staticAnalysis error in build?
Besides that, can we show staticAnalysis error in build log, or upload the output to some public blob that can share?
This can make the build failure investigation much more efficiently.

blueww:
Add a static analysis exception for parameter name "RetentionDays" of cmdlet "Enable-AzStorageBlobDeleteRetentionPolicy".
This parameter name is aligned with the related parameter name of "Enable-AzStorageDeleteRetentionPolicy" https://github.com/Azure/azure-powershell/blob/3ef406acd59be744bcc345631b10f464a1d53bf9/src/Storage/Storage/Common/Cmdlet/EnableAzureStorageDeleteRetentionPolicy.cs#L32. The parameter name in this comlet has been reviewed and agreed with Mark before.

src/Storage/Storage.Management/Blob/DisableAzureStorageBlobDeleteRetentionPolicy.cs
```diff
@@ -0,0 +1,122 @@
...
+        [Alias(AccountNameAlias)]
+        [ValidateNotNullOrEmpty]
+        public string StorageAccountName { get; set; }
+
+        [Parameter(Mandatory = true,
+            HelpMessage = "Storage account object",
+            ValueFromPipeline = true,
+            ParameterSetName = AccountObjectParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public PSStorageAccount StorageAccount { get; set; }
```
markcowl: If you have this parameter set (that is a parameter set that takes PSStorageAccount), you should not have ValueFromPipelineByProeprtyName for the other properties.

src/Storage/Storage.Management/Blob/DisableAzureStorageBlobDeleteRetentionPolicy.cs
```diff
@@ -0,0 +1,122 @@
...
+        [ValidateNotNullOrEmpty]
+        public string ResourceGroupName { get; set; }
+
+        [Parameter(
+            Position = 1,
+            Mandatory = true,
+            HelpMessage = "Storage Account Name.",
+           ParameterSetName = AccountNameParameterSet)]
+        [ResourceNameCompleter("Microsoft.Storage/storageAccounts", nameof(ResourceGroupName))]
+        [Alias(AccountNameAlias)]
```
markcowl: Is this alias 'Name'?

src/Storage/Storage.Management/Blob/EnableAzureStorageBlobDeleteRetentionPolicy.cs
```diff
@@ -0,0 +1,126 @@
...
+        [Alias(AccountNameAlias)]
+        [ValidateNotNullOrEmpty]
+        public string StorageAccountName { get; set; }
+
+        [Parameter(Mandatory = true,
+            HelpMessage = "Storage account object",
+            ValueFromPipeline = true,
+            ParameterSetName = AccountObjectParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public PSStorageAccount StorageAccount { get; set; }
```
markcowl: same comment - this parameter should take values from the pipeline, and ther should be no parameters that take values from pieline by property name.

src/Storage/Storage.Management/Blob/GetAzureStorageBlobServiceProperties.cs
```diff
@@ -0,0 +1,109 @@
...
+        [Alias(AccountNameAlias)]
+        [ValidateNotNullOrEmpty]
+        public string StorageAccountName { get; set; }
+
+        [Parameter(Mandatory = true,
+            HelpMessage = "Storage account object",
+            ValueFromPipeline = true,
+            ParameterSetName = AccountObjectParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public PSStorageAccount StorageAccount { get; set; }
```
markcowl: Same comment - there shoudl be no parameters with 'ValueFromPiepelineByProeprtyName'

src/Storage/Storage.Management/Blob/GetAzureStorageBlobServiceProperties.cs
```diff
@@ -0,0 +1,109 @@
...
+            Position = 0,
+            Mandatory = true,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Blob Service Properties Resource Id.",
+           ParameterSetName = PropertiesResourceIdParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public string ResourceId { get; set; }
+
+        [Parameter(Mandatory = false, HelpMessage = "Display ServiceProperties")]
+        public SwitchParameter PassThru { get; set; }
```
markcowl: Why does this have a passthru parameter?  The Get should always return the BlobServiceProperties.

src/Storage/Storage.Management/Blob/GetAzureStorageBlobServiceProperties.cs
```diff
@@ -0,0 +1,109 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Management.Automation;
+    using Microsoft.Azure.Commands.ResourceManager.Common.ArgumentCompleters;
+    using Microsoft.Azure.Management.Internal.Resources.Utilities.Models;
+
+    /// <summary>
+    /// Modify Azure Storage service properties
+    /// </summary>
+    [Cmdlet("Get", ResourceManager.Common.AzureRMConstants.AzureRMPrefix + StorageBlobServiceProperty, DefaultParameterSetName = AccountNameParameterSet), OutputType(typeof(PSBlobServiceProperties))]
```
markcowl: Shouldn't thsi just be 'Get-AzStorageBlobService'?

src/Storage/Storage.Management/Blob/UpdateAzureStorageBlobServiceProperties.cs
```diff
@@ -0,0 +1,148 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Management.Automation;
+    using Microsoft.Azure.Commands.ResourceManager.Common.ArgumentCompleters;
+    using Microsoft.Azure.Management.Internal.Resources.Utilities.Models;
+
+    /// <summary>
+    /// Modify Azure Storage service properties
+    /// </summary>
+    [Cmdlet("Update", ResourceManager.Common.AzureRMConstants.AzureRMPrefix + StorageBlobServiceProperty, SupportsShouldProcess = true, DefaultParameterSetName = AccountNameParameterSet), OutputType(typeof(PSBlobServiceProperties))]
```
markcowl: Same comment, shouldn't this be Update-AzStorageBlobService?

src/Storage/Storage.Management/Blob/UpdateAzureStorageBlobServiceProperties.cs
```diff
@@ -0,0 +1,148 @@
...
+        [Alias(AccountNameAlias)]
+        [ValidateNotNullOrEmpty]
+        public string StorageAccountName { get; set; }
+
+        [Parameter(Mandatory = true,
+            HelpMessage = "Storage account object",
+            ValueFromPipeline = true,
+            ParameterSetName = AccountObjectParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public PSStorageAccount StorageAccount { get; set; }
```
markcowl: Same comment - this is the only parameter that should take its value from the pipeline

src/Storage/Storage.Management/Blob/UpdateAzureStorageBlobServiceProperties.cs
```diff
@@ -0,0 +1,148 @@
...
+        [ValidateNotNullOrEmpty]
+        public string ResourceId { get; set; }
+
+        [Parameter(Mandatory = false, HelpMessage = "Default Service Version to Set")]
+        [ValidateNotNull]
+        public string DefaultServiceVersion { get; set; }
+
+        [Parameter(Mandatory = true, HelpMessage = "Blob service properties object to Set", ParameterSetName = AccountNamePropertiesObjectParameterSet, ValueFromPipeline = true)]
+        [ValidateNotNullOrEmpty]
+        public PSBlobServiceProperties BlobServiceProperty { get; set; }
```
markcowl: Shouldn't this be the parameter that takes its value from the pipeline?  In that case, the verb for this cmdlet should eb 'Set'.  'Update' only makes sense if you are using PATCH semantcis, meaning that you can set one blob service property and leave others unchanged.

src/Storage/Storage.Management/Blob/DisableAzureStorageBlobDeleteRetentionPolicy.cs
blueww: the alias is "AccountName"
`protected const string AccountNameAlias = "AccountName";`

src/Storage/Storage.Management/Blob/GetAzureStorageBlobServiceProperties.cs
blueww: Removed. Thanks for catching it!

src/Storage/Storage.Management/Blob/GetAzureStorageBlobServiceProperties.cs
blueww: Blob Service support Container, blob, blob service priperties. Add properties might be more specific, and this is also aligned with dataplane cmdlet name: Get-AzureStorageServiceProperties

src/Storage/Storage.Management/Blob/UpdateAzureStorageBlobServiceProperties.cs
blueww: Blob Service support Container, blob, blob service priperties. Add "properties" might be more specific, to let user know the blob/container won't updated. only service properties are updated.
And this is also aligned with dataplane cmdlet name: Update-AzureStorageServiceProperties

src/Storage/Storage.Management/Blob/UpdateAzureStorageBlobServiceProperties.cs
blueww: We can set one blob service property and leave others unchanged, so "Update" might be more proper. 
Actually, this is also aligned with dataplane cmdlet name: Update-AzureStorageServiceProperties. And "Update" verb is suggest by you when design the dataplane cmdlet

blueww: As we talked, I have removed parameter "PSBlobServiceProperties BlobServiceProperty" from this cmdlet. We might can add a Set-* cmdlet for this in the future, but not in this release.

src/Storage/Storage.Management/Blob/DisableAzureStorageBlobDeleteRetentionPolicy.cs
blueww: As we discussed, the only parameter has ValueFromPipelineByProeprtyName os "-ResourceId" and this is acceptable. 
As we discussed, I have change the parameter name in all 4 new cmdlets to "ResourceId", and all will accept both Storage account resourceId and Blob Service ResourceId.

src/Storage/Storage.Management/Blob/EnableAzureStorageBlobDeleteRetentionPolicy.cs
blueww: As we discussed, the only parameter has ValueFromPipelineByProeprtyName os "-ResourceId" and this is acceptable. 
As we discussed, I have change the parameter name in all 4 new cmdlets to "ResourceId", and all will accept both Storage account resourceId and Blob Service ResourceId.

src/Storage/Storage.Management/Blob/GetAzureStorageBlobServiceProperties.cs
blueww: As we discussed, the only parameter has ValueFromPipelineByProeprtyName os "-ResourceId" and this is acceptable. 
As we discussed, I have change the parameter name in all 4 new cmdlets to "ResourceId", and all will accept both Storage account resourceId and Blob Service ResourceId.

src/Storage/Storage.Management/Blob/GetAzureStorageBlobServiceProperties.cs
blueww: As we discussed ,will still use Get-AzureStorageServiceProperty

src/Storage/Storage.Management/Blob/UpdateAzureStorageBlobServiceProperties.cs
blueww: As we discussed ,will still use Update-AzureStorageServiceProperty

src/Storage/Storage.Management/Blob/UpdateAzureStorageBlobServiceProperties.cs
blueww: As we discussed, the only parameter has ValueFromPipelineByProeprtyName os "-ResourceId" and this is acceptable.
As we discussed, I have change the parameter name in all 4 new cmdlets to "ResourceId", and all will accept both Storage account resourceId and Blob Service ResourceId.

blueww:
@markcowl 
I have updated the PR per our discusstion:
1. Make the parameter name to -ResouceId, and support both Storage account Id and Blob Service properties Id
2. Remove "PSBlobServiceProperties BlobServiceProperty" parameter from Update-AzureStorageBlobServiceProperty. so we can keep the name as Update-*
3. remove -passthru  from Get-* cmdlet
4. other changelog and help related change.

Please help to check and merge it if no other issues.
Thanks!

markcowl:
@blueww One remaining issue, otherwise lgtm

src/Storage/Storage.Management/Blob/DisableAzureStorageBlobDeleteRetentionPolicy.cs
blueww: Fixed by add alias "Name".
See commit https://github.com/Azure/azure-powershell/pull/8874/commits/8749ddd484c4ae299796c4019ee0843c90249192

blueww:
@markcowl 
I have fixed the last comment from you, in commit https://github.com/Azure/azure-powershell/pull/8874/commits/8749ddd484c4ae299796c4019ee0843c90249192
Please help to check and merge the PR.

blueww:
@markcowl 
It seems the build failed, but not caused by my code change:
> The command "sudo apt-get update" failed and exited with 100 during .
Would you please help to fix the build? 
Is there any easy way to rerun build of the PR?

PR #8905, opened 2019-03-31
AsJob support for storage upload and download cmdlets

markcowl:
## Checklist

### DIFF
--- a/src/Storage/Storage.Management/ChangeLog.md
+++ b/src/Storage/Storage.Management/ChangeLog.md
@@ -25,6 +25,11 @@
     - Get-AzStorageBlobServiceProperty
     - Enable-AzStorageBlobDeleteRetentionPolicy
     - Disable-AzStorageBlobDeleteRetentionPolicy
+* -AsJob support for Blob and file upload and download cmdlets
+    - Get-AzStorageBlobContent
+    - Set-AzStorageBlobContent
+    - Get-AzStorageFileContent
+    - Set-AzStorageFileContent
 
 ## Version 1.1.0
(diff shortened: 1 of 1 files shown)


blueww:
@markcowl 
It seems upload blob without -asjob will hang with the code.
I just test upload blob with the code:. with asjob, it looks like works fine. But without -asjob, it hangs.

src/Storage/Storage/Blob/Cmdlet/GetAzureStorageBlobContent.cs
```diff
@@ -280,12 +282,33 @@ internal string GetFullReceiveFilePath(string fileName, string blobName, DateTim
...
+                InitChannelCurrentSubscription();
+                this.ExecuteSynchronouslyOrAsJob();
+            }
+            catch (Exception ex) when (!IsTerminatingError(ex))
+            {
+                WriteExceptionError(ex);
+            }
+        }
+
+
```
blueww: Following code should be added here, or the DoBeginProcessing() will be run 2 times when use -asjob parameter. (once in BeginProcessing(), once in  ExecuteCmdlet())

I have tested before/after adding the code, the cmdlets still runs successfully with/without -asjob.

```
        protected override void BeginProcessing()
        {
            if (!AsJob.IsPresent)
            {
                DoBeginProcessing();
            }
        }
```

markcowl: this is intentional - DoBeginProcessing is idempotent in this case

PR #8923, opened 2019-04-02
3842798: Automatic DNS Nameserver delegation in parent when child zon…

ShwethaRaghupathi:
…e created

## Description

- Enable DNS name server delegation when a DNS zone is created whose parent zone is already hosted in Azure.
- For successful resolution of the newly added child zone, an NS record must exist in the parent containing the name servers of the newly created child zone.
- Achieve this additional step as part of zone creation when user indicates the parent zone details as part of -ParentZoneName parameter

**Current state**

It supports parent zone present in same resource group & subscription as the child zone being created.

**Intended future state**

Since the parent zone might reside in another subscription, we propose to accept FQDN (fully qualified domain name) of the parent and extract resource group , subscription details

**Open Question ?**
Would it be possible to create two resources in different subscriptions as part of this single cmd ?
If so how ?

## Checklist


adxsdkps:

ShwethaRaghupathi:
+Team
@SoSain

cormacpayne:
@ShwethaRaghupathi as mentioned in our offline conversation, you will need to fill out a [cmdlet design review](https://github.com/Azure/azure-powershell-cmdlet-review-pr) before these changes can be made

ShwethaRaghupathi:
@cormacpayne Design PR for this change - https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/274 . Please review the same.

maddieclayton:
@azuresdkci Add to whitelist

src/Dns/Dns.Test/ScenarioTests/ZoneTests.ps1
```diff
@@ -80,6 +80,59 @@ function Test-ZoneCrud
...
+<#
+.SYNOPSIS
+Zone creation with delegation
+#>
+function Test-ZoneWithDelegation
+{
+	#Parent zone creation
+	$parentZoneName = Get-RandomZoneName
+    $resourceGroup = TestSetup-CreateResourceGroup
+	$createdParentZone = New-AzDnsZone -Name $parentZoneName -ResourceGroupName $resourceGroup.ResourceGroupName -Tags @{tag1="value1"}
```
cormacpayne: @ShwethaRaghupathi I would that this test is restructured in the following way:

```powershell
function Test-ZoneWithDelegation
{
    # Parent zone creation
    $parentZoneName = Get-RandomZoneName
    $resourceGroup = TestSetup-CreateResourceGroup
    try
    {
        # (Almost) all of the remaining test code here
    }
    finally
    {
        Remove-AzResourceGroup -Name $resourceGroup.ResourceGroupName -Force
    }
}
```

This structure allows you to do the necessary setup, run the cmdlets to be tested with the assertions in the `try` block, and at the end in the `finally` block, you delete the resource group containing all of the resources created during the test. This ensures that the resource group is _always_ deleted when the test is recorded and doesn't leave any additional resources hanging around.

src/Dns/Dns/Zones/NewAzureDnsZone.cs
```diff
@@ -102,7 +115,48 @@ public override void ExecuteCmdlet()
                         ? string.Format(ProjectResources.Success_NewPrivateZone, this.Name, this.ResourceGroupName)
                         : string.Format(ProjectResources.Success_NewZone, this.Name, this.ResourceGroupName));
                     this.WriteObject(result);
+
+                    if (!string.IsNullOrEmpty(this.ParentZoneName) && this.Name.EndsWith(this.ParentZoneName))
```
cormacpayne: @ShwethaRaghupathi I think that we should make these series of calls before we call `this.WriteObject(result)` above -- this will ensure that if there were any problems with adding the delegation, the cmdlet will fail before an object is written to the output stream

src/Dns/Dns/help/New-AzDnsZone.md
```diff
@@ -98,6 +105,51 @@ Accept pipeline input: True (ByPropertyName)
 Accept wildcard characters: False
 ```
 
+### -ParentZone
```
cormacpayne: @ShwethaRaghupathi for each of these parameters that have been added in the PR, would you mind adding an example to show users what values should be provided to the parameters and what the corresponding output is?

src/Dns/Dns/Zones/NewAzureDnsZone.cs
ShwethaRaghupathi: The requirement is that zone creation and delegation need not be atomic. Zone creation if successful and delegation does not go through, indicate success of zone creation following which indicate that delegation could not be added. Hence not moving it to the end.

src/Dns/Dns/help/New-AzDnsZone.md
ShwethaRaghupathi: Added

markcowl:
@ShwethaRaghupathi Test error on ubuntu worker

src/Dns/Dns/Zones/NewAzureDnsZone.cs
cormacpayne: @ShwethaRaghupathi this pattern can be confusing to users as there is a scenario where the cmdlet will return the zone that they created, but could also throw an exception after that if the delegation was unsuccessful -- this is the issue with having two completely separate operations put into a single cmdlet. I think what would be best is if we caught any errors when trying to do the delegation and wrote a warning saying that the delegation was unsuccessful rather than throwing an exception.

src/Dns/Dns/Zones/NewAzureDnsZone.cs
```diff
@@ -101,8 +114,107 @@ public override void ExecuteCmdlet()
...
+            }
+            return recordSet;
+        }
+
+        /// <summary>
+        /// This method parses the parent zone from the command arguments, depending
+        /// on the format of the parent zone details passed.
+        /// Supports resource id, parent zone name and the zone object.
+        /// </summary>
+        private DnsZone ParseParentZoneFromArguments()
```
cormacpayne: @ShwethaRaghupathi if you'd like, we have our own resource id parsing class [here](https://github.com/Azure/azure-powershell-common/blob/master/src/ResourceManager/Version2016_09_01/Utilities/Models/ResourceIdentifier.cs) that you can use to grab certain parts of the id rather than having to re-implement the below logic. Your project already has a reference to the package containing this class, so you'd just need to use the namespace.

src/Dns/Dns/Zones/NewAzureDnsZone.cs
ShwethaRaghupathi: Yes that makes sense to indicate via warning msg. Updated the PR with this change

src/Dns/Dns/Zones/NewAzureDnsZone.cs
ShwethaRaghupathi: Updated PR with this change.

PR #8943, opened 2019-04-04
Time zone parameter

NMijat1024:
## Description
This change introduces optional parameter time zone for the Managed Instance create commandlet.

## Checklist


NMijat1024:
@azuresdkci retest this please.

maddieclayton:
@NMijat1024 Please also add a changelog entry for this change

src/Sql/Sql/ManagedInstance/Cmdlet/NewAzureSqlManagedInstance.cs
```diff
@@ -167,6 +167,15 @@ public class NewAzureSqlManagedInstance : ManagedInstanceCmdletBase
         [PSArgumentCompleter(ManagedInstanceProxyOverride.Proxy, ManagedInstanceProxyOverride.Redirect, ManagedInstanceProxyOverride.Default)]
         public string ProxyOverride { get; set; }
 
+        /// <summary>
+        /// Gets or sets the instance time zone
+        /// </summary>
+        [Parameter(Mandatory = false,
+            HelpMessage = "The time zone id for the instance to set. A list of time zone ids is exposed through the sys.time_zone_info (Transact-SQL) view.")]
+        [ValidateNotNullOrEmpty]
+        [PSArgumentCompleter(Constants.TimeZoneUtc)]
```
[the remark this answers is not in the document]
NMijat1024: Internally, we made an agreement not to have any arg completer.

src/Sql/Sql/ManagedInstance/Cmdlet/NewAzureSqlManagedInstance.cs
```diff
@@ -167,6 +167,14 @@ public class NewAzureSqlManagedInstance : ManagedInstanceCmdletBase
         [PSArgumentCompleter(ManagedInstanceProxyOverride.Proxy, ManagedInstanceProxyOverride.Redirect, ManagedInstanceProxyOverride.Default)]
         public string ProxyOverride { get; set; }
 
+        /// <summary>
+        /// Gets or sets the instance time zone
+        /// </summary>
+        [Parameter(Mandatory = false,
+            HelpMessage = "The time zone id for the instance to set. A list of time zone ids is exposed through the sys.time_zone_info (Transact-SQL) view.")]
+        [ValidateNotNullOrEmpty]
```
maddieclayton: Why would you not want to add an argument completer when there are by definition a limited number of arguments that are unlikely to change in value?

NMijat1024: Yes, it is a limited number, but the list is getting new values and we want to avoid updating it manually.

maddieclayton: The reason that we added the argument completer is so that users can tab through a static list of valid values, but are not restricted to that set of values (like with a ValidateSet). I think that having a slightly out of date list of values as an argument completer is much more helpful than forcing the user to do a separate search just to find any possible value.

NMijat1024: I can agree with that.
We were against it because user could assume that new time zone is not supported because it cannot be auto completed. If you think that would not be the case, I think we can add the list.

maddieclayton: I do agree that that could be an issue, but I think that forcing the user to look up the correct value each time is a more impactful negative. Is there any way to dynamically get this value from the server? We could also add a dynamic completer here if so. It should be relatively straightforward.

NMijat1024: That sounds great. I am planning to implement getting these values from the server, but implementation won't be done before code complete deadline for PS. I think we should check this in without completer, and add dynamic completer when it is ready. What do you think?

jaredmoo: Can you just implement the completer as a static list of string values?

NMijat1024: I'll implement static list and replace it with dynamic in some of the next releases.

maddieclayton: Perfect! Let me know if you need any pointer for the dynamic completer.

maddieclayton: As discussed offline we will need to add small quotes around these - example 'Dateline Standard Time'

NMijat1024: @maddieclayton After your update, completer works well. I guess we can merge this now.

Issue #8352, opened 2019-01-18
Issues with Azure Data-lake store applying the permissions.


thunderfalcon:
Issues with Azure Data-lake store applying the permissions.

The new Az module have issues applying permissions.

Set-AzDataLakeStoreItemAclEntry -AccountName "storename" -Path /-AceType User -Id (ID)-Permissions All -Default -Recurse

AzDatalakeStoreItemAclEntry is throwing error

<# Set-AzDataLakeStoreItemAclEntry : Error in modifying ACL entries default:user::rwx for 
path /.
Operation: MODIFYACLENTRIES failed with   Unknown Error: The 'User-Agent' header must be modified using the appropriate 
property or method.
Parameter name: name Source: System StackTrace:    at System.Net.WebHeaderCollection.ThrowOnRestrictedHeader(String 
headerName)
   at System.Net.WebHeaderCollection.Set(String name, String value)
   at Microsoft.Azure.DataLake.Store.WebTransport.AssignCommonHttpHeaders(HttpWebRequest webReq, AdlsClient client, 
RequestOptions req, String token, String opMethod, IDictionary`2 customHeaders, Int32 postRequestLength)
   at Microsoft.Azure.DataLake.Store.WebTransport.<MakeSingleCallAsync>d__22.MoveNext().
Last encountered exception thrown after 5 tries. [The 'User-Agent' header must be modified using the appropriate property or 
method.
Parameter name: name,The 'User-Agent' header must be modified using the appropriate property or method.
Parameter name: name,The 'User-Agent' header must be modified using the appropriate property or method.
Parameter name: name,The 'User-Agent' header must be modified using the appropriate property or method.
Parameter name: name,The 'User-Agent' header must be modified using the appropriate property or method.
Parameter name: name]
[ServerRequestId:]
At line:1 char:1
+ Set-AzDataLakeStoreItemAclEntry -AccountName "" -Pa ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : CloseError: (:) [Set-AzDataLakeStoreItemAclEntry], AdlsException
    + FullyQualifiedErrorId : Microsoft.Azure.Commands.DataLakeStore.SetAzureDataLakeStoreItemAclEntry #>

I reverted back to AzureRm module and everything works smooth.
This fix needs to be deployed.

migrande:
[...]

PR #8945, opened 2019-04-04
Update the ADLS sdk to use httpclient, integrate with azure framework

rahuldutta90:
## Description

Update the ADLS sdk to use httpclient. This will fix issue https://github.com/Azure/azure-powershell/issues/8352
Integrate ADLS testing with Azure Test framework so that issues can be caught in validation
Minor fixes to tests

## Checklist


rahuldutta90:
Added the private nuget package temporarily under local feed. Will remove it once my changes are published in nuget.

src/DataLakeStore/DataLakeStore/DataLakeStore.csproj
```diff
@@ -12,7 +12,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.DataLake.Store" Version="1.1.15" />
+    <PackageReference Include="Microsoft.Azure.DataLake.Store" Version="1.2.2-beta" />
```
rahuldutta90: this will change to 1.2.1-alpha once my nuget package gets publichsed. TODO.

src/DataLakeStore/DataLakeStore.Test/ScenarioTests/AdlMockDelegatingHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+            CancellationToken cancellationToken)
+        {
+            var response = await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
+
+            var str = await response.Content.ReadAsStringAsync();
+            if (IsJson(str))
+            {
+                try
+                {
+                    object parsedJson = JsonConvert.DeserializeObject(str);
```
akharit: Can you add a comment here documenting why this is necessary?

src/DataLakeStore/DataLakeStore.Test/ScenarioTests/UrlDecodingRecordMatcher.cs
```diff
@@ -43,11 +45,11 @@ public override string GetMatchingKey(HttpRequestMessage request)
                 path = path.Replace("?&", "?");
             }
 
-            if (path.Contains("%3A"))
-            {
-                path = path.Replace("%3A", ":");
-            }
-
+            //if (path.Contains("%3A"))
```
akharit: Remove these lines?

src/DataLakeStore/DataLakeStore/DataPlaneModels/AdlsClientFactory.cs
```diff
@@ -43,25 +45,16 @@ internal static AdlsClient GetAdlsClient(string accntNm, IAzureContext context)
                 {
                     return ClientFactory[accntNm];
                 }
-                ServiceClientCredentials creds;
+                AdlsClient client = null;
```
markcowl: How does IsTest work?  Generally there should not be any dependency on test code in shipped cmdlets.  At least this needs to be protected by a #if for DEBUG versus Release builds.  Ideally you would use a Provider (We provide a registry for factories in AzureSession), so the factory can be provided in tests, but has a default value in the code.

src/DataLakeStore/DataLakeStore.Test/ScenarioTests/AdlsTests.ps1
```diff
@@ -699,16 +699,15 @@ function Test-DataLakeStoreFileSystem
 		Assert-AreEqual $result.FileCount 1
 
 		# Export DiskUsage
-		$targetFile = "/DuOutputFile"
-		Export-AdlStoreChildItemProperties -Account $accountName -Path $summaryFolder -OutputPath $targetFile -GetDiskUsage -IncludeFile -SaveToAdl
-		$result = Get-AzDataLakeStoreItem -Account $accountName -path $targetFile
+		$targetFile = "./ScenarioTests/DuOutput"
```
markcowl: Generally, you want to be careful to use the correct base directory (based on the executing assembly) to avoid issues between tests when using paths.  ALso, use Join-Path to join path segments - this guarantees that your path uses the correct platform-specific separators

src/DataLakeStore/DataLakeStore.Test/ScenarioTests/AdlsTests.ps1
```diff
@@ -857,10 +856,11 @@ function Test-DataLakeStoreFileSystemPermissions
 		Assert-AreEqual $($currentCount+1) $result.Count
 
 		# Export Acl
-		$targetFile = "/aclOutputFile"
-		Export-AzDataLakeStoreChildItemProperties -Account $accountName -Path "/" -OutputPath $targetFile -GetAcl -IncludeFile -SaveToAdl
-		$result = Get-AzDataLakeStoreItem -Account $accountName -path $targetFile
+		$targetFile = "./ScenarioTests/acloutput"
```
markcowl: same comment

markcowl: Assuming this specifies a local path an not an ADLS path

src/DataLakeStore/DataLakeStore.Test/ScenarioTests/AdlsTests.ps1
markcowl: Assuming this specifies a local path and not an ADLS path

src/DataLakeStore/DataLakeStore/DataLakeStore.csproj
```diff
@@ -12,7 +12,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.DataLake.Store" Version="1.1.15" />
+    <PackageReference Include="Microsoft.Azure.DataLake.Store" Version="1.2.1-alpha" />
```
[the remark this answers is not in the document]
rahuldutta90: We named it as alpha but it would be same as beta only. https://www.nuget.org/packages/Microsoft.Azure.DataLake.Store/1.2.1-alpha Do you want this to beta only? Do you want it to be beta only? We can make it beta since the difference would be the name.

src/DataLakeStore/DataLakeStore.Test/ScenarioTests/AdlsTestsBase.cs
```diff
@@ -29,14 +29,15 @@
...
+using Microsoft.Azure.Commands.DataLake.Test.ScenarioTests;
 
 namespace Microsoft.Azure.Commands.DataLakeStore.Test.ScenarioTests
 {
     public class AdlsTestsBase : RMTestBase
     {
         private readonly EnvironmentSetupHelper _helper;
 
-        internal const string ResourceGroupLocation = "eastus2";
+        internal const string ResourceGroupLocation = "westus";
```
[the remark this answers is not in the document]
rahuldutta90: I agree, however this has been hardcoded all along. eastus2/westus are allowed regions. I changed to westus, because, in our dogfood environment, eastus2 ADLS backend was giving issues. That is why changed it. I can try making this change as a part of a separate PR.

src/DataLakeStore/DataLakeStore/DataPlaneModels/AdlsClientFactory.cs
rahuldutta90: I agree provider is the best way, however I cannot directly use ICLientFactory (AzureSession.Instance.ClientFactory) to generate my client in a generic way because my client does not inherit from ServiceClient. but I agree way we can think of a better way to do this. I will create an issue related to this, but I dont think this will be a blocking issue.

 For now here is what we can do. I willmake this IsTest variables internal and use it only in "internalvisible" way to run only in my tests. In that way we do not expose these test specific stuff for use.

src/DataLakeStore/DataLakeStore.Test/ScenarioTests/AdlsTestsBase.cs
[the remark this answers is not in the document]
rahuldutta90: will create an issue so that I can do this as a part of separate pr.

src/DataLakeStore/DataLakeStore.Test/ScenarioTests/AdlsTests.ps1
rahuldutta90: fixed.

src/DataLakeStore/DataLakeStore.Test/ScenarioTests/AdlsTests.ps1
rahuldutta90: fixed.

PR #8966, opened 2019-04-08
ADO changes to follow guidelines

MiYanni:
## Description
See the guidelines here: https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki?wikiVersion=GBwikiMaster&pagePath=%2FAzure%20DevOps%20Pipeline%20Guidance

### Changes
- Moved yaml files to `.azure-pipelines` folder
- Moved core job logic to steps files in `.azure-pipelines/util`
- Removed `install-pwsh` steps since the Windows machines now have PowerShell Core installed
- Moved the `after.Azure.PowerShell.sln.targets` to `tools` so the `build` folder is no longer needed
- Used shared `Repo.props` values throughout the `build.proj`
- Removed duplicated and deprecated properties in `build.proj`
- Changed ADO build process to use a matrix to define the OS differentiation
- Changed ADO build process to have a `netcoreapp21` (called `powershell-core`) and `net472` (called `windows-powershell`) pipeline (based on test framework and intent)
- Made `TestNet472` target use the `Build` target to run properly (same as `Build` and `Test` for netcoreapp21)

## Checklist

### DIFF
--- a/.azure-pipelines/powershell-core.yml
+++ b/.azure-pipelines/powershell-core.yml
@@ -16,21 +16,21 @@ jobs:
   strategy:
     matrix:
       windows:
-        osName: ${{ variables.WindowsName }}
-        imageName: ${{ variables.WindowsImage }}
+        OSName: ${{ variables.WindowsName }}
+        ImageName: ${{ variables.WindowsImage }}
       linux:
-        osName: ${{ variables.LinuxName }}
-        imageName: ${{ variables.LinuxImage }}
+        OSName: ${{ variables.LinuxName }}
+        ImageName: ${{ variables.LinuxImage }}
       macOS:
-        osName: ${{ variables.MacOSName }}
-        imageName: ${{ variables.MacOSImage }}
+        OSName: ${{ variables.MacOSName }}
+        ImageName: ${{ variables.MacOSImage }}
   pool:
-    vmImage: $(imageName)
+    vmImage: $(ImageName)
 
   steps:
   - template: util/build-steps.yml
     parameters:
-      osName: $(osName)
+      osName: $(OSName)
       testFramework: ${{ variables.TestFramework }}
       configuration: ${{ variables.Configuration }}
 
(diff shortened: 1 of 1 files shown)


MiYanni:
CI runs: 
https://dev.azure.com/azure-sdk/public/_build/results?buildId=15836
https://dev.azure.com/azure-sdk/public/_build/results?buildId=15837

.azure-pipelines/util/download-build-steps.yml
```diff
@@ -0,0 +1,9 @@
+parameters:
+  osName: ''
+
+steps:
```
weshaggard: Looks like a number of these templates can be folded into the one yaml file to avoid some levels of complexity.

.azure-pipelines/net472.yml
```diff
@@ -0,0 +1,36 @@
+variables:
+  WindowsName: windows
+  WindowsImage: VS2017-Win2016
+  TestFramework: net472
+  TestTarget: TestNet472
+  Configuration: Debug
```
weshaggard: If this is a configuration you want to enable at queue time then you will need to remove it from here and put it only in the build pipeline variables tab.

.azure-pipelines/util/download-build-steps.yml
MiYanni: Since these are reused between different jobs in different pipelines, it makes sense to have a separate template. The primary issue is that you cannot template a job matrix. Meaning, I can't make a job template where a strategy is conditionally defined.

.azure-pipelines/net472.yml
MiYanni: This is not intended to be build-time variables. I use these simply as global variables to all jobs in the yaml. If there was a concept of hard-coded values (constants), I would have used them here. Also, it is convenient to have all values that may change in the future listed at the top of the file.

.azure-pipelines/util/download-build-steps.yml
weshaggard: Yes that is one of the limitations of the templates. However I don't think that would be necessary if you had one matrix strategy for both and just control the framework as one of the matrix variables.

Issue #8926, opened 2019-04-02
Regex problem in Set-AzVMAEMExtension causes installation failure when managed disk name contains hyphen.


tinodo:
## Description

Due to a RegEx problem in https://github.com/Azure/azure-powershell/blob/master/src/Compute/Compute/Extension/AEM/AEMHelper.cs (line 93 and 105) you cannot install the AEM Extension when you use a managed disk and the disk name contains hyphen. Example disk id that would fail but should succeed:

/subscriptions/xxx/resourcegroups/yyy/providers/Microsoft.Compute/disks/DRIVE0-SRV1

This is because the RegEx uses the 'w' character, which doesn't match a hyphen. It should be included.

There is another issue around line 105;
            var matcher = new Regex("/subscriptions/([^/]+)/resourceGroups/([^/]+)/providers/([^/]+)/([^/]+)/([^/]+)(/\\w+)?");
            var result = matcher.Match(id);
            if (!result.Success || result.Groups == null || **result.Groups.Count < 3**)
            {
                throw new InvalidOperationException(string.Format("Cannot find resource group name and storage account name from resource identity {0}", id));
            }
            return result.Groups[5].Value;

The **result.Groups.Count <3** in the if should read **result.Groups.Count < 6**

But that one is less important for now.

## Steps to reproduce

Install the extension on a VM with Managed Disks where the disk name contains a hyphen. It will fail:

Set-AzVMAEMExtension : Cannot find resource group name and storage account name from resource identity /subscriptions/xxx/resourcegroups/yyy/providers/Microsoft.Compute/disks/DRIVE0-SRV1 

## Environment data

Name                           Value
----                           -----
PSVersion                      5.1.17763.316
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.17763.316
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1

## Module versions

ModuleType Version    Name                                ExportedCommands
---------- -------    ----                                ----------------
Script     1.6.0      az

## Error output

Set-AzVMAEMExtension : Cannot find resource group name and storage account name from resource identity /subscriptions/xxx/resourcegroups/yyy/providers/Microsoft.Compute/disks/DRIVE0-SRV1

cormacpayne:
[...]

PR #8972, opened 2019-04-09
AEM: Fix regex to be case insensitive

MSSedusch:
## Description

The regular expressions to extract the resource group and the resource name from a resource id were case sensitive. In some cases, the cmdlet was not able to install the extension because the resource id of e.g. disks had a lowercase resourcegroups in the resource id.

Fix for issue #8926 

## Checklist


src/Compute/Compute/Extension/AEM/AEMHelper.cs
```diff
@@ -90,26 +90,26 @@ internal StorageAccount GetStorageAccountFromCache(string accountName)
         }
         internal string GetResourceGroupFromId(string id)
         {
-            var matcher = new Regex("/subscriptions/([^/]+)/resourceGroups/([^/]+)/providers/(\\w+)");
```
maddieclayton: @MSSedusch Rather than update this code, is there any reason you can't use our common code implementation found here: https://github.com/Azure/azure-powershell-common/blob/master/src/ResourceManager/Version2016_09_01/Utilities/Models/ResourceIdentifier.cs? If there is any functionality missing from that common code that will cause this to fail, please let me know and we can see about updating it.

MSSedusch: no reason. I was not aware of it :) I replaced it

PR #9020, opened 2019-04-16
Rename Advanced Threat Protection cmdlets to Advanced Data Security and enable VA by default

talhers:
## Description

Rename Advanced Threat Protection cmdlets to Advanced Data Security and enable VA by default

## Checklist


talhers:
All the tests pass, except Build macOS:
The build succeeds, but it keeps failing on:

Artifact build-macOS already exists for build 18985.

Can someone please advise?
Thanks!

talhers:
@maddieclayton @yaakoviyun @yoavfr

talhers:
> Since I do not see a design review linked for these changes, the only question I have is the one parameter, `DoNotAutoEnableVulnerabilityAssessment`. Any reason this isn't named `DisableVulnerabilityAssessment`? I would think that `do not auto enable` is the same as `disable`, since only 2 options are available.

I think that "disable" is not suitable here because VA is not enabled.
Do you have any suggestion to what we can name the parameter? Thanks!

MiYanni:
@talhers 
Let's go through this so I can understand. From your comment, you make it sound like there are 3 states, `Enabled`, `Disabled`, and `Not Enabled but Not Disabled`. But, that third state, what does it represent?

From my understanding, this PR intends to make `VulnerabilityAssessment` *enabled by default* when running `Enable-AzSqlInstanceAdvancedDataSecurity`. But, considering the opposite situation, if you wanted it to *not be enabled by default*, wouldn't that be *disabled by default*?

> I think that "disable" is not suitable here because VA is not enabled.
So, if VA is 'enabled' when this cmdlet is run (without this switch), then it *will be* enabled, correct? So, when you provide `DisableVulnerabilityAssessment`, that's staying the state of `VulnerabilityAssessment` after running this cmdlet. Here, `Disable` is used as a noun describing the state of `VulnerabilityAssessment`, not as a verb. So, providing `DisableVulnerabilityAssessment` means that `VulnerabilityAssessment` **will not be enabled**. That is the intent, correct?

MiYanni:
Reviewed via voice conversation. Name will be changed to `DoNoConfigureVulnerabilityAssessment`. Also, changelog entry (in `src/Sql/changelog.md`) will be added under the `## Upcoming Release` header.

PR #8607, opened 2019-02-23
Updated Get-AzAutomationJobOutputRecord to handle JSON and Text record values

lwajswaj:
## Description

This PR addresses issue #7977 and #8600. Root cause of the issue was identified as stream outputs always expecting to be a JSON file; however, this ain't true in all cases as String or PSObjects are not returned as JSON. To fix it, a RegEx was put in place to identify JSON output from text output. 

## Checklist

- [X] I have read the [_Submitting Changes_](https://github.com/Azure/azure-powershell/blob/master/CONTRIBUTING.md#submitting-changes) section of [`CONTRIBUTING.md`](https://github.com/Azure/azure-powershell/blob/master/CONTRIBUTING.md)
- [X] The title of the PR is clear and informative
- [X] The appropriate [change log has been updated](https://github.com/Azure/azure-powershell/blob/master/CONTRIBUTING.md#updating-the-change-log)
- [X] The PR does not introduce [breaking changes](https://github.com/Azure/azure-powershell/blob/master/documentation/breaking-changes/breaking-changes-definition.md)
- [ ] If applicable, the changes made in the PR have proper test coverage
- [ ] For public API changes to cmdlets:
    - [ ] the changes have gone through a [cmdlet design review](https://github.com/Azure/azure-powershell-cmdlet-review-pr) (Microsoft internal partners only)
    - [ ] the cmdlet markdown files were [generated using the `platyPS` module](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/help-generation.md)

Issue #8600, opened 2019-02-22
Get-AzAutomationJobOutputRecord fails with JSON conversion error


rwaal:
## Description
I am trying to get the full output of an Automation job output record. But it consistently fails with the following error:

"Conversion from JSON failed with error: Unexpected character encountered while parsing value: R. Path '', line 0, position 0". 

I am using the cmdlet exactly as described in the [online docs](https://docs.microsoft.com/en-us/powershell/module/az.automation/get-azautomationjoboutputrecord?view=azps-1.3.0&viewFallbackFrom=azps-1.0.0). So I don't understand why this is happening.

## Steps to reproduce

```powershell
Get-AzAutomationJobOutput -ResourceGroupName $rgName –AutomationAccountName $aaName -Id $job.JobId –Stream Any | Get-AzAutomationJobOutputRecord
```
or
```powershell
Get-AzAutomationJobOutputRecord -JobId $jobId -Id $id -ResourceGroupName $rgName -AutomationAccountName $aaName
```

## Environment data

```
Name                           Value
----                           -----
PSVersion                      6.1.3
PSEdition                      Core
GitCommitId                    6.1.3
OS                             Microsoft Windows 10.0.17763 
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

## Module versions

```powershell


    Directory: C:\Program Files\PowerShell\Modules


ModuleType Version    Name                                PSEdition ExportedCommands
---------- -------    ----                                --------- ----------------
Script     1.2.1      Az.Accounts                         Core,Desk {Disable-AzDataCollection, Disable-AzContextAutosave, Enable-AzDataCollection, Enable-AzContextAutosave...}
Script     1.2.0      Az.Accounts                         Core,Desk {Disable-AzDataCollection, Disable-AzContextAutosave, Enable-AzDataCollection, Enable-AzContextAutosave...}
Script     1.0.0      Az.Accounts                         Core,Desk {Disable-AzDataCollection, Disable-AzContextAutosave, Enable-AzDataCollection, Enable-AzContextAutosave...}
Script     1.0.1      Az.Aks                              Core,Desk {Get-AzAks, New-AzAks, Remove-AzAks, Import-AzAksCredential...}
[...]

Issue #7977, opened 2018-12-01
[Az 0.6.1] Start-AzAutomationRunbook ends with a JSON conversion error


alexandair:
Every time I start a runbook using the Start-AzAutomationRunbook, job completes as expected, but I get an error:
```
Start-AzAutomationRunbook : Conversion from JSON failed with error: Unexpected character encountered while parsing value: S. Path '', line 0, position 0.
```

cormacpayne:
@alexandair Hey Aleksandar, would you mind providing the following information:
- the debug trace from executing the command with `$DebugPreference = "Continue"`
- the full error output from running `Resolve-AzError`

alexandair:
```
Start-AzAutomationRunbook : Conversion from JSON failed with error: Unexpected character encountered while parsing value: S. Path '', line 0, position 0.
At line:1 char:1
+ Start-AzAutomationRunbook @startParams -Wait -Verbose
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : CloseError: (:) [Start-AzAutomationRunbook], CmdletInvocationException
+ FullyQualifiedErrorId : Microsoft.Azure.Commands.Automation.Cmdlet.StartAzureAutomationRunbook

DEBUG: AzureQoSEvent: CommandName - Start-AzAutomationRunbook; IsSuccess - False; Duration - 00:01:00.6955491; Exception - System.Management.Automation.CmdletInvocationException: Conversion from JSON failed with error: Unexpected character encountered while parsing value: S. Path '', line 0, position 0. ---> System.ArgumentException: Conversion from JSON failed with error: Unexpected character encountered while parsing value: S. Path '', line 0, position 0. ---> Newtonsoft.Json.JsonReaderException: Unexpected character encountered while parsing value: S. Path '', line 0, position 0.
   at Newtonsoft.Json.JsonTextReader.ParseValue()
   at Newtonsoft.Json.JsonTextReader.Read()
   at Newtonsoft.Json.JsonReader.ReadForType(JsonContract contract, Boolean hasConverter)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)
   at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)
[...]

PR #9031, opened 2019-04-18
Updated Get-AzAutomationJobOutputRecord to handle JSON and Text record values

lwajswaj:
This is a re-open of [PR 8607](https://github.com/Azure/azure-powershell/pull/8607). 

Answering @NarayanThiru comments: Current logic doesn't work as the error message and code is different (regardless that it's easier and cleaner to check on the output format rather than attempt and failing). Regarding the issue and how to repro/test, refer to the linked issues they contain a lot of good information.

CC: @vladimir-shcherbakov, @MiYanni 

## Description

This PR addresses issue #7977 and #8600. Root cause of the issue was identified as stream outputs always expecting to be a JSON file; however, this ain't true in all cases as String or PSObjects are not returned as JSON. To fix it, a RegEx was put in place to identify JSON output from text output. 

## Checklist

### DIFF
--- a/src/Automation/Automation.Test/ScenarioTests/AutomationTests.cs
+++ b/src/Automation/Automation.Test/ScenarioTests/AutomationTests.cs
@@ -77,7 +77,7 @@ namespace Commands.Automation.Test
         [Trait(Category.Service, Category.Automation)]
         public void TestAutomationRunbookWithParameter()
         {
-            TestRunner.RunTestScript("Test-RunbookWithParameter -runbookPath ScenarioTests\\Resources\\fastJob.ps1 -type 'PowerShell' -parameters @{'nums'='[1,2,3,4,5,6,7]'} -expectedResult 28");
+            TestRunner.RunTestScript("Test-RunbookWithParameter -runbookPath ScenarioTests\\Resources\\Test-PowershellRunbook.ps1 -type 'PowerShell' -parameters @{'nums'=@(1,2,3,4,5,6,7)} -expectedResult 28");
         }
 
         [Fact(Skip = "Need x64 test framework.")]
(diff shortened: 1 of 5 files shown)


markcowl:
@vrdmr @jemex Can one of you review this PR?

NarayanThiru:
The changes look good.  @thomasyip-msft, @vrdmr, @safeermohammed, can any of you merge this PR?

src/Automation/Automation/ChangeLog.md
```diff
@@ -31,6 +31,9 @@
 
 ## Version 1.1.2
 * Fixed issue when retreiving certain monthly schedules in several Azure Automation cmdlets
+* Updated Get-AzAutomationJobOutputRecord to handle JSON and Text record values.
```
markcowl: Can we move this change under the 'Upcoming Release' heading above?  Otherwise, LGTM

markcowl:
@lwajswaj if you can make the changelog change, we can fix this.  Is there a simple way we can add test automation for this fix, to prevent future regressions?

src/Automation/Automation/ChangeLog.md
lwajswaj: Just moved to the right spot, not sure how I got it there in the first place.

lwajswaj:
> 
> 
> @lwajswaj if you can make the changelog change, we can fix this. Is there a simple way we can add test automation for this fix, to prevent future regressions?

I might be able to add a quick test under the scenario tests. Kind of like starting the mock up runbook and retrieving its output stream (that's the current error condition)

lwajswaj:
> 
> 
> @lwajswaj if you can make the changelog change, we can fix this. Is there a simple way we can add test automation for this fix, to prevent future regressions?

@markcowl, Just added tests for this scenario...... along the way, I found the tests for runbook execution were not executing properly, for example expected result was never evaluated, that's why I included a couple of extra changes:

* Renamed **fastjob.ps1** to **Test-PowershellRunbook.ps1** (name is meaningful now) also migrated this script from using workflow to plain powershell. Added additional outputs since that's the error condition for  'Get-AzAutomationJobOutputRecord' otherwise test was successful with a single integer output.
* Renamed **fastjob.py** to **TestPythonRunbook.py** (name is meaningful too). Also added additional outputs (same reason as before)
* I found 'Test-RunbookWithParameter' was not using its parameter $expectedResult, I changed it to include this assertion as part of the test

Of course, added the tests for 'Get-AzAutomationJobOutputRecord' for preventing regressions.

PR #9127, opened 2019-05-02
Add ManagedServices Module

rajivnandivada:
## Description

This is the powershell module for ManagedServices RP.
This feature is used by customers of Managed Service Providers (MSPs). Customers give an MSP the ability to manage their subscription or resource group. In addition to granting access, the customer can also remove access or view existing access delegations. MSPs are able able to view the list of customers who have granted them access to subscriptions/resource groups. There are two objects involved in this process: A registration definition which identifies the MSP and the role definitions granted to the MSP. The MSP can optionally define this object using a Managed Services marketplace offering Access assignments which associate a subscription or resource group with the definition.

## Checklist

### DIFF
--- a/src/Websites/Websites.Test/ScenarioTests/WebAppBackupRestoreTests.cs
+++ b/src/Websites/Websites.Test/ScenarioTests/WebAppBackupRestoreTests.cs
@@ -116,21 +116,21 @@ namespace Microsoft.Azure.Commands.Websites.Test.ScenarioTests
             WebsitesController.NewInstance.RunPsTest(_logger, "Test-RestoreWebAppSnapshot");
         }
 
-        [Fact]
+        [Fact(Skip="Get-AzDeletedWebApp returning null in recorded tests #9191")]
         [Trait(Category.AcceptanceType, Category.CheckIn)]
         public void TestGetDeletedWebApp()
         {
             WebsitesController.NewInstance.RunPsTest(_logger, "Test-GetDeletedWebApp");
         }
 
-        [Fact]
+        [Fact(Skip="Get-AzDeletedWebApp returning null in recorded tests #9191")]
         [Trait(Category.AcceptanceType, Category.CheckIn)]
         public void TestRestoreDeletedWebAppToExisting()
         {
             WebsitesController.NewInstance.RunPsTest(_logger, "Test-RestoreDeletedWebAppToExisting");
         }
 
-        [Fact]
+        [Fact(Skip="Get-AzDeletedWebApp returning null in recorded tests #9191")]
         [Trait(Category.AcceptanceType, Category.CheckIn)]
         public void TestRestoreDeletedWebAppToNew()
         {


adxsdkps:

src/ManagedServices/ManagedServices/Commands/GetAzureRmManagedServicesAssignment.cs
```diff
@@ -0,0 +1,79 @@
...
+        [Parameter(Position = 0, ParameterSetName = DefaultParameterSet, HelpMessage = "The subscription scope.")]
+        [ScopeCompleter]
+        public string Scope { get; set; }
+
+        [Parameter(Position = 1, ParameterSetName = DefaultParameterSet, ValueFromPipeline = true, HelpMessage = "The Registration Assignment Id.")]
+        [ValidateNotNullOrEmpty]
+        [Alias("Id")]
+        public string ResourceId { get; set; }
+
+        [Parameter(ParameterSetName = DefaultParameterSet, HelpMessage = "Whether to include the Registration Definiton in the response.")]
```
markcowl: This should be in both parameter sets, and so no parameter set name is necessary

src/ManagedServices/ManagedServices/Commands/GetAzureRmManagedServicesAssignment.cs
```diff
@@ -0,0 +1,79 @@
...
+                        }
+                        else
+                        {
+                            // validate assignmentId.
+                            if (!assignmentId.IsGuid())
+                            {
+                                throw new ApplicationException("RegistrationAssignment must be a valid GUID.");
+                            }
+
+                            var result = this.PSManagedServicesClient.GetRegistrationAssignment(
```
markcowl: Thsi is not the intended usage of the resource id parameter set - the idea is to be able to pass in a resourceId for the assignment, without having to separately pass in a scope.

src/ManagedServices/ManagedServices/Commands/GetAzureRmManagedServicesAssignment.cs
```diff
@@ -0,0 +1,79 @@
...
+    using Microsoft.Azure.Commands.ResourceManager.Common.ArgumentCompleters;
+    using Microsoft.Azure.PowerShell.Cmdlets.ManagedServices.Models;
+    using System;
+    using System.Management.Automation;
+
+    [Cmdlet(
+        VerbsCommon.Get,
+        Microsoft.Azure.Commands.ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "ManagedServicesAssignment",
+        DefaultParameterSetName = DefaultParameterSet,
+        SupportsShouldProcess = true), OutputType(typeof(PSRegistrationAssignment))]
```
markcowl: No Need for SupprotsSHouldProcess

src/ManagedServices/ManagedServices/Commands/GetAzureRmManagedServicesDefinition.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Management.Automation;
+using System.Text;
+
+namespace Microsoft.Azure.PowerShell.Cmdlets.ManagedServices.Commands
+{
+    [Cmdlet(
+    VerbsCommon.Get,
+    Microsoft.Azure.Commands.ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "ManagedServicesDefinition",
+    DefaultParameterSetName = DefaultParameterSet,
+    SupportsShouldProcess = true), OutputType(typeof(PSRegistrationDefinition))]
```
markcowl: No need for SupportsSHouldProcess

src/ManagedServices/ManagedServices/Commands/GetAzureRmManagedServicesDefinition.cs
```diff
@@ -0,0 +1,77 @@
...
+    public class GetAzureRmManagedServicesDefinition : ManagedServicesCmdletBase
+    {
+        protected const string DefaultParameterSet = "Default";
+
+        [Parameter(Position = 0, ParameterSetName = DefaultParameterSet, HelpMessage = "The subscription scope.")]
+        [ScopeCompleter]
+        public string Scope { get; set; }
```
markcowl: This was supposed top be removed.  Why do we need scope for a Definition?

rajivnandivada:
@adxsdkps could you please review this? thank you.

src/ManagedServices/ManagedServices/Commands/GetAzureRmManagedServicesDefinition.cs
markcowl: The user shoudlk eb able to list all defintions, or get an individual definition.  I *think* they should eb able to do both with the Id /ResourceId parameter.  Scope doesn't seem to make sense, as a definition can apply at any scope it is assgiend to.   In your cmdlet review, you indicated that this parameter was removed.

src/ManagedServices/ManagedServices/Commands/GetAzureRmManagedServicesDefinition.cs
```diff
@@ -0,0 +1,77 @@
...
+            {
+                ConfirmAction(MyInvocation.InvocationName,
+                    "Microsoft.ManagedServices/registrationDefinitions",
+                    () =>
+                    {
+                        var scope = this.Scope ?? this.GetDefaultScope();
+
+                        if (string.IsNullOrEmpty(definitionId))
+                        {
+                            var results = this.PSManagedServicesClient.ListRegistrationDefinitions(scope);
```
markcowl: This should eb listing by subscription

src/ManagedServices/ManagedServices/Commands/GetAzureRmManagedServicesDefinition.cs
```diff
@@ -0,0 +1,77 @@
...
+                        else
+                        {
+                            // validate definitionId.
+                            if (!definitionId.IsGuid())
+                            {
+                                throw new ApplicationException("RegistrationDefinitionId must be a valid GUID.");
+                            }
+
+                            var result = this.PSManagedServicesClient.GetRegistrationDefinition(
+                                scope: scope,
```
markcowl: Again, no, Scope cannot be required - either the subscriptionId comes from the resource id, or, if the user supplies a guid, then the current subscription shoudl eb used.

src/ManagedServices/ManagedServices/Commands/NewAzureRmManagedServicesDefinition.cs
```diff
@@ -0,0 +1,125 @@
...
+    Microsoft.Azure.Commands.ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "ManagedServicesDefinition",
+    DefaultParameterSetName = DefaultParameterSet,
+    SupportsShouldProcess = true), OutputType(typeof(PSRegistrationDefinition))]
+    public class NewAzureRmManagedServicesDefinition : ManagedServicesCmdletBase
+    {
+        protected const string DefaultParameterSet = "Default";
+
+        [Parameter(ParameterSetName = DefaultParameterSet, Mandatory = false, HelpMessage = "The subscription scope.")]
+        [ScopeCompleter]
+        public string Scope { get; set; }
```
markcowl: This was supposed to be removed.  Please remove it.

src/ManagedServices/ManagedServices/Commands/NewAzureRmManagedServicesDefinition.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        [Parameter(ParameterSetName = DefaultParameterSet, Mandatory = true, HelpMessage = "The name of the Registration Definition.")]
+        public string RegistrationDefinitionName { get; set; }
+
+        [Parameter(ParameterSetName = DefaultParameterSet, Mandatory = true, HelpMessage = "The Managedby Tenant Identifier.")]
+        [ValidatePattern("^({){0,1}[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}(}){0,1}$")]
+        public string ManagedByTenantId { get; set; }
+
+        [Parameter(ParameterSetName = DefaultParameterSet, Mandatory = true, HelpMessage = "The Managedby Principal Identifier.")]
+        [ValidatePattern("^({){0,1}[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}(}){0,1}$")]
```
markcowl: These validatepatterns don't rhow very good errors - ypou are much better off throwing your own message at runtime if the id is not a guid

src/ManagedServices/ManagedServices/Commands/NewAzureRmManagedServicesDefinition.cs
```diff
@@ -0,0 +1,125 @@
...
+        public override void ExecuteCmdlet()
+        {
+            this.TryCatchExecuteCmdlet(() =>
+            {
+                ConfirmAction(MyInvocation.InvocationName,
+                    "Microsoft.ManagedServices/registrationDefinitions",
+                    () =>
+                    {
+                        Plan plan = null;
+                        if (!string.IsNullOrEmpty(this.PlanName) ||
```
markcowl: You want to do this if any of these are not null - don't some of them *have* to be non-null?

Also, you are better off using the IsParameterBound mechanism to determine if the user supplied this value (see: )  and using a ValidateNotNullOrEmpty attribute on the parameter

Finally, if any of these are actually required to define a plan, you should add a parameter set with these Plan parameters that makes the ones reuired for definiing a plan mandatory.

src/ManagedServices/ManagedServices/Commands/RemoveAzureRmManagedServcicesAssignment.cs
```diff
@@ -0,0 +1,71 @@
...
+    public class RemoveAzureRmManagedServcicesAssignment : ManagedServicesCmdletBase
+    {
+        protected const string DefaultParameterSet = "Default";
+        protected const string InputObjectParameterSet = "InputObject";
+
+        [Parameter(Position = 0, ParameterSetName = DefaultParameterSet, Mandatory = false, HelpMessage = "The subscription scope.")]
+        [ScopeCompleter]
+        public string Scope { get; set; }
+
+        [Parameter(ParameterSetName = DefaultParameterSet, Mandatory = true, ValueFromPipelineByPropertyName = true, HelpMessage = "The registration assignment Id.")]
```
markcowl: So, in this case, because you are either using Scope + Id to define the resource assign,ment, or you are piping from Get-AzResource, you should have separate parameters for Id and ResourceId, with Id being in the defualt parameter set, and ResourceId being in its own separate parameter set (and teking value from pipeline by property name)

src/ManagedServices/ManagedServices/Commands/RemoveAzureRmManagedServcicesAssignment.cs
```diff
@@ -0,0 +1,71 @@
...
+        [Parameter(Mandatory = true, ValueFromPipeline = true, HelpMessage = "The registration assignment object.", ParameterSetName = InputObjectParameterSet)]
+        [ValidateNotNull]
+        public PSRegistrationAssignment InputObject { get; set; }
+
+        [Parameter(Mandatory = false, HelpMessage = "Run cmdlet in the background")]
+        public SwitchParameter AsJob { get; set; }
+
+        public override void ExecuteCmdlet()
+        {
+            var resourceId = InputObject?.Id ?? this.ResourceId;
```
markcowl: No, this isn't going to work.  You need to get the value from Scope + Id, or extract these values from the resource id and  input object, you should not use default scopes below.

src/ManagedServices/ManagedServices/Commands/RemoveAzureRmManagedServicesDefinition.cs
```diff
@@ -0,0 +1,73 @@
...
+    SupportsShouldProcess = true), OutputType(typeof(PSRegistrationDefinition))]
+
+    public class RemoveAzureRmManagedServicesDefinition : ManagedServicesCmdletBase
+    {
+        protected const string InputObjectParameterSet = "InputObject";
+        protected const string DefaultParameterSet = "Default";
+
+        [Parameter(Position = 0, ParameterSetName = DefaultParameterSet, Mandatory = false, HelpMessage = "The subscription scope.")]
+        [ScopeCompleter]
+        public string Scope { get; set; }
```
markcowl: Again, what does scope have to do with a dfinition?  Scope is for assignments.  Remove this parameter

src/ManagedServices/ManagedServices/Commands/RemoveAzureRmManagedServicesDefinition.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        [Parameter(Position = 1, ParameterSetName = DefaultParameterSet, Mandatory = true, HelpMessage = "The registration definition identifier.")]
+        [Alias("Id")]
+        [ValidateNotNullOrEmpty]
+        public string ResourceId { get; set; }
```
markcowl: You need to have two parameter sets:
- one which breaks fown the resource id by its components (for most resources, this is resource group _ name, in this case, it may just be name, if these are tenant or subscription-level resources.

- one which takes ResourceId from a call to Get-AzResource with value from pipelinebypropertyname

src/ManagedServices/ManagedServices/Models/ManagedServicesCmdletBase.cs
```diff
@@ -0,0 +1,95 @@
...
+            return $"/subscriptions/{subscriptionId}";
+        }
+
+        protected void TryCatchExecuteCmdlet(Action action)
+        {
+            try
+            {
+                action();
+            }
+            catch (ErrorResponseException ere)
```
markcowl: Why do you need this on top of the existing handling of errors?

src/ManagedServices/ManagedServices/help/Az.ManagedServices.md
```diff
@@ -0,0 +1,31 @@
...
+Module Name: Az.ManagedServices
+Module Guid: {{ Update Module Guid }}
+Download Help Link: {{ Update Download Link }}
+Help Version: {{ Update Help Version }}
+Locale: {{ Update Locale }}
+---
+
+# Az.ManagedServices Module
+## Description
+{{ Fill in the Description }}
```
markcowl: You need to fill in a value

src/ManagedServices/ManagedServices/Commands/GetAzureRmManagedServicesAssignment.cs
rajivnandivada: good catch

src/ManagedServices/ManagedServices.Test/ScenarioTests/ManagedServicesTests.ps1
```diff
@@ -0,0 +1,152 @@
...
+
+    $cmdlet.ExecuteCmdlet()
+}
+
+function Test-ManagedServices_CRUD
+{
+    $roleDefinitionId = "acdd72a7-3385-48ef-bd42-f606fba81ae7";
+	$managedByTenantId = "bab3375b-6197-4a15-a44b-16c41faa91d7";
+	$principalId = "d6f6c88a-5b7a-455e-ba40-ce146d4d3671";
+	$scope = "subscriptions/38bd4bef-41ff-45b5-b3af-d03e55a4ca15"
```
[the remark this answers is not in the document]
rajivnandivada: https://github.com/Azure/azure-powershell/issues/9150

src/ManagedServices/ManagedServices/Commands/RemoveAzureRmManagedServicesDefinition.cs
rajivnandivada: Scope is required here. Registration definitions, unlike policy or role definitions, always exist at the subscription scope. I will rename it to SubscriptionId to help with the confusion.

src/ManagedServices/ManagedServices/Commands/NewAzureRmManagedServicesDefinition.cs
rajivnandivada: Thanks.

src/ManagedServices/ManagedServices/Commands/GetAzureRmManagedServicesDefinition.cs
rajivnandivada: RegistrationDefinitions are unlike Policy or Role definitions. They are a subscription level resource so they require a subscriptionid for CRUD operations. Scope, however, as I can see, is a very confusing. We are going to change this to say SubscriptionId.

src/ManagedServices/ManagedServices/Commands/GetAzureRmManagedServicesDefinition.cs
rajivnandivada: Yup.

src/ManagedServices/ManagedServices/Commands/NewAzureRmManagedServicesDefinition.cs
rajivnandivada: Using SubscriptionId instead.

src/ManagedServices/ManagedServices/Commands/NewAzureRmManagedServicesDefinition.cs
rajivnandivada: Done.

src/ManagedServices/ManagedServices/Commands/RemoveAzureRmManagedServcicesAssignment.cs
rajivnandivada: Fixed

src/ManagedServices/ManagedServices/Commands/GetAzureRmManagedServicesAssignment.cs
```diff
@@ -0,0 +1,107 @@
...
+            else if (this.IsParameterBound(x => x.Scope))
+            {
+                scope = this.Scope;
+            }
+            else
+            {
+                scope = this.GetDefaultScope();
+            }
+
+            ConfirmAction(MyInvocation.InvocationName,
```
markcowl: You should not use ConfirmAction here - this is only appropriate when implementing ShouldProcess, and Gets should not support ShouldProcess

src/ManagedServices/ManagedServices/Commands/GetAzureRmManagedServicesDefinition.cs
```diff
@@ -0,0 +1,85 @@
...
+                scope = this.GetDefaultScope();
+                definitionId = this.Id;
+            }
+            else if (this.IsParameterBound(x => x.ResourceId))
+            {
+                definitionId = this.ResourceId.GetResourceName();
+                scope = this.ResourceId.GetSubscriptionId().ToSubscriptionResourceId();
+            }
+
+            ConfirmAction(MyInvocation.InvocationName,
```
markcowl: See above - no ConfirmAction for Get cmdlets

src/ManagedServices/ManagedServices/Commands/GetAzureRmManagedServicesDefinition.cs
```diff
@@ -0,0 +1,85 @@
...
+    VerbsCommon.Get,
+    Microsoft.Azure.Commands.ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "ManagedServicesDefinition",
+    DefaultParameterSetName = DefaultParameterSet), OutputType(typeof(PSRegistrationDefinition))]
+    public class GetAzureRmManagedServicesDefinition : ManagedServicesCmdletBase
+    {
+        protected const string DefaultParameterSet = "Default";
+        protected const string ByResourceIdParameterSet = "ByResourceId";
+        protected const string ByIdParameterSet = "ById";
+
+        [Parameter(ParameterSetName = ByIdParameterSet, Mandatory = true, HelpMessage = "The registration definition identifier.")]
```
markcowl: Can't you list Managed Service Definitions at the subscription level?  If so, this should be non-mandatory

src/ManagedServices/ManagedServices/Commands/NewAzureRmManagedServicesAssignment.cs
```diff
@@ -0,0 +1,102 @@
...
+        public string Scope { get; set; }
+
+        [Parameter(ParameterSetName = DefaultParameterSet, Mandatory = true, HelpMessage = "The registration definition identifier.")]
+        [ValidateNotNullOrEmpty]
+        public string RegistrationDefinitionName { get; set; }
+
+        [Parameter(ParameterSetName = ByResourceIdParameterSet, ValueFromPipelineByPropertyName = true, Mandatory = true, HelpMessage = "The fully qualified resource id of the registration definition.")]
+        [ValidateNotNullOrEmpty]
+        [Alias("ResourceId")]
+        public string RegistrationDefinitionId { get; set; }
```
markcowl: This is confusing - is registrationDefinition Id the ResourceId of the registration definition, or just the id - this should be explicit,  RegistrationDefinitionResourceId (if it is a resource id)

src/ManagedServices/ManagedServices/Commands/GetAzureRmManagedServicesDefinition.cs
markcowl: Also, if this is called Name below, it should be Name here

src/ManagedServices/ManagedServices/Commands/NewAzureRmManagedServicesDefinition.cs
```diff
@@ -0,0 +1,145 @@
...
+    DefaultParameterSetName = DefaultParameterSet,
+    SupportsShouldProcess = true), OutputType(typeof(PSRegistrationDefinition))]
+    public class NewAzureRmManagedServicesDefinition : ManagedServicesCmdletBase
+    {
+        protected const string DefaultParameterSet = "Default";
+        protected const string ByPlanParameterSet = "ByPlan";
+
+        [Parameter(ParameterSetName = DefaultParameterSet, Mandatory = true, HelpMessage = "The name of the Registration Definition.")]
+        [Parameter(ParameterSetName = ByPlanParameterSet, Mandatory = true, HelpMessage = "The name of the Registration Definition.")]
+        public string RegistrationDefinitionName { get; set; }
```
markcowl: Name.  SInce this is the name matching the cmdlet noun, it should be called Name.  You can have an alias to the longer name, if you like

src/ManagedServices/ManagedServices/Commands/NewAzureRmManagedServicesDefinition.cs
```diff
@@ -0,0 +1,145 @@
...
+            {
+                throw new ApplicationException("RoleDefinitionId must be a valid GUID.");
+            }
+
+            if (this.RegistrationDefinitionId == default(Guid))
+            {
+                this.RegistrationDefinitionId = Guid.NewGuid();
+            }
+
+            ConfirmAction(MyInvocation.InvocationName,
```
markcowl: This message should be specific about what is being created at what scope - people use these messages to debug scripts, for example.  - at least the name of the registration definition, and where it is being created

src/ManagedServices/ManagedServices/Commands/NewAzureRmManagedServicesAssignment.cs
```diff
@@ -0,0 +1,102 @@
...
+                definitionId = this.RegistrationDefinition.Id;
+                scope = this.Scope ?? definitionId.GetSubscriptionId().ToSubscriptionResourceId();
+            }
+
+            if (this.RegistrationAssignmentId == default(Guid))
+            {
+                this.RegistrationAssignmentId = Guid.NewGuid();
+            }
+
+            ConfirmAction(MyInvocation.InvocationName,
```
markcowl: This should identify the type of resource and identifying information (in this case, scope and managedservice definition)

src/ManagedServices/ManagedServices/Commands/RemoveAzureRmManagedServcicesAssignment.cs
```diff
@@ -0,0 +1,93 @@
...
+        SupportsShouldProcess = true), OutputType(typeof(PSRegistrationDefinition))]
+    public class RemoveAzureRmManagedServcicesAssignment : ManagedServicesCmdletBase
+    {
+        protected const string DefaultParameterSet = "Default";
+        protected const string ByResourceIdParameterSet = "ByResourceId";
+        protected const string ByInputObjectParameterSet = "ByInputObject";
+
+        [Parameter(Position = 0, ParameterSetName = DefaultParameterSet, Mandatory = false, HelpMessage = "The scope where the registration assignment should be created.")]
+        [ScopeCompleter]
+        public string Scope { get; set; }
```
markcowl: What does scope do here if it is not mandatory?  Why would the user want to provide a scope?

src/ManagedServices/ManagedServices/Commands/RemoveAzureRmManagedServcicesAssignment.cs
```diff
@@ -0,0 +1,93 @@
...
+            else if (this.IsParameterBound(x => x.ResourceId))
+            {
+                assignmentId = this.ResourceId.GetResourceName();
+                if (!ManagedServicesUtility.TryParseAssignmentScopeFromResourceId(this.ResourceId, out scope))
+                {
+                    throw new ApplicationException($"Unable to parse the scope from [{this.ResourceId}]");
+                }
+            }
+
+            ConfirmAction(MyInvocation.InvocationName,
```
markcowl: Identify the type fo resource being removed, and identifying information - think of this as a way for a user to validate a script, or to check the effects of a complex command before executing

src/ManagedServices/ManagedServices/Commands/RemoveAzureRmManagedServicesDefinition.cs
```diff
@@ -0,0 +1,87 @@
...
+            {
+                definitionId = this.ResourceId.GetResourceName();
+                if (!ManagedServicesUtility.TryParseDefinitionScopeFromResourceId(this.ResourceId, out scope))
+                {
+                    throw new ApplicationException($"Unable to parse the scope from [{this.ResourceId}]");
+                }
+            }
+
+            ConfirmAction(MyInvocation.InvocationName,
+                "Microsoft.ManagedServices/registrationDefinitions",
```
markcowl: same comment on the ShouldProcess message

src/ManagedServices/ManagedServices/Extensions/ManagedServicesExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+// limitations under the License.
+// ----------------------------------------------------------------------------------
+
+using System;
+using System.Linq;
+using System.Management.Automation;
+using Microsoft.Azure.PowerShell.Cmdlets.ManagedServices.Models;
+using Microsoft.Azure.Management.ManagedServices.Models;
+
+namespace Microsoft.Azure.PowerShell.Cmdlets.ManagedServices
```
markcowl: Note that this common library:  https://github.com/Azure/azure-powershell-common/blob/master/src/ResourceManager/Version2016_09_01/Utilities/Models/ResourceIdentifier.cs  allows you to split a resource by components

src/ManagedServices/ManagedServices/Microsoft.Azure.PowerShell.Cmdlets.ManagedServices.generated.format.ps1xml
```diff
@@ -0,0 +1,85 @@
...
+        <TableRowEntries>
+          <TableRowEntry>
+            <TableColumnItems>
+              <TableColumnItem>
+                <Alignment>Left</Alignment>
+                <PropertyName>Name</PropertyName>
+              </TableColumnItem>
+              <TableColumnItem>
+                <Alignment>Left</Alignment>
+                <ScriptBlock>$_.Properties.RegistrationDefinitionId</ScriptBlock>
```
markcowl: nice

src/ManagedServices/ManagedServices/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+//
+//      Major Version
+//      Minor Version
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Build and Revision Numbers
+// by using the '*' as shown below:
+// [assembly: AssemblyVersion("1.0.*")]
+[assembly: AssemblyVersion("1.0.0.0")]
```
markcowl: These should match the actual module version

src/ManagedServices/ManagedServices/help/Az.ManagedServices.md
```diff
@@ -0,0 +1,31 @@
+---
+Module Name: Az.ManagedServices
+Module Guid: {{ Update Module Guid }}
```
markcowl: Actual module Guid

src/ManagedServices/ManagedServices/help/Az.ManagedServices.md
```diff
@@ -0,0 +1,31 @@
...
+Download Help Link: {{ Update Download Link }}
```
markcowl: Please follow the pattern in other module help files for these

src/ManagedServices/ManagedServices/help/Az.ManagedServices.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Az.ManagedServices Cmdlets
+### [Get-AzManagedServicesAssignment](Get-AzManagedServicesAssignment.md)
+Gets a list of the registration assignments.
+
+### [Get-AzManagedServicesDefinition](Get-AzManagedServicesDefinition.md)
+Gets a list of the registration definitions.
+
+### [New-AzManagedServicesAssignment](New-AzManagedServicesAssignment.md)
+Creates or updates a registration assignment.
```
markcowl: Can it actually update?  If so, the verb should be set

src/ManagedServices/ManagedServices/help/Az.ManagedServices.md
```diff
@@ -0,0 +1,31 @@
...
+
+### [New-AzManagedServicesDefinition](New-AzManagedServicesDefinition.md)
+Creates or updates a registration definition.
```
markcowl: same comment

src/ManagedServices/ManagedServices/help/Get-AzManagedServicesAssignment.md
```diff
@@ -0,0 +1,232 @@
...
+
+Required: True
+Position: Named
+Default value: None
+Accept pipeline input: False
+Accept wildcard characters: False
+```
+
+### -ResourceId
+The Registration Assignment Id.
```
markcowl: I think here we need to be a bit more descriptive in disambiguating these two ids - it should be clear which one is a simple Guid, and which one is a resource Id (an example value in this description would likely be in order).

src/ManagedServices/ManagedServices/help/Get-AzManagedServicesDefinition.md
```diff
@@ -0,0 +1,137 @@
...
+
+Required: False
+Position: Named
+Default value: None
+Accept pipeline input: False
+Accept wildcard characters: False
+```
+
+### -Id
+The registration definition identifier.
```
markcowl: Again, we should add some additional info and a value example here in the description to make the difference between these clear.

When using Get-Help, examples are not shown unless the user passes `-Full` or `-Examples`, so it's important that these parameter descriptions are descriptive enough

src/ManagedServices/ManagedServices/help/New-AzManagedServicesAssignment.md
```diff
@@ -0,0 +1,206 @@
...
+Aliases: ResourceId
+
+Required: True
+Position: Named
+Default value: None
+Accept pipeline input: True (ByPropertyName)
+Accept wildcard characters: False
+```
+
+### -RegistrationDefinitionName
```
markcowl: Per the comment above, we should be consistent in out use of 'Name' and 'Id'

src/ManagedServices/ManagedServices/help/Remove-AzManagedServicesAssignment.md
```diff
@@ -0,0 +1,206 @@
...
+
+Required: False
+Position: Named
+Default value: None
+Accept pipeline input: False
+Accept wildcard characters: False
+```
+
+### -Id
+The registration assignment Id.
```
markcowl: Same comment - whenever we have 'Id' we should make it absolutely clear whether it is a GUID or a resource identifier

src/ManagedServices/ManagedServices/Commands/NewAzureRmManagedServicesAssignment.cs
rajivnandivada: It is a ResourceId. renamed it accordingly.

src/ManagedServices/ManagedServices/Commands/RemoveAzureRmManagedServcicesAssignment.cs
rajivnandivada: If the Scope is not provided, we default to /subs/id but assignment could have been created under /subs/id/rgs/myRG. If Scope is not provided as a parameter we will never to able to delete the assignment.

src/ManagedServices/ManagedServices/Extensions/ManagedServicesExtensions.cs
rajivnandivada: I tried to use that one but it doesnt work when the resourceid doesnt contain resourcegroups

src/ManagedServices/ManagedServices/help/Az.ManagedServices.md
rajivnandivada: Module Name: Az.ManagedServices
Module Guid: d2a8f744-8dcb-4a13-ba80-eb181ff365ad
Download Help Link: https://docs.microsoft.com/en-us/powershell/module/az.managedservices
Help Version: 0.0.1
Locale: en-US

src/ManagedServices/ManagedServices/help/Az.ManagedServices.md
rajivnandivada: No it doesnt. updated the text accordingly.

src/ManagedServices/ManagedServices/help/Get-AzManagedServicesAssignment.md
rajivnandivada: updated to The Registration Assignment ResourceId. And added examples.

rajivnandivada:
@markcowl Thanks for the quick review. I have addressed all your comments. Please check when you get a chance. thank you.

src/ManagedServices/ManagedServices/help/Get-AzManagedServicesDefinition.md
```diff
@@ -0,0 +1,142 @@
...
+### Example
```
markcowl: I don't think this will work when generatign help - instead, you should just provide this in text:

The registration Identifier  (for example `<guid>`)

src/ManagedServices/ManagedServices/help/Get-AzManagedServicesDefinition.md
```diff
@@ -0,0 +1,142 @@
...
+Required: True
+Position: Named
+Default value: None
+Accept pipeline input: False
+Accept wildcard characters: False
+```
+
+### -ResourceId
+The fully qualified resource id of the registration definition.
+### Example
```
markcowl: same comment - and this applies through the markdown help files

rajivnandivada:
> Just one repeated issue - the markdown format has to be sued to generate xml files, and can be a bit sensitive, so please don't add the extra header for a parameter value example, just provide the example value in the inline description text

I fixed this issue.

PR #9250, opened 2019-05-20
Update Network SDK to use API version 2019-04-01

anton-evseev:
## Description

* Update Network SDK to use API version 2019-04-01
* Change type of `PSRouteFilter` in peerings to account for changes in SDK. `PSResourceId` is one of base classes for `PSRouteFilter` so it shouldn't be a breaking change.

## Checklist


number213:
@cormacpayne please take a look

src/Network/Network.Test/ScenarioTests/ExpressRoutePortsLocationTests.cs
```diff
@@ -25,7 +25,7 @@ public ExpressRoutePortsLocationTests(Xunit.Abstractions.ITestOutputHelper outpu
         {
         }
 
-        [Fact]
+        [Fact(Skip="No bandwidth available yet")]
```
cormacpayne: @number213 same comment here

src/Network/Network/Models/PSPeering.cs
```diff
@@ -74,7 +74,7 @@ public class PSPeering : PSChildResource
         public string LastModifiedBy { get; set; }
 
         [JsonProperty(Order = 1)]
-        public PSRouteFilter RouteFilter { get; set; }
+        public PSResourceId RouteFilter { get; set; }
```
cormacpayne: @number213 this wouldn't be a breaking change if the type was being changed from `PSResourceId` to `PSRouteFilter` (since all of the properties accessible in `PSResourceId` would be accessible in `PSRouteFilter`), but since we're doing the opposite, there are now properties in `PSRouteFilter` (such as `Rules`, `Peerings`) that are no longer accessible by the user since the type is now `PSResourceId`, so this is a breaking change. Is there any reason that this change needs to be made?

number213: @cormacpayne

>  Is there any reason that this change needs to be made?

this reflects recent bugfix in Swagger. Even though `routeFilter` property was defined as a whole `RouteFilter` resource, the server always returned it as a simple id-reference (`{ "id": "..." }`) and it doesn't support `$expand`. Type of `routeFilter` property was changed in Swagger to id-reference, which also causes a change in .NET SDK. I've updated model in PowerShell to resolve error with SDK<->Model mapping.

> there are now properties in PSRouteFilter (such as Rules, Peerings) that are no longer accessible by the user

Because the server always returned an id-object, `PSRouteFilter` on `PSPeering` object didn't have any properties defined (except `Id`). Would that be considered breaking change?

```powershell
PS C:\Users\v-anevse> $circuit = Get-AzExpressRouteCircuit -ResourceGroupName "v-anevse" -Name "er-rf"
PS C:\Users\v-anevse> $circuit.Peerings[0].RouteFilter


Name              :
ResourceGroupName :
Location          :
Id                : /subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxx/resourceGroups/v-anevse/providers/Microsoft.Network/routeFilters/filter-name
Etag              :
ProvisioningState :
Tags              :
Rules             : []
Peerings          : []
```

src/Network/Network.Test/ScenarioTests/ExpressRoutePortTests.cs
```diff
@@ -25,7 +25,7 @@ public ExpressRoutePortTests(Xunit.Abstractions.ITestOutputHelper output)
         {
         }
 
-        [Fact]
+        [Fact(Skip="No bandwidth available yet")]
```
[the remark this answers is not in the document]
number213: @cormacpayne Tests for `ExpressRoutePorts` are failing because no bandwidth is available from provider in tests. I'm trying to reach out to the person who worked on them so they could take a look.

Is it possible to merge this PR with two skipped tests to unblock other PRs? We'll make sure they are re-enabled before merging into `master`

Issue #9225, opened 2019-05-16
Search-AzGraph is not correctly formatting output


artisticcheese:
## Description
When making query using Search-AzGraph complex object are not correctly output

## Steps to reproduce

```powershell

$searchResults = Search-AzGraph -Query "where type =~ 'Microsoft.Compute/virtualmachines' |
extend disk = properties.storageProfile.osDisk.managedDisk |
extend datadisk = properties.storageProfile.dataDisks
|project name,  disk, datadisk" 
$searchResults | select -first 1 | fl

```

## Environment data

```

Name                           Value                                                                                                                                                                          
----                           -----                                                                                                                                                                          
PSVersion                      5.1.17763.316                                                                                                                                                                  
PSEdition                      Desktop                                                                                                                                                                        
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                                                                                                        
BuildVersion                   10.0.17763.316                                                                                                                                                                 
CLRVersion                     4.0.30319.42000                                                                                                                                                                
WSManStackVersion              3.0                                                                                                                                                                            
PSRemotingProtocolVersion      2.3                                                                                                                                                                            
SerializationVersion           1.1.0.1                                                                                                                                                                        



```

## Module versions

```powershell


    Directory: C:\Program Files\WindowsPowerShell\Modules


ModuleType Version    Name                                ExportedCommands                                                                                                                                    
---------- -------    ----                                ----------------                                                                                                                                    
Script     1.5.2      Az.Accounts                         {Disable-AzDataCollection, Disable-AzContextAutosave, Enable-AzDataCollection, Enable-AzContextAutosave...}                                         
Script     1.3.1      Az.Accounts                         {Disable-AzDataCollection, Disable-AzContextAutosave, Enable-AzDataCollection, Enable-AzContextAutosave...}                                         
Script     1.0.1      Az.Aks                              {Get-AzAks, New-AzAks, Remove-AzAks, Import-AzAksCredential...}                                                                                     
Script     1.1.0      Az.AnalysisServices                 {Resume-AzAnalysisServicesServer, Suspend-AzAnalysisServicesServer, Get-AzAnalysisServicesServer, Remove-AzAnalysisServicesServer...}               
Script     1.0.2      Az.AnalysisServices                 {Resume-AzAnalysisServicesServer, Suspend-AzAnalysisServicesServer, Get-AzAnalysisServicesServer, Remove-AzAnalysisServicesServer...}               
[...]

PR #9277, opened 2019-05-23
Resource Graph: Normalize top level arrays as PS arrays

ilidemi:
## Description

This fix addresses #9225.

When the client prepares the dynamic server response for output, it recursively transforms complex objects to PSCustomObjects. The issue was that the transformation is only invoked for JObjects but not JArrays, so top-level array fields get returned as JArrays. Transformation logic already handles arrays nested in objects so the fix is very simple.

## Checklist


src/ResourceGraph/ResourceGraph/Az.ResourceGraph.psd1
```diff
@@ -12,7 +12,7 @@
 # RootModule = ''
 
 # Version number of this module.
-ModuleVersion = '0.7.1'
+ModuleVersion = '0.7.2'
```
cormacpayne: @demyanenko you can revert this change -- we'll version this module and publish it during our next release

src/ResourceGraph/ResourceGraph/ChangeLog.md
```diff
@@ -17,4 +17,10 @@
     * Overview of change #1
         - Additional information about change #1
 -->
-## Upcoming Release
\ No newline at end of file
+## Upcoming Release
+
+## Version 0.7.2
```
cormacpayne: @demyanenko you can remove this `## Version 0.7.2` header since we'll do the versioning later. The line below will go under the `## Upcoming Release` header. The content you have below with `## Version 0.7.1` and the line below that can stay.

src/ResourceGraph/ResourceGraph/Properties/AssemblyInfo.cs
```diff
@@ -25,5 +25,5 @@
 [assembly: CLSCompliant(false)]
 [assembly: Guid("cdfd4e1b-bb5f-439b-8843-e6127cd9b8ad")]
 
-[assembly: AssemblyVersion("0.7.1")]
-[assembly: AssemblyFileVersion("0.7.1")]
+[assembly: AssemblyVersion("0.7.2")]
```
cormacpayne: @demyanenko you can revert the changes made to this file since we'll do the versioning later

src/ResourceGraph/ResourceGraph/Utilities/TableExtensions.cs
```diff
@@ -67,7 +67,7 @@ public static IEnumerable<PSObject> ToPsObjects(this Table table)
                 {
                     var value = row[columnIndex];
                     var normalizedValue =
-                        (value as JObject)?.ToPsObject() ?? value;
+                        (value as JToken)?.ToPsObject() ?? value;
```
cormacpayne: @demyanenko if we have examples in the markdown files with the corresponding output from running the cmdlets, we should update the output to reflect the changes being made here

demyanenko: It is a fairly corner case so the markdown examples are not covering it. I appreciate the concern.

PR #9285, opened 2019-05-23
EventGrid: PowerShell work for 2019-06-01 API Version

ahamad-MS:
## Description

Add new cmdlets and update some existing ones to accommodate for the new features GA'ed in 2019-06-01 API Version. These features include: Domain, Advanced Filtering, Pagination and Search Filter, ServiceBus Queue as destination, Removal of "all" for event types.

These features were already reviewed and were in preview in 2018-09-15-preview and 2019-02-01-preview.

- Link to 2019-02-01-preview Review Document: https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/257

- Link to 2019-02-01-preview Review Document: https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/164

## Checklist

    - [X] the changes have gone through a [cmdlet design review](https://github.com/Azure/azure-powershell-cmdlet-review-pr) (Microsoft internal partners only)

### DIFF
--- a/tools/StaticAnalysis/Exceptions/Az.EventGrid/BreakingChangeIssues.csv
+++ b/tools/StaticAnalysis/Exceptions/Az.EventGrid/BreakingChangeIssues.csv
@@ -1,4 +1,3 @@
 "AssemblyFileName","ClassName","Target","Severity","ProblemId","Description","Remediation"
 "Microsoft.Azure.PowerShell.Cmdlets.EventGrid.dll","Microsoft.Azure.Commands.EventGrid.NewAzureEventGridSubscription","New-AzEventGridSubscription","0","1050","The parameter set 'ResourceIdEventSubscriptionParameterSet' for cmdlet 'New-AzEventGridSubscription' has been removed.","Add parameter set 'ResourceIdEventSubscriptionParameterSet' back to cmdlet 'New-AzEventGridSubscription'."
 "Microsoft.Azure.PowerShell.Cmdlets.EventGrid.dll","Microsoft.Azure.Commands.EventGrid.NewAzureEventGridSubscription","New-AzEventGridSubscription","0","1050","The parameter set 'ResourceGroupNameParameterSet' for cmdlet 'New-AzEventGridSubscription' has been removed.","Add parameter set 'ResourceGroupNameParameterSet' back to cmdlet 'New-AzEventGridSubscription'."
-


ahamad-MS:
Hello @MiYanni , it seems you are the Sprint leader for this period according to 
https://github.com/Azure/azure-powershell-pr/wiki/PowerShell-Rotation

Can you please provide some help regarding static analysis failure I am hitting. In specific:
1. It looks like we moved away from Jenkins and hence I am not able to find the location of the artificats and static analysis results  according to this document https://github.com/Azure/azure-powershell/blob/master/documentation/Debugging-StaticAnalysis-Errors.md  .. Maybe it needs to be updated?

2. When I ran msbuild build.proj /t:StaticAnalysis locally, I got the static analysis result under azure-powershell\artifacts\StaticAnalysisResults. However, the errors I am getting are wrong.

For example, The following signature error is reported (Parameter set 'DomainEventSubscriptionParameterSet' of cmdlet 'New-AzEventGridSubscription' contains at least one parameter with a position larger than four, which is discouraged.) . However, 'DomainEventSubscriptionParameterSet' does not have more than 4 posistions. Same for the other breaking change errors. 

Thanks

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomainKeys.cs
```diff
@@ -0,0 +1,94 @@
...
+using Microsoft.Azure.Management.EventGrid.Models;
+using Microsoft.Azure.Commands.ResourceManager.Common.ArgumentCompleters;
+
+namespace Microsoft.Azure.Commands.EventGrid
+{
+    [Cmdlet(
+        VerbsCommon.Get,
+        ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "EventGridDomainKey",
+        DefaultParameterSetName = DomainNameParameterSet),
+    OutputType(typeof(DomainSharedAccessKeys))]
```
cormacpayne: @ahamad-MS it looks like this is a type from the SDK -- we should ensure that we're not exposing any types from the SDK as a parameter or output. We should wrap this type in a custom `PSDomainSharedAccessKeys` class (or however you'd like to name it) to help mitigate any breaking changes that may arise in the future in the SDK.

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomainKeys.cs
```diff
@@ -0,0 +1,94 @@
...
+
+    public class GetAzureEventGridDomainKeys : AzureEventGridCmdletBase
+    {
+        [Parameter(
+            Mandatory = true,
+            ValueFromPipelineByPropertyName = true,
```
cormacpayne: @ahamad-MS same comment about removing `ValueFromPipelineByProperty` from all parameters in this cmdlet other than `-ResourceId` (and `-DomainObject` should keep `ValueFromPipeline`)

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomainKeys.cs
```diff
@@ -0,0 +1,94 @@
...
+        public string ResourceId { get; set; }
+
+        public override void ExecuteCmdlet()
+        {
+            string resourceGroupName;
+            string domainName;
+
+            if (!string.IsNullOrEmpty(this.ResourceId))
+            {
+                EventGridUtils.GetResourceGroupNameAndDomainName(this.ResourceId, out resourceGroupName, out domainName);
```
cormacpayne: @ahamad-MS same comment about changing these if-statements:

```cs
if (!string.IsNullOrEmpty(this.ResourceId))
{
    EventGridUtils.GetResourceGroupNameAndDomainName(this.ResourceId, out this.ResourceGroupName, out this.Name);
}
else if (this.DomainObject != null)
{
    this.ResourceGroupName = this.DomainObject.ResourceGroupName;
    this.Name = this.DomainObject.DomainName;
}
```

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomainTopic.cs
```diff
@@ -0,0 +1,168 @@
...
+        ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "EventGridDomainTopic",
+        DefaultParameterSetName = DomainTopicNameParameterSet),
+    OutputType(typeof(PSDomainTopic),
+    typeof(PSDomainTopicListInstance))]
+
+    public class GetAzureEventGridDomainTopic : AzureEventGridCmdletBase
+    {
+        [Parameter(
+            Mandatory = true,
+            ValueFromPipelineByPropertyName = true,
```
cormacpayne: @ahamad-MS same comment about `ValueFromPipelineByPropertyName`

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomainTopic.cs
```diff
@@ -0,0 +1,168 @@
...
+        [ValidateRange(1, 100)]
+        public int Top { get; set; }
+
+        [Parameter(
+            Mandatory = false,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = EventGridConstants.NextLinkHelp,
+            ParameterSetName = NextLinkParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public string NextLink { get; set; }
```
cormacpayne: @ahamad-MS same question about whether or not we want to expose this to the user or do the paging automatically within the cmdlet

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomainTopic.cs
```diff
@@ -0,0 +1,168 @@
...
+            int? providedTop = null;
+
+            if (MyInvocation.BoundParameters.ContainsKey(nameof(this.Top)))
+            {
+                providedTop = this.Top;
+            }
+
+            if (!string.IsNullOrEmpty(this.ResourceId))
+            {
+                EventGridUtils.GetResourceGroupNameAndDomainNameAndDomainTopicName(this.ResourceId, out resourceGroupName, out domainName, out domainTopicName);
```
cormacpayne: @ahamad-MS same comment about changing the variables to the parameters rather than temporary variables and then removing the below if-statement

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomain.cs
```diff
@@ -0,0 +1,190 @@
...
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = EventGridConstants.ODataQueryHelp,
+            ParameterSetName = ResourceGroupNameParameterSet)]
+        [Parameter(
+            Mandatory = false,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = EventGridConstants.ODataQueryHelp,
+            ParameterSetName = ResourceIdEventSubscriptionParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public string ODataQuery { get; set; }
```
cormacpayne: @ahamad-MS do we want to expose this parameter to the user, or introduce new parameters that map to the properties that the user would be providing for `-ODataQuery`?

For example, if the filter supported filtering by some property `Foo`, rather than having the user provide a value to `-ODataQuery` that's `$filter=foo eq 'helloworld'`, should we introduce a `-Foo` parameter that we use to create a filter for the user? We follow this pattern in a few other cmdlets, so I can provide some examples if needed.

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomainTopic.cs
```diff
@@ -0,0 +1,168 @@
...
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = EventGridConstants.ODataQueryHelp,
+            ParameterSetName = DomainTopicNameParameterSet)]
+        [Parameter(
+            Mandatory = false,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = EventGridConstants.ODataQueryHelp,
+            ParameterSetName = ResourceIdDomainTopicParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public string ODataQuery { get; set; }
```
cormacpayne: @ahamad-MS same comment about if we want to expose this to users or implement custom filter parameters

src/EventGrid/EventGrid/Domain/NewAzureEventGridDomainKey.cs
```diff
@@ -0,0 +1,118 @@
...
+            HelpMessage = EventGridConstants.KeyNameHelp,
+            ParameterSetName = DomainInputObjectParameterSet)]
+        [Parameter(
+            Mandatory = true,
+            ValueFromPipelineByPropertyName = true,
+            Position = 1,
+            HelpMessage = EventGridConstants.KeyNameHelp,
+            ParameterSetName = ResourceIdEventSubscriptionParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public string KeyName { get; set; }
```
cormacpayne: @ahamad-MS this can be renamed to just `-Name` (you can add an alias to `-KeyName` if you'd like)

src/EventGrid/EventGrid/Domain/NewAzureEventGridDomainKey.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        [Parameter(
+            Mandatory = true,
+            ValueFromPipeline = true,
+            Position = 0,
+            HelpMessage = EventGridConstants.DomainInputObjectHelp,
+            ParameterSetName = DomainInputObjectParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public PSDomain InputObject { get; set; }
```
cormacpayne: @ahamad-MS I would rename this to be `-DomainObject`

src/EventGrid/EventGrid/Domain/NewAzureEventGridDomainKey.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        [Parameter(
+            Mandatory = true,
+            ValueFromPipelineByPropertyName = true,
+            Position = 0,
+            HelpMessage = EventGridConstants.DomainResourceIdHelp,
+            ParameterSetName = ResourceIdEventSubscriptionParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public string ResourceId { get; set; }
```
cormacpayne: @ahamad-MS this parameter isn't necessary, but if you'd like to keep it, I would rename it to `-DomainResourceId`

src/EventGrid/EventGrid/Domain/NewAzureEventGridDomainKey.cs
```diff
@@ -0,0 +1,118 @@
...
+        public override void ExecuteCmdlet()
+        {
+            if (this.ShouldProcess(this.DomainName, $"Regenerate key {this.KeyName} for domain {this.DomainName} in Resource Group {this.ResourceGroupName}"))
+            {
+                string resourceGroupName;
+                string domainName;
+
+                if (!string.IsNullOrEmpty(this.ResourceId))
+                {
+                    EventGridUtils.GetResourceGroupNameAndDomainName(this.ResourceId, out resourceGroupName, out domainName);
```
cormacpayne: @ahamad-MS same comment about the temporary variables

src/EventGrid/EventGrid/Domain/NewAzureEventGridDomainKey.cs
```diff
@@ -0,0 +1,118 @@
...
+using Microsoft.Azure.Commands.ResourceManager.Common.ArgumentCompleters;
+
+namespace Microsoft.Azure.Commands.EventGrid
+{
+    [Cmdlet(
+        VerbsCommon.New,
+        ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "EventGridDomainKey",
+        DefaultParameterSetName = DomainNameParameterSet,
+        SupportsShouldProcess = true), 
+    OutputType(typeof(DomainSharedAccessKeys))]
```
cormacpayne: @ahamad-MS same comment about exposing SDK types

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomain.cs
```diff
@@ -0,0 +1,104 @@
...
+        ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "EventGridDomain",
+        DefaultParameterSetName = DomainNameParameterSet,
+        SupportsShouldProcess = true),
+    OutputType(typeof(bool))]
+
+    public class RemoveAzureEventGridDomain : AzureEventGridCmdletBase
+    {
+        [Parameter(
+            Mandatory = true,
+            ValueFromPipelineByPropertyName = true,
```
cormacpayne: @ahamad-MS same comment about `ValueFromPipelineByPropertyName`

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomain.cs
```diff
@@ -0,0 +1,104 @@
...
+                string domainName = string.Empty;
+
+                if (!string.IsNullOrEmpty(this.Name))
+                {
+                    resourceGroupName = this.ResourceGroupName;
+                    domainName = this.Name;
+                }
+                else if (!string.IsNullOrEmpty(this.ResourceId))
+                {
+                    EventGridUtils.GetResourceGroupNameAndDomainName(this.ResourceId, out resourceGroupName, out domainName);
```
cormacpayne: @ahamad-MS same comment about temporary variables

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomainTopic.cs
```diff
@@ -0,0 +1,117 @@
...
+        ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "EventGridDomainTopic",
+        DefaultParameterSetName = DomainTopicNameParameterSet,
+        SupportsShouldProcess = true),
+    OutputType(typeof(bool))]
+
+    public class RemoveAzureEventGridDomainTopic : AzureEventGridCmdletBase
+    {
+        [Parameter(
+            Mandatory = true,
+            ValueFromPipelineByPropertyName = false,
```
cormacpayne: @ahamad-MS same comment about `ValueFromPipelineByPropertyName`

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomain.cs
```diff
@@ -0,0 +1,104 @@
...
+            ParameterSetName = DomainInputObjectParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public PSDomain InputObject { get; set; }
+
+        [Parameter(Mandatory = false)]
+        public SwitchParameter PassThru { get; set; }
+
+        public override void ExecuteCmdlet()
+        {
+            if (this.ShouldProcess(this.Name, $"Remove domain {this.Name} in resource group {this.ResourceGroupName}"))
```
cormacpayne: @ahamad-MS I would move the logic for parsing `this.ResourceGroupName` and `this.Name` to before this if-statement since they won't have a value if the user provides a value for `-ResourceId` or `-InputObject` until after this line is called and displayed to the user

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomainTopic.cs
```diff
@@ -0,0 +1,117 @@
...
+            ParameterSetName = DomainTopicInputObjectParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public PSDomainTopic InputObject { get; set; }
+
+        [Parameter(Mandatory = false)]
+        public SwitchParameter PassThru { get; set; }
+
+        public override void ExecuteCmdlet()
+        {
+            if (this.ShouldProcess(this.Name, $"Remove domain topic {this.Name} under domain {this.DomainName} in resource group {this.ResourceGroupName}"))
```
cormacpayne: @ahamad-MS same comment about parsing the value of the path parameters to before this if-statement

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomainTopic.cs
```diff
@@ -0,0 +1,117 @@
...
+
+                if (!string.IsNullOrEmpty(this.DomainName) && !string.IsNullOrEmpty(this.Name))
+                {
+                    resourceGroupName = this.ResourceGroupName;
+                    domainName = this.DomainName;
+                    domainTopicName = this.Name;
+                }
+                else if (!string.IsNullOrEmpty(this.ResourceId))
+                {
+                    EventGridUtils.GetResourceGroupNameAndDomainNameAndDomainTopicName(this.ResourceId, out resourceGroupName, out domainName, out domainTopicName);
```
cormacpayne: @ahamad-MS same comment about temporary variables

src/EventGrid/EventGrid/EventSubscription/GetAzureEventGridSubscription.cs
```diff
@@ -120,10 +169,101 @@ public class GetAzureRmEventGridSubscription : AzureEventGridCmdletBase
...
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = EventGridConstants.ODataQueryHelp,
+            ParameterSetName = EventSubscriptionDomainInputObjectParameterSet)]
+        [Parameter(
+            Mandatory = false,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = EventGridConstants.ODataQueryHelp,
+            ParameterSetName = EventSubscriptionDomainTopicInputObjectParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public string ODataQuery { get; set; }
```
cormacpayne: @ahamad-MS same comment about this parameter

src/EventGrid/EventGrid/EventSubscription/GetAzureEventGridSubscription.cs
```diff
@@ -120,10 +169,101 @@ public class GetAzureRmEventGridSubscription : AzureEventGridCmdletBase
...
+        [ValidateRange(1, 100)]
+        public int? Top { get; set; }
+
+        [Parameter(
+            Mandatory = false,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = EventGridConstants.NextLinkHelp,
+            ParameterSetName = NextLinkParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public string NextLink { get; set; }
```
cormacpayne: @ahamad-MS same comment about this parameter

src/EventGrid/EventGrid/help/Get-AzEventGridDomain.md
```diff
@@ -0,0 +1,229 @@
...
+If the domain name is provided, the details of a single Event Grid domain is returned.
+If the domain name is not provided, a list of domains is returned. The number of elements returned in this list is controlled by the Top parameter. If the Top value is not specified or $null, the list will contain all the domains items. Otherwise, Top will indicate the maximum number of elements to be returned in the list.
+If more domains are still available, the value in NextLink should be used in the next call to get the next page of domains.
+Finally, ODataQuery parameter is used to perform filtering for the search results. The filtering query follows OData syntax using the Name property only. The supported operations include: CONTAINS, eq (for equal), ne (for not equal), AND, OR and NOT.
+
+## EXAMPLES
+
+### Example 1
+```powershell
+PS C:\> Get-AzEventGridDomain -ResourceGroup MyResourceGroupName -Name Domain1
```
cormacpayne: @ahamad-MS for new examples, please make sure that the corresponding output that the user would see from running the command is provided in the example code block

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomain.cs
```diff
@@ -0,0 +1,190 @@
...
+        ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "EventGridDomain",
+        DefaultParameterSetName = ResourceGroupNameParameterSet),
+    OutputType(typeof(PSDomain),
+    typeof(PSDomainListInstance))]
+
+    public class GetAzureEventGridDomain : AzureEventGridCmdletBase
+    {
+        [Parameter(
+            Mandatory = true,
+            ValueFromPipelineByPropertyName = true,
```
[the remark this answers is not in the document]
ahamad-MS: Fixed.

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomain.cs
```diff
@@ -0,0 +1,190 @@
...
+        [ValidateRange(1, 100)]
+        public int Top { get; set; }
+
+        [Parameter(
+            Mandatory = false,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = EventGridConstants.NextLinkHelp,
+            ParameterSetName = NextLinkParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public string NextLink { get; set; }
```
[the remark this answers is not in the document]
ahamad-MS: We debated both options in the team earlier and we decided to go with this model where we expose both top and nextlink to the user so he is in control on how much he want to process and when. We did not hear any negative feedback from customer during the preview period hence we will keep this pattern which is consistant with some other cmdlets in PowerShell (e.g., storageblob)

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomain.cs
```diff
@@ -0,0 +1,190 @@
...
+            if (MyInvocation.BoundParameters.ContainsKey(nameof(this.Top)))
+            {
+                providedTop = this.Top;
+            }
+
+            if (!string.IsNullOrEmpty(this.ResourceId))
+            {
+                EventGridUtils.GetResourceGroupNameAndDomainName(this.ResourceId, out resourceGroupName, out domainName);
+            }
+            else if (!string.IsNullOrEmpty(this.Name))
```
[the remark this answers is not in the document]
ahamad-MS: Thanks for the suggestion but unfortunatelly this does not build as I am getting the following error when using this.ResourceGroupName and this.Name as out parameters:

"A Property or indexer may not be passed as as an out or ref parameter'. Keeping it as is.

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomainKeys.cs
ahamad-MS: Makes sense. Fixed.

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomainKeys.cs
ahamad-MS: Fixed.

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomainKeys.cs
ahamad-MS: Same as above, we cannot use indexer or property as out parameter. Keeping it as is.

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomainTopic.cs
ahamad-MS: Fixed.

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomainTopic.cs
ahamad-MS: Please refer to my previous answer/comment. No, we need it in this pattern/style.

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomainTopic.cs
ahamad-MS: Same here. Cannot change it as proposed.

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomain.cs
ahamad-MS: Thanks for the suggestion but it is not a preferred method for us as the problem is not with parameters themselves rather than the operators. Currently, only parameter/property we filter on is name. However, we have multiple operations with different layout. For example, 1. Contains(name, 'x') 2. name eq x 3. contains(name, 'xyz') or name ne 'z'

We want user to be free to have as complicated query as possible. We added example and enhanced help text for that. Actually, the feedback we got from users so far on having more operations to be supports rather than parameters. Hence, keeping this as is now.

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomainTopic.cs
ahamad-MS: check above.

src/EventGrid/EventGrid/Domain/NewAzureEventGridDomainKey.cs
ahamad-MS: Changed.

src/EventGrid/EventGrid/Domain/NewAzureEventGridDomainKey.cs
ahamad-MS: I rather keep this as InputObject to be consistent with corresponding topic one unless this is breaking something?

src/EventGrid/EventGrid/EventSubscription/GetAzureEventGridSubscription.cs
ahamad-MS: please check answer above.

src/EventGrid/EventGrid/EventSubscription/GetAzureEventGridSubscription.cs
ahamad-MS: please check answer above.

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomainTopic.cs
ahamad-MS: please check answer above.

src/EventGrid/EventGrid/Domain/NewAzureEventGridDomainKey.cs
ahamad-MS: can you please provide more context here on why you think it is not needed and why to rename it to DomainResourceId. The reason I am asking is because we would like to have same experienece between Topic and Domain

src/EventGrid/EventGrid/Domain/NewAzureEventGridDomainKey.cs
ahamad-MS: Please check asnwer above. thx

src/EventGrid/EventGrid/Domain/NewAzureEventGridDomainKey.cs
ahamad-MS: thx. fixed.

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomainTopic.cs
ahamad-MS: fixed.

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomain.cs
ahamad-MS: Thanks for the suggestion but I am assuming this is along the previous comments where I am not able to use indexer/property as out parameter. Hence, keeping it as is.

src/EventGrid/EventGrid.Test/ScenarioTests/Common.ps1
```diff
@@ -55,7 +64,7 @@ Get Hybrid Connection ResourceId
 #>
 function Get-HybridConnectionResourceId
 {
-    return "/subscriptions/$subscriptionId/resourceGroups/<ResourceGroupName>/providers/Microsoft.Relay/namespaces/<NameSpace>/hybridConnections/<HybridConnectionName>"
+    return "/subscriptions/5b4b650e-28b9-4790-b3ab-ddbd88d727c4/resourceGroups/DevExpRg/providers/Microsoft.Relay/namespaces/DevExpRelayNamespace/hybridConnections/hydbridconnectiondestination"
```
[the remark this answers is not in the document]
ahamad-MS: Actually, we did not change the behavior of the tests as it used to be, namely, we rely on existing resources in order to execute these tests. I realize the value of having creating resources as part of the tests so I will address this one separately in different checkin.

src/EventGrid/EventGrid/help/Get-AzEventGridDomain.md
ahamad-MS: Added output to the new examples/commands md files.

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomain.cs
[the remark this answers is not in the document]
cormacpayne: As an FYI -- there's a [`ResourceIdentifier`](https://github.com/Azure/azure-powershell-common/blob/05d3ed69c8674f750faa8962e8314a83f02cc441/src/ResourceManager/Version2016_09_01/Utilities/Models/ResourceIdentifier.cs) type in our common code that you can use to break down the components of a resource id to get the resource group name, resource type, resource name, parent name (if applicable), etc. You can use that here to avoid the use of `out` variables and fix the consolidation issue. Let me know if this sounds like a reasonable fix.

src/EventGrid/EventGrid/Domain/NewAzureEventGridDomainKey.cs
cormacpayne: Having the parameter named `-InputObject` won't break anything, but it may be confusing to users -- the value for `-InputObject` is usual the representation of the element that the cmdlet is acting upon. For example, a cmdlet called `Remove-AzFoo` would have the user provide the `Foo` object that they want to delete, and a cmdlet called `Update-AzFoo` would have the user provide the `Foo` object that they want to update. We don't advise having an `-InputObject` parameter for `New-*` cmdlets since you wouldn't pass the object representation of the resource you're trying to create (since it would have already been created). In this case specifically, the object that the user would provide is the parent object, which is an acceptable pattern, but we want the name to make this distinction clear. In [this example we created](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/examples/new-cmdlet-example.md#child-resource), you can see we follow a similar pattern by having a `-TopLevelResourceObject` parameter, which would be `-DomainObject` in this case.

Let me know if you have any questions about this pattern.

src/EventGrid/EventGrid/Domain/NewAzureEventGridDomainKey.cs
cormacpayne: The reasoning here is pretty similar to the one outlined above for `-InputObject` vs. `-DomainObject` -- the `-ResourceId` isn't typically included for `New-*` cmdlets (since you shouldn't be able to pass a resource id for a resource you're trying to create). To specify that this is the resource id for the domain you're creating the key for, I would advise renaming this to `-DomainResourceId`.

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomainTopic.cs
cormacpayne: It looks like this wasn't resolved for this file. Would you mind checking?

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomain.cs
cormacpayne: This should be fixed if you use the `ResourceIdentifier` type as mentioned in the initial comment -- I think making at least this change is important since we don't want to display a message to the user that says "Remove domain  in resource group  ".

src/EventGrid/EventGrid/EventSubscription/NewAzureEventGridSubscription.cs
```diff
@@ -194,7 +335,16 @@ public class NewAzureEventGridSubscription : AzureEventGridCmdletBase
         [Parameter(
             Mandatory = false,
             ValueFromPipelineByPropertyName = true,
-            Position = 5,
```
cormacpayne: @ahamad-MS make sure that the `Position` remains the same in the `ResourceGroupNameParameterSet` set

ahamad-MS: Fixed. This was missed somehow. thanks for the catch.

src/EventGrid/EventGrid/Domain/GetAzureEventGridDomain.cs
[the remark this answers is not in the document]
ahamad-MS: Thanks for the suggestion. This will work for the Domain Case but not the domain topic case as the domain topic case is a nest azure resource and not tracked azure resource. therefore, I changed this to use ResourceIdentifier in all Domain related cmdlets.

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomainTopic.cs
ahamad-MS: We cannot use ResourceIdentifier here as this is not a tracked resource. Keeping this as is.

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomainTopic.cs
ahamad-MS: We cannot use ResourceIdentifier here as this is not a tracked resource. Keeping this as is

src/EventGrid/EventGrid/Domain/NewAzureEventGridDomainKey.cs
```diff
@@ -0,0 +1,116 @@
...
+            HelpMessage = EventGridConstants.DomainResourceIdHelp,
+            ParameterSetName = ResourceIdEventSubscriptionParameterSet)]
+        [ValidateNotNullOrEmpty]
+        public string ResourceId { get; set; }
+
+        public override void ExecuteCmdlet()
+        {
+            if (this.ShouldProcess(this.DomainName, $"Regenerate key {this.Name} for domain {this.DomainName} in Resource Group {this.ResourceGroupName}"))
+            {
+                if (!string.IsNullOrEmpty(this.ResourceId))
```
cormacpayne: @ahamad-MS bringing up this comment again about moving this logic of assigning the values of `ResourceGroupName` and `DomainName` to before the `ShouldProcess` block so they have values when the message is displayed to the user when they provide `-WhatIf`

ahamad-MS: Fixed

src/EventGrid/EventGrid/Domain/NewAzureEventGridDomainKey.cs
ahamad-MS: OK. Changed.

src/EventGrid/EventGrid/Domain/NewAzureEventGridDomainKey.cs
ahamad-MS: Changed.

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomainTopic.cs
ahamad-MS: Fixed now.

src/EventGrid/EventGrid/Domain/RemoveAzureEventGridDomain.cs
ahamad-MS: My bad. There are two fixes requested here:

1. Moving the parsing above the if statement ---> this is done now. Somehow, I missed it in previous iteration.

2. Using ResourceIdentified for parsing the domain topic nested resource which we discussed offline after I tried it and proved to require additional changes to remove the provider as well as additional parameters that we don't believe needed. This part is kept as is and we will continue using our custom util for domain topic

src/EventGrid/EventGrid.Test/SessionRecords/Microsoft.Azure.Commands.EventGrid.Tests.ScenarioTests.DomainTests/EventGrid_DomainsGetKey.json
```diff
@@ -378,7 +378,7 @@
           "-1"
         ]
       },
-      "ResponseBody": "{\r\n  \"key1\": \"04c2gmm9FgEzK1Y7CsC28OgkPTaFv0Lb/nVm+JWcSHM=\",\r\n  \"key2\": \"KQ91yZDP1hLFlB1K1RJkdw87rt97Js0GKcg7fWu4A94=\"\r\n}",
+      "ResponseBody": "{\r\n  \"key1\": \"<HIDDEN>\",\r\n  \"key2\": \"<HIDDEN>\"\r\n}",
```
markcowl: So, if these are actual secrets that persist in Azure after this test, then you need to report their discosure on https://reportitnow

If these are just credentials that are used during the test and do not persist, you can just add a CredScan Suppression

ahamad-MS: Thanks for the info @markcowl .. These don't get persisted after the tests as we delete these topics/domains and all their SAS keys are gone. We remove them to please the static analysis but we will address suppression next time.

ahamad-MS:
@markcowl Can you please help with this remaining failures? it seems related to static analysis but I cannot find the failures/files reported under artifacts/StaticAnalysisResults. Also, we already added all related suppressions in previous uiterations.. Thanks

ahamad-MS:
@markcowl  @cormacpayne  All comments are handled and all checks had passed. Can you please complete the merge of this PR? thanks

src/EventGrid/EventGrid.Test/ScenarioTests/Common.ps1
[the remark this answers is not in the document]
ahamad-MS: I ended up doing this change for 95% of the created resources except for the Azure function as it does not seem we have direct/easy way in PS to get Event enabled Function in PS. Further investigation or alternative way needed.

PR #9299, opened 2019-05-24
Add Zones property to nat gateway

khannarheams:
## Checklist


niander:
@khannarheams @number213  Hi! There are a lot of cherry picked changes here. What is exactly the change for this PR? just the zone parameter? Do we need any tests for that? Were the help files updated with the new parameter?

EvgenyAgafonchikov:
Hi @niander, yes just zone, other are cherry-picked to check all CI checks pass.
@khannarheams, please update on the question above about tests.

src/Network/Network/NatGateway/NewAzureRMNatGatewayCommand.cs
```diff
@@ -50,6 +50,11 @@ public partial class NewAzureRmNatGateway : NetworkBaseCmdlet
             HelpMessage = "The idle timeout of the nat gateway.")]
         public int IdleTimeoutInMinutes { get; set; } = 4;
 
+        [Parameter(
+            Mandatory = false,
+            HelpMessage = "A list of availability zones denoting the zone in which Nat Gateway should be deployed.")]
+        public string[] Zone { get; set; }
```
MiYanni: Are availability zones numbers (like ints) or are there zones that have alpha characters? It would make sense to be int if they are only numbers.

EvgenyAgafonchikov: [Per API specs](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/network/resource-manager/Microsoft.Network/stable/2019-04-01/natGateway.json#L400) it is array of strings. Also it is consistent across Network PS: [sample](https://github.com/Azure/azure-powershell/blob/master/src/Network/Network/PublicIpAddress/NewAzurePublicIpAddressCommand.cs#L124)

MiYanni: If availability zones are simply numbers, I'd recommend the cmdlet use numbers, since it correctly specifies the expected input. I will likely make this change for the new Network generated cmdlets too.

MikhailTryakhov:
@khannarhea  can you please discuss requested chagnes?

src/Network/Network/NatGateway/NewAzureRMNatGatewayCommand.cs
number213: Currently there are three zones represented as "1", "2", "3" (they are not Network-specific, [Compute](https://github.com/Azure/azure-powershell/blob/master/src/Compute/Compute/VirtualMachine/Operation/NewAzureVMCommand.cs#L107) uses them too) but there is always a chance more zones are introduced that could use alphabet characters.

khannarheams:
@MikhailTryakhov @MiYanni as mentioned above "string" for zones is the standard method used in network and compute. This is what we use in ARM templates as well.

src/Network/Network/NatGateway/NewAzureRMNatGatewayCommand.cs
MiYanni: @number213 After some discussion, using `string` is fine as long as there is some example documentation showing what values make sense. In the test above, the usage of this parameter is `-Zone 1` and not `-Zone '1'` or `-Zone "1"`. That's what lead to this discussion to begin with. A completer would also be appropriate too, whichever you think is better.

MikhailTryakhov:
@khannarheams can you please help to figure out again :)

number213:
Updated help & test to use `"1"` to avoid confusion, added argument completer.
I've also removed `Zone` parameter from `Set` cmdlet because you can't update zone once the resource is created:

```powershell
PS D:\Projects\azure-powershell\artifacts\Debug\Az.Network> Set-AzNatGateway -ResourceGroupName "v-anevse" -Name "nat_gateway" -Zone "2"
Set-AzNatGateway : Resource /subscriptions/947d47b4-7883-4bb9-9d85-c5e8e2f572ce/resourceGroups/v-anevse/providers/Micro
soft.Network/natGateways/nat_gateway has an existing availability zone constraint 1 and the request has availability
zone constraint 2, which do not match. Zones cannot be added/updated/removed once the resource is created. The
resource cannot be updated from regional to zonal or vice-versa.
StatusCode: 400
ReasonPhrase: Bad Request


PS D:\Projects\azure-powershell\artifacts\Debug\Az.Network> Set-AzNatGateway -ResourceGroupName "v-anevse" -Name "nat_gateway"
Set-AzNatGateway : Resource /subscriptions/947d47b4-7883-4bb9-9d85-c5e8e2f572ce/resourceGroups/v-anevse/providers/Micro
soft.Network/natGateways/nat_gateway has an existing availability zone constraint 1 and the request has availability
zone constraint , which do not match. Zones cannot be added/updated/removed once the resource is created. The resource
cannot be updated from regional to zonal or vice-versa.
StatusCode: 400
ReasonPhrase: Bad Request
```
@khannarheams please confirm

khannarheams:
@number213 thanks! LGTM

number213:
@MiYanni @cormacpayne feedback has been addressed, could you please take a look?

PR #9364, opened 2019-06-05
Fix DateTimeExtraction model - Bug Fix

arolshan:
## Checklist

- [v] I have read the [_Submitting Changes_](https://github.com/Azure/azure-powershell/blob/master/CONTRIBUTING.md#submitting-changes) section of [`CONTRIBUTING.md`](https://github.com/Azure/azure-powershell/blob/master/CONTRIBUTING.md)
- [v] The title of the PR is clear and informative
- [v] The appropriate [change log has been updated](https://github.com/Azure/azure-powershell/blob/master/CONTRIBUTING.md#updating-the-change-log)
- [v] The PR does not introduce [breaking changes](https://github.com/Azure/azure-powershell/blob/master/documentation/breaking-changes/breaking-changes-definition.md)
- [v] If applicable, the changes made in the PR have proper test coverage
- [v] For public API changes to cmdlets:
    - [v] the changes have gone through a [cmdlet design review](https://github.com/Azure/azure-powershell-cmdlet-review-pr) (Microsoft internal partners only)
    - [v] the cmdlet markdown files were [generated using the `platyPS` module](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/help-generation.md)

### DIFF
--- a/src/OperationalInsights/OperationalInsights/Models/DataSourcePropertiesPerKind/PSCustomLogDataSourceProperties.cs
+++ b/src/OperationalInsights/OperationalInsights/Models/DataSourcePropertiesPerKind/PSCustomLogDataSourceProperties.cs
@@ -66,7 +66,7 @@ namespace Microsoft.Azure.Commands.OperationalInsights.Models
         /// Gets or sets the file system locations.
         /// </summary>
         [JsonProperty(PropertyName = "fileSystemLocations")]
-        public FileSystemLocations fileSystemLocations { get; set; }
+        public FileSystemLocations FileSystemLocations { get; set; }
 
         #endregion
     }


src/OperationalInsights/OperationalInsights/Models/DataSourcePropertiesPerKind/PSCustomLogDataSourceProperties.cs
```diff
@@ -163,7 +163,12 @@ public class DateTimeExtraction
         /// <summary>
         /// Gets or sets the regex.
         /// </summary>
-        public string regex { get; set; }
+        public RegexDelimiter[] regex { get; set; }
```
markcowl: I think thsi model will need to be populated by cmdlet code as well

markcowl:
Also, please add a changelog entry under src\Monitor\Monitor for this

src/OperationalInsights/OperationalInsights/Models/DataSourcePropertiesPerKind/PSCustomLogDataSourceProperties.cs
arolshan: Why?
This feature is in deprecation mode in its current form, therefore we don't want to invest too much in this. Currently just fix the bug, we will discuss our PMs if this is needed.

arolshan:
> Also, please add a changelog entry under src\Monitor\Monitor for this

sure

src/OperationalInsights/OperationalInsights/Models/DataSourcePropertiesPerKind/PSCustomLogDataSourceProperties.cs
markcowl: First, it is a breaking change, if this is returned from your cmdlet, which is not allowed.  Second, how is this value actually populated?

markcowl:
@arolshan Please respond - if this is a model that is part of cmdlet output (or input), then removing a property is a breaking change.  Also, it is unclear to me how this property would be populated, can you explain?

If something has changed in the underlying API, then you need to mitigate the break, which would mean keeping the current string property,  adding a new property with the new value, and ensuring that the properties are populated by the cmdlet. Adding a CmdletBreakingChangeAttribute to the model property.

src/OperationalInsights/OperationalInsights/Models/DataSourcePropertiesPerKind/PSCustomLogDataSourceProperties.cs
```diff
@@ -163,7 +163,12 @@ public class DateTimeExtraction
...
+
+        ///<summary>
+        /// Gets or sets the FormatString
+        /// </summary>
+        public string formatString { get; set; }
```
markcowl: Also, why is this camelCase?  The appropriate casing for .Net properties is PascalCase

src/OperationalInsights/OperationalInsights/Models/DataSourcePropertiesPerKind/PSCustomLogDataSourceProperties.cs
arolshan: 1. This is not a breaking change, this is a bug fix. You can see [here the ICM related incident](https://icm.ad.msft.net/imp/v3/incidents/details/121625584/home).  
2. The following object is used in two locations:
a. **In New-AzOperationalInsightsCustomLogDataSource cmdlet:** 
This cmdlet fails because of BadRequest because the backend expects a different contract. After this fix (validated of course) the request succeeds. This is populated by supplying a raw json of the CustomLog model. Example:
```powershell
$customLogRawJson = Get-Content 'C:\Users\arolshan\Desktop\customLogRaw.json' | Out-String
New-AzOperationalInsightsCustomLogDataSource -ResourceGroupName "arolshan-eus-rg" -WorkspaceName "arolshan-eus-ws2" -Name "arolshan_datetime_CL" -CustomLogRawJson $customLogRawJson
```
where $customLogRaw.json holds:
```json
{
    "CustomLogName": "arolshan_datetime_CL",
    "Description": "this is a description",
    "Extractions": [
        {
            "extractionName": "TimeGenerated",
            "extractionProperties": {
                "dateTimeExtraction": {
                    "regex": [
                        {
                            "matchIndex": 0,
                            "numberdGroup": null,
                            "pattern": "((\\d{2})|(\\d{4}))-([0-1]\\d)-(([0-3]\\d)|(\\d))\\s((\\d)|([0-1]\\d)|(2[0-4])):[0-5][0-9]:[0-5][0-9]"
                        }
                    ]
                }
            },
            "extractionType": "DateTime"
        }
    ],
    "Inputs": [
        {
            "location": {
                "fileSystemLocations": {
                    "linuxFileTypeLogPaths": null,
                    "windowsFileTypeLogPaths": [
                        "c:\\tmp\\datetimelog.txt"
                    ]
                }
            },
            "recordDelimiter": {
                "regexDelimiter": {
                    "matchIndex": 0,
                    "numberdGroup": null,
                    "pattern": "(^.*((\\d{2})|(\\d{4}))-([0-1]\\d)-(([0-3]\\d)|(\\d))\\s((\\d)|([0-1]\\d)|(2[0-4])):[0-5][0-9]:[0-5][0-9].*$)"
                }
            }
        }
    ]
}
```


b. **In Get-AzOperationalInsightsDataSource cmdlet:**
This also fails when trying to perform the following:
```powershell
Get-AzOperationalInsightsDataSource -ResourceGroupName "arolshan-eus-rg" -WorkspaceName "arolshan-eus-ws2" -Kind "CustomLog" 
```
on deserialization, because the backend sends a different model than the one the powershell module expects.

src/OperationalInsights/OperationalInsights/Models/DataSourcePropertiesPerKind/PSCustomLogDataSourceProperties.cs
arolshan: aligning with the contract the backend expects.

src/OperationalInsights/OperationalInsights/Models/DataSourcePropertiesPerKind/PSCustomLogDataSourceProperties.cs
markcowl: Removing a property or changing its type from the public API surface is, **by definition** a breaking change.  Is the contention that this never worked, or that the service changed its behavior for this api-version?

You are deciding in this cmdlet to simply serialize or deserialize using this model, that is an implementation choice - it should be possible for you to fix the bug without breaking the public API surface.

Also, if this is  a bug fix, you should add a regression test, and this will also help to demonstrate the actual usage of this model type in the cmdlet.

arolshan: > Removing a property or changing its type from the public API surface is, by definition a breaking change. Is the contention that this never worked, or that the service changed its behavior for this api-version?

Unfortunately this never worked and it took one customer to try it to notice the bug. 

> You are deciding in this cmdlet to simply serialize or deserialize using this model, that is an implementation choice - it should be possible for you to fix the bug without breaking the public API surface.

You are absolutely right, but I don't see the point of allowing backward compatibility for something that doesn't work. 

> Also, if this is a bug fix, you should add a regression test, and this will also help to demonstrate the actual usage of this model type in the cmdlet.

Your'e right, I'll modify current tests to actually test the functionality properly. 

> If something has changed in the underlying API, then you need to mitigate the break, which would mean keeping the current string property, adding a new property with the new value, and ensuring that the properties are populated by the cmdlet. Adding a CmdletBreakingChangeAttribute to the model property.

a. See point I made above. 
b. Sure I'll add CmdletBreakingChangeAttribute to the model.

markcowl: The point is that breaking changes are not allowed.  A customer can stiull write a script that uses this propety, even if it is null.  You need to add a property and not remove this one.  ALso, if it is possible to recreate this string based on the returned values (and looks to me like it might be), you should do that.

src/OperationalInsights/OperationalInsights/Models/DataSourcePropertiesPerKind/PSCustomLogDataSourceProperties.cs
markcowl: This is irrelevant.  You *must* use the appropriate .Net conventions, which means that properties are `PascalCase`

src/OperationalInsights/OperationalInsights/Models/DataSourcePropertiesPerKind/PSCustomLogDataSourceProperties.cs
```diff
@@ -207,4 +216,26 @@ public class CustomLogExtraction
 
         #endregion
     }
+
+    public class RegexDelimiterJsonConverter : JsonConverter
```
markcowl: If this serialization is broken in your swagger spec and/or .Net SDK, you must fix it there, and then move this fix into PowerShell.  Fixing it only in PowerShell leaves it broken in other ploaces that consume this swagger spec.

arolshan: okay

arolshan: its not documented in our swagger. the api expects a generic raw json.

src/OperationalInsights/OperationalInsights/Models/DataSourcePropertiesPerKind/PSCustomLogDataSourceProperties.cs
arolshan: mark, I added a json converter specifically for this case. If a user uses the script using the old property then the json converter will convert it to the right property. please see **RegexDelimiterJsonConverter**

src/OperationalInsights/OperationalInsights/Models/DataSourcePropertiesPerKind/PSCustomLogDataSourceProperties.cs
arolshan: sure, i will allign with .net conventions

src/OperationalInsights/OperationalInsights/Models/DataSourcePropertiesPerKind/PSCustomLogDataSourceProperties.cs
arolshan: I added the old property back. Please see.

PR #9384, opened 2019-06-07
Remove RuleSetVersion validation

meetug:
## Checklist

### DIFF
--- a/src/Network/Network/ChangeLog.md
+++ b/src/Network/Network/ChangeLog.md
@@ -22,6 +22,7 @@
 ## Version 1.9.0
 * Update ResourceId and InputObject for Nat Gateway
     - Add alias for ResourceId and InputObject
+* Removed WAF RuleSetVersion validation
 
 ## Version 1.8.1
 * Add DisableBgpRoutePropagation flag to Effective Route Table output


adxsdkps:

src/Network/Network/ApplicationGateway/FirewallConfiguration/AzureApplicationGatewayWebApplicationFirewallConfigurationBase.cs
```diff
@@ -43,7 +43,6 @@ public class AzureApplicationGatewayWebApplicationFirewallConfigurationBase : Ne
         [Parameter(
                Mandatory = false,
                HelpMessage = "The version of the rule set type.")]
-        [ValidateSet("3.0", "2.2.9")]
```
markcowl: Does it make sense to make this a PSArgumentCompleter, (which would provide tab completion but allow values outside the set) or are there no known values for this?

markcowl:
@MikhailTryakhov @chandrasekarsrinivasan Not familiar with this user, please LMK if this change looks OK

src/Network/Network/ApplicationGateway/FirewallConfiguration/AzureApplicationGatewayWebApplicationFirewallConfigurationBase.cs
meetug: No, because this list is dynamic. We don't want to keep track of PS changes. Also, there is a cmdlet to get all the versions.

meetug:
> Also, please add a Changelog entry for this change

src/Network/Network/ApplicationGateway/FirewallConfiguration/AzureApplicationGatewayWebApplicationFirewallConfigurationBase.cs
markcowl: If there is a cmdlet to get all the versions, there should be a resource completer here that uses that API call.  This is one reason we have design reviews.

PR #9415, opened 2019-06-11
Add support to regenerate policy keys

anusapan:
## Description

Add support to regenerate authorization policy keys of an azure iothub.

## Checklist

### DIFF
--- a/src/IotHub/IotHub/IotHub/AddAzureRmIotHubKey.cs
+++ b/src/IotHub/IotHub/IotHub/AddAzureRmIotHubKey.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Azure.Commands.Management.IotHub
             ValueFromPipelineByPropertyName = true,
             HelpMessage = "IotHub Resource Id")]
         [ValidateNotNullOrEmpty]
-        public string ResourceId { get; set; }
+        public string HubId { get; set; }
 
         [Parameter(
             Position = 0,
@@ -93,8 +93,8 @@ namespace Microsoft.Azure.Commands.Management.IotHub
             {
                 if (ParameterSetName.Equals(ResourceIdParameterSet))
                 {
-                    this.ResourceGroupName = IotHubUtils.GetResourceGroupName(this.ResourceId);
-                    this.Name = IotHubUtils.GetIotHubName(this.ResourceId);
+                    this.ResourceGroupName = IotHubUtils.GetResourceGroupName(this.HubId);
+                    this.Name = IotHubUtils.GetIotHubName(this.HubId);
                 }
 
                 var psAuthRule = new PSSharedAccessSignatureAuthorizationRule()
--- a/src/IotHub/IotHub/IotHub/GetAzureRmIotHubKey.cs
+++ b/src/IotHub/IotHub/IotHub/GetAzureRmIotHubKey.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Azure.Commands.Management.IotHub
             ValueFromPipelineByPropertyName = true,
             HelpMessage = "IotHub Resource Id")]
         [ValidateNotNullOrEmpty]
-        public string ResourceId { get; set; }
+        public string HubId { get; set; }
 
         [Parameter(
             Position = 0,
@@ -75,8 +75,8 @@ namespace Microsoft.Azure.Commands.Management.IotHub
         {
             if (ParameterSetName.Equals(ResourceIdParameterSet))
             {
-                this.ResourceGroupName = IotHubUtils.GetResourceGroupName(this.ResourceId);
-                this.Name = IotHubUtils.GetIotHubName(this.ResourceId);
+                this.ResourceGroupName = IotHubUtils.GetResourceGroupName(this.HubId);
+                this.Name = IotHubUtils.GetIotHubName(this.HubId);
             }
 
             if (KeyName != null)
--- a/src/IotHub/IotHub/IotHub/NewAzureRmIotHubKey.cs
+++ b/src/IotHub/IotHub/IotHub/NewAzureRmIotHubKey.cs
@@ -40,7 +40,7 @@ namespace Microsoft.Azure.Commands.Management.IotHub
             ValueFromPipelineByPropertyName = true,
             HelpMessage = "IotHub Resource Id")]
         [ValidateNotNullOrEmpty]
-        public string ResourceId { get; set; }
+        public string HubId { get; set; }
 
         [Parameter(
             Position = 0,
@@ -95,8 +95,8 @@ namespace Microsoft.Azure.Commands.Management.IotHub
             {
                 if (ParameterSetName.Equals(ResourceIdParameterSet))
                 {
-                    this.ResourceGroupName = IotHubUtils.GetResourceGroupName(this.ResourceId);
-                    this.Name = IotHubUtils.GetIotHubName(this.ResourceId);
+                    this.ResourceGroupName = IotHubUtils.GetResourceGroupName(this.HubId);
+                    this.Name = IotHubUtils.GetIotHubName(this.HubId);
                 }
 
                 var regeneratedAuthRule = new PSSharedAccessSignatureAuthorizationRule();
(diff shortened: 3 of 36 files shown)


cormacpayne:
@anusapan Hey Sapan, has a cmdlet design review been completed for the proposed changes? If not, please open an issue in [this repository](https://github.com/Azure/azure-powershell-cmdlet-review-pr), following the provided template, to begin this process so we can discuss the proposed API changes to your cmdlets.

cormacpayne:
@anusapan ping

anusapan:
@cormacpayne cmdlet design review PR https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/327

src/IotHub/IotHub/IotHub/NewAzureRmIotHubKey.cs
```diff
@@ -0,0 +1,123 @@
...
+    using System.Linq;
+    using System.Management.Automation;
+    using System.Text;
+    using Microsoft.Azure.Commands.Management.IotHub.Common;
+    using Microsoft.Azure.Commands.Management.IotHub.Models;
+    using Microsoft.Azure.Management.IotHub;
+    using Microsoft.Azure.Management.IotHub.Models;
+    using ResourceManager.Common.ArgumentCompleters;
+
+    [Cmdlet("New", ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "IotHubKey", SupportsShouldProcess = true)]
```
cormacpayne: @anusapan I think we should add two additional parameter sets to this cmdlet to allow users to provide the identity values of the key in different forms:

(1) A parameter set with an `-InputObject` parameter that's the same type as the returned object from `Get-AzIotHubKey`. This will allow users to get the key that they want to generate a new value for using `Get-AzIotHubKey` and then pipe it to this cmdlet without needing to provide a value for `-ResourceGroupName`, `-Name`, and `-KeyName`

(2) A parameter set with a `-ResourceId` parameter of type `string`. This will allow users to get a generic resource object from `Get-AzResource` and pipe it to this cmdlet without needing to provide a value for the same three parameters.

Let me know if you have any questions about these scenarios. I can provide example implementations of either of these.

src/IotHub/IotHub/help/New-AzIotHubKey.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+## DESCRIPTION
+Generate an Azure IoT Hub key.
+
+## EXAMPLES
+
+### Example 1 Regenerate primary key
+```
+PS C:\> New-AzIotHubKey -ResourceGroupName "myresourcegroup" -Name "myiothub" -KeyName "testKey" -RenewKey "primary"
```
cormacpayne: @anusapan for each example, please provide the corresponding output that the user would see if they were to run the command

src/IotHub/IotHub/IotHub/NewAzureRmIotHubKey.cs
anusapan: Added ResourceId parameter set and for InputObject parameter set the expected values aren't exposed by the related cmdlet

src/IotHub/IotHub/help/New-AzIotHubKey.md
anusapan: Added output

src/IotHub/IotHub/IotHub/AddAzureRmIotHubKey.cs
```diff
@@ -27,9 +27,22 @@ namespace Microsoft.Azure.Commands.Management.IotHub
     [OutputType(typeof(PSSharedAccessSignatureAuthorizationRule))]
```
cormacpayne: @anusapan you'll need to set the `DefaultParameterSetName` in the above `Cmdlet` attribute since you now have multiple parameter sets for this cmdlet. It should be set to `ResourceParameterSet` since that was the previous default (only) parameter set.

src/IotHub/IotHub/IotHub/GetAzureRmIotHubKey.cs
```diff
@@ -27,9 +27,22 @@ namespace Microsoft.Azure.Commands.Management.IotHub
     [OutputType(typeof(PSSharedAccessSignatureAuthorizationRule))]
```
cormacpayne: @anusapan same comment about adding the `DefaultParameterSetName` above

src/IotHub/IotHub/IotHub/GetAzureRmIotHubKey.cs
```diff
@@ -39,20 +52,26 @@ public class GetAzureRmIotHubKey : IotHubBaseCmdlet
...
             Position = 1,
             Mandatory = true,
+            ParameterSetName = ResourceParameterSet,
             ValueFromPipelineByPropertyName = true,
             HelpMessage = "Name of the Iot Hub")]
         [ValidateNotNullOrEmpty]
         public string Name { get; set; }
 
         [Parameter(
-            Position = 2,
```
cormacpayne: @anusapan removing this `Position` is the correct thing to do when it comes to PowerShell patterns, but removing it now will be a breaking change, so I'd advise reverting this change.

src/IotHub/IotHub/IotHub/GetAzureRmIotHubKey.cs
```diff
@@ -39,20 +52,26 @@ public class GetAzureRmIotHubKey : IotHubBaseCmdlet
...
             HelpMessage = "Name of the Key")]
         [ValidateNotNullOrEmpty]
         public string KeyName { get; set; }
 
         public override void ExecuteCmdlet()
         {
+            if (ParameterSetName.Equals(ResourceIdParameterSet))
+            {
+                this.ResourceGroupName = IotHubUtils.GetResourceGroupName(this.ResourceId);
+                this.Name = IotHubUtils.GetIotHubName(this.ResourceId);
```
cormacpayne: @anusapan is `KeyName` not something that we can pull from the provide resource id?

src/IotHub/IotHub/IotHub/NewAzureRmIotHubKey.cs
```diff
@@ -0,0 +1,155 @@
...
```
cormacpayne: @anusapan same comment about adding a `DefaultParameterSetName` here

src/IotHub/IotHub/IotHub/RemoveAzureRmIotHubKey.cs
```diff
@@ -30,10 +30,22 @@ namespace Microsoft.Azure.Commands.Management.IotHub
     [OutputType(typeof(PSSharedAccessSignatureAuthorizationRule))]
```
cormacpayne: @anusapan same comment about adding a `DefaultParameterSetName` above

anusapan: Done

src/IotHub/IotHub/IotHub/NewAzureRmIotHubKey.cs
anusapan: Done

src/IotHub/IotHub/IotHub/GetAzureRmIotHubKey.cs
anusapan: No, we can't retrieve `KeyName` from resourceId

src/IotHub/IotHub/IotHub/GetAzureRmIotHubKey.cs
anusapan: Agreed... reverted

src/IotHub/IotHub/IotHub/GetAzureRmIotHubKey.cs
anusapan: Done

src/IotHub/IotHub/IotHub/AddAzureRmIotHubKey.cs
anusapan: Done

src/IotHub/IotHub/IotHub/GetAzureRmIotHubKey.cs
cormacpayne: @anusapan if the key isn't a tracked resource and can't be represented by a resource id, then we shouldn't have a `-ResourceId` parameter -- however, if you'd like, since the resource id currently represents the id of the hub, you can rename the parameter to be `-HubId` instead, which would be a better indicator of what the parameter is.

anusapan: Fixed

Issue #8912, opened 2019-04-01
Update Get-AzProfile and Select-AzProfile cmdlets


markcowl:
## Description
- [x] Create protected branch for Profile cmdlets
- [x] Update Get-AzProfile per feedback
  - [x] Return an object instead of a string, which includes profile type and description (this can be boiler plate based on the name)
  - [x] Return all profiles by default
  - [ ] Remove ListAvailable and add ```Selected``` parameter that returns the currently selected profile
  - [ ] Investigate using S.M.A apis instead of PowerShell script for reading profile information
  - [ ] Return prifles for both loaded and PSModulePath cmdlets
- [x] Update Select-AzProfile per feedback
  - [x] Build completer over existing profiles
  - [x] Add PassThru parameter

markcowl:
Fixed

Issue #9276, opened 2019-05-22
Update Az.Accounts for data plane module support


markcowl:
## Description
- [x] Add lookup tables for environemnt endpoint and resource id from input endpoint
- [x] Add handler on UrlCreated event to update Uri with environment-specific Uri
- [x] Add functionality to authentication pipeline that looks up uri-> resource Id and adjusts authentication call accordingly
- [x] Provide pattern for BaseUri PSDefaultParameterValue
- [x] Move changes into generated module dependencies

markcowl:
Fixed

PR #9461, opened 2019-06-19
Data Plane and Environment Support for Generated Code

markcowl:
## Checklist

### DIFF
--- a/src/Automation/Automation/help/New-AzAutomationModule.md
+++ b/src/Automation/Automation/help/New-AzAutomationModule.md
@@ -22,11 +22,11 @@ New-AzAutomationModule [-Name] <String> [-ContentLinkUri] <Uri> [-ResourceGroupN
 The **New-AzAutomationModule** cmdlet imports a module into Azure Automation.
 This command accepts a compressed file that has a .zip file name extension.
 The file contains a folder that includes a file that is one of the following types: 
-- wps_2 module, which has a .psm1 or .dll file name extension 
-- wps_2 module manifest, which has a .psd1 file name extension
+- Windows PowerShell module, which has a .psm1 or .dll file name extension 
+- Windows PowerShell module manifest, which has a .psd1 file name extension
 The name of the .zip file, the name of the folder, and the name of the file in the folder must be the same.
 Specify the .zip file as a URL that the Automation service can access.
-If you import a wps_2 module into Automation by using this cmdlet or the Set-AzAutomationModule cmdlet, the operation is asynchronous.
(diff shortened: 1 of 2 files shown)


markcowl:
Fixes #8912, #9276

src/Accounts/Accounts/CommonModule/GetAzProfile.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            try
+            {
+                var includesModuleNames = this.IsBound(nameof(ModuleName));
+                var isListAvailable = this.IsBound(nameof(ListAvailable));
+                var moduleNames = includesModuleNames ? ModuleName : new string[] { };
+                var profiles = includesModuleNames || isListAvailable
+                    ? GetProfiles(InvokeCommand, isListAvailable, moduleNames)
+                    : new []{ ContextAdapter.Instance.SelectedProfile };
+                if (profiles.Any() && !profiles.All(String.IsNullOrEmpty))
```
cormacpayne: @markcowl for the `!profiles.All(string.IsNullOrEmpty)` check, are we making sure that _at least one_ profile isn't null/empty, or that _all_ profiles aren't null/empty?

If it's the first case (at least one isn't empty), then do we want to write it to the output stream? We could filter out the null/empty values before passing it to the `.Select(...)` below.

If it's the second case (all profiles have a value), then we should change the check to `!profiles.Any(string.IsNullOrEmpty)`.

src/Accounts/Accounts/CommonModule/SelectAzProfile.cs
```diff
@@ -0,0 +1,60 @@
...
+        public string Name { get; set; }
+
+        [Parameter]
+        public SwitchParameter PassThru { get; set; }
+
+        protected override void ProcessRecord()
+        {
+            try
+            {
+                if(this.IsBound(nameof(Name)) && !String.IsNullOrEmpty(Name))
```
cormacpayne: @markcowl nit: the `!string.IsNullOrEmpty(Name)` check isn't needed since the `-Name` parameter has the `ValidateNotNullOrEmpty` check and you're already ensuring that it's bound before getting to this check.

src/Accounts/Accounts/help/Get-AzProfile.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+## DESCRIPTION
+Get the service profiles supported by installed modules.
+
+## EXAMPLES
+
+### Example 1
+```powershell
+PS C:\> Get-AzProfile -Module Az.Compute
```
cormacpayne: @markcowl please add the corresponding output for this example (you'll also want to regenerate the help when removing the `SupportsShouldProcess`)

PR #9462, opened 2019-06-19
[Blueprints] Update Export and Import cmdlets with improvements

filizt:
## Checklist

### DIFF
--- a/src/Blueprint/Blueprint/Az.Blueprint.psd1
+++ b/src/Blueprint/Blueprint/Az.Blueprint.psd1
@@ -12,7 +12,7 @@
 # RootModule = ''
 
 # Version number of this module.
-ModuleVersion = '0.2.1'
+ModuleVersion = '0.2.0'
 
 # Supported PSEditions
 CompatiblePSEditions = 'Core', 'Desktop'
--- a/src/Blueprint/Blueprint/ChangeLog.md
+++ b/src/Blueprint/Blueprint/ChangeLog.md
@@ -18,9 +18,7 @@
         - Additional information about change #1
 -->
 ## Upcoming Release
-
-## Version 0.2.1
-*Bug fixes and improvements
+* Bug fixes and improvements
 
 ## Version 0.2.0
 * Added new cmdlets:
--- a/src/Blueprint/Blueprint/Cmdlets/BlueprintDefinition/BlueprintDefinitionCmdletBase.cs
+++ b/src/Blueprint/Blueprint/Cmdlets/BlueprintDefinition/BlueprintDefinitionCmdletBase.cs
@@ -173,7 +173,10 @@ namespace Microsoft.Azure.Commands.Blueprint.Cmdlets
 
             var artifactsPath = GetValidatedFolderPath(inputPath, artifacts);
 
-            if (artifactsPath == null) return; // if blueprint doesn't contain artifacts don't proceed.
+            if (artifactsPath == null)
+            {
+                return; // if blueprint doesn't contain artifacts don't proceed.
+            }
 
             var artifactFiles = AzureSession.Instance.DataStore.GetFiles(artifactsPath, "*.json", SearchOption.TopDirectoryOnly);
 


src/Blueprint/Blueprint/Az.Blueprint.psd1
```diff
@@ -12,7 +12,7 @@
 # RootModule = ''
 
 # Version number of this module.
-ModuleVersion = '0.2.0'
```
cormacpayne: @filizt you can revert the changes made in this file -- we're version the module before release

src/Blueprint/Blueprint/ChangeLog.md
```diff
@@ -19,6 +19,9 @@
 -->
 ## Upcoming Release
 
+## Version 0.2.1
```
cormacpayne: @filizt you can get rid of this header and put the below line below `## Upcoming Release`

src/Blueprint/Blueprint/Cmdlets/BlueprintDefinition/BlueprintDefinitionCmdletBase.cs
```diff
@@ -149,8 +150,17 @@ protected string CreateFolderIfNotExist(string path, string folderName)
         {
             var folderPath = Path.Combine(path, folderName);
 
-            if (!AzureSession.Instance.DataStore.DirectoryExists(folderPath))
+            if (AzureSession.Instance.DataStore.DirectoryExists(folderPath))
             {
+                AzureSession.Instance.DataStore.EmptyDirectory(folderPath);
```
cormacpayne: @filizt is there any reason we want to delete all of the files in this user's folder?

src/Blueprint/Blueprint/Cmdlets/BlueprintDefinition/BlueprintDefinitionCmdletBase.cs
```diff
@@ -163,11 +173,9 @@ protected void ImportArtifacts(string blueprintName, string scope, string inputP
 
             var artifactsPath = GetValidatedFolderPath(inputPath, artifacts);
 
-            var artifactFiles = AzureSession.Instance.DataStore.GetFiles(artifactsPath, "*.json", SearchOption.TopDirectoryOnly);
+            if (artifactsPath == null) return; // if blueprint doesn't contain artifacts don't proceed.
```
cormacpayne: @filizt nit:

```cs
if (artifactsPath == null)
{
    return;
}
```

src/Blueprint/Blueprint/Cmdlets/BlueprintDefinition/ExportAzureRMBlueprint.cs
```diff
@@ -48,38 +49,45 @@ public class ExportAzureRmBlueprint : BlueprintDefinitionCmdletBase
...
+                foreach (var artifact in artifacts)
+                {
+                    string serializedArtifact =
+                        BlueprintClient.GetBlueprintArtifactJsonFromObject(Blueprint.Scope, Blueprint.Name, artifact,
+                            Version);
 
-                var artifactFilePath = Path.Combine(artifactsPath, artifact.Name + ".json");
+                    var artifactFilePath = Path.Combine(artifactsPath, artifact.Name + ".json");
 
-                this.ConfirmAction(
```
cormacpayne: @filizt do we not need this check `ConfirmAction` check anymore?

src/Blueprint/Blueprint/Cmdlets/BlueprintDefinition/BlueprintDefinitionCmdletBase.cs
filizt: We wanted to make sure the the latest exported blueprint and its artifacts are the only blueprint and artifacts in the folder. So, if the user exported a blueprint and then made some changes(say deleted artifacts) through the portal and exported again, in the current implementation the export operation will keep the "deleted" artifact in the Artifacts folder since we are not doing any diffing between the local copy and the portal version of the blueprint. So, we thought we could delete the files in this case and re-create the blueprint and its artifact in this folder. That being said, if there is a way to avoid deleting files this way and still satisfy the above case, it'd be good to know. 

I guess we can always just delete the "Artifacts" folder, as a potentially less impactful option.

src/Blueprint/Blueprint/Cmdlets/BlueprintDefinition/ExportAzureRMBlueprint.cs
filizt: We don't since the "Artifacts" folder would have been deleted in the first blueprint creation step, which is the first confirm action, should it already existed locally.

src/Blueprint/Blueprint/Cmdlets/BlueprintDefinition/BlueprintDefinitionCmdletBase.cs
filizt: We also prompt the user of the folder deletion in the caller of the CreateFolderIfNotExist function here: https://github.com/Azure/azure-powershell/blob/7c8d702511580458b22c28506b4e7f18110155fa/src/Blueprint/Blueprint/Cmdlets/BlueprintDefinition/ExportAzureRMBlueprint.cs#L62

PR #9496, opened 2019-06-25
Update WebSites SDK, add Location parameter to Deleted Web Apps, fix clone slots bug

nking-1:
## Description

The user-facing changes in this pull request are a new Location parameter for the deleted web apps cmdlets, and an internal bug fix for cloning slots. The implementation for the deleted web apps cmdlets has also been updated to use a new location-based API. We are sharding our global endpoints to improve fault-tolerance and reliability, and as a result some APIs have been modified to require location information.

In order to make these changes, the WebSites SDK was updated. Likewise all tests have been recorded again, and some of the test code has been rewritten to be more reliable and not rely on pre-allocated resources.

## Checklist

### DIFF
--- a/src/Websites/Websites.Test/ScenarioTests/WebAppSlotTests.ps1
+++ b/src/Websites/Websites.Test/ScenarioTests/WebAppSlotTests.ps1
@@ -131,37 +131,31 @@ function Test-StartStopRestartWebAppSlot
 		$slot = $slot | Stop-AzWebAppSlot
 
 		Assert-AreEqual "Stopped" $slot.State
-		#$ping = PingWebApp $slot
 
 		# Start web app
 		$slot = $slot | Start-AzWebAppSlot
 
 		Assert-AreEqual "Running" $slot.State
-		#$ping = PingWebApp $slot
 
 		# Stop web app
 		$slot = Stop-AzWebAppSlot -ResourceGroupName $rgname -Name $appname -Slot $slotname
 
 		Assert-AreEqual "Stopped" $slot.State
-		#$ping = PingWebApp $slot
 
 		# Start web app
 		$slot = Start-AzWebAppSlot -ResourceGroupName $rgname -Name $appname -Slot $slotname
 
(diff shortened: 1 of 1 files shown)


Nking92:
This is a large change. It's fine if it's merged in the next sprint. No need to rush this for the upcoming release.

src/Websites/Websites/Utilities/WebsitesClient.cs
```diff
@@ -780,8 +781,20 @@ public IList<Snapshot> GetSiteSnapshots(string resourceGroupName, string webSite
             }
         }
 
-        public IList<DeletedSite> GetDeletedSites()
+        public IList<DeletedSite> GetDeletedSitesFromLocations(IEnumerable<string> locations)
         {
+            List<Task<IPage<DeletedSite>>> tasks = new List<Task<IPage<DeletedSite>>>();
+            foreach (string location in locations)
+            {
+                tasks.Add(WrappedWebsitesClient.DeletedWebApps.ListByLocationAsync(location));
```
cormacpayne: @Nking92 should we follow the `nextLink` returned from each of these operations, or are we only concerned about getting the first page for each location?

src/Websites/Websites.Test/ScenarioTests/WebAppSlotTests.ps1
```diff
@@ -131,37 +131,37 @@ function Test-StartStopRestartWebAppSlot
 		$slot = $slot | Stop-AzWebAppSlot
 
 		Assert-AreEqual "Stopped" $slot.State
-		$ping = PingWebApp $slot
+		#$ping = PingWebApp $slot
```
cormacpayne: @Nking92 nit: for these lines you're commenting out, would you mind just removing them?

src/Websites/Websites/Utilities/WebsitesClient.cs
Nking92: Good point, but internally we don't page the response, so this code should technically be ok for now. The first page will have all the sites.

PR #9506, opened 2019-06-26
Update Policy API versions to 2019-01-01

RamyasreeChakka:
Update policy api versions to use latest version '2019-01-01'


pilor:
Update the changelog in src/ResourceManager/Resources/Commands.Resources/ChangeLog.md

pilor:
Help files for New/Set definition need to be updated (src/ResourceManager/Resources/Commands.Resources/help/New-AzureRmPolicyDefinition.md)

pilor:
Should add one of the new modes to one of the unit tests.  You'll also need to re-record all the tests with the new api-version

src/Resources/ResourceManager/Implementation/Policy/NewAzurePolicyDefinition.cs
```diff
@@ -75,7 +75,6 @@ public class NewAzurePolicyDefinitionCmdlet : PolicyCmdletBase
         /// Gets or sets the new policy definition mode parameter.
         /// </summary>
         [Parameter(Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = PolicyHelpStrings.NewPolicyDefinitionModeHelp)]
-        [PSArgumentCompleter(PolicyDefinitionMode.All, PolicyDefinitionMode.Indexed)]
```
markcowl: You cannot remove the argument completer for this.  Are these values still allowed?  If not, this is a breaking change.  If there are additional known allowed values, please add them.  PSArgumentCompleter just allows the user to tab through known values, but the user can also provide values outside this set.

src/Resources/ResourceManager/Implementation/Policy/SetAzurePolicyDefinition.cs
```diff
@@ -85,7 +85,6 @@ public class SetAzurePolicyDefinitionCmdlet : PolicyCmdletBase
         /// Gets or sets the policy definition mode parameter.
         /// </summary>
         [Parameter(Mandatory = false, ValueFromPipelineByPropertyName = true, HelpMessage = PolicyHelpStrings.NewPolicyDefinitionModeHelp)]
-        [PSArgumentCompleter(PolicyDefinitionMode.All, PolicyDefinitionMode.Indexed)]
```
markcowl: You cannot remove the argument completer for this.  Are these values still allowed?  If not, this is a breaking change.  If there are additional known allowed values, please add them.  PSArgumentCompleter just allows the user to tab through known values, but the user can also provide values outside this set.

src/Resources/ResourceManager/Components/Constants.cs
```diff
@@ -77,17 +77,17 @@ public static class Constants
         /// <summary>
         /// The default policy definition API version.
         /// </summary>
-        public static readonly string PolicyDefinitionApiVersion = "2018-05-01";
+        public static readonly string PolicyDefinitionApiVersion = "2019-01-01";
```
markcowl: How does this not impact any tests?  Are there ary tests for Policy definition, and if so, why does this not invalidate those recordings?

src/Resources/ResourceManager/Components/Constants.cs
```diff
@@ -77,17 +77,17 @@ public static class Constants
         /// <summary>
         /// The default policy definition API version.
         /// </summary>
-        public static readonly string PolicyDefinitionApiVersion = "2018-05-01";
```
markcowl: Need to add a description of this change in the module Changelog

markcowl:
@RamyasreeChakka Please respond to review comments.  This PR will be closed next week if there is no further response.

src/Resources/ResourceManager/Implementation/Policy/SetAzurePolicyDefinition.cs
RamyasreeChakka: Thanks, will remove argument completer as the current values are still valid.

src/Resources/ResourceManager/Components/Constants.cs
RamyasreeChakka: I think the tests didn't break as the schema of the API didn't change with new 2019-01-01 version. We have exposed new api version to enable data plane policies and to protect the existing clients from querying data plane policies. I am updating the recordings.

RamyasreeChakka:
> @RamyasreeChakka Please respond to review comments.  This PR will be closed next week if there is no further response.

src/Resources/ResourceManager/Implementation/Policy/NewAzurePolicyDefinition.cs
RamyasreeChakka: Thanks, will remove argument completer as the current values are still valid.

src/Resources/Resources/ChangeLog.md
```diff
@@ -19,6 +19,7 @@
 -->
 ## Upcoming Release
 - Remove missing cmdlet referenced in `New-AzResourceGroupDeployment` documentation
```
RamyasreeChakka: @markcowl Should I need to move the existing content in Upcoming release to Version 1.7.0? Or is it something that PowerShell team update after updating the PS modules?

markcowl: Just puttign it in Upcoming Release is correct - we have automated versioning tools that update the version based on semver.

PR #9597, opened 2019-07-11
Use the NextLink when fetching alerts, until maximum of 1500 alerts

yibirnba:
## Description

When fetching alerts , use the NextLink to fetch more alerts if exist, until maximum of 1500 alerts

## Checklist

### DIFF
--- a/src/Security/Security/ChangeLog.md
+++ b/src/Security/Security/ChangeLog.md
@@ -19,6 +19,9 @@
 -->
 ## Upcoming Release
 
+## Version 0.7.5
+* Use NextLink for return maximum of 1500 security alerts in command Get-AzSecurityAlert
+
 ## Version 0.7.4
 * Deprecate ResourceGroupName parameter in commands: Get-AzSecurityPricing and Set-AzSecurityPricing
 


adxsdkps:

src/Security/Security/Cmdlets/Alerts/GetAlerts.cs
```diff
@@ -26,6 +27,8 @@ namespace Microsoft.Azure.Commands.Security.Cmdlets.Alerts
     [Cmdlet(VerbsCommon.Get, ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "SecurityAlert", DefaultParameterSetName = ParameterSetNames.SubscriptionScope), OutputType(typeof(PSSecurityAlert))]
     public class GetAlerts : SecurityCenterCmdletBase
     {
+        private const int MaxAlertsToFetch = 1500;
```
markcowl: Instead of setting an arbitrary maximum, you should follow the SupportsPaging pattern, so the user can specify where the list should start and end.  Youc an see an example of thsi pattern here:  https://github.com/Azure/azure-powershell/blob/master/src/Resources/Resources/ActiveDirectory/Cmdlets/GetAzureADUserCommand.cs

src/Security/Security/Cmdlets/Alerts/GetAlerts.cs
```diff
@@ -12,6 +12,7 @@
 // limitations under the License.
 // ------------------------------------
 
+using System;
```
markcowl: You need to add an entry to the Az.Security Changelog for this change

src/Security/Security/Cmdlets/Alerts/GetAlerts.cs
yibirnba: The reason i'm using this maximum number is that the RP does not support the Skip and First parameters

src/Security/Security/Cmdlets/Alerts/GetAlerts.cs
yibirnba: Done

src/Security/Security/Cmdlets/Alerts/GetAlerts.cs
orparnes: @markcowl, this is our long-term plan, across all Microsoft.Security PS API.

This change is just a temporally one, as we have customers complaining about this specific API and the better solution is much more expansive for us for now.

markcowl: You can support paging without support for First and Skip in your API, there are examples of this in the repo.  Since this is a preview module however, the quick fix is OK for now.  You will need to fix this before inclusion in the 'Az' module

PR #9627, opened 2019-07-12
typo: usein -> use in

nschonni:


adxsdkps:

src/Compute/Compute/help/New-AzVmss.md
```diff
@@ -303,7 +303,7 @@ Accept wildcard characters: False
 ```
 
 ### -FrontendPoolName
-The name of the frontend address pool to usein the Scale Set locad balancer.  If no value is supplied, a new Frontend Address Pool will be created, with the same name as the scale set.
+The name of the frontend address pool to use in the Scale Set load balancer.  If no value is supplied, a new Frontend Address Pool will be created, with the same name as the scale set.
```
markcowl: Requires a change to the Compute changelog

PR #9634, opened 2019-07-12
Merge with private repo

erich-wang:
merge with master branch from private repo


adxsdkps:

markcowl:
We should not bring commits from the private repo into the public repo.

PR #9680, opened 2019-07-18
Update to R4.5 RP standard (#22)

leonardbf:
Addition of protocolTypes
Internal ticket NFSAAS-2898

## Checklist


leonardbf:
Static analysis failures but these don't look related to the changes being made. Please advise.

src/NetAppFiles/NetAppFiles/Volume/NewNetAppFilesVolume.cs
```diff
@@ -83,7 +84,7 @@ public class NewAzureRmNetAppFilesVolume : AzureNetAppFilesCmdletBase
             Mandatory = true,
             HelpMessage = "The maximum storage quota allowed for a file system in bytes")]
         [ValidateNotNullOrEmpty]
-        public long? UsageThreshold { get; set; }
+        public long UsageThreshold { get; set; }
```
niander: This is one of the places causing the breaking change issue

src/NetAppFiles/NetAppFiles/Pool/NewNetAppFilesPool.cs
```diff
@@ -71,7 +71,7 @@ public class NewAzureRmNetAppFilesPool : AzureNetAppFilesCmdletBase
             Mandatory = true,
             HelpMessage = "The size of the ANF pool")]
         [ValidateNotNullOrEmpty]
-        public long? PoolSize { get; set; }
+        public long PoolSize { get; set; }
```
niander: This is one of the places causing the breaking change issue

niander:
@leonardbf I made comments on where is causing the breaking change issue. Please take a look. You were removing the nullable property of these type (which is the `?` suffix for the type names)

leonardbf:
> @leonardbf I made comments on where is causing the breaking change issue. Please take a look. You were removing the nullable property of these type (which is the `?` suffix for the type names)

@niander the nullable property in the generated code changed because the swagger changed such that these became required fields. I therefore matched this in the cmdlet and actually this was necessary to resolve a build error in assigning the non-nullable to nullable in the cmdlet body.  I suppose I could provide a default but this is contrary to what expected. What's the normal solution to resolve this. Also, can you confirm, is this the sole cause of the failing checks?

niander:
@leonardbf Hi! I am sorry but I forgot that NetApp is still in preview. Thus, its is ok to have this breaking change. The only thing we need here is to suppress it. Check our docs [here ](https://github.com/Azure/azure-powershell/blob/master/documentation/Debugging-StaticAnalysis-Errors.md) to see how to do it.

I am not sure if this is the sole cause of all failing checks but according to the `BreakingChangeIssues.csv` those two are the only static analysis breaking change issues.

src/NetAppFiles/NetAppFiles.Test/Properties/AssemblyInfo.cs
```diff
@@ -45,6 +45,6 @@
 //
 // You can specify all the values or you can default the Build and Revision Numbers 
 // by using the '*' as shown below:
-[assembly: AssemblyVersion("0.9.0")]
-[assembly: AssemblyFileVersion("0.9.0")]
+[assembly: AssemblyVersion("1.1.0")]
```
niander: You don't need to change assembly versions for test projects

markcowl:
@leonardbf Ping waiting for your changes to fix the anlysis breaks, per Niander's comments

src/NetAppFiles/NetAppFiles.Test/Properties/AssemblyInfo.cs
leonardbf: ok, I will revert this

leonardbf:
> @leonardbf Hi! I am sorry but I forgot that NetApp is still in preview. Thus, its is ok to have this breaking change. The only thing we need here is to suppress it. Check our docs [here ](https://github.com/Azure/azure-powershell/blob/master/documentation/Debugging-StaticAnalysis-Errors.md) to see how to do it.
> 
> I am not sure if this is the sole cause of all failing checks but according to the `BreakingChangeIssues.csv` those two are the only static analysis breaking change issues.

@niander ok, thanks. I have looked at the page and see that I need to add the breaking changes file to the exceptions area. However, I'm not sure where I find the original BreakingChangeIssues.csv. It says it is under the Jenkin's build artifacts. Can you give me the link to that please. Also, I'd like to be sure on the preview status. The ANF service went GA a few weeks ago so does that not also mean that the PowerShell cmdlets are actually now GA?

leonardbf:
@niander assistance with the couple of points in my previous comment would be appreciated. Many thanks.

src/NetAppFiles/NetAppFiles/Properties/AssemblyInfo.cs
```diff
@@ -24,5 +24,5 @@
 [assembly: ComVisible(false)]
 [assembly: CLSCompliant(false)]
 [assembly: Guid("42656543-77AD-4968-BA4B-BE0778705625")]
-[assembly: AssemblyVersion("0.1.1")]
-[assembly: AssemblyFileVersion("0.1.1")]
+[assembly: AssemblyVersion("1.1.0")]
```
niander: You don't that to change this either

niander:
@leonardbf Hi. You can find the `BreakingChangeIssues.csv` file in your computer when you build your project and run the static analysis. It is the same one generated by the CI (azure devops, since we are not using jenkins anymore. Sorry about that). About the PowerShell module going from Preview to Release it should happen after we merge these changes and release a new version of it. After that we can talk about going to Release and adding the module to Az

leonardbf:
@niander I updated the PR with the BreakingChangeIssues.csv file just to check the validation but the pipeline now seems to be stuck. I don't see any links to the process to try and restart.

leonardbf:
Replaced with https://github.com/Azure/azure-powershell/pull/9762

PR #9706, opened 2019-07-23
Fix hanging of Get-DataLakeStoreDeletedItem for any errors or remote exceptions

rahuldutta90:
## Description

Earlier for any errors or exceptions, Get-DataLakeStoreDeletedItem used to hang. The reason is the finishing of the task depended on a flag set in the progresstracker. However if a error occurs the progresstracker wont be updated so it will be on an infinite loop.

## Checklist


src/DataLakeStore/DataLakeStore/DataPlaneModels/DataLakeStoreFileSystemClient.cs
```diff
@@ -777,49 +777,24 @@ public void GetFileProperties(string accountName, string path, bool getAclUsage,
...
-
-            // Keep printing (?) the updates returned by successive calls by the SDK to the backend, until the whole query completes
-            while(true)
+                IEnumerable<TrashEntry> result = null;
+                // Api call below consists of multiple rest calls so multiple debug statements will be posted
+                // so since we want to the debug lines to be updated while the command runs, we have to flush the debug statements in queue and thats why we want to do it this way
+                var enumerateTask = Task.Run(() => {
+                result = client.EnumerateDeletedItems(filter, "", count, null, cmdletCancellationToken);
+                }, cmdletCancellationToken);
+                TrackTaskProgress(enumerateTask, cmdlet, null, cmdletCancellationToken);
```
cormacpayne: @rahuldutta90 what does the `TrackTaskProcess` method do differently than the `while` loop that was previously being used?

rahuldutta90: @cormacpayne  TrackTaskProcess will exit if the task throws a exception, where as the while loop depends on "completed" to exit and "completed" will be true when "e.NumFound >= count || String.IsNullOrEmpty(e.NextListAfter)". Now imagine customer does not have permissions to run this operation and they will get a remote exception from ADLS, so "client.EnumerateDeletedItems" will throw an exception, however progresschanged will never be triggered and "completed" wil remain false.

TrackTaskProcess checks the status of task itself.

PR #9705, opened 2019-07-23
Update permissive record matcher to allow ignoring api versions at resource type level.

johnpaulkee:
## Description

- Bump Management.Azure.Management.Sql to 1.33.0-preview
- Ignore Microsoft.Sql api versions
- Fix PermissiveRecordMatcherWithApiExclusion uri escaping

## Checklist

- [x] I have read the [_Submitting Changes_](https://github.com/Azure/azure-powershell/blob/master/CONTRIBUTING.md#submitting-changes) section of [`CONTRIBUTING.md`](https://github.com/Azure/azure-powershell/blob/master/CONTRIBUTING.md)
- [x] The title of the PR is clear and informative
- [x] The appropriate [change log has been updated](https://github.com/Azure/azure-powershell/blob/master/CONTRIBUTING.md#updating-the-change-log)
- [x] The PR does not introduce [breaking changes](https://github.com/Azure/azure-powershell/blob/master/documentation/breaking-changes/breaking-changes-definition.md)
- [x] If applicable, the changes made in the PR have proper test coverage
- [n/a] For public API changes to cmdlets:
    - [n/a] the changes have gone through a [cmdlet design review](https://github.com/Azure/azure-powershell-cmdlet-review-pr) (Microsoft internal partners only)
    - [n/a] the cmdlet markdown files were [generated using the `platyPS` module](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/help-generation.md)

PR #9744, opened 2019-07-30
Add instance pool, instance pool usages cmdlets, and update managed instance cmdlets to support instance pools.

johnpaulkee:
## Description

This pull request adds cmdlets for Azure SQL Instance pools, Instance pool usages, and updates existing managed instance cmdlets to support placement of an instance within an instance pool.

- New cmdlets:
  - New-AzSqlInstancePool
  - Set-AzSqlInstancePool
  - Remove-AzSqlInstancePool
  - Get-AzSqlInstancePool
  - Get-AzSqlInstancePoolUsage
- Updated cmdlets:
  - New-AzSqlInstance
  - Set-AzSqlInstance
  - Get-AzSqlInstance

Design review: 
https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/329

Related pull request:
https://github.com/Azure/azure-powershell/pull/9705
- This pull request includes the above PR. I'm setting up a new PR for the instance pool cmdlets as there a lot more files that will need to be added for this review, so having both will help making reviewing easier.
- Secondly, since we have new recordings for managed instance tests in this PR, if the proposed solution for allowing ignoring api versions per top level resources is allowed we can ignore the top level resource tests for standalone managed instances as we've tested and recorded new tests for managed instances within instance pools.

## Checklist

### DIFF
--- a/tools/ScenarioTest.ResourceManager/PermissiveRecordMatcherWithApiExclusion.cs
+++ b/tools/ScenarioTest.ResourceManager/PermissiveRecordMatcherWithApiExclusion.cs
@@ -1,36 +1,33 @@
-﻿// ----------------------------------------------------------------------------------
-//
-// Copyright Microsoft Corporation
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-// http://www.apache.org/licenses/LICENSE-2.0
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-// ----------------------------------------------------------------------------------
+﻿// ---------------------------------------------------------------------------------- 
+// 
+// Copyright Microsoft Corporation 
+// Licensed under the Apache License, Version 2.0 (the "License"); 
+// you may not use this file except in compliance with the License. 
+// You may obtain a copy of the License at 
+// http://www.apache.org/licenses/LICENSE-2.0 
+// Unless required by applicable law or agreed to in writing, software 
+// distributed under the License is distributed on an "AS IS" BASIS, 
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. 
+// See the License for the specific language governing permissions and 
+// limitations under the License. 
+// ---------------------------------------------------------------------------------- 
 
-using Microsoft.Azure.Management.Internal.Resources.Utilities.Models;
 using Microsoft.Azure.Test.HttpRecorder;
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
 
 namespace Microsoft.WindowsAzure.Commands.ScenarioTest
 {
-    // Excludes api version when matching mocked records.
+    // Excludes api version when matching mocked records. 
     // If alternate api version is provided, uses that to match records else removes the api-version matching.
     public class PermissiveRecordMatcherWithApiExclusion : IRecordMatcher
     {
         protected bool _ignoreGenericResource;
         protected Dictionary<string, string> _providersToIgnore;
         protected Dictionary<string, string> _userAgentsToIgnore;
-        protected string[] _resourcesToIgnore;
 
         public PermissiveRecordMatcherWithApiExclusion(bool ignoreResourcesClient, Dictionary<string, string> providers)
         {
@@ -48,30 +45,18 @@ namespace Microsoft.WindowsAzure.Commands.ScenarioTest
             _userAgentsToIgnore = userAgents;
         }
 
-        public PermissiveRecordMatcherWithApiExclusion(
-            bool ignoreResourcesClient,
-            Dictionary<string, string> providers,
-            Dictionary<string, string> userAgents,
-            string[] resourcesToIgnore)
-        {
-            _ignoreGenericResource = ignoreResourcesClient;
-            _providersToIgnore = providers;
-            _userAgentsToIgnore = userAgents;
-            _resourcesToIgnore = resourcesToIgnore;
-        }
-
         public virtual string GetMatchingKey(System.Net.Http.HttpRequestMessage request)
         {
-            var requestUri = request.RequestUri.PathAndQuery;
-            if (requestUri.Contains("?&"))
+            var path = request.RequestUri.PathAndQuery;
+            if (path.Contains("?&"))
             {
-                requestUri = requestUri.Replace("?&", "?");
+                path = path.Replace("?&", "?");
             }
 
             string version;
-            if (ContainsIgnoredProvider(requestUri, out version))
+            if (ContainsIgnoredProvider(path, out version))
             {
-                requestUri = RemoveOrReplaceApiVersion(requestUri, version);
+                path = RemoveOrReplaceApiVersion(path, version);
             }
             else if (_userAgentsToIgnore != null && _userAgentsToIgnore.Any())
             {
@@ -82,62 +67,46 @@ namespace Microsoft.WindowsAzure.Commands.ScenarioTest
                     {
                         if (agent.Value.Any(v => v.StartsWith(userAgnet.Key, StringComparison.OrdinalIgnoreCase)))
                         {
-                            requestUri = RemoveOrReplaceApiVersion(requestUri, userAgnet.Value);
+                            path = RemoveOrReplaceApiVersion(path, userAgnet.Value);
                             break;
                         }
                     }
                 }
             }
 
-            var encodedPath = Convert.ToBase64String(Encoding.UTF8.GetBytes(requestUri));
+            var encodedPath = Convert.ToBase64String(Encoding.UTF8.GetBytes(path));
             return string.Format("{0} {1}", request.Method, encodedPath);
         }
 
         public virtual string GetMatchingKey(RecordEntry recordEntry)
         {
             var encodedPath = recordEntry.EncodedRequestUri;
-            var requestUri = recordEntry.RequestUri;
+            var path = recordEntry.RequestUri;
             var changed = false;
-            if (requestUri.Contains("?&"))
+            if (path.Contains("?&"))
             {
-                requestUri = recordEntry.RequestUri.Replace("?&", "?");
+                path = recordEntry.RequestUri.Replace("?&", "?");
                 changed = true;
             }
 
             string version;
-            if (ContainsIgnoredProvider(requestUri, out version))
+            if (ContainsIgnoredProvider(path, out version))
             {
-                requestUri = RemoveOrReplaceApiVersion(requestUri, version);
+                path = RemoveOrReplaceApiVersion(path, version);
                 changed = true;
             }
 
             if (changed)
             {
-                encodedPath = Convert.ToBase64String(Encoding.UTF8.GetBytes(requestUri));
+                encodedPath = Convert.ToBase64String(Encoding.UTF8.GetBytes(path));
             }
 
             return string.Format("{0} {1}", recordEntry.RequestMethod, encodedPath);
         }
 
-        public bool ContainsIgnoredProvider(string requestUri, out string version)
+        protected bool ContainsIgnoredProvider(string requestUri, out string version)
         {
-            return ContainsIgnoredProvider(requestUri, _ignoreGenericResource, _providersToIgnore, _resourcesToIgnore, apiVersion: out version);
-        }
-
-        /// <summary>
-        /// Helper method to determine whether or not this request api version should be ignored
-        /// </summary>
-        /// <param name="requestUri">The request uri</param>
-        /// <param name="apiVersion">The api verison to use</paraam>
-        /// <returns></returns>
-        public static bool ContainsIgnoredProvider(
-            string requestUri,
-            bool shouldIgnoreGenericResource,
-            Dictionary<string, string> providersToIgnore,
-            string[] resourcesToIgnore,
-            out string apiVersion)
-        {
-            if (shouldIgnoreGenericResource &&
+            if (_ignoreGenericResource &&
                 !requestUri.Contains("providers/") &&
                 !requestUri.StartsWith("/certificates?", StringComparison.InvariantCultureIgnoreCase) &&
                 !requestUri.StartsWith("/pools", StringComparison.InvariantCultureIgnoreCase) &&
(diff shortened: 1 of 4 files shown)


adxsdkps:

src/Sql/Sql/help/Get-AzSqlInstancePool.md
```diff
@@ -0,0 +1,214 @@
+---
+external help file: Microsoft.Azure.PowerShell.Cmdlets.Sql.dll-Help.xml
+Module Name: Az.Sql
+online version:
```
cormacpayne: @johnpaulkee please make sure that all of the new cmdlet markdown files you're adding have the correct URL for the `online version` field in the header of the file -- the build is currently failing since this value is missing from the new files. You can find what the exact URL should be from the artifacts of the CI.

src/Sql/Sql.Test/ScenarioTests/AdvancedDataSecurityManagedInstanceTests.cs
```diff
@@ -32,6 +33,9 @@ protected override void SetupManagementClients(RestTestFramework.MockContext con
 
         public AdvancedDataSecurityManagedInstanceTests(ITestOutputHelper output) : base(output)
         {
+            base.resourceTypesToIgnoreApiVersion = new string[] {
```
cormacpayne: @johnpaulkee TBD on this change 😁

src/Sql/Sql.Test/ScenarioTests/Common.ps1
```diff
@@ -738,6 +738,56 @@ function Create-ManagedInstanceForTest ($resourceGroup, $subnetId)
...
+}
+
+
+function Get-InstancePoolTestProperties()
+{
+    $tags = @{ instance="Pools" };
+    $instancePoolTestProperties = @{
+        resourceGroup = "instancePoolCSSdemo"
+        name = "cssinstancepool0"
+        subnetId = "/subscriptions/2e7fe4bd-90c7-454e-8bb6-dc44649f27b2/resourceGroups/instancePoolCSSdemo/providers/Microsoft.Network/virtualNetworks/vnet-cssinstancepool0/subnets/InstancePool"
```
cormacpayne: @johnpaulkee we should avoid hard-coding resource ids when we can -- with this strings, it may not be possible for other users to re-record this test if needed since they may not have access to the given subscription and/or resource(s)

src/Sql/Sql.Test/ScenarioTests/InstancePoolCrudTests.ps1
```diff
@@ -0,0 +1,452 @@
...
+    Tests getting an instance pool
+#>
+function Test-GetInstancePool
+{
+    # Setup
+    $instancePool = Create-InstancePoolForTest
+    Assert-InstancePoolProperties $instancePool
+
+    # Test get single - default params
+    $instancePool = Get-AzSqlInstancePool -ResourceGroupName $instancePool.ResourceGroupName -Name $instancePool.InstancePoolName
```
cormacpayne: @johnpaulkee a quick refactoring comment: for tests where you're making multiple cmdlet calls, I would advise to follow a pattern similar to the following:

```powershell
# Create necessary resources for test before try-finally
try
{
    # Call cmdlets to test and verify behavior with assertions
}
finally
{
    # Delete resource(s) and/or resource group(s) created during this test
}
```

With this model, you're guaranteed to delete all resources when recording the test regardless of if there's an exception thrown or not.

src/Sql/Sql.Test/UnitTests/PermissiveRecordMatcherWithApiExclusionUnitTests.cs
```diff
@@ -0,0 +1,74 @@
+﻿using Microsoft.WindowsAzure.Commands.ScenarioTest;
```
cormacpayne: @johnpaulkee TBD on this file 😁

src/Sql/Sql/Instance Pools/Cmdlet/GetAzureSqlInstancePool.cs
```diff
@@ -0,0 +1,127 @@
+﻿using System.Collections.Generic;
```
cormacpayne: @johnpaulkee it looks like a lot of the `.cs` files you've added are missing license headers -- please make sure to add this to the top of each of these files

src/Sql/Sql/Instance Pools/Cmdlet/GetAzureSqlInstancePool.cs
```diff
@@ -0,0 +1,127 @@
...
+using Microsoft.Azure.Commands.Sql.Instance_Pools.Model;
+using Microsoft.Azure.Management.Internal.Resources.Utilities.Models;
+using Microsoft.WindowsAzure.Commands.Utilities.Common;
+
+namespace Microsoft.Azure.Commands.Sql.Instance_Pools.Cmdlet
+{
+    /// <summary>
+    /// Defines the Get-AzRmSqlInstancePool cmdlet
+    /// </summary>
+    [Cmdlet(VerbsCommon.Get, ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "SqlInstancePool")]
```
cormacpayne: @johnpaulkee since this cmdlet has multiple parameter sets, please make sure that a `DefaultParameterSetName` is defined in this cmdlet attribute (it should be `ListBySubOrResourceGroupParameterSet`)

src/Sql/Sql/Instance Pools/Cmdlet/GetAzureSqlInstancePool.cs
```diff
@@ -0,0 +1,127 @@
...
+        private const string GetInstancePoolDefaultParameterSet = "ListByInstancePoolDefaultsParameterSet";
+        private const string GetInstancePoolByInstancePoolResourceIdentifierParameterSet = "GetInstancePoolByInstancePoolResourceIdentifierParameterSet";
+
+        /// <summary>
+        /// Gets or sets the resource group name
+        /// </summary>
+        [Parameter(
+            ParameterSetName = ListBySubOrResourceGroupParameterSet,
+            Mandatory = false,
+            Position = 0,
```
cormacpayne: @johnpaulkee please remove the position attribute from all _optional_ parameters

This comment applies to all instances of this in this PR.

src/Sql/Sql/Instance Pools/Cmdlet/GetAzureSqlInstancePool.cs
```diff
@@ -0,0 +1,127 @@
...
+
+        /// <summary>
+        /// Gets or sets the instance pool resource identifier.s
+        /// </summary>
+        [Parameter(
+            ParameterSetName = GetInstancePoolByInstancePoolResourceIdentifierParameterSet,
+            Position = 0,
+            Mandatory = true,
+            HelpMessage = "The instance pool resource identifier.")]
+        public string ResourceId { get; set; }
```
cormacpayne: @johnpaulkee please make sure that this parameter has the property `ValueFromPipelineByPropertyName = true` to enable piping

src/Sql/Sql/Instance Pools/Cmdlet/RemoveAzureSqlInstancePool.cs
```diff
@@ -0,0 +1,150 @@
...
+using Microsoft.Azure.Management.Internal.Resources.Utilities.Models;
+using Microsoft.Rest.Azure;
+using Microsoft.WindowsAzure.Commands.Utilities.Common;
+
+namespace Microsoft.Azure.Commands.Sql.Instance_Pools.Cmdlet
+{
+    /// <summary>
+    /// Defines the Remove-AzSqlInstancePool cmdlet
+    /// </summary>
+    [Cmdlet(VerbsCommon.Remove, ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "SqlInstancePool",
```
cormacpayne: @johnpaulkee same comment about adding `DefaultParameterSetName` here (it should be `DeleteByNameParameterSet`)

src/Sql/Sql/Instance Pools/Cmdlet/SetAzureSqlInstancePool.cs
```diff
@@ -0,0 +1,193 @@
...
+using Microsoft.Azure.Commands.Sql.Instance_Pools.Model;
+using Microsoft.Azure.Management.Internal.Resources.Utilities.Models;
+using Microsoft.WindowsAzure.Commands.Utilities.Common;
+
+namespace Microsoft.Azure.Commands.Sql.Instance_Pools.Cmdlet
+{
+    /// <summary>
+    /// Defines the Set-AzSqlInstancePool cmdlet
+    /// </summary>
+    [Cmdlet(VerbsCommon.Set, ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "SqlInstancePool",
```
cormacpayne: @johnpaulkee same comment about adding `DefaultParameterSetName` here (it should be `DefaultSetInstancePoolParameterSet`)

tools/ScenarioTest.ResourceManager/PermissiveRecordMatcherWithApiExclusion.cs
```diff
@@ -1,33 +1,36 @@
-﻿// ----------------------------------------------------------------------------------
```
cormacpayne: @johnpaulkee TBD on the changes in this file

src/Sql/Sql/Usages/Cmdlet/GetAzureSqlInstancePoolUsage.cs
```diff
@@ -0,0 +1,126 @@
...
+using Microsoft.Azure.Commands.Sql.Instance_Pools.Model;
+using Microsoft.Azure.Management.Internal.Resources.Utilities.Models;
+
+namespace Microsoft.Azure.Commands.Sql.Usages
+{
+    /// <summary>
+    /// Defines the Get-AzSqlInstancePoolUsage cmdlet
+    /// </summary>
+    [Cmdlet(VerbsCommon.Get, ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "SqlInstancePoolUsage",
+         DefaultParameterSetName = InstancePoolUsageDefaultParameterSet, SupportsShouldProcess = true)]
```
cormacpayne: @johnpaulkee we shouldn't need `SupportsShouldProcess` for a `Get` command

src/Sql/Sql/help/Get-AzSqlInstance.md
```diff
@@ -137,34 +311,106 @@ Accept pipeline input: False
...
 
 ```yaml
 Type: System.String
-Parameter Sets: (All)
+Parameter Sets: GetInstanceByNameAndResourceGroup
 Aliases: InstanceName
 
-Required: False
-Position: 0
+Required: True
```
cormacpayne: @johnpaulkee it looks like the changes to this command have reversed the positional order of `-Name` and `-ResourceGroupName`, which is a breaking change. Would you mind changing this back?

Also, it looks like `-Name` is now mandatory in its parameter set when it was optional before -- this would be a breaking change if the user was able to previously run the command with just `-Name` and no `-ResourceGroupName`. If the command previously failed in this case, then this isn't a breaking change.

src/Sql/Sql/help/New-AzSqlInstance.md
```diff
@@ -242,7 +379,7 @@ Parameter Sets: (All)
 Aliases: InstanceName
 
 Required: True
-Position: 0
+Position: 1
```
cormacpayne: @johnpaulkee same comment about reversing the positions of `-Name` and `-ResourceGroupName`

src/Sql/Sql/help/Remove-AzSqlInstance.md
```diff
@@ -98,7 +98,7 @@ Parameter Sets: RemoveInstanceFromInputParameters
 Aliases: InstanceName
 
 Required: True
-Position: 0
+Position: 1
```
cormacpayne: @johnpaulkee same comment about reversing the position of `-Name` and `-ResourceGroupName`

src/Sql/Sql/help/Set-AzSqlInstance.md
```diff
@@ -182,7 +224,7 @@ Parameter Sets: SetInstanceFromInputParameters
 Aliases: InstanceName
 
 Required: True
-Position: 0
+Position: 1
```
cormacpayne: @johnpaulkee same comment about reversing the position of `-Name` and `-ResourceGroupName`

src/Sql/Sql/help/Get-AzSqlInstance.md
johnpaulkee: Sure I'll revert this :)

src/Sql/Sql.Test/UnitTests/PermissiveRecordMatcherWithApiExclusionUnitTests.cs
johnpaulkee: Acked 👍 I'm hoping we can come to a solution sometime this week.

src/Sql/Sql/Usages/Cmdlet/GetAzureSqlInstancePoolUsage.cs
johnpaulkee: Will remove.

src/Sql/Sql/help/New-AzSqlInstance.md
johnpaulkee: Acked.

src/Sql/Sql/help/Remove-AzSqlInstance.md
johnpaulkee: Acked.

src/Sql/Sql/help/Set-AzSqlInstance.md
johnpaulkee: Acked.

src/Sql/Sql/help/Get-AzSqlInstance.md
johnpaulkee: >  If the command previously failed in this case, then this isn't a breaking change.
Will get back to you on this.

src/Sql/Sql/Instance Pools/Cmdlet/GetAzureSqlInstancePool.cs
johnpaulkee: Will add license headers for these files.

src/Sql/Sql/Instance Pools/Cmdlet/GetAzureSqlInstancePool.cs
johnpaulkee: Acked.

src/Sql/Sql.Test/ScenarioTests/Common.ps1
johnpaulkee: @cormacpayne for this case, Instance pool creation requires a preconfigured subnet similar to managed instance.

If I gave some instructions here on how to pre-configure their subnet, is this enough?
The ability for us to re-use the subnet would be great.
We do something similar in this method - CreateAndGetVirtualNetworkForManagedInstance

jaredmoo: Why aren't you just calling CreateAndGetVirtualNetworkForManagedInstance here?

johnpaulkee: We can update to use this.
We can reuse subnet if existing with developer's subscription, but will create if not existing.

src/Sql/Sql/help/Get-AzSqlInstance.md
johnpaulkee: I can't seem to run:
image
Given this, I think it isn't a breaking change.
@cormacpayne let me know what you think.

src/Sql/Sql.Test/ScenarioTests/Common.ps1
cormacpayne: If there isn't a way to do the creation of the resource(s) during test setup, then providing details as a comment in the script on how to construct the necessary resource(s) is sufficient. We just need to make sure that anyone can run the test in record mode, or has the proper instructions on how to do so.

src/Sql/Sql/help/Get-AzSqlInstance.md
cormacpayne: Weird, I just ran the same thing on Cloud Shell and it didn't prompt me for `-ResourceGroupName`. I'm not sure what would be causing it to prompt you 😕 Would you mind installing the latest `Az.Sql` module on your machine and testing it locally?

Regardless, I think the case we want to be cautious of is when the user only provided `-ResourceGroupName`, because with the new parameter set, it will prompt them to provide a value for `-Name` where it didn't previously. That would be the breaking change in this case.

johnpaulkee: Just kidding, locally in master we can do Name by itself.
image

johnpaulkee: Updated so that Name and ResourceGroup are non mandatory again. Also removed positional values since they were both non mandatory.

johnpaulkee:
Friendly ping to @markcowl to help take a look at the permissive record matcher changes when possible. We're really trying to catch the 8/5 code completion date for release by 8/13 and also to unblock the other SQL folks for the failing MI tests due to version bump.

johnpaulkee:
thank you @cormacpayne and @markcowl for reviewing and merging.

PR #9945, opened 2019-08-30
Making the Monitor PS SDK to point to the version 0.24.1-preview of the .Net SDK

gucalder:
## Description

* Pointing to the most recent Monitor SDK, i.e. 0.24.1-preview
   - Adds non-braking changes to the Metrics cmdlets, i.e. the Unit enumeration supports several new values. These are read-only cmdlets, so there would be no change in the input of the cmdlets.
   - The api-version of the **ActionGroups** requests is now **2019-06-01**, before it was **2018-03-01**. The scenario tests have been updated to accommodate for this change.
   - The constructors for the classes **EmailReceiver** and **WebhookReceiver** added one new mandatory argument, i.e. a Boolean value called **useCommonAlertSchema**. Currently, the value is fixed to **false** to hide this breaking change from the cmdlets. **NOTE**: this is a temporary change that must be validated by the Alerts team.
   - The order of the arguments for the constructor of the class **Source** (related to the **ScheduledQueryRuleSource** class) changed from the previous SDK. This change required two unit tests to the be fixed: they compiled, but failed to pass the tests.
   - The order of the arguments for the constructor of the class **AlertingAction** (related to the **ScheduledQueryRuleSource** class) changed from the previous SDK. This change required two unit tests to the be fixed: they compiled, but failed to pass the tests.

## Checklist

- [x] The PR does not introduce [breaking changes](../documentation/breaking-changes/breaking-changes-definition.md)


isra-fel:
Static analysis failed because it detected a possible breaking change:

> "Microsoft.Azure.PowerShell.Cmdlets.Monitor.dll","Microsoft.Azure.Commands.Insights.Metrics.GetAzureRmMetricCommand","Get-AzMetric","0","3030","The generic type for 'property Count' has been changed from 'System.Nullable`1[System.Int64]' to 'System.Nullable`1[System.Double]'. ","Change the generic type for 'property Count' back to 'System.Nullable`1[System.Int64]'."

I found that the change was introduced by [this PR](`https://github.com/Azure/azure-rest-api-specs/pull/6485/files#diff-e3f0bd57dbd517abfc194f2848d47528`). One property of the output type of `Get-AzMetric` cmdlet -- `Count` was changed from a `long?` to `double?`.

Though technically changing the output type is considered a breaking change, I suppose it is OK in this case, because it's safe to place a `long` where it should be a `double`.

@cormacpayne what is your opinion? Thanks!

gucalder:
That is correct. Noe I remember that change in the Swagger spec. The reason for this change is historic: actually the best type for these data should long, but the API has been returning a double for a long time. So we actually changed the type to reflect current reality. BTW: there was a request from some other team for us to do that too.

cormacpayne:
@isra-fel that's correct -- because previous values that the user could've provided for `long?` fit inside of the new `double?` type, there shouldn't be a script that would break because of this change ☺️ 

@gucalder you will need to suppress this exception to get the `Analyze` step of the pipeline to pass. You can follow the steps outlined in the [Suppressing Exceptions](https://github.com/Azure/azure-powershell/blob/af5f11a13b3ba161275c8380014d1ea5584263cf/documentation/tooling/static-analysis.md#suppressing-exceptions) section of this static analysis document.

gucalder:
Ok, I am suppressing the exception. Thanks!

gucalder:
Ok, it failed again, but I cannot find why since the logs shown here are a bit limited (they only say ti failed) and I cannot run it locally because it fails all the time.

I think I suppressed the issue in the right place.

Can you guys help let me know what happened this time?

tools/StaticAnalysis/Exceptions/Az.Monitor/SignatureIssues.csv
```diff
@@ -6,4 +6,5 @@
...
\ No newline at end of file
+"Microsoft.Azure.PowerShell.Cmdlets.Monitor.dll","Microsoft.Azure.Commands.Insights.ScheduledQueryRules.NewScheduledQueryRuleTriggerConditionCommand","New-AzScheduledQueryRuleTriggerCondition","1","8100","New-AzScheduledQueryRuleTriggerCondition Does not support ShouldProcess but the cmdlet verb New indicates that it should.","Determine if the cmdlet should implement ShouldProcess and if so determine if it should implement Force / ShouldContinue"
+"Microsoft.Azure.PowerShell.Cmdlets.Monitor.dll","Microsoft.Azure.Commands.Insights.Metrics.GetAzureRmMetricCommand","Get-AzMetric","0","3030","The generic type for 'property Count' has been changed from 'System.Nullable1[System.Int64]' to 'System.Nullable1[System.Double]'. ","Change the generic type for 'property Count' back to 'System.Nullable`1[System.Int64]'."
```
cormacpayne: @gucalder you added this line to the `SignatureIssues.csv` file when it should have gone in the `BreakingChangeIssues.csv` file -- this would cause the build to fail since the files are parsed differently due to the differences in their columns

gucalder:
Analysis is working fine now, but tests fail and also the powershell-core. I do not know which tests (they all ran fine locally.)

cormacpayne:
@gucalder the version of the Monitor SDK being used by the SQL test project was different than the version you upgraded to in the Monitor project, which was causing assembly loading issues since there were two different versions trying to be loaded during the SQL tests. I updated the version in the SQL test project, so we should be good now.

gucalder:
I just resolved the conflicts.

isra-fel:
LGTM
@cormacpayne any thoughts? 😃

PR #9982, opened 2019-09-05
[Preview] MSAL and shared token cache support

cormacpayne:
## Checklist

- [ ] The PR does not introduce [breaking changes](../documentation/breaking-changes/breaking-changes-definition.md)


src/Accounts/Authenticators/DefaultAuthenticatorBuilder.cs
```diff
@@ -0,0 +1,48 @@
...
+using Microsoft.Azure.Commands.Common.Authentication;
+using Microsoft.Azure.Commands.Common.Authentication.Abstractions;
+
+namespace Microsoft.Azure.PowerShell.Authenticators
+{
+    public class DefaultAuthenticatorBuilder : IAuthenticatorBuilder
+    {
+        public DefaultAuthenticatorBuilder()
+        {
+            AppendAuthenticator(() => { return new InteractiveUserAuthenticator(); });
```
markcowl: Is the oredeing here important at all?

src/Accounts/Authenticators/DeviceCodeAuthenticator.cs
```diff
@@ -0,0 +1,60 @@
...
+using Microsoft.Azure.Commands.ResourceManager.Common;
+using Microsoft.Identity.Client;
+
+namespace Microsoft.Azure.PowerShell.Authenticators
+{
+    public class DeviceCodeAuthenticator : DelegatingAuthenticator
+    {
+        public override Task<IAccessToken> Authenticate(AuthenticationParameters parameters)
+        {
+            var cancellationTokenSource = new CancellationTokenSource();
```
markcowl: Should we be passing in a cancellation token to all of these methods?  The new cmdlets are a cancellation token source, and it would be nice if someone htting ctrl-c could automatically stop authentication.

markcowl: This would mean we'd have to plumb through a cancellation token, but we could default it in the method signatures, so that no cancellation token is necessarily required.

src/Accounts/Authenticators/DeviceCodeAuthenticator.cs
```diff
@@ -0,0 +1,60 @@
...
+        public override bool CanAuthenticate(AuthenticationParameters parameters)
+        {
+            return parameters is DeviceCodeParameters;
+        }
+
+        private void WriteWarning(string message)
+        {
+            if (AzureSession.Instance.TryGetComponent(AzureRMCmdlet.WriteWarningKey, out EventHandler<StreamEventArgs> writeWarningEvent))
+            {
+                writeWarningEvent(this, new StreamEventArgs() { Message = message });
```
markcowl: I am worried here that we could end up in the wrong thread with a cmdlet.WriteWarning

src/Accounts/Authenticators/InteractiveUserAuthenticator.cs
```diff
@@ -26,23 +29,56 @@ namespace Microsoft.Azure.PowerShell.Authenticators
...
+
+                if (!string.IsNullOrEmpty(replyUrl))
+                {
+                    var clientId = AuthenticationHelpers.PowerShellClientId;
+                    var authority = AuthenticationHelpers.GetAuthority(parameters.Environment, parameters.TenantId);
+                    publicClient = _authenticationClientFactory.CreatePublicClient(clientId: clientId, authority: authority, redirectUri: replyUrl);
+
+                    var interactiveResponse = publicClient.AcquireTokenInteractive(scopes)
+                        .WithCustomWebUi(new CustomWebUi(interactiveParameters.PromptAction))
+                        .ExecuteAsync();
```
markcowl: cancellationtoken?

src/Accounts/Authenticators/InteractiveUserAuthenticator.cs
```diff
@@ -26,23 +29,56 @@ namespace Microsoft.Azure.PowerShell.Authenticators
...
+            }
+
+            return null;
         }
 
-        public override bool CanAuthenticate(IAzureAccount account, IAzureEnvironment environment, string tenant, SecureString password, string promptBehavior, Task<Action<string>> promptAction, IAzureTokenCache tokenCache, string resourceId)
+        public override bool CanAuthenticate(AuthenticationParameters parameters)
         {
-            return (account?.Type == AzureAccount.AccountType.User && environment != null && !string.IsNullOrWhiteSpace(tenant) && password == null && promptBehavior != ShowDialog.Never && tokenCache != null  && account != null && !account.IsPropertySet("UseDeviceAuth"));
+            return parameters is InteractiveParameters;
```
markcowl: nit: it is slightly more efficient to use
``` c#
return (parameters as InteractiveParameters) != null;

```

src/Accounts/Authenticators/ManagedServiceIdentityAuthenticator.cs
```diff
@@ -0,0 +1,91 @@
...
+            }
+
+            if (string.IsNullOrWhiteSpace(tenant))
+            {
+                tenant = environment.AdTenant ?? CommonAdTenant;
+            }
+
+            if (account.IsPropertySet(AppServiceManagedIdentityFlag))
+            {
+                return new ManagedServiceAppServiceAccessToken(account, environment, GetFunctionsResourceId(resourceId, environment), tenant);
```
markcowl: Would it be cleaner to have one authenticator for AppServiceMSI first in the chain, and the authenticator for Compute MSI after?

src/Accounts/Authenticators/ServicePrincipalAuthenticator.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            var spParameters = parameters as ServicePrincipalParameters;
+            var scopes = new string[] { string.Format(AuthenticationHelpers.DefaultScope, spParameters.ResourceEndpoint) };
+            var clientId = spParameters.ApplicationId;
+            var authority = AuthenticationHelpers.GetAuthority(spParameters.Environment, spParameters.TenantId);
+            var redirectUri = spParameters.Environment.ActiveDirectoryServiceEndpointResourceId;
+            IConfidentialClientApplication confidentialClient = null;
+            if (spParameters.Secret == null)
+            {
+                var certificate = AzureSession.Instance.DataStore.GetCertificate(spParameters.Thumbprint);
```
markcowl: What if Thumbprint wasn't provided?  Shouldn't we have a tri state (HasCert / HasSecret / Throw)?

src/Accounts/Authenticators/SilentAuthenticator.cs
```diff
@@ -0,0 +1,45 @@
...
+        public override Task<IAccessToken> Authenticate(AuthenticationParameters parameters)
+        {
+            var silentParameters = parameters as SilentParameters;
+            var scopes = new string[] { string.Format(AuthenticationHelpers.DefaultScope, silentParameters.ResourceEndpoint) };
+            var clientId = AuthenticationHelpers.PowerShellClientId;
+            var authority = AuthenticationHelpers.GetAuthority(silentParameters.Environment, silentParameters.TenantId);
+            var publicClient = _authenticationClientFactory.CreatePublicClient(clientId: clientId, authority: authority);
+            var accounts = publicClient.GetAccountsAsync()
+                .ConfigureAwait(false).GetAwaiter().GetResult();
+            var response = publicClient.AcquireTokenSilent(scopes, accounts.FirstOrDefault(a => a.Username == silentParameters.UserId)).ExecuteAsync();
```
markcowl: again, I really think we need to make the cancellation token part of each of these calls, and pass it through to MSAL

src/Accounts/Authentication/Authentication/Clients/AuthenticationClientFactory.cs
```diff
@@ -0,0 +1,211 @@
...
+using System.Security;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.Azure.Commands.Common.Authentication.Authentication.Clients
+{
+    public abstract class AuthenticationClientFactory
+    {
+        public static readonly string AuthenticationClientFactoryKey = nameof(AuthenticationClientFactory);
+        protected readonly string PowerShellClientId = "1950a258-227b-4e31-a9cf-717495945fc2";
+        private static readonly string CommonTenant = "organizations";
```
markcowl: Does this same common tenant work in the MSA flow?

src/Accounts/Authentication/Factories/AuthenticationFactory.cs
```diff
@@ -14,21 +14,22 @@
...
+using Microsoft.Azure.Commands.Common.Authentication.Authentication.Clients;
 
 namespace Microsoft.Azure.Commands.Common.Authentication.Factories
 {
     public class AuthenticationFactory : IAuthenticationFactory
     {
         public const string AppServiceManagedIdentityFlag = "AppServiceManagedIdentityFlag";
 
-        public const string CommonAdTenant = "Common",
+        public const string CommonAdTenant = "organizations",
```
markcowl: Seems like we should have this constant in one place, and use it evwerywhere.  Also, same question as before about whether thsi works with MSA

src/Accounts/Authentication/Factories/AuthenticationFactory.cs
```diff
@@ -94,79 +94,46 @@ public IServicePrincipalKeyStore KeyStore
...
-                configuration.ClientRedirectUri,
-                configuration.ResourceClientUri,
-                configuration.ValidateAuthority);
-            if (account != null && account.Type == AzureAccount.AccountType.ManagedService)
-            {
-                token = GetManagedServiceToken(account, environment, tenant, resourceId);
-            }
-            else if (account != null && environment != null
-                && account.Type == AzureAccount.AccountType.AccessToken)
+            Task<IAccessToken> authToken;
```
markcowl: I am worried that we are not doing anough tracing here.  When there is an authentication error, it is useful to know what values were passed to the authentication library, and what the results were - without this, pinpointing authentication errors can be difficult

src/Accounts/Authentication/Factories/AuthenticationFactory.cs
```diff
@@ -502,37 +469,107 @@ private string GetEndpointToken(IAzureAccount account, string targetEndpoint)
...
+            IAzureAccount account,
+            IAzureEnvironment environment,
+            string tenant,
+            SecureString password,
+            string promptBehavior,
+            Action<string> promptAction,
+            IAzureTokenCache tokenCache,
+            string resourceId = AzureEnvironment.Endpoint.ActiveDirectoryServiceEndpointResourceId)
+        {
+            if (account.Type == AzureAccount.AccountType.User)
```
markcowl: Use a switch here

src/Accounts/Authentication.ResourceManager/AzureRmProfile.cs
```diff
@@ -188,7 +205,7 @@ private void Initialize(AzureRmProfile profile)
                     this.Contexts.Add(context.Key, context.Value);
                 }
 
-                DefaultContextKey = profile.DefaultContextKey ?? "Default";
+                DefaultContextKey = profile.DefaultContextKey ?? (profile.Contexts.Any() ? null : "Default");
```
markcowl: Setting the defautl key to null if there are contexts but none selected - when would this occur?

src/Accounts/Authentication.ResourceManager/AzureRmProfile.cs
```diff
@@ -631,5 +648,133 @@ static string GetJsonText(string text)
 
             return result;
         }
+
+        private void WriteWarning(string message)
+        {
+            EventHandler<StreamEventArgs> writeWarningEvent;
+            if (AzureSession.Instance.TryGetComponent(AzureRMCmdlet.WriteWarningKey, out writeWarningEvent))
+            {
+                writeWarningEvent(this, new StreamEventArgs() { Message = message });
```
markcowl: This is fine, but I think the implementation on the other end may lead to issues.

src/Accounts/Authentication.ResourceManager/AzureRmProfile.cs
```diff
@@ -631,5 +648,133 @@ static string GetJsonText(string text)
...
+            var accounts = authenticationClientFactory.ListAccounts();
+            if (!accounts.Any())
+            {
+                if (!Contexts.Any(c => c.Key != "Default" && c.Value.Account.Type == AzureAccount.AccountType.User))
+                {
+                    // If there are no accounts in the cache, but we never had any existing contexts, return
+                    return;
+                }
+
+                WriteWarning($"No accounts found in the shared token cache; removing all user contexts.");
```
markcowl: resource string

src/Accounts/Authentication.ResourceManager/AzureRmProfile.cs
```diff
@@ -631,5 +648,133 @@ static string GetJsonText(string text)
...
+        {
+            EventHandler<StreamEventArgs> writeWarningEvent;
+            if (AzureSession.Instance.TryGetComponent(AzureRMCmdlet.WriteWarningKey, out writeWarningEvent))
+            {
+                writeWarningEvent(this, new StreamEventArgs() { Message = message });
+            }
+        }
+
+        private void RefreshContextsFromCache()
+        {
```
markcowl: We need to add some more tracing in here, so we can debug issues with migration

src/Accounts/Authentication.ResourceManager/AzureRmProfile.cs
```diff
@@ -631,5 +648,133 @@ static string GetJsonText(string text)
...
+                // Check to see if each account has at least one context
+                foreach (var account in accounts)
+                {
+                    if (Contexts.Values.Where(v => v.Account != null && v.Account.Type == AzureAccount.AccountType.User )
+                                       .Any(v => string.Equals(v.Account.Id, account.Username, StringComparison.OrdinalIgnoreCase)))
+                    {
+                        continue;
+                    }
+
+                    WriteWarning($"Creating context for each subscription accessible by account '{account.Username}'.");
```
markcowl: All these user-facing strings need to be in resource strings

src/Accounts/Authentication.ResourceManager/AzureRmProfile.cs
```diff
@@ -631,5 +648,133 @@ static string GetJsonText(string text)
...
+            }
+        }
+
+        private void RefreshContextsFromCache()
+        {
+            var authenticationClientFactory = new SharedTokenCacheClientFactory();
+            var accounts = authenticationClientFactory.ListAccounts();
+            if (!accounts.Any())
+            {
+                if (!Contexts.Any(c => c.Key != "Default" && c.Value.Account.Type == AzureAccount.AccountType.User))
```
markcowl: I think this method could use a bit of breaking up.

RemoveUserContexts()
RemoveUnmatchedUserContexts()
etc.

Also, as stated above, we need soem tracing here so we can tell why a context was removed if soemthign goes wrong.

src/Accounts/Accounts/Common/AzureContextModificationCmdlet.cs
```diff
@@ -43,7 +44,11 @@ protected virtual void ModifyContext(Action<AzureRmProfile, RMProfileClient> con
         {
             using (var profile = GetDefaultProfile())
             {
-                contextAction(profile.ToProfile(), new RMProfileClient(profile));
+                var client = new RMProfileClient(profile)
+                {
+                    WarningLog = (s) => WriteWarning(s)
```
markcowl: Per above, this will cause threaqding problems

src/Accounts/Accounts/Common/AzureContextModificationCmdlet.cs
```diff
@@ -194,5 +199,41 @@ protected bool TryGetExistingContextNameParameter(string name, string parameterS
...
+        {
+            WriteDebug(args.Message);
+        }
+
+        private void WriteVerboseSender(object sender, StreamEventArgs args)
+        {
+            WriteVerbose(args.Message);
+        }
+
+        private void WriteWarningSender(object sender, StreamEventArgs args)
```
markcowl: Well, actually these are the part that will cause threading sisues.

src/Accounts/Authentication/Authentication.csproj
```diff
@@ -12,7 +12,24 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.IdentityModel.Clients.ActiveDirectory" Version="3.19.2" />
+    <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="2.2.0" />
```
markcowl: We need to replace the AdalLogger above with some kind of adapter to get logging messages out of MSAL.

src/Accounts/Authentication/Authentication/ProtectedFileTokenCache.cs
```diff
@@ -83,130 +113,86 @@ public ProtectedFileTokenCache(string cacheFile, IDataStore store = null)
...
-                }
-
-                // Eagerly create cache file.
-#if !NETSTANDARD
-                var dataToWrite = ProtectedData.Protect(Serialize(), null, DataProtectionScope.CurrentUser);
-#else
-                var dataToWrite = Serialize();
-#endif
-                _store.WriteFile(cacheFileName, dataToWrite);
+                _store.DeleteFile(CacheFileName);
```
markcowl: Do we ever want to delete the shared cache file?

src/Accounts/Accounts/Models/RMProfileClient.cs
```diff
@@ -692,9 +692,11 @@ private string GetCommonTenant(IAzureAccount account)
                                 return tenant;
                             }).ToList();
                 }
+
                 if (!result.Any())
                 {
-                    throw new RuntimeException("Error occurred when attempting to acquire the common tenant token. Please run 'Connect-AzAccount` again to authenticate.");
+                    WriteWarningMessage("Error occurred when attempting to acquire the common tenant token. Please run 'Connect-AzAccount` again to authenticate.");
```
markcowl: `<nit>`resource string `</nit>`

src/Accounts/Authentication/Authentication/Clients/AuthenticationClientFactory.cs
```diff
@@ -67,6 +68,11 @@ public abstract class AuthenticationClientFactory
                 builder = builder.WithRedirectUri(redirectUri);
             }
 
+            builder.WithLogging((level, message, pii) =>
+            {
+                TracingAdapter.Information(string.Format("[MSAL] {0}: {1}", level, message));
```
markcowl: I can't remember, is this written to the debbug log, or is there a separate method for Debug messages in the Tracing Adapter?

src/Accounts/Authenticators/DesktopAuthenticatorBuilder.cs
```diff
@@ -17,20 +17,15 @@ namespace Microsoft.Azure.PowerShell.Authenticators
...
-            session.RegisterComponent(AuthenticatorBuilder.AuthenticatorBuilderKey, () =>
-            {
-                var userPassword = new UsernamePasswordAuthenticator();
-                userPassword.Next = new InteractiveUserAuthenticator();
-                var authenticator = new DesktopAuthenticatorBuilder();
-                authenticator.Authenticator = userPassword;
-                return authenticator as IAuthenticatorBuilder;
-            });
+            AppendAuthenticator(() => { return new InteractiveUserAuthenticator(); });
+            var defaultBuilder = new DefaultAuthenticatorBuilder();
```
isra-fel: Wouldn't that be 2 `InteractiveUserAuthenticator`s 🤔 ?

src/Accounts/Authentication.ResourceManager/AzureRmProfile.cs
isra-fel: Needs validate: If logged in in VS (not PS), when first time runing PS command (not select-context), should be prompted with "please select a context"

PR #9997, opened 2019-09-09
Merge latest master into Az.Batch-preview

matthchr:
This just contains the latest master changes as well as a few conflict resolutions


matthchr:
@cormacpayne - Take a look at this when you get a chance

matthchr:
Also had to fix a few issues with the breaking changes file due to cmdlet renames which took place previously

markcowl:
@matthchr Is there a reason why we need to use this old branch?  Can we remove it and create another?

matthchr:
@markcowl - we can/could, but the changes in this branch have:
1. Not been merged to master yet
2. Are being built upon for the PR I will send once this PR merges (or a new branch is set).

This is me just trying to get the base state of a branch to contain both the changes I already did in `Az.Batch-preview` as well as the latest `master`.

Issue #9926, opened 2019-08-28
[Documentation] Overview of ARM/Graph concepts and scenarios


cormacpayne:
## Description

Outline important concepts and scenarios for the ARM and Graph cmdlets.

- [x] `ResourceManager` project
    - [x] Resources / ResourceGroups
    - [x] Deployments
    - [x] Policies
- [x] `Resources` project
    - [x] ActiveDirectory
        - [x] Applications
        - [x] Groups
        - [x] ServicePrincipals
        - [x] Users
    - [x] RoleAssignments
    - [x] RoleDefinitions

markcowl:
Fixed

PR #10012, opened 2019-09-11
Add documentation on Resource Manager and Active Directory concepts and cmdlets

cormacpayne:
## Description

Resolves issue https://github.com/Azure/azure-powershell/issues/9926

## Checklist

- [x] The PR does not introduce [breaking changes](../documentation/breaking-changes/breaking-changes-definition.md)

### DIFF
--- a/documentation/internal/resource-manager-active-directory.md
+++ b/documentation/internal/resource-manager-active-directory.md
@@ -125,7 +125,7 @@ The `Resources` project contains all of the Active Directory, Authorization, and
 
 ### Active Directory Applications
 
-From the [_Application and service principal objects in Azure Active Directory_](https://github.com/Azure/azure-powershell/blob/master/src/Resources/Resources/help/Test-AzResourceGroupDeployment.md) document, an Azure AD application is defined as the following:
+From the [_Application and service principal objects in Azure Active Directory_](https://docs.microsoft.com/en-us/azure/active-directory/develop/app-objects-and-service-principals#application-object) document, an Azure AD application is defined as the following:
 
 _An Azure AD application is defined by its one and only application object, which resides in the Azure AD tenant where the application was registered, known as the application's "home" tenant._
 


cormacpayne:
@markcowl let me know if you have any feedback on the document and if there's anything else I should include or outline.

CC: @erich-wang @isra-fel @dolauli @msJinLei @wyunchi-ms please take a look as well and let me know if you have any questions ☺️

documentation/internal/resource-manager-active-directory.md
```diff
@@ -0,0 +1,272 @@
...
+- [`Stop-AzDeployment`](https://github.com/Azure/azure-powershell/blob/master/src/Resources/Resources/help/Stop-AzDeployment.md)
+    - Cancels a deployment in the current subscription with the given name or ID.
+- [`Stop-AzResourceGroupDeployment`](https://github.com/Azure/azure-powershell/blob/master/src/Resources/Resources/help/Stop-AzResourceGroupDeployment.md)
+    - Cancels a deployment in the specified resource group with the given name or ID.
+- [`Test-AzDeployment`](https://github.com/Azure/azure-powershell/blob/master/src/Resources/Resources/help/Test-AzDeployment.md)
+    - Determines whether the given deployment template and its parameter values are valid in the current subscription.
+- [`Test-AzResourceGroupDeployment`](https://github.com/Azure/azure-powershell/blob/master/src/Resources/Resources/help/Test-AzResourceGroupDeployment.md)
+    - Determines whether the given deployment template and its parameter values are valid in the specified resource group.
+
+#### Policy Definition
```
markcowl: So, I was thinking this would mostly be about the parts of this that we own, particularly the generic resource cmdlets, and the new graph.rabc cmdlets.  This other stuff about policy and role assignments is also useful, but not the intended focus.

documentation/internal/resource-manager-active-directory.md
```diff
@@ -0,0 +1,272 @@
...
+
+From the [_Azure Resource Manager overview_](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-overview#terminology) document, a resource group is defined as the following:
+
+_A container that holds related resources for an Azure solution. The resource group includes those resources that you want to manage as a group. You decide which resources belong in a resource group based on what makes the most sense for your organization._
+
+The cmdlets currently found in the `Az.Resources` module that act upon resource groups are the following:
+
+- [`Export-AzResourceGroup`](https://github.com/Azure/azure-powershell/blob/master/src/Resources/Resources/help/Export-AzResourceGroup.md)
+    - Converts the contents of a resource group to a template and saves it to a file on the user's machine.
+- [`Get-AzResourceGroup`](https://github.com/Azure/azure-powershell/blob/master/src/Resources/Resources/help/Get-AzResourceGroup.md)
```
markcowl: Can we expand here a bit on the implementation of the Get/New/Set Resource cmdlets, how they are implemented, how piping works, and current known issues?

documentation/internal/resource-manager-active-directory.md
```diff
@@ -0,0 +1,272 @@
...
+
+_When an application is given permission to access resources in a tenant (upon registration or consent), a service principal object is created._
+
+_A service principal must be created in each tenant where the application is used, enabling it to establish an identity for sign-in and/or access to resources being secured by the tenant. A single-tenant application has only one service principal (in its home tenant), created and consented for use during application registration. A multi-tenant Web application/API also has a service principal created in each tenant where a user from that tenant has consented to its use._
+
+The cmdlets currently found in the `Az.Resources` module that act upon service principals are the following:
+
+- [`Get-AzADServicePrincipal`](https://github.com/Azure/azure-powershell/blob/master/src/Resources/Resources/help/Get-AzADServicePrincipal.md)
+     - Retrieves all service principals in the current tenant. The user can also filter results by object ID, application ID, display name and service principal name.
+- [`New-AzADServicePrincipal`](https://github.com/Azure/azure-powershell/blob/master/src/Resources/Resources/help/New-AzADServicePrincipal.md)
```
markcowl: Please explain a bit more about the intended scenario - why do we have c amcldte that does all these things, how is it implemented, what are its dependecies, and what are any current known issues?

cormacpayne: @markcowl updated -- called out why we want to enable this scenario, how it works with the Authorization SDK, gave some examples, and left a note about what we can do in the future to reach parity with the CLI scenario.

documentation/internal/resource-manager-active-directory.md
cormacpayne: @markcowl updated -- stripped out all of the sections that we don't own, included examples for piping scenarios per cmdlet (where applicable) and called out any user scenarios we've implemented.

documentation/internal/resource-manager-active-directory.md
cormacpayne: @markcowl updated -- talked about the implementation of the three cmdlets, whether they are using the `HttpClient` or `ResourceManagerClient` (or both), some piping examples, and notes about what we can look into in the future. Let me know if I should expand anymore.

documentation/internal/resource-manager-active-directory.md
```diff
@@ -0,0 +1,292 @@
...
+
+The last scenario that we would possibly use the generic `HttpClient` for this cmdlet would be if the user provided either the `-ApiVersion` or `-ExpandProperties` parameters. The `-ApiVersion` parameter is used to define which API version of the service should be used when retrieving the resource(s), which can change the values in the `Properties` property on the returned object(s). The `-ExpandProperties` switch parameter is used to signal that the `Properties` property on the returned object(s) should be expanded and include all resource specific information. The `ResourceManagerClient` doesn't allow for either of these scenarios (outside of providing an API version for a resource ID, which was covered previously), so the `HttpClient` must be used to provide users with this functionality.
```
markcowl:
```suggestion
The last scenario that we would possibly use the generic `HttpClient` for this cmdlet would be if the user provided either the `-ApiVersion` or `-ExpandProperties` parameters. The `-ApiVersion` parameter is used to define which API version of the service should be used when retrieving the resource(s), which can change the values in the `Properties` property on the returned object(s).  By default the Get-AzResource cmdlet is handled by the ARM service and returns only the generic parts of the resource managed by ARM which are common to all resources (e.g. ResourceId, SKU, Tags, etc.). The `-ExpandProperties` switch parameter is used to signal that the call should be handled by the RP directly and the  `Properties` property on the returned object(s) should be expanded and include all resource specific information. The `ResourceManagerClient` doesn't allow for either of these scenarios (outside of providing an API version for a resource ID, which was covered previously), so the `HttpClient` must be used to provide users with this functionality.
```

documentation/internal/resource-manager-active-directory.md
```diff
@@ -0,0 +1,292 @@
...
...h the portal to manually create the permissions. One of the scenarios we added to this cmdlet was to give users the ability to also create a role assignment for the service principal within this cmdlet using new `-Role` and `-Scope` parameters. If the user provided a value for either parameter, once the service principal was created, a [call to the authorization SDK](https://github.com/Azure/azure-powershell/blob/master/src/Resources/Resources/ActiveDirectory/Cmdlets/NewAzureADServicePrincipalCommand.cs#L346-L380) would be made to assign the specified role to the service principal over the given scope. If neither parameter was provided, or the `-SkipAssignment` switch was set, then a role assignment would not be created. If only `-Role` was provided, `-Scope` would be [set to the user's current subscription](https://github.com/Azure/azure-powershell/blob/master/src/Resources/Resources/ActiveDirectory/Cmdlets/NewAzureADServicePrincipalCommand.cs#L252-L256), and if only `-Scope` was provided, `-Role` would be [set to the "Contributor"](https://github.com/Azure/azure-powershell/blob/master/src/Resources/Resources/ActiveDirectory/Cmdlets/NewAzureADServicePrincipalCommand.cs#L260-L264).
```
markcowl: A common scenario is for a user to create a service principal in order to automate Azure management tasks. When a Service Principal is used to automate management tasks, it must have a credential, and it must be given permissions to manage the resources that are part of its task.  The `New-AzADServicePricnipal` cmdlet allows the user to create a service principal, give it secure credentials, and allow the service principal the appropriate level of access in Azure through issuing a single cmdlet.

documentation/internal/resource-manager-active-directory.md
```diff
@@ -0,0 +1,292 @@
...
+New-AzADServicePrincipal -Role Reader @Properties
+
+# Create a service principal with "Contributor" permissions over the "MyResourceGroup" resource group
+New-AzADServicePrincipal -Scope /subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/resourceGroups/MyResourceGroup @Properties
+
+# Create a service principal with "Reader" permissions over the "MyResourceGroup" resource group
+New-AzADServicePrincipal -Role Reader -Scope /subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/resourceGroups/MyResourceGroup @Properties
+```
+
+_Note_: for parity with Azure CLI, parameters should be added in the future that would allow for the user to create a self-signed certificate and create or retrieve a certificate from a specified Azure KeyVault.
```
markcowl: add in: .Net standard framework does not contain cross-platform libraries for certificate creation, so this would mean identifying platform-specific mechanisms for creating certificates.

documentation/internal/resource-manager-active-directory.md
```diff
@@ -0,0 +1,292 @@
...
+
+In the future, it should be determined whether or not the `ResourceManagerClient` can be used to make the subscription level calls as it has two functions, `CreateOrUpdate()` and `CreateOrUpdateById()`, which should be able to perform the same steps outlined above.
+
+**Set-AzResource**
```
isra-fel: Excuse me, the behavior that a cmdlet could perform 2 different operations, and the operations are differenciated by a switch parameter - is it a common practice? Didn't see that before 🤔

documentation/internal/resource-manager-active-directory.md
```diff
@@ -0,0 +1,292 @@
...
+For example, a virtual machined called "MyVM" found in the resource group "MyResourceGroup" for subscription "830afc1a-b2c2-49fa-b36d-c05c335ea986" would have the following ID:
+
+```
+/subscriptions/830afc1a-b2c2-49fa-b36d-c05c335ea986/resourceGroups/MyResourceGroup/Microsoft.Compute/virtualMachines/MyVM
+```
+
+Child resources have IDs that follow the below pattern:
+
+```
+/subscriptions/{SUBSCRIPTION_ID}/resourceGroups/{RESOURCE_GROUP}/{PARENT_PROVIDER_NAMESPACE}/{PARENT_RESOURCE_TYPE}/{PARENT_RESOURCE_NAME}/{CHILD_RESOURCE_TYPE}/{CHILD_RESOURCE_NAME}
```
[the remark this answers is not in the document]
cormacpayne: @isra-fel that's correct -- each additional child nested will add `{CHILD_RESOURCE_TYPE}/{CHILD_RESOURCE_NAME}` to the end of the resource ID

documentation/internal/resource-manager-active-directory.md
cormacpayne: @isra-fel this isn't common practice and is advised against in almost every scenario -- we want to follow the pattern `PATCH` --> `Update-*` and `PUT` --> `Set-*`.

This is a very rare case where the cmdlet will handle both operations based on a switch parameter and should not be followed moving forward. This cmdlet was implemented long before we put our current patterns and standards in place, and this cmdlet uses the generic `HttpClient` to make it's updates to the resources, so my guess is that, at the time, it was easier to support both `PUT` and `PATCH` in the cmdlet by just switching what operation was used by the `HttpClient` since it didn't require a separate SDK call.

Let me know if you still have any questions about this.

documentation/internal/resource-manager-active-directory.md
```diff
@@ -0,0 +1,292 @@
...
+$Resource | Set-AzResource -Force
+```
+
+## `Resources` project
+
+The `Resources` project contains all of the Active Directory, Authorization, and Management Groups cmdlets.
+
+### Active Directory Applications
+
+From the [_Application and service principal objects in Azure Active Directory_](https://github.com/Azure/azure-powershell/blob/master/src/Resources/Resources/help/Test-AzResourceGroupDeployment.md) document, an Azure AD application is defined as the following:
```
wyunchi-ms: Excuse me, I think the link is not about _Application and service principal objects in Azure Active Directory_

cormacpayne: @wyunchi-ms yep, you're right, looks like a copy-paste error 😁

isra-fel:
Changes made. I'll just merge it. Thanks 😁

PR #10155, opened 2019-09-30
Azure AZ PowerShell Key Vault creation 

v-rasaa:
## Description
PowerShell module does not allow the creation of Key Vault with Bypass or to add IP Ranges or IP rules during creation as parameter. The command for creating a new key vault does not have a parameter that can be added in order to assign a IP Range, this is by default but if Customer uses CLI it gives him the option to add a Bypass but not for IP rules or Vnet which customer is trying to implement. The situation in here is that customer has an Azure policy that enforces this behavior and he can not turn it off. 
why I can not use PowerShell to achieve his goal and why PowerShell does not have the same Bypass option as CLI? 

PowerShell for creating a new Key Vault
https://docs.microsoft.com/en-us/powershell/module/az.keyvault/new-azkeyvault?view=azps-2.7.0#targetText=Enables%20the%20Azure%20disk%20encryption,keys%20from%20this%20key%20vault.&targetText=Enables%20Azure%20Resource%20Manager%20to,referenced%20in%20a%20template%20deployment.&targetText=Specifies%20the%20Azure%20region%20in%20which%20to%20create%20the%20key%20vault.

Cli for creating a new key Vault https://docs.microsoft.com/en-us/cli/azure/keyvault?view=azure-cli-latest#az-keyvault-create

### DIFF
--- a/src/StackAdmin/Azs.Backup.Admin/Module/Azs.Backup.Admin.csproj
+++ b/src/StackAdmin/Azs.Backup.Admin/Module/Azs.Backup.Admin.csproj
@@ -40,8 +40,8 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Microsoft.AzureStack.Management.Admin.Backup, Version=0.3.0-preview">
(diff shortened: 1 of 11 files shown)


adxsdkps:

markcowl:
@v-rasaa Please file an issue https://github.com/azure/azure-powershell/issues

PR #10291, opened 2019-10-15
Nullreferenceexception in new azure ad service principal#9478

VeryEarly:
## Checklist

- [ ] I have read the [_Submitting Changes_](../CONTRIBUTING.md#submitting-changes) section of [`CONTRIBUTING.md`](../CONTRIBUTING.md)
- [ ] The PR does not introduce [breaking changes](../documentation/breaking-changes/breaking-changes-definition.md)


src/Resources/Resources/ActiveDirectory/Cmdlets/NewAzureADServicePrincipalCommand.cs
```diff
@@ -383,7 +391,12 @@ private void CreateSimpleServicePrincipal()
 
         private bool SkipRoleAssignment()
         {
-            return this.IsParameterBound(c => c.SkipAssignment) || (!this.IsParameterBound(c => c.Role) && !this.IsParameterBound(c => c.Scope));
+            return this.IsParameterBound(c => c.SkipAssignment) || (!this.IsParameterBound(c => c.Role) && !this.IsParameterBound(c => c.Scope) && !HasSubscription());
+        }
+
+        private bool HasSubscription()
+        {
+            return DefaultContext.Subscription != null && DefaultContext.Subscription.Id != null ;
```
markcowl:
```suggestion
            return DefaultContext.Subscription?.Id != null ;
```

VeryEarly:
> One minor suggestion. It would be nice to replumb the logic of this so that we didn't have to check for roe assignment skipping more than once, but changing minimum code mitigates risk.
> 
> What are your thoughts about testing this? Seems like we should add a test to prevent us from regressing in this scenario. We can add automation after the release if you file an issue and place it in the next milestone.

I reduce one of the usages of "SkipRoleAssignment", there are still two left and seems not likely to remove either one of them.

I created test case for it but test scenario required a connection to a service principal without subscription which I cannot find how to set up in record mode.

Do you have any idea about these?

VeryEarly:
> > One minor suggestion. It would be nice to replumb the logic of this so that we didn't have to check for roe assignment skipping more than once, but changing minimum code mitigates risk.
> > What are your thoughts about testing this? Seems like we should add a test to prevent us from regressing in this scenario. We can add automation after the release if you file an issue and place it in the next milestone.
> 
> I reduce one of the usages of "SkipRoleAssignment", there are still two left and seems not likely to remove either one of them.
> 
> I created test case for it but test scenario required a connection to a service principal without subscription which I cannot find how to set up in record mode.
> 
> Do you have any idea about these?

git issue, close this pull request, will try fix them and open another one

PR #10392, opened 2019-10-25
Revert "Create Managed Network Powershell"

isra-fel:
Reverts Azure/azure-powershell#10195

This is for public preview.
Code is now on branch `Az.ManagedNetwork-Preview`.


adxsdkps:

isra-fel:
@yanfa317 FYI

tools/StaticAnalysis/Exceptions/Az.ManagedNetwork/SignatureIssues.csv
```diff
@@ -1,2 +0,0 @@
-"AssemblyFileName","ClassName","Target","Severity","ProblemId","Description","Remediation"
```
markcowl: seems liek this file should be completedlt removed.  Also, the 'PS-VSPrompt.lnk file above should not be included in this PR.

isra-fel: @markcowl
Yes, `SignatureIssues.csv` is indeed deleted.

"PS-VSPrompt.lnk" was deleted by accident in the original PR. This PR will recover it back.

Issue #10053, opened 2019-09-18
Update release script for version bump


erich-wang:
## Description
1. The folder `C:\Program Files\PowerShell\Modules` doesn't exist in some cases; need to update script RunVersionController.ps1

2. The version 2.1.3 of PowerShellGet fails to detect manifest change (silently), however version 2.2.1 works well. We need to alert user

3. The serialized Cmdlets json file for preview modules with version 0.* should not be put the folder 
Tools.Common/SerializedCmdlets/

PR #10412, opened 2019-10-29
Fix the bug of PowerShellGet in tools/RunVersionController.ps1

wyunchi-ms:
## Description

1. Fix the bug that can not find the right version PowerShellGet in some system.
2. Make sure will not contain the moudle which's version is lower than 1.0.0. And remove used json files in SerializedCmdlets.

## Checklist


adxsdkps:

tools/RunVersionController.ps1
```diff
@@ -151,11 +151,13 @@ function Update-ChangeLog
     ($Content + $ChangeLogContent) | Set-Content -Path $ChangeLogFile.FullName -Encoding UTF8
 }
 
-if (!(Test-Path "C:/Program Files/PowerShell/Modules/PowerShellGet"))
+$PowerShellGetModuleList = gmo -l PowerShellGet | where version -ge '2.2.1'
+if ($PowerShellGetModuleList.Count -ne 0)
 {
+    $PowerShellGetPath = $PowerShellGetModuleList.path.Split('PowerShellGet')[0]
```
markcowl: What if `$PowerShellGetModuleList` is empty?  I think we just want to use Install-Module, which will automatically add this to the path if not selected, rather than downloading to a particular path.

`Install-Module -Scope CurrentUser` will automatically place the module first in the search order.

Alternately, we could just start the script with

``` powershell
#Requires -Modules @{ModuleName="PowerShellGet"; ModuleVersion="2.2.1"}
```

This will throw at the start of execution if it cannot find/load a version of PowerSgellGet >= 2.2.1

wyunchi-ms:
https://github.com/Azure/azure-powershell/issues/10053

PR #10464, opened 2019-11-03
update document

erich-wang:
Update document with examples


adxsdkps:

PR #10513, opened 2019-11-11
Fix the problem of api version with sub-resource.

wyunchi-ms:
## Description

To fix issue https://github.com/Azure/azure-powershell/issues/9797

## Checklist

### DIFF
--- a/src/Resources/ResourceManager/Components/ApiVersionHelper.cs
+++ b/src/Resources/ResourceManager/Components/ApiVersionHelper.cs
@@ -26,6 +26,7 @@ namespace Microsoft.Azure.Commands.ResourceManager.Cmdlets.Components
     using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
+    using Microsoft.Azure.Commands.ResourceManager.Cmdlets.Utilities;
 // TODO: Remove IfDef
 #if NETSTANDARD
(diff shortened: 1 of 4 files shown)


adxsdkps:

src/Resources/ResourceManager/Components/ApiVersionHelper.cs
```diff
@@ -127,11 +128,12 @@ private static string[] GetApiVersionsForResourceType(IAzureContext context, str
                     cancellationToken: cancellationToken),
                 cancellationToken: cancellationToken);
 
+            string topLevelResourceType = ResourceTypeUtility.GetTopLevelResourceType(resourceType);
```
markcowl: Don't we want to do this only in the case where the resource type is not found initially (that is, if the original query returns nothing?

src/Resources/ResourceManager/Utilities/ResourceTypeUtility.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.Commands.ResourceManager.Cmdlets.Utilities
+{
+    public static class ResourceTypeUtility
```
markcowl: Please add some unit tests for this new method

PR #10683, opened 2019-12-10
Retrieve policy aliases at tenant level

mentat9:
## Checklist

- [X] The appropriate `ChangeLog.md` file(s) has been updated:

### DIFF
--- a/src/Resources/Resources/ChangeLog.md
+++ b/src/Resources/Resources/ChangeLog.md
@@ -21,7 +21,6 @@
 * Fix an error in help document of `Remove-AzTag`.
 * Fix for aliases missing from output of Get-AzPolicyAlias
 * Update resource client to new version that retrieves providers and aliases at tenant level
-* Fix the bug that the output of some sub-resource is empty when using `Get-AzResource`.
 * Update Get-AzPolicyAlias to retrieve aliases at tenant level
 * Update -Policy parameter of New-AzPolicyDefinition and Set-AzPolicyDefinition to allow full policy object
 
@@ -29,6 +28,7 @@
 * Update references in .psd1 to use relative path
 * Fix an issue where template deployment fails to read a template parameter if its name conflicts with some built-in parameter name.
 * Updated policy cmdlets to use new api version 2019-09-01 that introduces grouping support within policy set definitions.
+* Fix the bug that the output of some sub-resource is empty when using `Get-AzResource`.
 
 ## Version 1.8.0
 - Updated policy cmdlets to use new api version 2019-06-01 that has new EnforcementMode property in policy assignment.


mentat9:
Note: this isn't a breaking change. No inputs, parameters, types or api versions have changed. This is to fix a bug where a few aliases are missing from the subscription-level query. The correct set of aliases to return from this cmdlet is retrieved from the tenant level.

msJinLei:
@mentat9 Please fix the failed test case.

mentat9:
@msJinLei: test is fixed. Also added test coverage for new object type support.

src/Resources/ResourceManager/Implementation/Policy/NewAzurePolicyDefinition.cs
```diff
@@ -141,17 +141,37 @@ private JToken GetResource()
...
-                    PolicyRule = this.GetObjectFromParameter(this.Policy, nameof(this.Policy)),
-                    Metadata = this.Metadata == null ? null : this.GetObjectFromParameter(this.Metadata, nameof(this.Metadata)),
-                    Parameters = this.Parameter == null ? null : this.GetObjectFromParameter(this.Parameter, nameof(this.Parameter)),
-                    Mode = string.IsNullOrEmpty(this.Mode) ? PolicyDefinitionMode.All : this.Mode
-                }
+                Properties = new PolicyDefinitionProperties()
             };
 
+            var policyObject = this.GetObjectFromParameter(this.Policy, nameof(this.Policy));
+            if (policyObject["policyRule"] != null)
```
pilor: Have we had requests for supporting this?  If this starts being supported it should be supported in Set-AzPolicyDefinition as well

pilor: Should be "properties.policyRule" (and properties.metadata, etc...) right)

src/Resources/Resources/ChangeLog.md
```diff
@@ -21,6 +21,7 @@
 * Update references in .psd1 to use relative path
 * Fix an issue where template deployment fails to read a template parameter if its name conflicts with some built-in parameter name.
 * Updated policy cmdlets to use new api version 2019-09-01 that introduces grouping support within policy set definitions.
+* Update resource client to new version that retrieves providers and aliases at tenant level
```
pilor: The only change was to retrieve policy aliases at tenant level right?  Not providers for Get-AzResourceProvider

src/Resources/ResourceManager/Implementation/Policy/GetAzurePolicyAliasCmdlet.cs
```diff
@@ -116,12 +115,12 @@ private bool IsMatch(string input, string match)
         private IEnumerable<Provider> GetAllProviders()
         {
             var returnList = new List<Provider>();
-            var tempResult = this.ResourceManagerSdkClient.ResourceManagementClient.Providers.List(expand: "resourceTypes/aliases");
+            var tempResult = this.ResourceManagerSdkClient.ResourceManagementClient.Providers.ListAtTenantScope(expand: "resourceTypes/aliases");
```
pilor: No recorded sessionrecords needed to be updated for tests after making this change?

mentat9: I made a recorded session file and added it. Need to investigate why it isn't showing here.

mentat9: I expected it to, but the tests all passed on playback.

src/Resources/ResourceManager/Implementation/Policy/NewAzurePolicyDefinition.cs
mentat9: Working from a work item you opened a while back. I think there were requests.

src/Resources/Resources/ChangeLog.md
mentat9: Correct. The alias cmdlet gets aliases from the providers call results. I'll reword this to avoid confusion on that point.

src/Resources/ResourceManager/Implementation/Policy/NewAzurePolicyDefinition.cs
mentat9: I chose a policy format without the property bag (you can see the sample file in this PR). This matches what's displayed by the VSCode extension and the portal (when creating a new policy). It also seems more consistent with the format used for the policyRule-flavored input (the policy rule isn't inside ```"policyRule": {}``` or ```"properties": { "policyRule": {} }```). Since this input comes from users, I figured supporting copy-and-paste from VSCode and the portal was a good idea. Let me know if you think I should change it.

src/Resources/ResourceManager/Implementation/Policy/GetAzurePolicyAliasCmdlet.cs
mentat9: The tests for this cmdlet aren't live scenario tests so it doesn't need recordings.

msJinLei:
@mentat9 Please resolve the conflicts

mentat9:
@msJinLei conflicts and checks are fixed.

msJinLei:
@mentat9 
- You need to rebase your commits to the latest master manually to prevent some merge errors.
- Why you have to change / add resource manager dependent to other modules? Most of them don't reply on resource manager SDK directly

mentat9:
Believe me, I didn't want to. Unfortunately, the test projects in several modules are using the resource SDK indirectly via two common test projects (below tools folder). I had to make an explicit reference to 3.0.0 version of the resource SDK to get a clean build (version downgrade caused build breaks).

msJinLei:
> Believe me, I didn't want to. Unfortunately, the test projects in several modules are using the resource SDK indirectly via two common test projects (below tools folder). I had to make an explicit reference to 3.0.0 version of the resource SDK to get a clean build (version downgrade caused build breaks).

Could you try this way
- revert all the changes to csproj other than "src/Resources/ResourceManager/ResourceManager.csproj"
- update the reference to "Microsoft.Azure.Management.ResourceManager" in tools/Common.Netcore.Dependencies.Test.targets from 2.2.0 to 3.0 and see whether the tests can pass.

mentat9:
OK, I rebased on official master, updated the .targets file and removed subsequently unneeded changes to .csproj files. All checks have passed.

PR #10807, opened 2019-12-30
Resolved break changes in DataBoxEdge

dingmeng-xue:
Resolved break changes in DataBoxEdge

1. Add previous parameter name DeviceObject as alias
2. Remove parameter set name on parameter name. Name should have parameter set name otherwise it shows in all parameter sets. A bug will logged to fix this problem in the next major release.


adxsdkps:

PR #10813, opened 2019-12-31
Fixed Static Analysis issue

dingmeng-xue:
Static analysis does break change for each PR but it doesn't work now. 

This PR intends to fix it.


adxsdkps:

isra-fel:
Static analyze still passes. And the logs you added are nowhere to be seen 🤔

PR #10892, opened 2020-01-14
Nemanjas/mi change hw generation

v-nestan:
## Description

This change adds support for changing existing managed instance hardware generation.
It can be changed using SkuName and ComputeGeneration parameters.

TFS item [link](http://sqlbuvsts01:8080/Main/SQL%20Server/_workitems#_a=edit&id=13293296)

## Checklist

### DIFF
--- a/src/Sql/Sql/ChangeLog.md
+++ b/src/Sql/Sql/ChangeLog.md
@@ -19,6 +19,7 @@
 -->
 ## Upcoming Release
 * Added support for cross subscription point in time restore on Managed Instances.
+* Add support for changing existing Sql Managed Instance hardware generation
 
 ## Version 2.2.0
 Fix New-AzSqlDatabaseSecondary cmdlet to check for PartnerDatabaseName existence instead of DatabaseName existence.
@@ -31,7 +32,6 @@ Fix New-AzSqlDatabaseSecondary cmdlet to check for PartnerDatabaseName existence
 ## Version 2.1.1
 * Update references in .psd1 to use relative path
 * Upgraded storage creation in Vulnerability Assessment auto enablement to StorageV2
-* Adding support for changing existing Sql Managed Instance hardware generation
 
 ## Version 2.1.0
 * Added support for database ReadReplicaCount.


src/Sql/Sql/ChangeLog.md
```diff
@@ -27,6 +27,7 @@
 ## Version 2.1.1
 * Update references in .psd1 to use relative path
 * Upgraded storage creation in Vulnerability Assessment auto enablement to StorageV2
+* Adding support for changing existing managed instance hardware generation
```
NMijat1024: Sql Managed Instance

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
```diff
@@ -178,6 +179,24 @@ public class SetAzureSqlManagedInstance : ManagedInstanceCmdletBase
         [Parameter(HelpMessage = "Skip confirmation message for performing the action")]
         public SwitchParameter Force { get; set; }
 
+        /// <summary>
+        /// Gets or sets the instance compute generation
+        /// </summary>
+        [Parameter(Mandatory = false,
+            HelpMessage = "The compute generation for the instance.")]
+        [ValidateNotNullOrEmpty]
```
NMijat1024: Is it possible to add custom validators as attributes above fields?

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
```diff
@@ -246,6 +420,7 @@ protected override IEnumerable<AzureSqlManagedInstanceModel> PersistChanges(IEnu
         /// </summary>
         public override void ExecuteCmdlet()
         {
+            // System.Diagnostics.Debugger.Launch();
```
stefanb995: Remove this

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
```diff
@@ -187,6 +206,151 @@ protected override IEnumerable<Model.AzureSqlManagedInstanceModel> GetEntity()
...
         }
 
+        /// <summary>
+        /// Validate requested edition.
+        /// </summary>
+        protected void ValidateRequestedEdition()
+        {
+            ModelAdapter = InitModelAdapter();
+            AzureSqlManagedInstanceModel existingInstance = ModelAdapter.GetManagedInstance(this.ResourceGroupName, this.Name);
+            Management.Internal.Resources.Models.Sku Sku = new Management.Internal.Resources.Models.Sku();
```
stefanb995: Not used

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
```diff
@@ -187,6 +206,151 @@ protected override IEnumerable<Model.AzureSqlManagedInstanceModel> GetEntity()
...
+        /// <summary>
+        /// Validate requested Hardware family.
+        /// </summary>
+        protected void ValidateRequestedHardwareFamily(out bool shouldConfirmHardwareFamilyChange)
+        {
+            shouldConfirmHardwareFamilyChange = false;
+
+            ModelAdapter = InitModelAdapter();
+            AzureSqlManagedInstanceModel existingInstance = ModelAdapter.GetManagedInstance(this.ResourceGroupName, this.Name);
+            Management.Internal.Resources.Models.Sku Sku = new Management.Internal.Resources.Models.Sku();
```
stefanb995: Not used

src/Sql/Sql/help/Set-AzSqlInstance.md
```diff
@@ -290,6 +325,21 @@ Accept pipeline input: True (ByPropertyName)
 Accept wildcard characters: False
 ```
 
+### -SkuName
+The SKU name for the instance e.g. 'BC_Gen4'.
```
stefanb995: As Gen4 is becoming deprecated, maybe it's better to have BC_Gen5 here

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
```diff
@@ -178,6 +179,24 @@ public class SetAzureSqlManagedInstance : ManagedInstanceCmdletBase
...
+            HelpMessage = "The compute generation for the instance.")]
+        [ValidateNotNullOrEmpty]
+        [PSArgumentCompleter(Constants.ComputeGenerationGen5)]
+        public string ComputeGeneration { get; set; }
+
+        /// <summary>
+        /// Gets or sets the instance SKU name
+        /// </summary>
+        [Parameter(Mandatory = false,
+            HelpMessage = "The SKU name for the instance e.g. 'GP_Gen4','BC_Gen4'.")]
```
stefanb995: I suggest adding 'GP_Gen5','BC_Gen5' in help message as Gen4 is becoming deprecated

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
```diff
@@ -197,12 +361,22 @@ protected override IEnumerable<AzureSqlManagedInstanceModel> ApplyUserInputToMod
...
+            // Get current hardware family
+            string currentComputeGeneration = existingInstance.Sku.Name.Contains(Constants.ComputeGenerationGen4) ? Constants.ComputeGenerationGen4 : Constants.ComputeGenerationGen5;
+
+            if (SkuName != null)
             {
-                string computeGeneration = existingInstance.Sku.Name.Contains(Constants.ComputeGenerationGen4) ? Constants.ComputeGenerationGen4 : Constants.ComputeGenerationGen5;
-                string editionShort = AzureSqlManagedInstanceAdapter.GetInstanceSkuPrefix(Edition);
-                Sku.Name = editionShort + "_" + computeGeneration;
-                Sku.Tier = Edition;
+                Sku.Name = SkuName;
```
stefanb995: Why not populating Sku.Tier and Sku.Family here?

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
```diff
@@ -197,12 +361,22 @@ protected override IEnumerable<AzureSqlManagedInstanceModel> ApplyUserInputToMod
             AzureSqlManagedInstanceModel existingInstance = ModelAdapter.GetManagedInstance(this.ResourceGroupName, this.Name);
             Management.Internal.Resources.Models.Sku Sku = new Management.Internal.Resources.Models.Sku();
 
-            if (Edition != null)
+            // Get current edition
+            string currentEdition = existingInstance.Sku.Tier;
+
+            // Get current hardware family
+            string currentComputeGeneration = existingInstance.Sku.Name.Contains(Constants.ComputeGenerationGen4) ? Constants.ComputeGenerationGen4 : Constants.ComputeGenerationGen5;
```
stefanb995: Are these always populated for existing managed instances?

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
v-nestan: I've checked SQL DB, instance pools and managed instance new and set cmdlets and couldn't find the example of that so I decided to go this way. If it is possible, it is not common.

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
v-nestan: Nice catch. Fixed

src/Sql/Sql/ChangeLog.md
v-nestan: Fixed

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
v-nestan: It is used in ExecuteCmdlet() method:

`            // Validate requested edition
            if (!string.IsNullOrWhiteSpace(this.SkuName) || !string.IsNullOrWhiteSpace(this.Edition))
            {
                this.ValidateRequestedEdition();
            }`

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
```diff
@@ -187,6 +206,151 @@ protected override IEnumerable<Model.AzureSqlManagedInstanceModel> GetEntity()
...
+            bool isEditionChanged = !string.IsNullOrWhiteSpace(requestedEdition) && !currentEdition.Equals(requestedEdition, StringComparison.InvariantCultureIgnoreCase);
+
+            if (isEditionChanged)
+            {
+                // Check whether managed instance is in instance pool because it is not possible to change hardware family in that case.
+                if (!string.IsNullOrWhiteSpace(existingInstance.InstancePoolName))
+                {
+                    throw new PSArgumentException(
+                        string.Format(Microsoft.Azure.Commands.Sql.Properties.Resources.InstancePoolInstanceCannotChangeHardwareFamily, this.Name, existingInstance.InstancePoolName),
+                        "Edition");
```
jaredmoo: Why all these validations on client side? Isn't server side doing the same validation? Having business logic on the client side is generally a bad idea because PowerShell/CLI is installed on user's machine, so they need to download updates when our policies change - this is really bad experience.

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
```diff
@@ -187,6 +206,151 @@ protected override IEnumerable<Model.AzureSqlManagedInstanceModel> GetEntity()
...
+        protected string GetEditionFromSkuName(string skuName)
+        {
+            if (string.IsNullOrWhiteSpace(skuName))
+            {
+                return null;
+            }
+
+            string edition = null;
+
+            if (skuName.Equals(Constants.GeneralPurposeGen4, StringComparison.InvariantCultureIgnoreCase)
```
jaredmoo: This hardcoding of sku names / edition names on client side will add maintenance cost every time a new edition / gen is released. I think we need a different strategy here. For example, in CLI code we use capabilities API to find SKU based on requested parameters. Alternatively, perhaps you can reuse some of the existing edition/sku handling logic from db/elastic pool.

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
v-nestan: Great catch. Thanks. Fixed

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
v-nestan: Fixed

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
v-nestan: Sku.Name is always populated

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
v-nestan: I've omitted for brevity because Sku.Name is enough for upsert

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
v-nestan: Leftover from debugging. Removed.

src/Sql/Sql/help/Set-AzSqlInstance.md
v-nestan: It is much better. Fixed

src/Sql/Sql.Test/UnitTests/AzureSqlInstanceUnitTests.cs
```diff
@@ -42,5 +42,33 @@ public void GetInstanceSkuPrefix()
                 "BC",
                 AzureSqlManagedInstanceAdapter.GetInstanceSkuPrefix("BusinessCritical"));
         }
+
+        [Fact]
+        [Trait(Category.AcceptanceType, Category.CheckIn)]
+        public void GetHardwareGenerationFromSkuName()
```
jaredmoo: Nice to see!

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
```diff
@@ -178,6 +179,24 @@ public class SetAzureSqlManagedInstance : ManagedInstanceCmdletBase
...
+        public string ComputeGeneration { get; set; }
+
+        /// <summary>
+        /// Gets or sets the instance SKU name
+        /// </summary>
+        [Parameter(Mandatory = false,
+            HelpMessage = "The SKU name for the instance e.g. 'GP_Gen5','BC_Gen5'.")]
+        [ValidateNotNullOrEmpty]
+        [PSArgumentCompleter(Constants.GeneralPurposeGen5, Constants.BusinessCriticalGen5)]
+        public string SkuName { get; set; }
```
jaredmoo: I think this should be a different parameter set than compute&family params.

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
jaredmoo: AFAIK there is a completer for resource group name, if that exists then it must just be an ARM call so you could also call capabilities to find edition, sku name, compute gen, etc. So probably possible. But let's not block release on it :)

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
```diff
@@ -187,6 +206,60 @@ protected override IEnumerable<Model.AzureSqlManagedInstanceModel> GetEntity()
...
+
+            // Get current hardware family
+            string currentHardwareFamily = existingInstance.Sku.Family;
+
+            // Get requested hardware family
+            string requestedHardwareFamily = !string.IsNullOrWhiteSpace(this.SkuName) ? AzureSqlManagedInstanceAdapter.GetHardwareGenerationFromSkuName(this.SkuName) : this.ComputeGeneration;
+
+            bool isHardwareFamilyChanged = !string.IsNullOrWhiteSpace(requestedHardwareFamily) && !currentHardwareFamily.Equals(requestedHardwareFamily, StringComparison.InvariantCultureIgnoreCase);
+
+            // Check whether hardware family is being changed to a deprecated hardware family
```
jaredmoo: Don't you mean, check whether it's being changed **away from** a deprecated family?

jaredmoo: I think this code is wrong. Instead of checking if requested == Gen5, it should be checking if current == Gen4. Moving Gen7->Gen5 shouldn't require confirmation, right?

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
```diff
@@ -187,6 +206,60 @@ protected override IEnumerable<Model.AzureSqlManagedInstanceModel> GetEntity()
...
+        {
+            if (string.IsNullOrWhiteSpace(skuName))
+            {
+                return null;
+            }
+
+            string edition = null;
+
+            if (skuName.Equals(Constants.GeneralPurposeGen4, StringComparison.InvariantCultureIgnoreCase)
+                || skuName.Equals(Constants.GeneralPurposeGen5, StringComparison.InvariantCultureIgnoreCase))
```
jaredmoo: Can you just check StartsWith("GP_") or StartsWith("BC_"), instead of having every combination programmed ?

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
```diff
@@ -197,12 +270,22 @@ protected override IEnumerable<AzureSqlManagedInstanceModel> ApplyUserInputToMod
             AzureSqlManagedInstanceModel existingInstance = ModelAdapter.GetManagedInstance(this.ResourceGroupName, this.Name);
             Management.Internal.Resources.Models.Sku Sku = new Management.Internal.Resources.Models.Sku();
 
-            if (Edition != null)
+            // Get current edition
+            string currentEdition = existingInstance.Sku.Tier;
+
+            // Get current hardware family
+            string currentComputeGeneration = existingInstance.Sku.Name.Contains(Constants.ComputeGenerationGen4) ? Constants.ComputeGenerationGen4 : Constants.ComputeGenerationGen5;
```
jaredmoo: This looks like it would cause problems when next compute gen is released. Please distinguish between Gen4, Gen5, and "unknown"

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
jaredmoo: If you need to explain in PR, it's worth adding a code comment :)

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
jaredmoo: Even better would be if you implemented a function which is the opposite of GetInstanceSkuPrefix()

src/Sql/Sql/ManagedInstance/Services/AzureSqlManagedInstanceAdapter.cs
```diff
@@ -228,5 +229,27 @@ public static string GetInstanceSkuPrefix(string tier)
 
             return SqlSkuUtils.GetVcoreSkuPrefix(tier) ?? "Unknown";
         }
+
+        /// <summary>
+        /// Get hardware family from SKU name.
+        /// </summary>
+        /// <returns>The model to send to the update</returns>
+        public static string GetHardwareGenerationFromSkuName(string skuName)
```
jaredmoo: consider moving to SqlSkuUtils.cs

src/Sql/Sql/Properties/Resources.resx
```diff
@@ -580,4 +580,16 @@
   <data name="VulnerabilityAssessmentOperationIsNotAvailableOnMasterDatabase" xml:space="preserve">
     <value>Vulnerability Assessment operation is not available on master database. Skipping...</value>
   </data>
+  <data name="DoYouWantToProceed" xml:space="preserve">
+    <value>Do you want to proceed?</value>
+  </data>
+  <data name="ChangingHardwareFamilyIsIrreversable" xml:space="preserve">
+    <value>For this update operation provided compute generation parameter is Gen5. Once the hardware generation is upgraded, it is not possible to switch back to Gen4 compute generation as this hardware is being deprecated.</value>
```
jaredmoo: This message will not be accurate when next gen is released and you upgrade Gen4->Gen7

src/Sql/Sql/Properties/Resources.resx
```diff
@@ -580,4 +580,16 @@
...
+  </data>
+  <data name="CannotChangeHardwareFamily" xml:space="preserve">
+    <value>You cannot change hardware family.</value>
+  </data>
+  <data name="InvalidManagedInstanceSkuName" xml:space="preserve">
+    <value>Provided SkuName '{0}' is invalid '{1}'. Please enter a valid SkuName value and execute command again.</value>
```
jaredmoo: It would be nice to give example sku name so that customer has an better idea of what we're asking for

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
v-nestan: Nice tip. Thanks for that. Adding comment

src/Sql/Sql/Properties/Resources.resx
v-nestan: Updated to fill it in dinamically

src/Sql/Sql.Test/UnitTests/AzureSqlInstanceUnitTests.cs
v-nestan: Thanks :)

src/Sql/Sql/Properties/Resources.resx
v-nestan: This message is no longer needed. Removing it.

src/Sql/Sql/ManagedInstance/Services/AzureSqlManagedInstanceAdapter.cs
v-nestan: Done

v-nestan: Note that, in order to run unit tests, SqlSkuUtils had to be changed to public

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
v-nestan: I will use SqlSkuUtils.GetHardwareGenerationFromSkuName instead this if.

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
v-nestan: You are absolutely correct. Brilliant catch. Fixing this

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
v-nestan: Actually, this method is also redundant. I'll remove it

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
v-nestan: As we discussed offline, we will remove SkuName from parameters.

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
v-nestan: Removed all validation except prompting from confirmation that the changing hardware family is irreversible.

src/Sql/Sql/ManagedInstance/Cmdlet/SetAzureSqlManagedInstance.cs
v-nestan: Not needed any more since SkuName parameter is removed.

src/Sql/Sql/ChangeLog.md
```diff
@@ -31,6 +31,7 @@ Fix New-AzSqlDatabaseSecondary cmdlet to check for PartnerDatabaseName existence
 ## Version 2.1.1
 * Update references in .psd1 to use relative path
 * Upgraded storage creation in Vulnerability Assessment auto enablement to StorageV2
+* Adding support for changing existing Sql Managed Instance hardware generation
```
isra-fel: Changes for the upcoming release should go under the section titled "Upcoming Release".

isra-fel:
Hi @v-nestan please fix the problem in Changelog.md . Besides, in order for CI to be triggered, please join Azure orgnization: https://aka.ms/AzureGithub

PR #10956, opened 2020-01-22
Adding support to Stop execution if an exception is thrown when adding a custom domain to a website. #10422 

### LINKED ISSUE #10422: webapp: Stop executing if an exception is thrown when adding a custom domain to a website.
## Description
When a user makes a mistake in their powershell script. It sometimes results in all their existing custom domains being removed. As part of the WebsiteClient in chaning customDomains the client does a GET to see existing domains, adds new domains, and then Deletes domains that are not included. If a customDomain fails to be added, we should not remove custom domains and should stop that part of the update customDomains code.

## Steps to reproduce
I understand that this isnt the newest version, but I forgot to save those commands
```powershell
$W=Get-AzureRmWebApp -ResourceGroupName "prod-rg" -Name "prod-rg"
$X = $W.hostnames
$X.add("mycustom.domain.com")
$HN = $X.tostring()
Set-AzureRmWebApp -ResourceGroupName "prod-rg" -Name "prod-rg" -HostNames  $HN

```
At this point, X = [oldDomain1, OldDomain2, New domain]
And HN = system.collections.generic.list`1[system.string]
When customer tries to send HN the websites cmdlet sends that one new domain, this fails because this 'domain' is not valid. and then the websites cmdlet continues 
## Environment data

```

```

## Module versions

```powershell

```

## Debug output

```

```

## Error output

```

```

ThejaChoudary:
Fixes #10422

## Checklist

### DIFF
--- a/src/Websites/Websites.Test/ScenarioTests/AppServicePlanTests.ps1
+++ b/src/Websites/Websites.Test/ScenarioTests/AppServicePlanTests.ps1
@@ -24,16 +24,14 @@ function Test-CreateNewAppServicePlan
 	$location = Get-Location
 	$capacity = 2
 	$skuName = "S2"
-	#Tags
-	$tagsDict = New-Object "System.Collections.Generic.Dictionary``2[System.String,System.String]"
-	$tagsDict.Add('TagName','TagValue')
+
 	try
 	{
 		#Setup
 		New-AzResourceGroup -Name $rgname -Location $location
 
 		# Test
-		$job = New-AzAppServicePlan -ResourceGroupName $rgname -Name  $whpName -Location  $location -Tier "Standard" -WorkerSize Medium -NumberOfWorkers $capacity -Tags $tagsDict -AsJob
+		$job = New-AzAppServicePlan -ResourceGroupName $rgname -Name  $whpName -Location  $location -Tier "Standard" -WorkerSize Medium -NumberOfWorkers $capacity -AsJob
 		$job | Wait-Job
 		$createResult = $job | Receive-Job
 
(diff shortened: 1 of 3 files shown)


panchagnula:
@ThejaChoudary the linked GitHub issue has repro steps can we use this & add as a test case to make sure this scenario is covered to make sure we have no future regressions?

src/Websites/Websites/Utilities/WebsitesClient.cs
```diff
@@ -145,6 +146,7 @@ public void AddCustomHostNames(string resourceGroupName, string location, string
                             {
                                 SiteName = webAppName,
                             });
+                        successCount++;
```
panchagnula: I don't think we need to keep track of successcount , the problem here is that the hostname value is an invalid format from the example in the issue, the value is "system.collections.generic.list`1[system.string]" as a result the API fails to add - & throws an exception, as a result when an exception occurs we should stop execution of code or return.

ThejaChoudary:
> @ThejaChoudary the linked GitHub issue has repro steps can we use this & add as a test case to make sure this scenario is covered to make sure we have no future regressions?

sure will include and send a new PR.

src/Websites/Websites/Utilities/WebsitesClient.cs
ThejaChoudary: I consider the scenario, when they try to add multiple Host Names in a single object and the 1st HN is added successfully where the 2nd HN is failed to add.In this case we should consider either of the following 
1. Remove the successfully added HN and stop the execution. With this the old Host Names will remain as it is.
2. Continue the code execution if at least one custom domain is added and remove the old custom domains.

Here I choose the 2nd option. Please suggest.

markcowl:
@panchagnula Can you review the responses and comment or sign off?

panchagnula:
closing this since this is covered by https://github.com/Azure/azure-powershell/pull/10972

PR #11003, opened 2020-01-30
adding support for firewall policy on vnet firewall

saisujithreddym:
## Checklist

### DIFF
--- a/src/Network/Network.Test/ScenarioTests/AzureFirewallTests.cs
+++ b/src/Network/Network.Test/ScenarioTests/AzureFirewallTests.cs
@@ -81,7 +81,6 @@ namespace Commands.Network.Test.ScenarioTests
         {
             TestRunner.RunTestScript("Test-AzureFirewallPrivateRangeCRUD");
         }
-
         [Fact]
         [Trait(Category.AcceptanceType, Category.CheckIn)]
         [Trait(Category.Owner, NrpTeamAlias.azurefirewall)]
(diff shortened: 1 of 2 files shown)


src/Network/Network/AzureFirewall/NewAzureFirewallCommand.cs
```diff
@@ -255,22 +255,46 @@ private PSAzureFirewall CreateAzureFirewall()
             }
             else
             {
+
+                if (FirewallPolicyId != null && (this.ApplicationRuleCollection != null || this.NetworkRuleCollection != null || this.NatRuleCollection != null))
```
markcowl: This seems like something that should be enforced through parameter sets - FirewallPolicyId should not be in the same parameter sets that include ApplicationRUleCollection or NetworkRuleCollection

saisujithreddym: I am having trouble with the mutual exclusion scenario @markcowl Place firewall policy id in one parameter set and rule collections in another did not achieve the desired result. 

I have tried various combinations but could not get it done. Do you have an example where we can achieve mutual exclusion using paramter sets? I will keep exploring as well

saisujithreddym: We have decided to fail the call from NRP when there are both the policy and the rule collections. This way we need not make changes on the PS when we try to support policy+collections in case, show a message which is common across multiple platforms and track the usage. I have created another PR: https://github.com/Azure/azure-powershell/pull/11013. I am closing this one

Issue #10965, opened 2020-01-23
Update-AzADApplication overwrites AvailableToOtherTenants even when not provided


psignoret:
## Description

Running `Update-AzADApplication` to update any property of an Application (e.g. `IdentifierUri` or `DisplayName`) will reset `AvailableToOtherTenants` to `false` even if `AvailableToOtherTenants` is not passed as a parameter.

My guess: Because the `AvailableToOtherTenants` parameter is defined as a `bool` rather than `bool?` in [UpdateAzureADApplicationCommand.cs](https://github.com/Azure/azure-powershell/blob/master/src/Resources/Resources/ActiveDirectory/Cmdlets/UpdateAzureADApplicationCommand.cs), it defaults to `false` if the parameter is not set and will be passed as a property in the PATCH request (instead of simply being omitted).

## Steps to reproduce

1. Create an app with `AvailableToOtherTenants` set to `true`:
   
    ```powershell
    New-AzADApplication -DisplayName "Foo app" `
                        -AvailableToOtherTenants $true `
                        -IdentifierUris @("https://phsignor.onmicrosoft.com/foo")
    ```
    ```none
    DisplayName             : Foo app
    ObjectId                : 4bbf9ca2-557a-4ea5-94e8-11a402d0029c
    IdentifierUris          : {https://phsignor.onmicrosoft.com/foo}
    HomePage                :
    Type                    :
    ApplicationId           : c668613d-f01f-4bfc-87ac-aacf673afa67
    AvailableToOtherTenants : True
    AppPermissions          :
    ReplyUrls               : {}
    ObjectType              : Application    
    ```
2. Update the app's DisplayName:

    ```powershell
    Update-AzADApplication -ObjectId "4bbf9ca2-557a-4ea5-94e8-11a402d0029c" `
                           -DisplayName "Bar app"
    ```
    ```none
    DisplayName             : Bar app
    ObjectId                : 4bbf9ca2-557a-4ea5-94e8-11a402d0029c
    IdentifierUris          : {https://phsignor.onmicrosoft.com/foo}
    HomePage                :
    Type                    :
    ApplicationId           : c668613d-f01f-4bfc-87ac-aacf673afa67
    AvailableToOtherTenants : False
    AppPermissions          :
    ReplyUrls               : {}
    ObjectType              : Application
    ```

Expected: Only the DisplayName should change.

Observed: Both DisplayName and AvailableToOtherTenants change.

## Environment data

```
Name                           Value
----                           -----
[...]

PR #11024, opened 2020-02-04
Set null if parameter AvailableToOtherTenants is not set

dingmeng-xue:
Fix issue #10965
If AvailableToOtherTenants is not set, set null for update operation

### DIFF
--- a/src/Resources/Resources/ActiveDirectory/Cmdlets/UpdateAzureADApplicationCommand.cs
+++ b/src/Resources/Resources/ActiveDirectory/Cmdlets/UpdateAzureADApplicationCommand.cs
@@ -102,7 +102,7 @@ namespace Microsoft.Azure.Commands.ActiveDirectory
                     Homepage = HomePage,
                     IdentifierUris = IdentifierUri,
                     ReplyUrls = ReplyUrl,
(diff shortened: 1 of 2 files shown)


src/Resources/Resources/ActiveDirectory/Cmdlets/UpdateAzureADApplicationCommand.cs
```diff
@@ -102,7 +102,7 @@ public override void ExecuteCmdlet()
                     Homepage = HomePage,
                     IdentifierUris = IdentifierUri,
                     ReplyUrls = ReplyUrl,
-                    AvailableToOtherTenants = AvailableToOtherTenants
+                    AvailableToOtherTenants = MyInvocation.BoundParameters.ContainsKey("AvailableToOtherTenants") ? AvailableToOtherTenants : (bool?)null
```
erich-wang: Suggest to use IsParamaterBound to determine whether parameter is provided, just as line 90:
```
AvailableToOtherTenants = this.IsParameterBound(c => c.AvailableToOtherTenants ) ? AvailableToOtherTenants : (bool?)null;
```

PR #11033, opened 2020-02-04
[Compute] Allow empty value for ProximityPlacementGroupId during update

hyonholee:
## Checklist


src/Compute/Compute/Generated/VirtualMachineScaleSet/VirtualMachineScaleSetUpdateMethod.cs
```diff
@@ -251,7 +251,7 @@ public override void ExecuteCmdlet()
 
         [Parameter(
             Mandatory = false)]
-        [ValidateNotNullOrEmpty]
+        [AllowEmptyString]
```
VeryEarly: Can you help me understand why this class doesn't need to set ProximityPlacementGroupId to null like others?

hyonholee: The ask was allow an empty string value for ProximityPlacementGroupId to support the scenario of removing ProximityPlacementGroupId from VM.   In REST request, setting the ID to null or setting the ID to empty string have the same effect, but ProximityPlacementGroup should not be null.  (If ProximityPlacementGroup is null, then no change of ProximityPlacementGroupId).  So, I think we can accept null value here (for ID), but it may be better that a user to give an empty string explicitly to avoid unintended removal.

VeryEarly: understood, but in other two cmdlets you changed, after [AllowEmptyString], there are if conditions to set
"avSetParams.ProximityPlacementGroup.Id = null;"
and 
"parameters.ProximityPlacementGroup.Id = null;"

I'm just wondering why this cmdlet doesn't have such thing.

src/Compute/Compute/AvailabilitySets/UpdateAzureAvailabilitySetCommand.cs
```diff
@@ -79,6 +79,11 @@ public override void ExecuteCmdlet()
                                                 : this.AvailabilitySet.ProximityPlacementGroup
                     };
 
+                    if (avSetParams.ProximityPlacementGroup != null && string.IsNullOrEmpty(avSetParams.ProximityPlacementGroup.Id))
+                    {
+                        avSetParams.ProximityPlacementGroup.Id = null;
```
VeryEarly: correct me if I was wrong, this line is to ensure the value of Id is null not white spaces, am I right?

src/Compute/Compute/Generated/VirtualMachineScaleSet/VirtualMachineScaleSetUpdateMethod.cs
hyonholee: Thanks.  I updated the PR by adding the same logic to VMSS.

src/Compute/Compute/AvailabilitySets/UpdateAzureAvailabilitySetCommand.cs
hyonholee: Yes.  It ensures to send a null value for ID, which removes it from VM/VMSS/AvSet.

PR #11062, opened 2020-02-08
Add-AzServiceFabricClusterCertificate: Add support for new certificate with same common name

jcleavel:
## Description

Currently, the `Add-AzServiceFabricClusterCertificate` cmdlet fails to add a new certificate to the cluster if the supplied common name is the same as the one currently installed on the cluster. This does not align with how Service Fabric handles common name-based certificate rotation.

Cmdlet argument example:
```powershell
Add-AzServiceFabricClusterCertificate `
    -ResourceGroupName 'cluster-rg' `
    -Name 'fake-cluster' `
    -SecretIdentifier 'https://VAULT-NAME.vault.azure.net/secrets/CERT-NAME/VERSION' `
    -CertificateIssuerThumbprint 'XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX' `
    -CertificateCommonName 'common-name.fake-u.rl'
```

The cmdlet fails to run for the following reasons:
* A tangentially-related `NullReferenceException` would throw when the cmdlet is called on cluster without a `certificate` object in its resource properties. This object isn't necessary in clusters secured with common name-referenced certificates. Fixed by adding appropriate `null` handling and correcting what criteria constitute a "Secure" cluster.
* The ARM API would reject a PATCH operation to the cluster resource if the new `certificateCommonNames.commonNames[]` array contained multiple items with the same value in the `certificateCommonName` property. Fixed by de-duping and overwriting the duplicate value with the incoming value(s).

The following _problematic but non-fatal_ behavior was also corrected:
* The cmdlet would add duplicate values to the `settings.certificate.commonNames[]` array on the cluster VMSS's respective Service Fabric VMSS Extensions. This doesn't appear to cause a problem, but it's definitely confusing and messy. Fixed by case-insensitive de-duping.

## Checklist

### DIFF
--- a/src/ServiceFabric/ServiceFabric/ChangeLog.md
+++ b/src/ServiceFabric/ServiceFabric/ChangeLog.md
@@ -18,6 +18,10 @@
         - Additional information about change #1
 -->
 ## Upcoming Release
+* Modified the `Add-AzServiceFabricClusterCertificate` cmdlet to allow adding new certificates with the same common name as the existing certificate by fixing the following bugs:
+    - A `NullReferenceException` would throw when cmdlet called on cluster without a `certificate` object in its resource properties.
+    - Cmdlet would throw when attempting to add multiple `certificateCommonNames.commonNames[*]` items with the same `certificateCommonName` property value.
+    - Cmdlet would add duplicate values to the `settings.certificate.commonNames[]` array on the Service Fabric VMSS Extension.
 
 ## Version 2.0.1
 * Update references in .psd1 to use relative path


VeryEarly:
Thanks for your contribution, changes look good to me, it would be great if you can add some test cases to cover these new logic.

jcleavel:
@VeryEarly: I would love to, but I'm afraid that I can't quite figure out how to write more tests without disrupting the existing ones.

Altern1ty:
We are currently blocked from updating our SF clusters due to this bug in the cmdlet. Can this code please be merged into a release so customers like us can make use of it? I appreciate wanting test cases but I would like to see this merged sooner rather than later so it is available to customers like us.

a-santamaria:
@jcleavel sorry I just saw this pr. Thanks for finding and looking at this issue, the cmdlet was intended to add new common name certs to the cluster resources. If the common name is already defined in the cluster and the need is to roll it over, the cert should be deployed to the vmss resource using Add-AzVmssSecret as described [here](https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-cluster-rollover-cert-cn) which will install the new certs in the vms and since the common name is the same as the older one the cluster will automatically pick up the one with the furthest expiration date. The problem with this cmdlet is we should check this and throw with the correct message also update the documentation to explain better the use of this cmdlet for common name. Let me know if you want to continue with this pr otherwise I'll create a new one with these changes.

adxsdkps:

jcleavel:
What would be the point of throwing an error in that scenario? I can't think of a legitimate operational reason such an error would be useful to me as a user.
Why can't the command both perform the task that you describe above as well perform routine certificate renewals?

a-santamaria:
Because that operation would be only adding a cert to the vmss resource it would not make any change to the service fabric resource. That scenario is already covered by Add-AzVmssSecret.

jcleavel:
Then why does the command add the new cert to the VM Scale Sets?
Following that logic, the command would only add the new common name to the cluster definition and leave it to the user to later add the new cert to the VMSS resources using the `Add-AzVmssSecret` command.

a-santamaria:
yes, you're right. Then we can do it here but we shouldn't start a patch against the cluster. if the common name is already defined in the cluster it should only do the vmss update

dragav:
Add-AzServiceFabricClusterCertificate `
    -SecretIdentifier 'https://VAULT-NAME.vault.azure.net/secrets/CERT-NAME/VERSION' `
    -CertificateIssuerThumbprint 'XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX' `
    -CertificateCommonName 'common-name.fake-u.rl'

Rule #1 is you can't declare a cert by CN if it's provisioned by version, or vice-versa: don't provision a cert by instance (versioned KV uri) for a cert declared by CN.

Cluster certificates are not mere declarations - they are directives for the runtime on which certificate to use. Upcoming changes in SF71 will block cluster config upgrades when the cert declarations are not met by the VMSS state.

If you want to add/declare a _client_ certificate by CN, that's all fine.

dragav:
Sorry about closing/re-opening the PR, for a moment I thought this was an issue. While we appreciate the assist, I would strongly prefer that we retain control over what our cmdlets do. If they have shortcomings or bugs, please track them with issues - we do monitor those (and even fix some).

jcleavel:
That's alright, @dragav. I'll go ahead and close this PR since the prospect of me getting anywhere on it is very low.

We're getting ready to start the deprecation of our Service Fabric infrastructure because of stuff like this and our systems can get by with a combination of Terraform template updates and `az` CLI scripts until that time.

jcleavel:
I would hope that your team at least fix the `NullReferenceException`s / ARM API bugs and/or expand the documentation on this cmdlet so users know what this cmdlet is _actually for_.
The sparse documentation states that the cmdlet is for adding (or overwriting an existing secondary) cluster certificate, and that only seems to be true for some of all the potential operations it's capable of. Designing the cmdlet to do what the documentation/name indicates while being more "forgiving" of the input criteria leads to a better user experience.

dragav:
Thanks @jcleavel, we appreciate the input, and will most likely incorporate some of your changes. Documentation is, indeed, a problem, and so is the promise of this cmdlet - which is why we're trying to deprecate it.

Issue #10619, opened 2019-11-27
New-AzDeploymentWhatIf should be -WhatIf switch, not new cmdlet


oising:
Whomever suggested "WhatIf" behaviour for New-AzDeployment most likely did not envisage a new cmdlet, but rather the addition of the far more idiomatic -WhatIf switch to the existing cmdlet where 99% of PowerShell users would expect to find it. 

TL;DR

Surface this as a -WhatIf switch and parameterset on existing New-AzDeployment. Do not create a new cmdlet. 

UPDATE: Alternatively, as @bgelens has suggested below, rename to `Get-AzDeploymentPlan` if otherwise constrained.

vexx32:
Yes please! Following the established PowerShell patterns and practices from base PowerShell modules and the wider community will make for far less mistakes, especially where implementing `ShouldProcess` support comes in.

And just to be clear -- this should _not_ be a custom `-WhatIf` switch; the automatic switch that comes with `[Cmdlet(SupportsShouldProcess = true)]` or `[CmdletBinding(SupportsShouldProcess)]` should be used instead, along with proper checks for [`Cmdlet.ShouldProcess()` ](https://docs.microsoft.com/en-us/dotnet/api/system.management.automation.cmdlet.shouldprocess?view=pscore-6.2.0)

billkindle:
Gonna toss in my opinion here as well. Not using existing patterns and practices sets a bad precedent. What happens here is we are tossing out a decade of existing scripting knowledge that many a sysadmin like myself know by now but may not have begun using Azure just yet. I for one expect some consistency with my parameter sets, and -WhatIf is one I expect to always be there. 

Besides, if someone wanted a new cmdlet to do this, why couldn’t they just configure an alias?

VeryEarly:
Hi thanks for the feedback, we will take a look into this request.

bgelens:
How about `New-AzDeploymentPlan` or `Get-AzDeploymentPlan` (I like the Get a little bit better)? It would follow the Terraform terminology (what people who are looking for it are expecting I think).

As a side note, I'm not sure if the command currently returns only text or objects? I hope an object is part of what is returned as I imagine some programmatic usecases for it. E.g.

```powershell
$result = Get-AzDeploymentPlan
if ($result.PlannedForDeletion) {
  # maybe inquire if we can continue
} elseif ($result.PlannedForChange) {
  # start New-AzDeployment, no need to inquiry
} else {
[...]

PR #11228, opened 2020-03-03
[Resources] [ARM] Add cmdlets for ARM template What-If.

shenglol:
## Description

- Added two new cmdles for ARM template What-If operation:
    - `Get-AzDeploymentWhatIf` (alias: `Get-AzSubscriptionDeploymentWhatIf`)
    - `Get-AzResourceGroupDeploymentWhatIf`

- Overrode `-WhatIf` and `-Confirm` switch parameters for `New-AzDeploymentWhatIf` and `New-AzResourceGroupWhatIf` to use ARM template What-If output.

**Notes for code reviewers**:
We would like to get a preview release for `Az.Resources` at the end of this month. There are some improvements we need to make on our back-end before we can merge this PR.

## Checklist

### DIFF
--- a/.azure-pipelines/powershell-core.yml
+++ b/.azure-pipelines/powershell-core.yml
@@ -4,7 +4,7 @@ variables:
   LinuxName: linux
   LinuxImage: ubuntu-18.04
   MacOSName: macOS
-  MacOSImage: macOS-10.13
+  MacOSImage: macOS-10.14
   TestFramework: netcoreapp2.2
   TestTarget: Test
   Configuration: Debug


adxsdkps:

kilasuit:
Do not merge this

You should not ever suffix a command with WhatIf and the design of this, and its functionality realistically should be a component of the core Test commands `Test-AzDeployment` `Test-AzManagementGroupDeployment` `Test-AzResourceGroupDeployment` or `Test-AzTenantDeployment` not a Get Command

shenglol:
> Please Do not merge this
> 
> You should not ever suffix a command with WhatIf and the design of this, and its functionality realistically should be a component of the core Test commands `Test-AzDeployment` `Test-AzManagementGroupDeployment` `Test-AzResourceGroupDeployment` or `Test-AzTenantDeployment` not a Get Command
> 
> I would strongly urge you redesign this as functionality into the above mentioned commands cc @markcowl @sphibbs

Hi @kilasuit , thanks for the feedback.

For your reference, [ARM Template deployment What-If](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-deploy-what-if) is a feature currently in private preview. It basically tells you what would happen when you deploy a template if it is valid. The `Test-*AzDeployment` cmdlets serve a different purpose. It will only validate a template and its parameters.

We already had a internal discussion with @markcowl and decided to go with a separate cmdlet. There's an existing issue https://github.com/Azure/azure-powershell/issues/10619 about the cmdlet name, and the `Get` verb is actually based on suggestions we got there.

cc @alex-frankel @majastrz .

alex-frankel:
I don't have a strong opinion on Get vs Test, but this is not what Test returns today. Should we do Test-AzDeployment -WhatIf to return the object? Will that be confusing that -WhatIf for `Test-Az*Deployment` returns an object while -WhatIf for `New-Az*Deployment` returns text?

majastrz:
It sounds like the problem is with `WhatIf` not being a noun. @kilasuit, how about calling that cmdlet `Get-Az*DeploymentWhatIfResult`.

shenglol:
Hi @isra-fel , do you mind taking a look since you reviewed it before? There's not many changes compared to the version that got reverted a few months before.

kilasuit:
> It sounds like the problem is with WhatIf not being a noun. @kilasuit, how about calling that cmdlet Get-Az*DeploymentWhatIfResult.

I personally don't like that all the much more than what is proposed already, but I do prefer that to the current proposal as it reads better,
 
> I don't have a strong opinion on Get vs Test, but this is not what Test returns today. Should we do Test-AzDeployment -WhatIf to return the object? Will that be confusing that -WhatIf for `Test-Az*Deployment` returns an object while -WhatIf for `New-Az*Deployment` returns text?

See this makes makes more sense when you think what it is the end user is doing, in this case they would want to test their current templates will deploy and then optionally test what the template they are going to deploy are going to change from what is already deployed. Which to me is an extension to the Test cmdlets and not a new cmdlet completelty. Therefore I think that overloading the `WhatIf` on `Test` to return an object, at least to me feels more natural & optimal for those running this process in the console rather than having to resort to cmdlet change.

The optimal flow (simplified), in my opinion when working from the console would look like this

```powershell
Test-AzDeployment
Test-AzDeployment -WhatIf
New-AzDeployment
``
Which is only a few actual key presses (thanks to console history and tab completion)

 compared to this 
```powershell
Test-AzDeployment
Get-AzDeploymentWhatIfResult.
New-AzDeployment
```
Hope this helps explain my view on this

bergmeister:
The -WhatIf switch should be on the New-AzDeployment cmdlet, not its Test-AzDeployment counterpart. As a user the -WhatIf switch tells me what would happen if I executed the cmdlets without the -WhatIf switch, hence why it's more appropriate on the New- cmdlet rather than the Test- one.

kilasuit:
> The -WhatIf switch should be on the New-AzDeployment cmdlet, not its Test-AzDeployment counterpart. As a user the -WhatIf switch tells me what would happen if I executed the cmdlets without the -WhatIf switch, hence why it's more appropriate on the New- cmdlet rather than the Test- one.

Generally in other areas I would agree with you, however as we have `Test` cmdlets in this module I think this feels more appropriate on them than on `New` because you are not technically initiating a full new deployment but are testing what would change to already existing deployed resources if you did, hence why I think its should be on Test. 

Though I suppose that you could argue that this functionality it should actually be on both `New` & `Test` cmdlets, though I know that generally I would be using `WhatIf` in template testing stages (i.e. local development) not template deployment via my CD Pipelines. 
The difference in mindsets in the two, whilst subtle, make it worth while having this functionality in both for authors to pick which they feel is more appropriate.

But it definitely should not be another _unnecessary_ cmdlet

shenglol:
> > The -WhatIf switch should be on the New-AzDeployment cmdlet, not its Test-AzDeployment counterpart. As a user the -WhatIf switch tells me what would happen if I executed the cmdlets without the -WhatIf switch, hence why it's more appropriate on the New- cmdlet rather than the Test- one.
> 
> Generally in other areas I would agree with you, however as we have `Test` cmdlets in this module I think this feels more appropriate on them than on `New` because you are not technically initiating a full new deployment but are testing what would change to already existing deployed resources if you did, hence why I think its should be on Test.
> 
> Though I suppose that you could argue that this functionality it should actually be on both `New` & `Test` cmdlets, though I know that generally I would be using `WhatIf` in template testing stages (i.e. local development) not template deployment via my CD Pipelines.
> The difference in mindsets in the two, whilst subtle, make it worth while having this functionality in both for authors to pick which they feel is more appropriate.
> 
> But it definitely should not be another _unnecessary_ cmdlet

The separate cmdlet is useful for users who what to parse the what-if result programmatically, e.g.:
```pwsh
$whatIfResult = Get-AzResourceGroupWhatIf ...
# do whatever you want with $whatIfResult
```

brwilkinson:
I would just like one single cmdlet for everything.

New-AzDeployment

Then depending on the parameters it will deploy at different scopes.... then depending on -whatif or not it will actually deploy or tell me the results of the whatif.

I prefer no separate cmdlets for different deployment scope deployments and no different cmdlets for a test deploy.

It's MUCH easier to modify a set of parameters that can be splatted into the single Cmdlet than having to call the different Cmdlets for the different scenarios.

alex-frankel:
@kilasuit and @bergmeister: 

The clean, text-based WhatIf response (which includes the validation check that the Test* does) will trigger when you do `New-Az*Deployment -WhatIf` or `New-Az*Deployment -Confirm`. So we are expecting most dev/test loops to simply be:

`New-Az*Deployment -Confirm`

or

`New-Az*Deployment -WhatIf`
`New-Az*Deployment`

Personally, I have been enjoying just using `-Confirm` and, depending on the response, I select `Y` or `N`

Only if you need to *programmatically* interrogate the results, you would do `Get-Az*DeploymentWhatIfResult`


@brwilkinson - we did start by trying to unify this all into New-AzDeployment with different parameter sets, but it made the cmdlet autocomplete experience somewhat confusing in certain edge cases because the parameter sets are too different (e.g. Sub, Mg and Tenant deployments all require a location, but RG does not)

We're not against unifying, but we need to spend more time getting the usability right.

alex-frankel:
@isra-fel - can you or someone on your team please take a look at this?

src/Resources/Resources/ChangeLog.md
```diff
@@ -30,6 +30,10 @@
     - Update-AzTag -ResourceId
 * Brought ScopedDeployment from SDK 3.3.0 
 * Breaking changes for policy cmdlets to support typed pipeline objects
+* Add Get-AzDeploymentWhatIf cmdlet for getting ARM template What-If at subscription scope.
```
isra-fel: Please move new contents after "Upcoming Release", our tools for release depend on this structure

shenglol: Missed this one after resolving merge conflicts. Pushed out a fix.

isra-fel:
Hi, we have a commit https://github.com/Azure/azure-powershell/commit/bc5776362805015341494addfa9432633a2442c0 on master to update the mac os image version in our CI checks. The resources preview branch hasn't sync with master for quite some time, that's why the CI is constantly failing.
The simplest way I think is to cherry-pick that commit to your branch. Could you do that? Thanks

shenglol:
> Hi, we have a commit [bc57763](https://github.com/Azure/azure-powershell/commit/bc5776362805015341494addfa9432633a2442c0) on master to update the mac os image version in our CI checks. The resources preview branch hasn't sync with master for quite some time, that's why the CI is constantly failing.
> The simplest way I think is to cherry-pick that commit to your branch. Could you do that? Thanks

That's good to know! I've cherry-picked the commit.

shenglol:
Hi @isra-fel, can you please merge this?

PR #11260, opened 2020-03-06
Merge release-2020-03-10 to master

dingmeng-xue:


adxsdkps:

src/Attestation/Attestation/ChangeLog.md
```diff
@@ -20,9 +20,11 @@
 
 
 ## Upcoming Release
+* Added three additional required assemblies to Az.Attestation.psd1
```
isra-fel: This line is actually in S167 not S166, so you kept in under "upcoming release", right?

dingmeng-xue: yes

PR #11263, opened 2020-03-06
Erich/msal disable auto refresh context

erich-wang:
## Description
1. disable auto refresh context from token cache
2. allow to refresh context from token cache using Get-AzContext -ListAvailable -RefreshContextFromTokenCache

## Checklist


adxsdkps:

src/Accounts/Accounts/Context/GetAzureRMContext.cs
```diff
@@ -63,13 +64,32 @@ protected override IAzureContext DefaultContext
         [Parameter(Mandatory =true, ParameterSetName = ListAllParameterSet, HelpMessage ="List all available contexts in the current session.")]
         public SwitchParameter ListAvailable { get; set; }
 
+        [Parameter(Mandatory = false, ParameterSetName = ListAllParameterSet, HelpMessage = "Refresh contexts from token cache")]
+        public SwitchParameter RefreshContextFromTokenCache { get; set; }
```
isra-fel: I just want to confirm that eventually we shall get rid of this parameter. Is that correct?

isra-fel:
Hi Erich, thanks for the PR.
I'm just confused with one thing: I thought the idea is to disable refresh context unless one situation: user passes in a flag to force doing that, right?
If that's true, then I don't get the logic in `ConnectAzureRmAccount.cs` to store/restore the original refreshcontext...
Also there's still [one place](https://github.com/Azure/azure-powershell/blob/3e509a35938d673f9a1929aa8b88dca2660cb6a1/src/Accounts/Authentication.ResourceManager/AzureRmProfile.cs#L241) we assign `true` to refreshcontext.... I think we should change it to `false`.

erich-wang:
> Hi Erich, thanks for the PR.
> I'm just confused with one thing: I thought the idea is to disable refresh context unless one situation: user passes in a flag to force doing that, right?
> If that's true, then I don't get the logic in `ConnectAzureRmAccount.cs` to store/restore the original refreshcontext... because it will always be `false`.
> Also there's still [one place](https://github.com/Azure/azure-powershell/blob/3e509a35938d673f9a1929aa8b88dca2660cb6a1/src/Accounts/Authentication.ResourceManager/AzureRmProfile.cs#L241) we assign `true` to refreshcontext.... I think we should change it to `false`.

Hi Yeming, thanks for your comments. First of all, the change in this PR is just short-term solution, we finally need to figure out long term solution to support SSO.

After this PR, user could only refresh context from token cache by running
```
Get-AzContext -ListAvailable -RefreshContextFromTokenCache
```
For the property **ShouldRefreshContextFromCache**, I leave them on purpose:
1. Leave one tricky way for us for debug purpose
2. It is helpful information when implementing final SSO solution.

Issue #11018, opened 2020-02-03
Az.DataLakeStore New-AzDataLakeStoreItem doesn't allow uploading Byte[] content


tedhardyMSFT:
## Description
Migrating from AzureRm to Az for DataLakeStore operations. In AzureRm, we are able to upload compressed data using a byte[] for the -Value parameter. When using passing a byte[] for the -Value for the New-AzDataLakeStoreItem (or Add-AzDataLakeStoreItemContent) cmdlet, I get an invalid Encoding type.

## Steps to reproduce

```powershell
## login and get content, compress using Gzip into byte array.

[byte[]] $compressedData = Compress-StringGzip -inputString $FileData;

## upload to Adls
New-AzDataLakeStoreItem -Encoding Byte -Account $AzureDataLakeStorageAccount -Path $DestinationFileName -Value $compressedData -ErrorAction Stop;
```

## Environment data

```
Name                           Value
----                           -----
PSVersion                      5.1.18362.145
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.18362.145
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

## Module versions

```powershell
PS C:\> get-module -listavailable                                                                                       

    Directory: C:\Program Files\WindowsPowerShell\Modules


ModuleType Version    Name                                ExportedCommands
---------- -------    ----                                ----------------
Script     1.7.0      Az.Accounts                         {Disable-AzDataCollection, Disable-AzContextAutosave, Enab...
Script     1.6.2      Az.Accounts                         {Disable-AzDataCollection, Disable-AzContextAutosave, Enab...
Script     1.1.1      Az.Advisor                          {Get-AzAdvisorRecommendation, Enable-AzAdvisorRecommendati...
Script     1.0.1      Az.Advisor                          {Get-AzAdvisorRecommendation, Enable-AzAdvisorRecommendati...
Script     1.0.3      Az.Aks                              {Get-AzAks, New-AzAks, Remove-AzAks, Import-AzAksCredentia...
Script     1.0.2      Az.Aks                              {Get-AzAks, New-AzAks, Remove-AzAks, Import-AzAksCredentia...
Script     1.1.2      Az.AnalysisServices                 {Resume-AzAnalysisServicesServer, Suspend-AzAnalysisServic...
[...]

Issue #10667, opened 2019-12-06
One test case of DataLakeStore module constantly fails randomly at linux environment only.


VeryEarly:
## Description
This single test case `Microsoft.Azure.Commands.DataLakeStore.Test.ScenarioTests.AdlsAliasTests.TestAdlsFileSystem`, fails randomly time and time again. This affects our efficiency to review PRs.

Need to investigate.

Sample PR: https://github.com/Azure/azure-powershell/pull/10755
CI job: https://dev.azure.com/azure-sdk/public/_build/results?buildId=231953
Log: https://dev.azure.com/azure-sdk/public/_build/results?buildId=231953&view=logs&j=ad07b744-5a1a-5d54-7693-fcb02e666633&t=b0841cae-a6b1-5a5c-4ebd-3cc6b3d3163d&l=704

Log snippet:
```
  [xUnit.net 00:00:15.65]     Microsoft.Azure.Commands.DataLakeStore.Test.ScenarioTests.AdlsAliasTests.TestAdlsFileSystem [FAIL]
    X Microsoft.Azure.Commands.DataLakeStore.Test.ScenarioTests.AdlsAliasTests.TestAdlsFileSystem [2s 950ms]
EXEC : error Message:  [/home/vsts/work/1/s/build.proj]
     System.Management.Automation.ActionPreferenceStopException : The running command stopped because the preference variable "ErrorActionPreference" or common parameter is set to Stop: Object reference not set to an instance of an object.
    Stack Trace:
       at System.Management.Automation.Runspaces.PipelineBase.Invoke(IEnumerable input)
     at System.Management.Automation.Runspaces.Pipeline.Invoke()
     at System.Management.Automation.PowerShell.Worker.ConstructPipelineAndDoWork(Runspace rs, Boolean performSyncInvoke)
     at System.Management.Automation.PowerShell.Worker.CreateRunspaceIfNeededAndDoWork(Runspace rsToUse, Boolean isSync)
     at System.Management.Automation.PowerShell.CoreInvokeHelper[TInput,TOutput](PSDataCollection`1 input, PSDataCollection`1 output, PSInvocationSettings settings)
     at System.Management.Automation.PowerShell.CoreInvoke[TInput,TOutput](PSDataCollection`1 input, PSDataCollection`1 output, PSInvocationSettings settings)
     at System.Management.Automation.PowerShell.CoreInvoke[TOutput](IEnumerable input, PSDataCollection`1 output, PSInvocationSettings settings)
[...]

PR #11355, opened 2020-03-18
Added option of Byte encoding for New-AzDataLakeStoreItem, Add-AzDAtaLakeStoreItemContent, Get-AzDAtaLakeStoreItemContent

rahuldutta90:
Added option of Byte encoding for New-AzDataLakeStoreItem, Add-AzDAtaLakeStoreItemContent, Get-AzDAtaLakeStoreItemContent. Fixes https://github.com/Azure/azure-powershell/issues/11018

## Checklist

### DIFF
--- a/src/DataLakeStore/DataLakeStore/DataPlaneModels/DataLakeStoreFileSystemCmdletBase.cs
+++ b/src/DataLakeStore/DataLakeStore/DataPlaneModels/DataLakeStoreFileSystemCmdletBase.cs
@@ -101,7 +101,13 @@ namespace Microsoft.Azure.Commands.DataLakeStore.Models
                 {
                     return byteArray;
                 }
-
+                /*
+                 *  [byte[]] $byteData = 1,2,3,4,5
+                 * $MyList = [System.Collections.Generic.List[object]]::new()
+                 * $MyList.Add($byteData[0])
+                 * $MyList.Add($byteData[1])
+                 * And then pass $MyList.ToArray() this will pass object[] containing bytes
+                 */
                 // attempt to convert the object into an object array	
                 var contentArray = content as object[];
                 if (contentArray == null)
--- a/src/DataLakeStore/DataLakeStore/DataPlaneCommands/AddAzureRmDataLakeStoreItemContent.cs
+++ b/src/DataLakeStore/DataLakeStore/DataPlaneCommands/AddAzureRmDataLakeStoreItemContent.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Azure.Commands.DataLakeStore
             HelpMessage =
                 "Optionally indicates the encoding for the content being uploaded as part of 'Value'. Default is UTF8")]
         [ArgumentToEncodingTransformation]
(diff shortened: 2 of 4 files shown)


rahuldutta90:
design PR: https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/519

src/DataLakeStore/DataLakeStore/DataPlaneModels/DataLakeStoreFileSystemCmdletBase.cs
```diff
@@ -85,13 +93,46 @@ private static byte[] GetBytes(string content, Encoding encoding)
...
             {
-                throw new CloudException(Resources.InvalidContent);
+                // first attempt to convert it directly into a byte array	
+                var byteArray = content as byte[];
+                if (byteArray != null)
+                {
+                    return byteArray;
+                }
+
+                // attempt to convert the object into an object array
```
akharit: Can you add a comment on cases when this will work when above fails?

src/DataLakeStore/DataLakeStore.Test/ScenarioTests/AdlsTests.cs
```diff
@@ -66,13 +66,13 @@ public void TestAdlsAccountTiers()
             NewInstance.RunPsTest(_logger, string.Format("Test-DataLakeStoreAccountTiers -location '{0}'", AdlsTestsBase.ResourceGroupLocation));
         }
 
-        [Fact(Skip="This case has been unstable on Linux for approximately 20% of failure. Needs investigation.")]
+        [Fact(Skip = "This case has been unstable on Linux for approximately 20% of failure. Needs investigation.")]
```
isra-fel: This test case was previously skipped because it fails randomly in Linux environment. Details are in #10667
Please enable it and fix the error. Thanks

rahuldutta90: @isra-fel I did see you skipped it, however i could not reproduce it in my linux VM. So I still do not understand the exact root cause of it, however i feel why it might be happening. But thing is the failure is for adlsaliastest and you have skipped it for adlstest.

Now if you see there are two versions of test AdlsAliasTest and AdlsTest. Both are exact replicas functionally, only difference is one of them uses alias commandlet names. And in most of the logs i see "Export-AdlStoreChildItemProperties" failing for alias. As a part of this commandlet we write output to a local path. So i suspect adlstest and adlsaliastest are interfering with each other. Reason why i think is because adlsaliastest was the one failng, but you disabled adlstest and it started passing more. I still do not know why they are interfering. 

What I will do is try to comment out "Export-AdlStoreChildItemProperties" in adlsaliastest.testfilesystem (since thats the one failing) and enable the adlstest.testfilesystem. But I dont think this PR is the correct place to fix this since this adresses a different bug. 

I will send out a different PR after this with the proposed change.

isra-fel: Thanks for all your explanation, it makes sense to me.

PR #11368, opened 2020-03-19
Support Identity, Encryption, UserOwnedStorage

yangyuan:
## Description

Support Identity, Encryption, UserOwnedStorage

## Checklist

### DIFF
--- a/src/CognitiveServices/CognitiveServices/CognitiveServicesAccount/SetAzureCognitiveServicesAccount.cs
+++ b/src/CognitiveServices/CognitiveServices/CognitiveServicesAccount/SetAzureCognitiveServicesAccount.cs
@@ -17,6 +17,7 @@ using Microsoft.Azure.Commands.Management.CognitiveServices.Properties;
 using Microsoft.Azure.Commands.ResourceManager.Common.ArgumentCompleters;
 using Microsoft.Azure.Management.CognitiveServices;
 using Microsoft.Azure.Management.CognitiveServices.Models;
+using Microsoft.WindowsAzure.Commands.Utilities.Common;
 using Newtonsoft.Json.Linq;
 using System;
 using System.Collections;
@@ -30,9 +31,19 @@ namespace Microsoft.Azure.Commands.Management.CognitiveServices
     /// <summary>
     /// Update a Cognitive Services Account (change SKU, Tags)
     /// </summary>
-    [Cmdlet("Set", ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "CognitiveServicesAccount", SupportsShouldProcess = true), OutputType(typeof(CognitiveServicesModels.PSCognitiveServicesAccount))]
+    [Cmdlet("Set", ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "CognitiveServicesAccount", SupportsShouldProcess = true, DefaultParameterSetName = CognitiveServicesEncryptionParameterSet), OutputType(typeof(CognitiveServicesModels.PSCognitiveServicesAccount))]
     public class SetAzureCognitiveServicesAccountCommand : CognitiveServicesAccountBaseCmdlet
     {
+        /// <summary>
+        /// CognitiveServices Encryption parameter set name
+        /// </summary>
+        private const string CognitiveServicesEncryptionParameterSet = "CognitiveServicesEncryption";
+
+        /// <summary>
+        /// KeyVault Encryption parameter set name
+        /// </summary>
+        private const string KeyVaultEncryptionParameterSet = "KeyVaultEncryption";
+
         [Parameter(
             Position = 0,
             Mandatory = true,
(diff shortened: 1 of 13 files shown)


src/CognitiveServices/CognitiveServices/CognitiveServicesAccount/SetAzureCognitiveServicesAccount.cs
```diff
@@ -74,6 +85,51 @@ public class SetAzureCognitiveServicesAccountCommand : CognitiveServicesAccountB
...
+        [Parameter(
+            Mandatory = false,
+            HelpMessage = "Generate and assign a new Cognitive Services Account Identity for this storage account for use with key management services like Azure KeyVault.")]
+        public SwitchParameter AssignIdentity { get; set; }
+
+        [Parameter(
+            Mandatory = false,
+            ValueFromPipelineByPropertyName = false)]
+        [ValidateNotNullOrEmpty]
+        public IdentityType? IdentityType { get; set; }
```
isra-fel: It should just be of type `IdentityType`

isra-fel: `ValueFromPipelineByPropertyName = false` is the default, you don't need to specify that

src/CognitiveServices/CognitiveServices/CognitiveServicesAccount/SetAzureCognitiveServicesAccount.cs
```diff
@@ -74,6 +85,51 @@ public class SetAzureCognitiveServicesAccountCommand : CognitiveServicesAccountB
...
+
+        [Parameter(HelpMessage = "List of User Owned Storage Accounts.", Mandatory = false)]
+        [ValidateNotNull]
+        [AllowEmptyCollection]
+        public string[] StorageAccountIds { get; set; }
```
isra-fel: PowerShell is strict on the single/plural format of param name -- only when the length of it's value must > 1 can it be plural, so please use `StorageAccountId` instead

src/CognitiveServices/CognitiveServices/CognitiveServicesAccount/SetAzureCognitiveServicesAccount.cs
```diff
@@ -74,6 +85,51 @@ public class SetAzureCognitiveServicesAccountCommand : CognitiveServicesAccountB
...
+
+        [Parameter(HelpMessage = "Whether to set Cognitive Services Account Encryption KeySource to Microsoft.CognitiveServices or not.", Mandatory = false, ParameterSetName = CognitiveServicesEncryptionParameterSet)]
+        public SwitchParameter CognitiveServicesEncryption { get; set; }
+
+        [Parameter(HelpMessage = "Whether to set Cognitive Services Account encryption keySource to Microsoft.KeyVault or not. " +
+            "If you specify KeyName, KeyVersion and KeyVaultUri, Cognitive Services Account Encryption KeySource will also be set to Microsoft.KeyVault weather this parameter is set or not.",
+            Mandatory = false, ParameterSetName = KeyVaultEncryptionParameterSet)]
+        public SwitchParameter KeyVaultEncryption { get; set; }
```
isra-fel: These two switch params are used to distinguish param sets, so they are **mandatory** in each param set's perspective.

src/CognitiveServices/CognitiveServices/CognitiveServicesAccount/SetAzureCognitiveServicesAccount.cs
yangyuan: I removed `ValueFromPipelineByPropertyName = false`
IdentityType is an enum, and `IdentityType.None` doesn't equal to IdentityType is null.
(`IdentityType.None` means remove Identity, while IdentityType is null means leave it not changed.) I removed ValidateNotNullOrEmpty in case it cases confusing.

src/CognitiveServices/CognitiveServices/CognitiveServicesAccount/SetAzureCognitiveServicesAccount.cs
yangyuan: Made the change.

src/CognitiveServices/CognitiveServices/CognitiveServicesAccount/SetAzureCognitiveServicesAccount.cs
yangyuan: What we are trying to do is exact same as StoargeAccount: https://github.com/Azure/azure-powershell/blob/master/src/Storage/Storage.Management/StorageAccount/SetAzureStorageAccount.cs

CognitiveServicesEncryption and KeyVaultEncryption can both be false, if I make them mandatory then at least one of them has to be true.

isra-fel: Makes sense

src/CognitiveServices/CognitiveServices/CognitiveServicesAccount/SetAzureCognitiveServicesAccount.cs
isra-fel: As the parameter is optional, if user doesn't specify it then its value is `null`. That's why we don't need null-able type parameters.
Please add helpmessage too

src/CognitiveServices/CognitiveServices/help/Set-AzCognitiveServicesAccount.md
```diff
@@ -90,6 +131,82 @@ Accept pipeline input: False
 Accept wildcard characters: False
 ```
 
+### -IdentityType
+{{ Fill IdentityType Description }}
```
isra-fel: After adding help message, regenerate help doc to get rid of this

msJinLei:
@yangyuan Cannot review your commits. Please resolve the conflicts.

yangyuan:
@msJinLei  looks like the master has been force pushed!?

yangyuan:
@msJinLei, @isra-fel 
Looks like master has been force pushed. I've updated my branch accordingly, without any change.

PR #11644, opened 2020-04-20
[Resources] Add support for showing formatted deployment errors

filizt:
## Description

This PR has the changes to support showing formatted and relevant deployment errors. 

## Checklist

### DIFF
--- a/src/Resources/ResourceManager/Implementation/ResourceGroupDeployments/GetAzureResourceGroupDeploymentOperationCmdlet.cs
+++ b/src/Resources/ResourceManager/Implementation/ResourceGroupDeployments/GetAzureResourceGroupDeploymentOperationCmdlet.cs
@@ -78,7 +78,7 @@ namespace Microsoft.Azure.Commands.ResourceManager.Cmdlets.Implementation
                 getNextPage: nextLink => this.GetNextLink<JObject>(nextLink),
                 cancellationToken: this.CancellationToken,
                 action: resources => this.WriteObject(sendToPipeline: resources.CoalesceEnumerable().SelectArray(resource => 
-                    resource.ToDeepConvertedPsObject("System.Management.Automation.PSCustomObject#DeploymentOperation")), enumerateCollection: true));
+                    resource.ToPsObjectByDeepConvert("System.Management.Automation.PSCustomObject#DeploymentOperation")), enumerateCollection: true));
         }
 
         /// <summary>


filizt:
I looked at the test failures. Will talk to the test owner to get them to a resolution.

wyunchi-ms:
Please take a look at the failed test cases. @filizt

src/Resources/ResourceManager/Extensions/JTokenExtensions.cs
```diff
@@ -123,5 +123,91 @@ internal static object ConvertPropertyValueForPsObject(JToken propertyValue)
...
+        internal static PSObject ToPsObjectByDeepConvert(this JToken jtoken, string objectType = null)
+        {
+            if (jtoken == null)
+            {
+                return null;
+            }
+
+            if (jtoken.Type != JTokenType.Object)
+            {
+                return new PSObject(JTokenExtensions.ConvertPropertyValueForPsObject(propertyValue: jtoken));
```
mentat9, line 141: This branch isn't doing deep convert. For example if you pass this object to ToPsObjectByDeepConvert():
```
[
    {
        "obj1": {
            "val1": "abc",
            "obj2": {
                 "obj3": {
                    "val2": "def",
                    "val3": "ghi",
                    "arr1": [
                        "val4", "val5", "val6"
                    ]
                }
            }
        }
    },
    {
        "val5": "jkl"
    }
]
```

... it will call ToPSObject() on each element of the array, which will not do a deep convert on obj1, which means it will not convert below obj2 level (as you observed), which will skip obj3 and arr1 (turn them into strings).

src/Resources/ResourceManager/Extensions/JTokenExtensions.cs
```diff
@@ -123,5 +123,91 @@ internal static object ConvertPropertyValueForPsObject(JToken propertyValue)
 
             return propertyValue.ToString();
         }
+
```
mentat9, line 126: These two new methods are similar enough to the existing methods, it's worth looking for ways to reduce the duplication. For example, adding an optional boolean parameter to the ToPsObject method:

```ToPsObject(this JToken jtoken, string objectType = null, bool deepConvert = false)```

captures the semantics nicely without adding a method or breaking any existing uses of the current method. To implement it you can add that parameter and a conditional block to the existing ConvertPropertyValueForPsObject method like this:
```
        internal static object ConvertPropertyValueForPsObject(this JToken propertyValue, bool deepConvert)
        {
            if (propertyValue.Type == JTokenType.Object)
            {
                return propertyValue.ToPsObject(deepConvert: deepConvert);
            }

            if (propertyValue.Type == JTokenType.Array)
            {
                var jArray = (JArray)propertyValue;

                var array = new object[jArray.Count];

                for (int i = 0; i < array.Length; ++i)
                {
                    array[i] = jArray[i].ConvertPropertyValueForPsObject(deepConvert);
                }

                return array;
            }

>>>
            if (deepConvert && propertyValue is JProperty property)
            {
                return property.Value.ToPsObject(deepConvert: deepConvert);
            }
>>>

            Type primitiveType;
            if (JTokenExtensions.PrimitiveTypeMap.TryGetValue(propertyValue.Type, out primitiveType))
            {
                try
                {
                    return propertyValue.ToObject(primitiveType, JsonExtensions.JsonObjectTypeSerializer);
                }
                catch (FormatException)
                {
                }
                catch (ArgumentException)
                {
                }
                catch (JsonException)
                {
                }
            }

            return propertyValue.ToString();
        }
```

Note, I also changed it to an extension method for consistency.

I don't know for sure if this builds out the properties exactly the same way the two new methods did, but it's effectively the same and the details should be pretty easy to tweak. I think this idea is worth looking at: it also fixes the other issue in which one of the conversion paths doesn't do deep conversion.

File: src/Resources/ResourceManager/SdkModels/Deployments/DeploymentOperationErrorInfo.cs [hunk 3]
```diff
@@ -0,0 +1,142 @@
...
+            dynamic dynamicStatusMessage;
+
+            try
+            {
+                dynamicStatusMessage = JsonConvert.DeserializeObject(statusMessage);
+                error = JsonConvert.DeserializeObject<ErrorResponse>(dynamicStatusMessage?.error?.ToString());
+            }
+            catch
+            {
+                // We'll ignore the exception if we can't properly deserialize the JSON.
```
mentat9, line 133: I'd recommend putting something into the error string for this. Maybe the exception message, and/or a message indicating deserialization failed and including the raw JSON so someone could figure out what happened. At minimum, I think it's better not to return null (unless null means something special to the caller).

src/Resources/ResourceManager/SdkModels/Deployments/DeploymentOperationErrorInfo.cs
```diff
@@ -0,0 +1,142 @@
...
+                // The reason for that can be 1- statusMessage is not a valid JSON(it can be a string when it's the outer generic error message)
+                // which will result is DeserializeObject exception. 2- If error is not properly modeled as ErrorResponse.
+            }
+
+            return error;
```
mentat9, line 138: This will return null if you get an exception, which will crash at += with NullReferenceException in the recursive call.

src/Resources/ResourceManager/Extensions/JTokenExtensions.cs
filizt: This is a typo. I can fix it but I'll go with your suggestion of eliminating code duplication in a later comment. 

---
In reply to: [414313776](https://github.com/Azure/azure-powershell/pull/11644#discussion_r414313776) [](ancestors = 414313776)

src/Resources/ResourceManager/SdkModels/Deployments/DeploymentOperationErrorInfo.cs
filizt: The recursive call `GetErrorMessageWithDetails` has a check for null condition, it shouldn't be able to go into the += part of the code.

src/Resources/ResourceManager/SdkModels/Deployments/DeploymentOperationErrorInfo.cs
filizt: I might be wrong, but the assumption I made was if we can't deserialize an error returned from our service, then it's not clear to us and to the user what that error is. I could surface the exception message but I am not sure if the user could do anything about that exception. I might be overlooking something though - what do you think of this approach?

mentat9: Yes, that's true, but I think saying nothing is usually worse. Not only does the user not know what happened, but we will never find out it happened, either. If there is a common error mode that fails to deserialize, we could find and fix it if the errors are getting back to the user.

I think it's usually best to tell them everything you know. Maybe something like this:

Your deployment failed, but we were unable to get additional error information. The attempt to retrieve detailed information failed with this error:
Message: <message>
Stacktrace: <stacktrace>

Or something similar.

src/Resources/ResourceManager/SdkModels/Deployments/DeploymentOperationErrorInfo.cs
mentat9: Cool, just need to be sure every caller does the check. Since the null only happens on an error path, it's a slightly non-obvious return value. Not a big deal, but it might be safer to use empty string.

src/Resources/ResourceManager/SdkModels/Deployments/DeploymentOperationErrorInfo.cs
filizt: Created a new PR to address some of the feedback. [https://github.com/Azure/azure-powershell/pull/11763](https://github.com/Azure/azure-powershell/pull/11763)

---
In reply to: [414766481](https://github.com/Azure/azure-powershell/pull/11644#discussion_r414766481) [](ancestors = 414766481)

Issue #11335, opened 2020-03-16
What is the -TenantLevel parameter used for on the Get-AzResourceLock cmdlet?


theheatDK:
## Description
Hi,
I cannot find any examples of how the -TenantLevel parameter is used for the Get-AzResourceLock cmdlet.
The documentation just states "Indicates that this cmdlet operates at the tenant level."
I have had a quick look at the resource lock REST APIs and I cannot find anything that mentions tenant level. What does it do?
Regards Peter

## Steps to reproduce

```powershell

```

## Environment data

```

```

## Module versions

```powershell

```

## Debug output

```

```

## Error output

```

```

dingmeng-xue:
It needs service team to explain it further.

rthorn17:
Hi- @theheatDK  Let me research and get back to you on this.  Thanks

ifeoluwaokunoren:
@theheatDK you're correct - [resource (management) locks](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/lock-resources) are only valid at the Subscription, Resource Groups, Resource scopes. 
I'd add this to our backlog, so we can update the command to remove the -TenantLevel parameter since it can be used in the context of this command.

Thanks

theheatDK:
Thanks @ifeoluwaokunoren 
I assume that you are aware the parameter is used in several of the lock cmdlets? (New-AzResourceLock, Remove-AzResourceLock...)

theheatDK:
Hi @ifeoluwaokunoren,
You never answered my question from 19 of Marts? :-)
Should I close the issue?
Regards Peter

ifeoluwaokunoren:
Hi @theheatDK,
Thanks so much for following up.
Yes, the fix would address the other lock cmdlets as well.
I'd close this issue once we check-in the fix.

Thanks :)

theheatDK:
Super. Thanks @ifeoluwaokunoren

ifeoluwaokunoren:
PR fix: https://github.com/Azure/azure-powershell/issues/11335

PR #11689, opened 2020-04-24
Remove -TenantLevel parameter used for on the Get-AzResourceLock cmdlet

brucelin-511:
## Description
Remove -TenantLevel parameter used for on the Get-AzResourceLock cmdlet to fix the issue of 
What is the -TenantLevel parameter used for on the Get-AzResourceLock cmdlet? #11335

## Checklist

- [X] The appropriate `ChangeLog.md` file(s) has been updated:


adxsdkps:

isra-fel:
Hi @brucelin-511 , what would happen if the user used `-TenantLevel` before? The cmdlet returned nothing, or threw an error?

I ask because we cannot check in breaking changes without notifying our customers in advance, unless the function is already broken.

brucelin-511:
Previously, adding -TenantLevel, it don't throw the error. Actually, the effect of adding it or not, the result is the same. Resource (management) locks are only valid at the Subscription, Resource Group, Resource scopes. TenantLevel should not be a parameter for the command.

PR #11758, opened 2020-04-29
Fix bug in Add Certificate using --SecretIdentifier getting the wrong certificate thumbprint

a-santamaria:
## Description

Adding a certificate with secret identifier was failing some times because the cmdlet was retrieving the user's secret and taking one of the certs in the chain. the correct way to do it is to query de key vault cert to get the thumbprint.

## Checklist

### DIFF
--- a/src/ServiceFabric/ServiceFabric/ChangeLog.md
+++ b/src/ServiceFabric/ServiceFabric/ChangeLog.md
@@ -18,6 +18,7 @@
         - Additional information about change #1
 -->
 ## Upcoming Release
+* Fix bug in add certificate using --SecretIdentifier that was getting the wrong certificate thumbprint
 
 ## Version 2.0.2
 * Improved code formatting and usability of `New-AzServiceFabricCluster` examples


src/ServiceFabric/ServiceFabric/Commands/ServiceFabricClusterCertificateCmdlet.cs
```diff
@@ -97,6 +97,11 @@ public abstract class ServiceFabricClusterCertificateCmdlet : ServiceFabricClust
         [ValidateNotNullOrEmpty]
         public string SecretIdentifier { get; set; }
 
+        [Parameter(Mandatory = false, ValueFromPipeline = true, ParameterSetName = ByExistingKeyVault,
+                   HelpMessage = "The thumbprint for the certificate corresponding to the SecretIdentifier. Use this if the certificate is not managed as the key vault would only have the certificate stored as a secret and the cmdlet is unable to retrieve the thumbprint.")]
```
dragav, line 101: >Use this if the certificate is not managed as the key vault would only have the certificate stored as a secret and the cmdlet is unable to retrieve the thumbprint [](start = 108, length = 162)

Use this when:
- you do not have permissions to retrieve the certificate from the vault
- the certificate is not managed by KeyVault (was imported as a secret)
- you have the thumbprint or simply do not want to retrieve the certificate from the vault

src/ServiceFabric/ServiceFabric/Commands/ServiceFabricClusterCertificateCmdlet.cs
```diff
@@ -299,7 +304,7 @@ private void CreateSelfSignedCertificate(string subjectName, string keyVaultUrl,
                 var secretBundle = this.KeyVaultClient.GetSecretAsync(keyVaultUrl, this.keyVaultCertificateName).Result;
                 var kvSecretBytes = Convert.FromBase64String(secretBundle.Value);
                 var certCollection = new X509Certificate2Collection();
-                certCollection.Import(kvSecretBytes,null,X509KeyStorageFlags.Exportable);
+                certCollection.Import(kvSecretBytes, null, X509KeyStorageFlags.Exportable);
```
dragav, line 307: >X509KeyStorageFlags.Exportable [](start = 59, length = 30)

Maybe for later, but why must this be exportable?

src/ServiceFabric/ServiceFabric/Commands/ServiceFabricClusterCertificateCmdlet.cs
```diff
@@ -409,27 +414,15 @@ internal List<CertificateInformation> GetOrCreateCertificateInformation()
...
-                        certificateInformations.Add(new CertificateInformation()
-                        {
-                            KeyVault = vault,
-                            SecretUrl = this.SecretIdentifier,
-                            CertificateThumbprint = GetThumbprintFromSecret(this.SecretIdentifier),
-                            CertificateCommonName = GetCommonNameFromSecret(this.SecretIdentifier),
-                            SecretName = vaultSecretName,
-                            Version = version
-                        });
+                        CertificateInformation certInfor = GetCertificateInforamtionFromSecret(this.SecretIdentifier);
```
dragav, line 424: >Inforamtion [](start = 73, length = 11)

(typo)

src/ServiceFabric/ServiceFabric/Commands/ServiceFabricClusterCertificateCmdlet.cs
```diff
@@ -97,6 +97,11 @@ public abstract class ServiceFabricClusterCertificateCmdlet : ServiceFabricClust
         [ValidateNotNullOrEmpty]
         public string SecretIdentifier { get; set; }
 
+        [Parameter(Mandatory = false, ValueFromPipeline = true, ParameterSetName = ByExistingKeyVault,
```
dragav, line 100: >ParameterSetName = ByExistingKeyVault, [](start = 64, length = 38)

This looks a bit odd to me - both the SecretIdentifier and the Thumbprint parameter are:
- validate not null or empty
- corresponding to the same parameter set - byExistingKV

Either of them would be necessary and sufficient, but not both? Or did you mean to create a new parameter set?

src/ServiceFabric/ServiceFabric/Commands/ServiceFabricClusterCertificateCmdlet.cs
```diff
@@ -643,161 +636,78 @@ private static bool IsValidKeyVaultObjectChar(char name)
...
-            return cert.Thumbprint;
-        }
-
-        private string GetCommonNameFromSecret(string secretUrl)
-        {
-            if (RunningTest)
+            CertificateBundle certBundle = this.KeyVaultClient.GetCertificateAsync(vault.Properties.VaultUri, certName, certVersion).GetAwaiter().GetResult();
+            string thumbprint;
+            string commonName;
+            if (certBundle.Cer == null)
```
dragav, line 664: >if (certBundle.Cer == null) [](start = 12, length = 27)

This wouldn't happen, you'd get a KVClient exception indicating a 404..

src/ServiceFabric/ServiceFabric/Commands/ServiceFabricClusterCertificateCmdlet.cs
a-santamaria: validate not null or empty is a check whenever the parameters are used. I'm using the same parameter set and only having SecretIdentifier as mandatory since Thumbprint is only needed for the cases where we can't get the certificate bundle form the key vault. Only using Thumbprint parameter is not enough, that's why SecretIdentifier  is mandatory

---
In reply to: [418234139](https://github.com/Azure/azure-powershell/pull/11758#discussion_r418234139) [](ancestors = 418234139)

src/ServiceFabric/ServiceFabric/Commands/ServiceFabricClusterCertificateCmdlet.cs
a-santamaria: ok thanks, I'll change the message in a new pr

---
In reply to: [418230302](https://github.com/Azure/azure-powershell/pull/11758#discussion_r418230302) [](ancestors = 418230302)

PR #11984, opened 2020-05-25
VirtualWan: Use hubvnet CRUD operations for Hubvnet commands

shyamshd:
## Description

- Put/Delete hubvnet operations were using PUT on parent hub. Instead just use the newly exposed put/delete APIs on hubvnet connection.
- Deprecate hubvnet connections in virtual hub cmds. This aligns with the rest API changes to avoid having multiple ways of doing the same thing..
- Throw an error indicating that the user needs to use the alternate apis  instead of quietly removing the option. The option can be removed altogether in the future
- In, update hubvnet command, make enableInternetSecurity to be true by default unless specified otherwise. Similar change in New hubvnet connection would break backward compatibility. So leave it as is till I can find a good solution.

## Checklist

### DIFF
--- a/src/Network/Network/Cortex/HubVnetConnection/NewAzureRmHubVirtualNetworkConnectionCommand.cs
+++ b/src/Network/Network/Cortex/HubVnetConnection/NewAzureRmHubVirtualNetworkConnectionCommand.cs
@@ -158,6 +158,7 @@ namespace Microsoft.Azure.Commands.Network
 
             var hubVnetConnection = new MNM.HubVirtualNetworkConnection();
             hubVnetConnection.Name = this.Name;
+            // Fixme.. make it a boolean
             hubVnetConnection.EnableInternetSecurity = this.EnableInternetSecurity.IsPresent;
 
             //// Resolve the remote virtual network
--- a/src/Network/Network/Cortex/VirtualHub/NewAzureRmVirtualHubCommand.cs
+++ b/src/Network/Network/Cortex/VirtualHub/NewAzureRmVirtualHubCommand.cs
@@ -28,12 +28,14 @@ namespace Microsoft.Azure.Commands.Network
     using Microsoft.Azure.Commands.ResourceManager.Common.ArgumentCompleters;
     using System.Linq;
     using Microsoft.Azure.Management.Internal.Resources.Utilities.Models;
+    using Microsoft.WindowsAzure.Commands.Common.CustomAttributes;
 
     [Cmdlet(VerbsCommon.New,
         ResourceManager.Common.AzureRMConstants.AzureRMPrefix + "VirtualHub",
         DefaultParameterSetName = CortexParameterSetNames.ByVirtualWanObject,
         SupportsShouldProcess = true),
         OutputType(typeof(PSVirtualHub))]
+
     public class NewAzureRmVirtualHubCommand : VirtualHubBaseCmdlet
     {
         [Parameter(
@@ -135,6 +137,11 @@ namespace Microsoft.Azure.Commands.Network
                 throw new PSArgumentException(Properties.Resources.VirtualWanReferenceNeededForVirtualHub);
             }
 
+            if (this.HubVnetConnection != null)
+            {
+                throw new PSArgumentException(Properties.Resources.VirtualNetworkConnectionNotAllowedInHubOperation);
+            }
+
             PSVirtualWan resolvedVirtualWan = new VirtualWanBaseCmdlet().GetVirtualWan(virtualWanRGName, virtualWanName);
 
             ConfirmAction(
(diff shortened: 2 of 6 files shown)


VeryEarly:
Hi @shyamshd ,

Please fix the build issue

shyamshd:
@VeryEarly  This has a dependency on some swagger/sdk changes. I'll re-run the build once the sdk is available. The PR introduces some breaking changes, so I raised the request to check if reviewers are ok with it.

src/Network/Network/Cortex/HubVnetConnection/HubVnetConnectionBaseCmdlet.cs
```diff
@@ -22,7 +22,7 @@ namespace Microsoft.Azure.Commands.Network
     using System.Net;
     using MNM = Microsoft.Azure.Management.Network.Models;
 
-    public class HubVnetConnectionBaseCmdlet : VirtualHubBaseCmdlet
```
nagula-ritvika, line 25: can we add a VirtualHubVnetConenctionBaseCmdlet if possible ?

src/Network/Network/Cortex/HubVnetConnection/NewAzureRmHubVirtualNetworkConnectionCommand.cs
```diff
@@ -156,33 +156,22 @@ public override void Execute()
...
-            PSVirtualHub parentVirtualHub = this.GetVirtualHub(this.ResourceGroupName, this.ParentResourceName);
-            if (parentVirtualHub == null)
-            {
-                throw new PSArgumentException(Properties.Resources.ParentVirtualHubNotFound);
-            }
-
-            PSHubVirtualNetworkConnection hubVnetConnection = new PSHubVirtualNetworkConnection();
+            var hubVnetConnection = new MNM.HubVirtualNetworkConnection();
             hubVnetConnection.Name = this.Name;
+            // Fixme.. make it a boolean
```
nagula-ritvika, line 161: it is currenlty a switch parameter i suppose..

src/Network/Network/Cortex/HubVnetConnection/NewAzureRmHubVirtualNetworkConnectionCommand.cs
```diff
@@ -191,18 +180,13 @@ public override void Execute()
...
                 Properties.Resources.CreatingResourceMessage,
                 this.Name,
                 () =>
                 {
                     WriteVerbose(String.Format(Properties.Resources.CreatingLongRunningOperationMessage, this.ResourceGroupName, this.Name));
-                    this.CreateOrUpdateVirtualHub(this.ResourceGroupName, this.ParentResourceName, parentVirtualHub, parentVirtualHub.Tag, auxAuthHeader);
-                    var createdVirtualHub = this.GetVirtualHub(this.ResourceGroupName, this.ParentResourceName);
-
-                    WriteObject(createdVirtualHub.VirtualNetworkConnections.FirstOrDefault(hubConnection => hubConnection.Name.Equals(this.Name, StringComparison.OrdinalIgnoreCase)));
+                    WriteObject(this.CreateOrUpdateHubVirtualNetworkConnection(this.ResourceGroupName, this.ParentResourceName, this.Name, hubVnetConnection, auxAuthHeader));
```
nagula-ritvika, line 189: ❤️

src/Network/Network/Cortex/HubVnetConnection/UpdateAzureRmHubVirtualNetworkConnectionCommand.cs
```diff
@@ -127,6 +124,10 @@ public override void Execute()
             {
                 connectionToModify.EnableInternetSecurity = this.EnableInternetSecurity.Value;
             }
+            else
```
nagula-ritvika, line 127: dont do this, i think the main purpose of this cmdlet is to allow the user to be able to modify the enab;eInternetSecurity value , revert this

src/Network/Network/Cortex/HubVnetConnection/NewAzureRmHubVirtualNetworkConnectionCommand.cs
shyamshd: Yes. I did not want to break existing scripts which specify the parameter & punish users who were actually using the option. The other suggestion I got on the powershell channel was to simply introduce a new parameter.

src/Network/Network/Cortex/HubVnetConnection/UpdateAzureRmHubVirtualNetworkConnectionCommand.cs
shyamshd: What I was trying to do was make the option to be true by default if not specified (as you had indicated in our offline discussion). The user would still be able to set it to false if they do so explicitly.

nagula-ritvika: option to be true by default if not specified (as you had indicated in our offline discussion). -> this is only for new connections to be created not for existing connections. 
for update scenario : we dont want to modify the flag value in case the cx has forgotten to explicitly mention it.

shyamshd: Hmm, why make it asymmetric? If we are doing it for 'New', why not keep it consistent and do it on the 'Update' too? We can talk more offline..

VeryEarly:
> @VeryEarly This has a dependency on some swagger/sdk changes. I'll re-run the build once the sdk is available. The PR introduces some breaking changes, so I raised the request to check if reviewers are ok with it.

Breaking changes won't be allowed until next breaking change release which is in this October.

shyamshd:
> > @VeryEarly This has a dependency on some swagger/sdk changes. I'll re-run the build once the sdk is available. The PR introduces some breaking changes, so I raised the request to check if reviewers are ok with it.
> 
> Breaking changes won't be allowed until next breaking change release which is in this October.

I'll withdraw this PR. I have raised a new one which adds a warning and keeps the behavior backward compatible.

PR #12107, opened 2020-06-10
[AppService] Fix - bug that prevented changing Container settings in Set-AzWebApp and Set-AzWebAppSlot

vinisoto:
## Description

-Fix - bug that prevented changing Container settings in Set-AzWebApp and Set-AzWebAppSlot
-Fixed typos.

## Checklist

### DIFF
--- a/src/Websites/Websites.Test/ScenarioTests/WebAppTests.cs
+++ b/src/Websites/Websites.Test/ScenarioTests/WebAppTests.cs
@@ -47,6 +47,13 @@ namespace Microsoft.Azure.Commands.Websites.Test.ScenarioTests
 
         [Fact]
         [Trait(Category.AcceptanceType, Category.CheckIn)]
+        public void TestSetWebAppHyperVCredentials()
+        {
+            WebsitesController.NewInstance.RunPsTest(_logger, "Test-SetWebAppHyperVCredentials");
+        }
+
+        [Fact]
+        [Trait(Category.AcceptanceType, Category.CheckIn)]
         public void TestEnableContainerContinuousDeploymentAndGetUrl()
         {
             WebsitesController.NewInstance.RunPsTest(_logger, "Test-EnableContainerContinuousDeploymentAndGetUrl");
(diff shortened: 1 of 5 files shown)


src/Websites/Websites/Utilities/WebsitesClient.cs
```diff
@@ -539,14 +539,6 @@ public void UpdateWebAppConfiguration(string resourceGroupName, string location,
 
             if (useSlot)
             {
-                if (siteConfig != null)
```
Kotasudhakarreddy, line 542: @vinisoto - what was the scenario to move `siteConfig` configuration update to behind the `appSettings` configuration update. It is reported as regression issue in #14998 . please let me whether I can alter their sequence of updating the configuration or it should be like this for any of the other functionality.

vinisoto: We need to make sure that the app settings below are updated before or at the same time as WindowsFxVersion (which is in SiteConfig). Otherwise, server side validation for WindowsFxVersion will fail since it will be trying to use "old" container settings.

public const string DockerRegistryServerUrl = "DOCKER_REGISTRY_SERVER_URL";
public const string DockerRegistryServerUserName = "DOCKER_REGISTRY_SERVER_USERNAME";
public const string DockerRegistryServerPassword = "DOCKER_REGISTRY_SERVER_PASSWORD";

Kotasudhakarreddy: @vinisoto - For now I have reverted your changes. Because it has created a lot of mess in just few days. can you give me the #GitID for the same. I want to analyze it further to understand the issue and fix it if required. 

@panchagnula - FYI.

panchagnula: @Kotasudhakarreddy we should make sure to  fix the issue Vini was working on as well with your changes so that we don't trade one regression for another. i.e this bug https://github.com/Azure/azure-powershell/pull/13866

Kotasudhakarreddy: @vinisoto - I have reverted your changes only from **WebsitesClient.cs->UpdateWebAppConfiguration** method. But I don't see any server side validation for setting `WindowsFxVersion`. Could you please test  [this](https://github.com/Azure/azure-powershell/pull/15011) PR and confirm us the same.

PR #12205, opened 2020-06-18
Example was missing Update-AzVM cmdlet

michawets:
When running the example, the diagnostics settings were not applied to the VM, only to the variable stored in the Powershell windows at that moment. No real update was done on the VM.
For the diagnostics to be applied, the cmdlet Update-AzVM still needs to be executed

## Checklist


VeryEarly:
looks like the updated vm need to be passed to Update-AzVM, please check my commit and confirm.

michawets:
> looks like the updated vm need to be passed to Update-AzVM, please check my commit and confirm.

Hi @VeryEarly , there are multiple ways to do it 😊.

I double checked my commit, and that will work. I didn't want to change to much on the example, and keep the $VM variable as a single variable

VeryEarly:
> > looks like the updated vm need to be passed to Update-AzVM, please check my commit and confirm.
> 
> Hi @VeryEarly , there are multiple ways to do it 😊.
> 
> I double checked my commit, and that will work. I didn't want to change to much on the example, and keep the $VM variable as a single variable
> 
> So the $updatedVM variable is actually not needed 😊

You're right,I'll revert my commit.

PR #12221, opened 2020-06-19
Aladdin generated examples for Compute

aladdindoc:
- Add Aladdin crafted examples.


adxsdkps:

src/Compute/Compute/help/New-AzHostGroup.md
```diff
@@ -39,6 +39,14 @@ Tags                     : {[key1, val1]}
 
 This command creates a host group in the given location and zone.
 
+### Example 2
+
+Creates a host group. (autogenerated)
+
+```powershell <!-- Aladdin Generated Example --> 
+New-AzHostGroup -Location <String> -Name <String> -PlatformFaultDomain <Int32> -ResourceGroupName <String>
+```
```
mikefrobbins, line 48: We should omit examples where less than 50% of the specified parameters have actual values. Specifying the type for all of the parameters doesn't provide much value as the syntax can already be obtained from the cmdlet.

src/Compute/Compute/help/New-AzHost.md
```diff
@@ -44,6 +44,14 @@ Tags                 : {"key1":"val2"}
 
 This command creates a host of the given Sku in the given host group and the given location.
 
+### Example 2
+
+Creates a  host. (autogenerated)
+
+```powershell <!-- Aladdin Generated Example --> 
+New-AzHost -AutoReplaceOnFailure $false -HostGroupName <String> -Location <String> -Name <String> -PlatformFaultDomain <Int32> -ResourceGroupName <String> -Sku <String>
```
mikefrobbins, line 52: We should omit examples where less than 50% of the specified parameters have actual values. Specifying the type for all of the parameters doesn't provide much value as the syntax can already be obtained from the cmdlet.

PR #12229, opened 2020-06-22
Remove Sha256Checksum parameter of example of New-AzImageBuilderCustomizerObject.

### LINKED ISSUE #12200: example of New-AzImageBuilderCustomizerObject needs to remove Sha256Checksum parameter
## Description

`New-AzImageBuilderCustomizerObject’ cmdlet requires to use a Sha256 checksum however this is not possible in the case of an inline script. We need to docs as well.

## Steps to reproduce

```powershell

```

## Environment data

```

```

## Module versions

```powershell

```

## Debug output

```

```

## Error output

```

```

LucasYao93:
## Description

Fix #12200


adxsdkps:

PR #12268, opened 2020-06-25
Lower case server name in upsert server operation

skycckk:
WCF api calls will fail getting servers if server name is in upper case
due to the server name validation from the backend.

To mitigate the future pressure, this fix is going to lower case to
the server name while upserting servers in management service.

NOTE: Azure portal already has the server name constraint where lower
case is attached.

## Checklist

### DIFF
--- a/src/Sql/Sql/Properties/Resources.resx
+++ b/src/Sql/Sql/Properties/Resources.resx
@@ -632,6 +632,6 @@
     <value>Are you sure you want to failover the Azure Sql Managed Instance '{0}'?</value>
   </data>
   <data name="ServerNameInvalid" xml:space="preserve">
-    <value>Invalid server name: '{0}'. Server name can contain only lowercase letters, number.</value>
+    <value>The server name '{0}' cannot be empty or null. The server name can only be made up of lowercase letters a-z, the numbers 0-9 and the hyphen. The hyphen may not lead or trail in the server name. Please fix the server name and retry. Please contact Microsoft support if the issue persists.</value>
   </data>
 </root>
--- a/src/Sql/Sql.Test/ScenarioTests/VulnerabilityAssessmentTests.ps1
+++ b/src/Sql/Sql.Test/ScenarioTests/VulnerabilityAssessmentTests.ps1
@@ -105,7 +105,7 @@ function Test-VulnerabilityAssessmentServerSettingsTest
 		 -NotificationEmail $testNotificationEmail
 		
 		# Test piping of set settings from resource object
-		$newServerName = "newServerName" +$testSuffix;
+		$newServerName = "newservername" +$testSuffix;
 		$testNewNotificationEmail = @("test3@mailTest.com", "test4@mailTest.com")
 		$testStorageName = $params.storageAccount
 		$testBlobStorageSasUri = "https://" + $testStorageName +".blob.core.windows.net/" + $testScanResultsContainerName + "?st=XXXXXXXXXXXXXXX"
--- a/src/Sql/Sql.Test/SessionRecords/Microsoft.Azure.Commands.Sql.Test.ScenarioTests.VulnerabilityAssessmentTests/VulnerabilityAssessmentServerSettingsTest.json
+++ b/src/Sql/Sql.Test/SessionRecords/Microsoft.Azure.Commands.Sql.Test.ScenarioTests.VulnerabilityAssessmentTests/VulnerabilityAssessmentServerSettingsTest.json
@@ -4348,8 +4348,8 @@
       "StatusCode": 200
     },
     {
-      "RequestUri": "/subscriptions/cca24ec8-99b5-4aa7-9ff6-486e886f304c/resourceGroups/sql-va-cmdlet-test-rgps64/providers/Microsoft.Sql/servers/newServerNameps64?api-version=2015-05-01-preview",
-      "EncodedRequestUri": "L3N1YnNjcmlwdGlvbnMvY2NhMjRlYzgtOTliNS00YWE3LTlmZjYtNDg2ZTg4NmYzMDRjL3Jlc291cmNlR3JvdXBzL3NxbC12YS1jbWRsZXQtdGVzdC1yZ3BzNjQvcHJvdmlkZXJzL01pY3Jvc29mdC5TcWwvc2VydmVycy9uZXdTZXJ2ZXJOYW1lcHM2ND9hcGktdmVyc2lvbj0yMDE1LTA1LTAxLXByZXZpZXc=",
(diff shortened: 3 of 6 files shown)


adxsdkps:

skycckk:
Hi @VeryEarly , backend (management service) is already cased silently and this client update is just to prevent these ARM cache issues.

VeryEarly:
> Hi @VeryEarly , backend (management service) is already cased silently and this client update is just to prevent these ARM cache issues.

Having validation forces customers to use lowercase which guarantees input/output are identical.

You mentioned
"WCF api calls will fail getting servers if server name is in upper case
due to the server name validation from the backend."
What exactly is the server behavior?

skycckk:
> Having validation forces customers to use lowercase which guarantees input/output are identical.
> 
> You mentioned
> "WCF api calls will fail getting servers if server name is in upper case
> due to the server name validation from the backend."
> What exactly is the server behavior?

The server service codes validate and throw exception if incoming server name from ARM has upper case letter.
Backend WebAPI module always have the ability to do lower case for the server name when the request comes in.
And the ARM resource name supposed to be case insensitive, so GETs going forward should be fine with the server name regardless of the casing.
The issues are using a WebAPI based api version for creating upper case server name, then used the older WCF api version for GETs, which is not compatible.

So the intention here is to prevent inconsistent behavior between server/client side regarding the case of server name.

Still, adding validation might also a great solution but it might take more effort to do as I'm not an expert on the powershell infra here. To start with, I've added a version where replacing ToLower() with validation at [a553b1a](https://github.com/Azure/azure-powershell/commit/a553b1a4ba2e375000bd8557a118cb00099642f7). If that is what you're looking for, then it would be great. Or any suggestions would be appreciated.

VeryEarly:
> > Having validation forces customers to use lowercase which guarantees input/output are identical.
> > You mentioned
> > "WCF api calls will fail getting servers if server name is in upper case
> > due to the server name validation from the backend."
> > What exactly is the server behavior?
> 
> The server service codes validate and throw exception if incoming server name from ARM has upper case letter.
> Backend WebAPI module always have the ability to do lower case for the server name when the request comes in.
> And the ARM resource name supposed to be case insensitive, so GETs going forward should be fine with the server name regardless of the casing.
> The issues are using a WebAPI based api version for creating upper case server name, then used the older WCF api version for GETs, which is not compatible.
> 
> So the intention here is to prevent inconsistent behavior between server/client side regarding the case of server name.
> 
> Still, adding validation might also a great solution but it might take more effort to do as I'm not an expert on the powershell infra here. To start with, I've added a version where replacing ToLower() with validation at [a553b1a](https://github.com/Azure/azure-powershell/commit/a553b1a4ba2e375000bd8557a118cb00099642f7). If that is what you're looking for, then it would be great. Or any suggestions would be appreciated.

Hi @skycckk ,

Thanks for the explanation, your commit looks good to me except this line:
https://github.com/Azure/azure-powershell/commit/a553b1a4ba2e375000bd8557a118cb00099642f7#diff-e0a87086db113cd1509a633209c515a0R22

Is this line needed?

skycckk:
> Thanks for the explanation, your commit looks good to me except this line:
> [a553b1a#diff-e0a87086db113cd1509a633209c515a0R22](https://github.com/Azure/azure-powershell/commit/a553b1a4ba2e375000bd8557a118cb00099642f7#diff-e0a87086db113cd1509a633209c515a0R22)
> 
> Is this line needed?

good catch, fixed, thx.

VeryEarly:
> > Thanks for the explanation, your commit looks good to me except this line:
> > [a553b1a#diff-e0a87086db113cd1509a633209c515a0R22](https://github.com/Azure/azure-powershell/commit/a553b1a4ba2e375000bd8557a118cb00099642f7#diff-e0a87086db113cd1509a633209c515a0R22)
> > Is this line needed?
> 
> good catch, fixed, thx.

Hi @skycckk ,

Can you take a look at the failed test case: https://dev.azure.com/azure-sdk/public/_build/results?buildId=446883&view=logs&jobId=5264e576-3c6f-51f6-f055-fab409685f20&j=5264e576-3c6f-51f6-f055-fab409685f20&t=11b6e5a1-913e-50ac-35c6-54672a9d99ea

It looks like it used to accept upper case letters in server name.
You mentioned "WCF api calls will fail getting servers if server name is in upper case" if that's the case, this test case could not be recorded in the first place: https://github.com/skycckk/azure-powershell/blob/fixServerNameCaseInsensitive/src/Sql/Sql.Test/ScenarioTests/VulnerabilityAssessmentTests.ps1#L108 

Can you please double check?

src/Sql/Sql/Properties/Resources.resx
```diff
@@ -631,4 +631,7 @@
   <data name="FailoverAzureSqlInstanceWarning" xml:space="preserve">
     <value>Are you sure you want to failover the Azure Sql Managed Instance '{0}'?</value>
   </data>
+  <data name="ServerNameInvalid" xml:space="preserve">
+    <value>Invalid server name: '{0}'. Server name can contain only lowercase letters, number.</value>
```
johnpaulkee, line 635: Can update to this?

The server name '{0}' cannot be empty or null. The server name can only be made up of lowercase letters a-z, the numbers 0-9 and the hyphen. The hyphen may not lead or trail in the server name. Please fix the server name and retry. Please contact Microsoft support if the issue persists.

skycckk: Done

skycckk:
> > > Thanks for the explanation, your commit looks good to me except this line:
> > > [a553b1a#diff-e0a87086db113cd1509a633209c515a0R22](https://github.com/Azure/azure-powershell/commit/a553b1a4ba2e375000bd8557a118cb00099642f7#diff-e0a87086db113cd1509a633209c515a0R22)
> > > Is this line needed?
> > 
> > 
> > good catch, fixed, thx.
> 
> Hi @skycckk ,
> 
> Can you take a look at the failed test case: https://dev.azure.com/azure-sdk/public/_build/results?buildId=446883&view=logs&jobId=5264e576-3c6f-51f6-f055-fab409685f20&j=5264e576-3c6f-51f6-f055-fab409685f20&t=11b6e5a1-913e-50ac-35c6-54672a9d99ea
> 
> It looks like it used to accept upper case letters in server name.
> You mentioned "WCF api calls will fail getting servers if server name is in upper case" if that's the case, this test case could not be recorded in the first place: https://github.com/skycckk/azure-powershell/blob/fixServerNameCaseInsensitive/src/Sql/Sql.Test/ScenarioTests/VulnerabilityAssessmentTests.ps1#L108
> 
> Can you please double check?

Hi @VeryEarly , that failed test case was fixed by changing the resource server name from mixcased to lower as that test was testing with a legit server name, which becomes no-mixcased at this fix.

Regarding why the test was passed in the first place, it's because the using resource url API versions are not WCF, whereas version 2014-04-01 or GET/V2/MODULES/.../SERVERS is WCF. But in this fix, we don't want any upper cases in ARM cache, otherwise WCF call would get upper case server name and fails if doing an PUT after.

PR #12370, opened 2020-07-10
Add RemoteBgpCommunities property to the VirtualNetworkPeering resource

jarango:
## Description

This pull request adds the RemoteBgpCommunities property to the VirtualNetworkPeering resource. The value of this property is populated from the value of the BgpCommunities property of the remote VNET that is being peered to the local VNET.

This is an addition being done for important scalability reasons. ExpressRoute needs to read the BgpCommunities of all the spoke VNETs that are peered with a hub VNET. Performing GET operations on each peered VNET does not scale because the number of peered VNETs can be large. The efficient way to obtain this information is from the VirtualNetworkPeering child resources of the hub VNET.

## Checklist

- [ Yes] I have read the [_Submitting Changes_](../blob/master/CONTRIBUTING.md#submitting-changes) section of [`CONTRIBUTING.md`](../blob/master/CONTRIBUTING.md)
- [ Yes] The title of the PR is clear and informative
- [Yes ] The appropriate `ChangeLog.md` file(s) has been updated:
- [Yes ] The PR does not introduce [breaking changes](../blob/master/documentation/breaking-changes/breaking-changes-definition.md)
- [ Yes] If applicable, the changes made in the PR have proper test coverage
- [not applicable] For public API changes to cmdlets:
    - [not applicable ] a cmdlet design review was approved for the changes in [this repository](https://github.com/Azure/azure-powershell-cmdlet-review-pr) (_Microsoft internal only_)
    - [not applicable ] the markdown help files have been regenerated using the commands listed [here](../blob/master/documentation/development-docs/help-generation.md#updating-all-markdown-files-in-a-module)

### DIFF
--- a/src/Network/Network.Test/ScenarioTests/VirtualNetworkTests.cs
+++ b/src/Network/Network.Test/ScenarioTests/VirtualNetworkTests.cs
@@ -50,8 +50,8 @@ namespace Commands.Network.Test.ScenarioTests
             TestRunner.RunTestScript("Test-subnetCRUD");
         }
 
-        [Fact]
-        [Trait(Category.AcceptanceType, Category.CheckIn)]
+        [Fact(Skip = "Authentication failed for auxiliary token: The '1' auxiliary tokens contains duplicates which are from the same tenant.")]
+        [Trait(Category.AcceptanceType, Category.LiveOnly)]
         [Trait(Category.Owner, NrpTeamAlias.sdnnrp)]
         public void TestVirtualNetworkBgpCommunitiesCRUD()
         {


dingmeng-xue:
Scrum master, FYI. 

Talked with Jarango, we found `Add-AzVirtualNetworkPeering` cannot be supported by current test framework. So, I suggest Jarango to keep his test case in repo but mark it as liveonly because it can executed locally in his subscription.

src/Network/Network.Test/ScenarioTests/VirtualNetworkTests.cs
```diff
@@ -50,8 +50,8 @@ public void TestVirtualNetworkSubnetCRUD()
             TestRunner.RunTestScript("Test-subnetCRUD");
         }
 
-        [Fact]
-        [Trait(Category.AcceptanceType, Category.CheckIn)]
+        [Fact(Skip = "Authentication failed for auxiliary token: The '1' auxiliary tokens contains duplicates which are from the same tenant.")]
```
jarango, line 53: This test has to be "skipped" because any test that uses the "Add-AzVirtualNetworkPeering" commandlet throws the following error message:

"Authentication failed for auxiliary token: The '1' auxiliary tokens contains duplicates which are from the same tenant."

The only other test case that uses this commandlet is TestVirtualNetworkPeeringCRUD() and it is also marked as "skip" for the same reason.

src/Network/Network.Test/ScenarioTests/VirtualNetworkTests.cs
```diff
@@ -50,8 +50,8 @@ public void TestVirtualNetworkSubnetCRUD()
             TestRunner.RunTestScript("Test-subnetCRUD");
         }
```
jarango, line 52: This test has to be "skipped" because any test that uses the "Add-AzVirtualNetworkPeering" commandlet throws the following authentication error message:

"Authentication failed for auxiliary token: The '1' auxiliary tokens contains duplicates which are from the same tenant."

The only other test case that uses this commandlet is TestVirtualNetworkPeeringCRUD() and it is also marked as "skip" for the same reason.

PR #12463, opened 2020-07-22
Revert "Add new attributes to role assignment objects and related cmdlets"

dagoroz:
Reverts Azure/azure-powershell#12425

This feature should have been released as a preview.
This change will be reverted from the public release cycle (this PR) and re-published with the appropriate info/procedure (future PR)


adxsdkps:

PR #12660, opened 2020-08-13
Add test case of AppPlatform.

LucasYao93:
## Description

Add test case of AppPlatform.


adxsdkps:

PR #12664, opened 2020-08-13
Support build/publish one module via TargetModule property

dingmeng-xue:
## Description

TargetModule can be set by user to just build and publish one module and its dependency Az.Accounts. For instance
```
dotnet msbuild build.proj /t:"Build;Publish" /p:"Configuration=Release;NuGetKey=12345;TargetModule=Compute"
```

It cannot support the array of modules because existing code in publish has limitation and assumes Az always needs to be published. I check if Az is required according to number of module is built. If it is more than 2, it means user doesn't use TargetModule property. For long term, we need to find a smart way to check if Az needs to be published according to the content in artifact folder only.

## Checklist

### DIFF
--- a/tools/BuildPackagesTask/Microsoft.Azure.Build.Tasks/SetGenerator.cs
+++ b/tools/BuildPackagesTask/Microsoft.Azure.Build.Tasks/SetGenerator.cs
@@ -57,7 +57,6 @@ namespace Microsoft.WindowsAzure.Build.Tasks
             }
 
             var outputSet = new HashSet<string>();
-            var filesProvidedCount = filesChangedSet.Count;
             var filesFoundCount = 0;
             var useFullMapping = false;
             if (filesChangedSet.Count >= MaxFilesPossible || filesChangedSet.Count == 0)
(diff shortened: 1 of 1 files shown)


adxsdkps:

tools/BuildPackagesTask/Microsoft.Azure.Build.Tasks/SetGenerator.cs
```diff
@@ -152,5 +120,37 @@ public static HashSet<string> Generate(HashSet<string> filesChangedSet, Dictiona
...
+        /// <param name="moduleNames"></param>
+        /// <param name="mappingsDictionary"></param>
+        /// <returns></returns>
+        public static HashSet<string> Generate(string[] moduleNames, Dictionary<string, string[]> mappingsDictionary)
+        {
+            var outputSet = new HashSet<string>();
+
+            foreach (var module in moduleNames)
+            {
+                foreach (var pair in mappingsDictionary)
```
erich-wang, line 135: Suggest to use mappingDictionary.ContainsKey($"src/{module}/")

dingmeng-xue: Updated.

PR #12746, opened 2020-08-24
Generated docs using M4.

LucasYao93:
## Description

Generated docs using M4.


adxsdkps:

PR #12922, opened 2020-09-10
Improve Az.Tool.Predictor

kceiw:
## Description

This brings some improvement on the Az.Tools.Predictor
1. Rename the namespace, module name, project name, etc.
2. Collect more telemetry.
3. Remove the local NuGet feed. Use the one from nuget.org
4. Try to cancel previous HTTP request when we start a new one.
5. Fix an issue in loading the setting and change the PSReadLineOption when this module is imported.

## Checklist

### DIFF
--- a/tools/Az.Tools.Predictor/Az.Tools.Predictor.Test/Az.Tools.Predictor.Test.csproj
+++ b/tools/Az.Tools.Predictor/Az.Tools.Predictor.Test/Az.Tools.Predictor.Test.csproj
@@ -9,7 +9,7 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
     <RepoArtifacts>$(MSBuildThisFileDirectory)</RepoArtifacts>
-    <OutputPath>$(RepoArtifacts)..\..\..\artifacts\Tools\Az.Tools.Predictor\</OutputPath>
+    <OutputPath>$(RepoArtifacts)..\..\..\artifacts\Tools\Az.Tools.Predictor.Test\</OutputPath>
     <DocumentationFile>$(OutputPath)\Microsoft.Azure.PowerShell.AzPredictor.Tools.Test.xml</DocumentationFile>
   </PropertyGroup>
 
--- a/tools/Az.Tools.Predictor/build.proj
+++ b/tools/Az.Tools.Predictor/build.proj
@@ -20,7 +20,7 @@
     <MakeDir Directories="$(ArtifactFolder)" />
     <Exec Command="dotnet --version" />
 
-    <Exec Command="dotnet build AzPredictor.sln -c Release" />
+    <Exec Command="dotnet build Az.Tools.Predictor.sln -c Release" />
   </Target>
 
   <Target Name="Test" DependsOnTargets="Build">


kceiw:
@dingmeng-xue @daxian-dbw Can you review it?

tools/Az.Tools.Predictor/Az.Tools.Predictor/Az.Tools.Predictor.ps1
```diff
@@ -0,0 +1,23 @@
+$version = $PSVersionTable.PSVersion
+
+function PSConsoleHostPredictor
+{
+    Microsoft.PowerShell.Core\Set-StrictMode -Off
```
daxian-dbw, line 5: You are assuming the `PSReadLine` will always be the version that you are expecting, which may not be true. In that case, calling like `Set-PSReadLineOption -PredictionSource Plugin` and `Set-PSReadLineOption -PredictionViewStyle InlineView` will fail. It's okay to do so for testing purpose, but not for the official module.

kceiw: Yes. This script is to make it's possible to test the experience out of box. We need to address in long term what is the best way to enable the correct options if they ever change. Do you have any suggestion about enabling those options?

daxian-dbw: For the official module, it shouldn't interfere with the PSReadLine setting. The user controls the PSReadLine setting -- deciding what source to use and what view to use. Documentation is needed to show how it works end-to-end.
The module could write out a verbose message when loading to point to the documentation.

kceiw: Thanks. Let me remove this file. I think we can prepare the script separately for the testing.

tools/Az.Tools.Predictor/Az.Tools.Predictor/Az.Tools.Predictor.psd1
```diff
@@ -29,15 +29,15 @@ CompanyName = 'Microsoft Corporation'
...
 # Description of the functionality provided by this module
-Description = 'Microsoft Azure PowerShell: Provide prediction while user typing Azure PowerShell commands'
+Description = 'Microsoft Azure PowerShell Predictor: Provide prediction while user types Azure PowerShell commands'
 
 # Minimum version of the PowerShell engine required by this module
 PowerShellVersion = '7.1'
 
 # Modules that must be imported into the global environment prior to importing this module
-#RequiredModules = @(@{ModuleName = 'PSReadLine'; ModuleVersion = '2.0'; })
+RequiredModules = @(@{ModuleName = 'PSReadLine'; ModuleVersion = '2.1.0'; }, @{ModuleName = 'Az'; ModuleVersion = '3.0.0'; })
```
daxian-dbw, line 38: The PSReadLine version this module depends on should be `2.2.0-beta1` eventually, because our team has decided to move the prediction related feature to 2.2.0, and we will GA 2.1.0 with the `History` source only. But `2.2.0-beta1` is not released yet. So maybe for now, just remove the dependency of PSReadLine.

tools/Az.Tools.Predictor/Az.Tools.Predictor/Az.Tools.Predictor.psd1
```diff
@@ -62,7 +62,7 @@ PrivateData = @{
         ReleaseNotes = '* the first preview release'
 
         # Prerelease string of this module
-        # Prerelease = 'beta'
+        Prerelease = 'beta'
```
daxian-dbw, line 65:
```suggestion
        Prerelease = 'beta1'
```

PR #12965, opened 2020-09-15
Resolve dll loading conflict during help generation

isra-fel:
## Description

Unblock #12528

IotHub and DeviceProvisioningServices are depending on Newtonsoft.Json 12.0.3, however our tools are using 11.0.2. This caused dll loading conflict during help generation in static analysis.
Solution is to use a custom ALC (assembly loading context) to load each module and their dependencies, so that they won't conflict with each other or common tools.

(No changelog required because it's only code change to tools)

## Checklist

### DIFF
--- a/tools/Tools.Common/Loaders/CmdletLoader.cs
+++ b/tools/Tools.Common/Loaders/CmdletLoader.cs
@@ -20,47 +20,32 @@ using System.Management.Automation;
 using Tools.Common.Models;
 using Tools.Common.Extensions;
 using System.IO;
+using System.Runtime.Loader;
 
 namespace Tools.Common.Loaders
 {
-// TODO: Remove IfDef
-#if NETSTANDARD
     public class CmdletLoader
-#else
-    public class CmdletLoader : MarshalByRefObject
-#endif
     {
         public static ModuleMetadata ModuleMetadata;
 
-        public ModuleMetadata GetModuleMetadata(string assemblyPath, List<string> commonOutputFolders)
-        {
-            AppDomain.CurrentDomain.AssemblyResolve += (sender, args) =>
-            {
-                foreach (var commonOutputFolder in commonOutputFolders)
-                {
-                    var assemblyName = args.Name.Substring(0, args.Name.IndexOf(","));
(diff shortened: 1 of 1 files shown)


tools/Tools.Common/Loaders/CmdletLoader.cs
```diff
@@ -20,47 +20,32 @@
...
-                foreach (var commonOutputFolder in commonOutputFolders)
-                {
-                    var assemblyName = args.Name.Substring(0, args.Name.IndexOf(","));
-                    var dll = Directory.GetFiles(commonOutputFolder, "*.dll").FirstOrDefault(f => Path.GetFileNameWithoutExtension(f) == assemblyName);
-                    if (dll == null)
-                    {
-                        continue;
-                    }
-
-                    return Assembly.LoadFrom(dll);
```
isra-fel, line 48: `Assembly.LoadFrom()` will load the dll into the default ALC, so the dlls may conflict.

tools/Tools.Common/Loaders/CmdletLoader.cs
```diff
@@ -208,9 +193,40 @@ public ModuleMetadata GetModuleMetadata(string assemblyPath)
...
+        {
+            foreach (var assemblyPath in _assemblyDirectories)
+            {
+                var dll = Directory.GetFiles(assemblyPath, "*.dll").FirstOrDefault(f => Path.GetFileNameWithoutExtension(f) == assemblyName.Name);
+                if (dll == null)
+                {
+                    continue;
+                }
+
+                return LoadFromAssemblyPath(dll);
```
isra-fel, line 227: The assembly resolving logic is the same as before. I just use `AssemblyLoadContext.LoadFromAssemblyPath()` to load the dll into individual ALCs

isra-fel:
Root cause: we depend on Microsoft.PowerShell.SDK 6.1.0, which depends on Newtonsoft.Json 11.0.2.
Before loading the dll, we run a powershell script to read the psd1 file, which loads powershell sdk as well as newtonsoft.json 11.0.2, causing other versions not able to be loaded.

PR #12984, opened 2020-09-16
Predict parameter values in the suggestion

kceiw:
## Description
- The user may have used the values in some Az command. That is a hint that the user may want to use those values again in new commands. So we construct and record the parameter values in this way that we can use that in the new command.
   * We extract the resource name from the Az command name, for example: `New-AzResourceGroup`. The resource name is ResourceGroup. Then we use it to construct the a key for each parameter name. The key is in the form <Resource Name><Parameter Name>. We store the key and the parameter value.
   * When we provide a new suggestion, we try to find the parameter value by the parameter name of the new command. If we find one, then that is used in the suggestion.
- To assist testing, a new project MockPSConsole is added. This is a pseudo ps console which echoes back most commands instead of executing it. So that we can test with Az commands knowing that it won't really affect the Azure. There are some special commands which actually execute. These commands allow us to change the setting and importing new modules.

## Checklist


kceiw:
@albaloo @daxian-dbw This is the first step to predict parameter value. It uses the parameter value from previous command to predict the one in new command. We can improve the prediction in future PRs.

wyunchi-ms (approved):

PR #13098, opened 2020-09-28
return multiple suggestion

kceiw:
## Description
PSReadLine can show multiple suggestions on the list view. This change returns multiple suggestions. It also enable the flag to receive feedback on when the user accepts a suggestion.

## Checklist

### DIFF
--- a/tools/Az.Tools.Predictor/Az.Tools.Predictor/AzPredictor.cs
+++ b/tools/Az.Tools.Predictor/Az.Tools.Predictor/AzPredictor.cs
@@ -58,6 +58,8 @@ namespace Microsoft.Azure.PowerShell.Tools.AzPredictor
         private readonly ITelemetryClient _telemetryClient;
         private readonly Settings _settings;
 
+        private Queue<string> _lastTwoMaskedCommands = new Queue<string>(AzPredictorConstants.CommandHistoryCountToProcess);
+
         /// <summary>
         /// Constructs a new instance of <see cref="AzPredictor"/>
         /// </summary>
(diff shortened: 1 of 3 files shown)


kceiw:
@daxian-dbw and @albaloo Can you review this PR?

tools/Az.Tools.Predictor/Az.Tools.Predictor/AzPredictor.cs
```diff
@@ -56,16 +56,19 @@ public sealed class AzPredictor : ICommandPredictor
...
         /// <param name="service">The service that provides the suggestion</param>
         /// <param name="telemetryClient">The client to collect telemetry</param>
-        public AzPredictor(IAzPredictorService service, ITelemetryClient telemetryClient)
+        /// <param name="settings">The settings of the service</param>
+        public AzPredictor(IAzPredictorService service, ITelemetryClient telemetryClient, Settings settings)
         {
             this._service = service;
             this._telemetryClient = telemetryClient;
+            this._settings = settings;
         }
```
daxian-dbw, line 72: The `StartEarlyProcessing` seems to be doing a lot redundant work. The list of history passed in is a moving window, so every time this method gets called, the latest command line is added to the list and the oldest command line is removed, so ideally, you only need to parse the new one, which should be at the index 0 of the list.

kceiw: I agree with what you said that we have redundant work there. I'll try to reduce that part of work.

PR #13113, opened 2020-09-29
Remove dependencies

kceiw:
## Description
This module used to depend on Az module which makes the loading time a little long. The reason why that's needed is to solve conflict when they both are loaded. They both depend on Microsoft.Azure.PowerShell.Common.dll and Microsoft.Azure.PowerShell.Authentication.Abstractions.dll. That cause conflicts in loading the assemblies. Once one module is loaded, the other one won't work. Though technically this module only augments Az module experience and it won't add anything if that doesn't exist, we can usually have Az module loaded when this module is used. The problem is loading Az module also takes a long time. The fix here is to remove the dependencies on Microsoft.Azure.PowerShell.Common.dll and Microsoft.Azure.PowerShell.Authentication.Abstractions.dll. We only need to find the configuration file about data collection. So I copied piece of the code to do that. This also bring down the our module size when we package it since we don't need to package the dependencies.

## Checklist


kceiw:
@albaloo @daxian-dbw @BethanyZhou  Can you review it?

isra-fel:
Hi @kceiw thanks for the contribution. Our team are just back from national holiday so sorry for late response. Anyway the changes LGTM :)

kceiw:
@isra-fel Thanks.

Issue #13173, opened 2020-10-10
Update databricks workspace withouth -EncryptionKeyVersion will keep last key version


isra-fel:
## Description

If you have once updated a databricks workspace with one key version (for CMK purpose), and then update it again without key version, the last version is kept, but we expect it to be empty string.

## Repro

1. Created and prepared a new Azure Databricks workspace for encryption
```powershell
New-AzDatabricksWorkspace -Name LG-ADB_CMK-1 -ResourceGroupName ADBPowerShellRG  -Location westus2 -Sku Premium -PrepareEncryption
```
 

2. Configured DBFS encryption with customer-managed keys
```powershell
Update-AzDatabricksWorkspace -Name LG-ADB_CMK-1 -ResourceGroupName ADBPowerShellRG

               -EncryptionKeySource Microsoft.Keyvault -EncryptionKeyName LGADBCMKKey -EncryptionKeyVaultUri  https://lgadbcmkkv2.vault.azure.net:443 -EncryptionKeyVersion 4d554df209dc4feb8ab528233782d62d
```
 

3. Tried updating DBFS encryption configuration with the new key without specifying the Encryption Key Version
```powershell
Update-AzDatabricksWorkspace -Name LG-ADB_CMK-1 -ResourceGroupName ADBPowerShellRG

               -EncryptionKeySource Microsoft.Keyvault -EncryptionKeyName LGADBCMKKey -EncryptionKeyVaultUri  https://lgadbcmkkv2.vault.azure.net:443
```

## Root cause

In implementation of the `Update-` command, we first do a GET then PUT. If user doesn't specify key version, the key version from the GET request is reused in the PUT request. We should check and set it to "".

## work-around

For now you may forcely set key version to empty string. E.g.
```powershell
 Update-AzDatabricksWorkspace -Name LG-ADB_CMK-1 -ResourceGroupName yemingtemp -EncryptionKeySource Microsoft.Keyvault -EncryptionKeyName LGADBCMKKey -EncryptionKeyVaultUri https://yemingkv03.vault.azure.net:443 -EncryptionKeyVersion ""
```

LucasYao93:
Hi @isra-fel,
I changed the code. Please help me to check if it is correct.

## Changed code
```csharp
// custom/Update-AzDatabricksWorkspace.ps1
            if ($hasEncryptionKeyVersion) {
                $workspace.EncryptionKeyVersion = $EncryptionKeyVersion
            } else {
[...]

PR #13197, opened 2020-10-13
[Databricks] Fixed bug that withouth -EncryptionKeyVersion will keep last key version.

LucasYao93:
## Description
#13173

### DIFF
--- a/src/Databricks/.gitignore
+++ b/src/Databricks/.gitignore
@@ -12,3 +12,4 @@ test/*-TestResults.xml
 /*.snk
 /*.csproj
 /*.nuspec
+tools


src/Databricks/test/utils.ps1
```diff
@@ -8,7 +8,7 @@ function RandomString([bool]$allChars, [int32]$len) {
 $env = @{}
 function setupEnv() {
     Write-Host -ForegroundColor Yellow "WARNING: Need to use Az.KeyVault module, Please check if installed Az.KeyVault(2.0.0 or Greater)."
-    # Import-Module -Name Az.KeyVault
+    Import-Module -Name Az.KeyVault
```
isra-fel, line 11: Should this line be commented out?

src/Databricks/custom/Update-AzDatabricksWorkspace.ps1
```diff
@@ -184,6 +184,11 @@ function Update-AzDatabricksWorkspace {
             }
             if ($hasEncryptionKeyVersion) {
                 $workspace.EncryptionKeyVersion = $EncryptionKeyVersion
+            } else {
+                if ($workspace.PrepareEncryption)
```
isra-fel, line 188: Is it because that you cannot set `EncryptionKeyVersion` unless the workspace has prepared for encryption?

src/Databricks/test/utils.ps1
LucasYao93: I don't think it should be commented out because it will not execute when the test case is run through the playback model.

src/Databricks/custom/Update-AzDatabricksWorkspace.ps1
LucasYao93: Yes

isra-fel (approved):
LGTM. Thanks!
Please continue to move the code to master branch

PR #13307, opened 2020-10-22
Fix Error in Get-AzAutomationDscNode Example 1

DominicJ2:
In Example 1, there is a different string passed for the ResourceGroupName in the first function call and the second function call. This string should be the same value for both function calls. All of the other examples use the ResourceGroupName `ResourceGroup03`, so I swapped to use that as the correct value.

## Checklist


DominicJ2:
In Example 1, there is a different string passed for the ResourceGroupName in the first function call and the second function call. This string should be the same value for both function calls. All of the other examples use the ResourceGroupName ResourceGroup03, so I swapped to use that as the correct value.

isra-fel (approved):
Makes sense. Thanks for your contribution 👍

PR #13327, opened 2020-10-26
Update how we send request and collect telemetry

kceiw:
## Description
There are a few issues we address in this PR:
1. Add the right parameter to the service endpoint. The service actually expects those set. Though it also works if those are not set, we'll try to set it so that we don't need any workaround at the service side in future.
2. We don't want to send request for the same command history. We don't want to create too much unnecessary traffic to the service.
3. When we return the suggestion, we substitute the actual parameter values if the user changes it. We don't want to collect those in the telemetry. Instead, we return the parameter names-values pair from the prediction model we create. We collect the one from the model instead. So we don't collect user input except the command name and the parameter names.

## Checklist


kceiw:
@baopingz Can you review it?

dingmeng-xue (approved):

PR #13590, opened 2020-11-25
Pass principal type during role assignment put operations

dagoroz:
## Description

Added checks to ObjectPrincipal type to see if it's included in the PUT request, if not it adds it automatically

## Checklist

        - {Please put the link here}

### DIFF
--- a/src/Resources/Resources/Models.Authorization/AuthorizationClient.cs
+++ b/src/Resources/Resources/Models.Authorization/AuthorizationClient.cs
@@ -132,7 +132,7 @@ namespace Microsoft.Azure.Commands.Resources.Models.Authorization
 
             if (asignee == null)
             {
-                throw new ArgumentException("No AD object could be found with current parameters, please confirm the information provided is correct and try again");
+                throw new ArgumentException(ProjectResources.NoADObjectFound);
             }
 
             string principalId = asignee.Id;
--- a/src/Resources/Resources/Properties/Resources.Designer.cs
+++ b/src/Resources/Resources/Properties/Resources.Designer.cs
@@ -484,6 +484,15 @@ namespace Microsoft.Azure.Commands.Resources.Properties {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to No AD object could be found with current parameters, please confirm the information provided is correct and try again.
+        /// </summary>
+        internal static string NoADObjectFound {
+            get {
+                return ResourceManager.GetString("NoADObjectFound", resourceCulture);
+            }
+        }
+        
+        /// <summary>
         ///   Looks up a localized string similar to Cannot find any deployments in resource group &apos;{0}&apos;..
         /// </summary>
         internal static string NoDeploymentWereFound {
--- a/src/Resources/Resources/Properties/Resources.resx
+++ b/src/Resources/Resources/Properties/Resources.resx
@@ -450,4 +450,7 @@
   <data name="UpdatingRoleAssignment" xml:space="preserve">
     <value>Updating role assignment for AD object '{0}' on scope '{1}' with role definition '{2}'</value>
   </data>
+  <data name="NoADObjectFound" xml:space="preserve">
+    <value>No AD object could be found with current parameters, please confirm the information provided is correct and try again</value>
+  </data>
 </root>


src/Resources/Resources.Test/ScenarioTests/RoleAssignmentTests.cs
```diff
@@ -174,7 +174,7 @@ public void RaGetByUPNWithExpandPrincipalGroups()
             TestRunner.RunTestScript("Test-RaGetByUPNWithExpandPrincipalGroups");
         }
 
-        [Fact]
+        [Fact(Skip = "Breaking PR in 2 parts needs this test to be disabled temporarily")]
```
dagoroz, line 177: The change I'm making changes the codeflow intended for when creating role assignments through an SP (e.g. Azure app)
The flow is not broken but it does need to be updated. That will be done on a separate PR as it might be an extensive change by itself.

src/Resources/Resources/Models.Authorization/AuthorizationClient.cs
```diff
@@ -128,7 +128,15 @@ public IEnumerable<PSRoleDefinition> FilterRoleDefinitionsByCustom(string scope,
         /// <returns>The created role assignment object</returns>
         public PSRoleAssignment CreateRoleAssignment(FilterRoleAssignmentsOptions parameters, Guid roleAssignmentId = default(Guid))
         {
-            string principalId = ActiveDirectoryClient.GetObjectId(parameters.ADObjectFilter);
+            PSADObject asignee = ActiveDirectoryClient.GetADObject(parameters.ADObjectFilter);
+
+            if (asignee == null)
+            {
+                throw new ArgumentException("No AD object could be found with current parameters, please confirm the information provided is correct and try again");
```
isra-fel, line 135: Please use a resource string instead

src/Resources/Resources.Test/ScenarioTests/RoleAssignmentTests.ps1
```diff
@@ -707,17 +707,19 @@ function Test-RaCreatedBySP
     #Setup
     # Conect to azure with SP
     # If you need to re-record replace this setup
-    $passwd = ConvertTo-SecureString 'MCg1SVb3MX' -AsPlainText -Force
-    $pscredential = New-Object System.Management.Automation.PSCredential('c82180ed-2f4b-4cb2-965f-ec5088751710' , $passwd)
-    $tenantId = '1462fd46-afe5-491b-a340-31ebae81d1ce'
-    Connect-AzAccount -ServicePrincipal -Credential $pscredential -Tenant $tenantId
+    $passwd = ConvertTo-SecureString 'm7XHgb9cDVP60L4.3bT~VN1Rhjpo-.ZTU-' -AsPlainText -Force
```
isra-fel, line 710: Just want to double check if the credential/SP has been removed

src/Resources/Resources.Test/ScenarioTests/RoleAssignmentTests.cs
isra-fel: Sounds OK, but what about the above cases with `Skip = "Test is not deterministic, only works on record mode"`?

isra-fel (changes requested):
Generally LGTM. Added a few comments

src/Resources/Resources.Test/ScenarioTests/RoleAssignmentTests.cs
dagoroz: the tests themselves pass, in record mode they use my credentials and my tenant and everything goes well ✅
the problem is that the tests generate guids on the fly which break the tests on playback mode. When trying to read the guid from the session records, since it always change it's impossible to request the same guid twice.

src/Resources/Resources.Test/ScenarioTests/RoleAssignmentTests.ps1
dagoroz: I removed the credentials and the SP from subscription

src/Resources/Resources/Models.Authorization/AuthorizationClient.cs
dagoroz: changed

isra-fel (approved):

PR #13669, opened 2020-12-04
Az.Tools.Predictor refactor and performance improvement.

kceiw:
## Description
There are a few things going on during the refactor: telemetry, readability, and performance.

1. Replace Tuple/ValueTuple with more meaningful types that don't use 'Item1' and 'Item2'.
2. Replace Newtonsoft Json library with the built-in System.Text.Json.
3. Reformat the telemetry events. Combine the error events with the normal event. Just have a field in the event to indicate if there is an exception or not.
4. Collect the event if the http request to the server has been sent or not. The content isn't collected.
5. Transform data format and send telemetry event in a worker in the default thread pool. This can also improve the performance since that removes the burden on the thread that's handling callback from PSReadLine.
6. Validate parameter values and also catch http response error early instead of when we try to parse the response content.
7. Remove duplicate parsing on the user input.
8. Reduce string manipulation.

## Checklist

        - {Please put the link here}


tools/Az.Tools.Predictor/Az.Tools.Predictor/TelemetryData.cs
```diff
@@ -0,0 +1,179 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// ----------------------------------------------------------------------------------
+
+using System;
+using System.Management.Automation.Language;
+
+namespace Microsoft.Azure.PowerShell.Tools.AzPredictor
```
mirdaki, line 18: Is it normal to have many different classes in the same file?

kceiw: I remove `TelemetryData` and move the other interface/class to seperate files.

mirdaki:
Looks good to me!

mirdaki (approved):

tools/Az.Tools.Predictor/Az.Tools.Predictor/Telemetry/AzPredictorTelemetryClient.cs
```diff
@@ -0,0 +1,278 @@
...
+#endif
+        }
+
+        /// <summary>
+        /// Check whether the data collection is opted in from user.
+        /// </summary>
+        /// <returns>true if allowed.</returns>
+        private static bool IsDataCollectionAllowed()
+        {
+            if (AzurePSDataCollectionProfile.Instance.EnableAzureDataCollection == true)
```
baopingz, line 160: nit: just return AzurePSDataCollectionProfile.Instance.EnableAzureDataCollection

baopingz (approved):
looks good.

PR #13721, opened 2020-12-11
[Synapse] -  Fixed deserialization error when create Pipeline/Dataset/Trigger through DefinitionFile

idear1203:
## Description

<TL;DR>
This PR aims to fix deserialization issue for New-*Pipeline/Dataset/Trigger cmdlets.
<TL:DR>

Synapse has integrated some Azure Data Factory's fundamental features, for example pipeline, dataset, triggers, etc. We refer these resources as artifacts. One common use scenario for New-*Artifact is, users provide a definition file that contains request body, the PowerShell command reads the definition file, deserialize its content to some SDK models and call methods to send a PUT request to create some artifact resource.

Previously, these related PowerShell cmdlets suffer from a deserialization issue. In short, this is because, The PowerShell consumes  Azure.Analytics.Synapse.Artifacts C# SDK. The SDK is track 2 data plane SDK. Its serializer/deserializer (methods to serialize/deserialize SDK models) are not exposed. For example, see [PipelineResource.Serialization.cs](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/synapse/Azure.Analytics.Synapse.Artifacts/src/Generated/Models/PipelineResource.Serialization.cs) - methods in this file are not `public`.

Before this change, our deserialization approach in PowerShell is as below: we writes some wrapper classes, let's say PSArtifact.cs. First of call, we leverage Newtonsoft.Json library for serialization to an instance of `PSArtifact`, then we write and call `ToSdkObject` method inside `PSArtifact` to transform `PSArtifact` to some SDK model `SdkArtifact`. Then call SDK method with `SdkArtifact`. This approach is buggy and we have seen several customer reported deserialization failure issues recently. At code level, this is because we have to write lots of `ToSdkObject` methods by hand and provide parameterless constructor for many files (Otherwise the Newtonsoft.Json will complain that it cannot find proper ctor or call some unexpected ctor).

In this PR, we are trying to fix the deserialization issue by calling SDK's original serialization/de-serialization methods, rather than leverage Newtonsoft.Json library. Due to the invisibility of those SDK serialization methods, we cannot call them directly. So we turn to use reflection instead. The mainly related files are:
- We write a helper class to call non-public SDK methods through reflection: [JsonConvert](https://github.com/idear1203/azure-powershell/blob/3a7d4cfdadbe8c0014b7984c6b705ac5c2c8984d/src/Synapse/Synapse/Common/JsonConvert.cs)
- We use the [JsonConvert](https://github.com/idear1203/azure-powershell/blob/3a7d4cfdadbe8c0014b7984c6b705ac5c2c8984d/src/Synapse/Synapse/Common/JsonConvert.cs) to update the code in [SynapseAnalyticsArtifactsClient](https://github.com/idear1203/azure-powershell/blob/3a7d4cfdadbe8c0014b7984c6b705ac5c2c8984d/src/Synapse/Synapse/Models/SynapseAnalyticsArtifactsClient.cs#L49)

Besides the major change that is mentioned above, the PR also does the following items:
- Upgrade `Azure.Analytics.Synapse.Artifacts` to the latest version and update code in `Set-AzSynapseNotebook` to solve an payload validation issue due to a breaking change at backend side.
- Remove 200+ unused model wrapper classes

## Checklist

        - {Please put the link here}

### DIFF
--- a/src/Synapse/Synapse/Common/JsonConvert.cs
+++ b/src/Synapse/Synapse/Common/JsonConvert.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Azure.Commands.Synapse.Common
     {
         internal static T DeserializeObject<T>(string rawJsonContent)
         {
-            var document = JsonDocument.Parse(rawJsonContent, default);
+            var document = JsonDocument.Parse(rawJsonContent);
             MethodInfo deserializer = typeof(T).GetMethod($"Deserialize{typeof(T).Name}", BindingFlags.NonPublic | BindingFlags.Static);
             return (T) deserializer.Invoke(null, new object[] { document.RootElement });
         }


src/Synapse/Synapse/Common/JsonConvert.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Reflection;
+using System.Text.Json;
+
+namespace Microsoft.Azure.Commands.Synapse.Common
+{
+    internal static class JsonConvert
+    {
+        internal static T DeserializeObject<T>(string rawJsonContent)
+        {
+            var document = JsonDocument.Parse(rawJsonContent, default);
```
wyunchi-ms, line 11: Feature `default literal` is not available in C# 7.0.

idear1203: Thanks for the suggestion. Fixed.

idear1203:
@wyunchi-ms could you please help review again?

wyunchi-ms:
Could you please add these modification into `Changelog.md`.

idear1203:
@wyunchi-ms there is already an entry in the change log:
> Fixed deserialization error when create Pipeline/Dataset/Trigger through DefinitionFile

wyunchi-ms (approved):

PR #13840, opened 2020-12-25
[PostgreSql]Add parameter MinimalTlsVersion for the cmdlets

LucasYao93:
## Description

Add New/Update-AzPostgreSqlServer is missing parameter for minimal-tls-version.


msJinLei (changes requested):
The same as MySql

LucasYao93:
> The same as MySql

Update completed

msJinLei (approved):

PR #13863, opened 2020-12-31
Only use supported commands for prediction request

kceiw:
- We only request predictions for the last two commands. When the
  commands are not supported, we replace them with "start_of_snippet".
  The problem of that is when the user has inputted a few unsupported
  commands such as assignments, we'll start to request prediction for
  [ "start_of_snippet", "start_of_snippet" ]. That'll the same as
  resetting the prediction as the beginning of the session. While in
  fact there may be Az commands in the history.
- The change here is not to use those unsupported commands to request
  predictions. We'll skip the unsupported commands unless we don't have
  enough commands from history to use for the prediction.

## Checklist

        - {Please put the link here}

### DIFF
--- a/tools/Az.Tools.Predictor/Az.Tools.Predictor.Test/AzPredictorTests.cs
+++ b/tools/Az.Tools.Predictor/Az.Tools.Predictor.Test/AzPredictorTests.cs
@@ -38,9 +38,9 @@ namespace Microsoft.Azure.PowerShell.Tools.AzPredictor.Test
         public AzPredictorTests(ModelFixture modelFixture)
         {
             _fixture = modelFixture;
-            var starory = $"{AzPredictorConstants.CommandPlaceholder}{AzPredictorConstants.CommandConcatenator}{AzPredictorConstants.CommandPlaceholder}";
+            var startHistory = $"{AzPredictorConstants.CommandPlaceholder}{AzPredictorConstants.CommandConcatenator}{AzPredictorConstants.CommandPlaceholder}";
 
-            _service = new MockAzPredictorService(starory, _fixture.PredictionCollection[starory], _fixture.CommandCollection);
+            _service = new MockAzPredictorService(startHistory, _fixture.PredictionCollection[startHistory], _fixture.CommandCollection);
             _telemetryClient = new MockAzPredictorTelemetryClient();
             _azPredictor = new AzPredictor(_service, _telemetryClient, new Settings()
             {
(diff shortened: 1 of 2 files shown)


tools/Az.Tools.Predictor/Az.Tools.Predictor.Test/AzPredictorTests.cs
```diff
@@ -37,12 +37,12 @@ public sealed class AzPredictorTests
         /// </summary>
         public AzPredictorTests(ModelFixture modelFixture)
         {
-            this._fixture = modelFixture;
-            var startHistory = $"{AzPredictorConstants.CommandPlaceholder}{AzPredictorConstants.CommandConcatenator}{AzPredictorConstants.CommandPlaceholder}";
+            _fixture = modelFixture;
+            var starory = $"{AzPredictorConstants.CommandPlaceholder}{AzPredictorConstants.CommandConcatenator}{AzPredictorConstants.CommandPlaceholder}";
```
jjaguirre394, line 41: It seems this variable name was accidentally updated? If not, I am unclear what is a starory.

tools/Az.Tools.Predictor/Az.Tools.Predictor/AzPredictor.cs
```diff
@@ -83,11 +83,22 @@ public void StartEarlyProcessing(IReadOnlyList<string> history)
...
+                // We try to find the commands to request predictions for.
+                // We should only have "start_of_snippet" when there are no enough Az commands for prediction.
+                // We then ignore that when there are new "start_of_snippet".
+                // This is the scenario.
+                // 1. New-AzResourceGroup -Name ****
+                // 2. $resourceName="Test"
+                // 3. $resourceLocation="westus2"
+                // 4. New-AzVM -Name $resourceName -Location $resourceLocation
+                //
+                // We'll replace 2 and 3 with "start_of_snippet" but if we request prediction for 2 and 3, that'll reset the
```
jjaguirre394, line 95: Just wanted to say these comments are very helpful being new to the code base. I know the logic of this application can get complicated at times. Just wanted to say we should look for clarification comments like this when we can add them :)

jjaguirre394 (dismissed):

tools/Az.Tools.Predictor/Az.Tools.Predictor/AzPredictor.cs
mirdaki: Agreed!

mirdaki (dismissed):
Looks good!

mirdaki (approved):

PR #13864, opened 2021-01-01
Improve getting command ast and parse parameters

kceiw:
- We use the RelatedAsts from PSReadLine to get the command AST as the
  user input. This is what our prediction will try to match.
- Now we can parse the parameter in the format of "-Name:Value".
- We don't parse  positioinal parameters yet.

## Checklist

        - {Please put the link here}


jjaguirre394 (approved):

tools/Az.Tools.Predictor/Az.Tools.Predictor/AzPredictorService.cs
```diff
@@ -143,22 +144,46 @@ protected virtual void Dispose(bool disposing)
...
+            try
+            {
+                inputParameterSet = new ParameterSet(commandAst);
+            }
+            catch when (!IsSupportedCommand(commandName))
+            {
+                // We only ignore the exception when the command name is not supported.
+                // For the supported ones, this most likely happens when positional parameters are used.
+                // We want to collect the telemetry about the exception how common a positional parameter is used.
+                return null;
```
mirdaki, line 183: Is there something besides null we can return here? Just don't like giving "others" null if it can be helped. If not, please add it to the method comments.

tools/Az.Tools.Predictor/Az.Tools.Predictor/ParameterSet.cs
```diff
@@ -68,11 +74,22 @@ public ParameterSet(CommandAst commandAst)
...
+                if (parameter != null)
                 {
-                    parameters.Add(new Parameter(parameterName.ParameterName, (parameterValue == null) ? null : CommandLineUtilities.UnescapePredictionText(parameterValue.ToString())));
+                    var value = parameterValue?.ToString();
+                    if (value == null)
+                    {
+                        value = parameter.Argument?.ToString();
+                    }
+
+                    if (value != null)
```
mirdaki, line 87: I just want to make sure you want the first if to be checking `parameterValue?.ToString();` and the second if to be checking if `parameter.Argument?.ToString()`?

tools/Az.Tools.Predictor/Az.Tools.Predictor/ParameterValuePredictor.cs
```diff
@@ -27,7 +24,15 @@ public ParameterValuePredictor()
...
             var mappingFilePath = Path.Join(directory, "command_param_to_resource_map.json");
-            _command_param_to_resource_map = JsonSerializer.Deserialize<Dictionary<string, Dictionary<string, string>>>(File.ReadAllText(mappingFilePath), JsonUtilities.DefaultSerializerOptions);
+
+            try
+            {
+                _command_param_to_resource_map = JsonSerializer.Deserialize<Dictionary<string, Dictionary<string, string>>>(File.ReadAllText(mappingFilePath), JsonUtilities.DefaultSerializerOptions);
+            }
+            catch
+            {
+                // We don't want it to crash the module when the file doesn't exist or when it's mal-formatted.
```
mirdaki, line 34: Could (or should) we add telemetry to report a crash?

tools/Az.Tools.Predictor/Az.Tools.Predictor/ParameterValuePredictor.cs
```diff
@@ -55,18 +60,23 @@ public void ProcessHistoryCommand(CommandAst command)
         /// <returns>The parameter value from the history command. Null if that is not available.</returns>
         public string GetParameterValueFromAzCommand(string commandNoun, string parameterName)
         {
-            if (_command_param_to_resource_map.ContainsKey(commandNoun))
+            parameterName = parameterName.ToLower();
+            var key = parameterName;
+            Dictionary<string, string> commandNounMap = null;
+
+            if (_command_param_to_resource_map?.TryGetValue(commandNoun, out commandNounMap) == true)
```
mirdaki, line 67: Is this necessary? I'm OK with being explicit, but none of the other ifs are.

tools/Az.Tools.Predictor/Az.Tools.Predictor/ParameterValuePredictor.cs
```diff
@@ -55,18 +60,23 @@ public void ProcessHistoryCommand(CommandAst command)
...
+                    key = parameterNameMappedValue;
                 }
             }
+
+            if (_localParameterValues.TryGetValue(key, out var value))
+            {
+                return value;
+            }
+
             return null;
```
mirdaki, line 80: I realize this isn't a new change for this PR, but would an empty string work here instead of a null?

mirdaki (dismissed):
These are not picks or clarifications.

tools/Az.Tools.Predictor/Az.Tools.Predictor/AzPredictorService.cs
kceiw: I add to the method comments to indicate null is returned when the user input context isn't valid/supported. It's v.s. a default value (en empty collection) meaning that we try to find the suggestion but there are not.

tools/Az.Tools.Predictor/Az.Tools.Predictor/ParameterSet.cs
kceiw: The first is to check `parameterValue?.ToString()`. If that's null, then I'll set `value` to `parameter.Argument?.ToString()`. So the second `if` checks whether the `value` is null. `value` is either from `parameterValue` or `parameter.Argument`.

There are two formats of parameter name and value pair here.
1. "-name value". The value is in `parameterValue`.
2. "-name:value". The value is in `parameter.Argument`.

tools/Az.Tools.Predictor/Az.Tools.Predictor/ParameterValuePredictor.cs
kceiw: It is. It's using "?." which returns a "bool?". Cannot use "bool?" directly inside `if`.  The check means _command_param_to_resource_map` is not `null` and it has the key `commandNoun`.

tools/Az.Tools.Predictor/Az.Tools.Predictor/ParameterValuePredictor.cs
kceiw: I think the caller doesn't care the case if it's null or empty string and will treat them the same. But I also think that `null` has the meaning that the value it's looking for doesn't exist v.s. it exists and it's empty. In that case, I'll keep `null` for now.

tools/Az.Tools.Predictor/Az.Tools.Predictor/ParameterValuePredictor.cs
kceiw: Good point. I add it.

tools/Az.Tools.Predictor/Az.Tools.Predictor/ParameterSet.cs
mirdaki: OK, just wanted to make sure since it looked like something that could be mixed up.

tools/Az.Tools.Predictor/Az.Tools.Predictor/ParameterValuePredictor.cs
mirdaki: Oh I see now, thank you for clarifying.

mirdaki (dismissed):
Thank you for the updates and clarifications!

kceiw:
@dingmeng-xue Looks like the build failure isn't related to this change. Can you take a look?

dingmeng-xue:
It's random failure in ADO. Is this PR ready for merge? I will force merge it.

PR #14437, opened 2021-03-04
Updating & enhancing Service Fabric cmdlets to facilitate Ubuntu 18 & Linux OS migration scenarios.

maburlik:
## Description

* Add parameters to `Add-AzServiceFabricNodeType` to facilitate easy alternate OS image creation for new node type.
    Parameters: `VMImagePublisher`, `VMImageOffer`, `VMImageSku`, `VMImageVersion`
    - Example usage: For a Ubuntu 16.04 cluster, add a new Ubuntu 18 node type: <br/>
    ```powershell
    $resouceGroup="Group1"
    $clusterName="Contoso01SFCluster"
    $nodeTypeName="nt2"
    $securePassword = ConvertTo-SecureString -String 'Password$123456' -AsPlainText -Force

    Add-AzServiceFabricNodeType -ResourceGroupName $resouceGroup  -ClusterName $clusterName -NodeType $nodeTypeName -Capacity 5 -VmUserName adminuser -VmPassword $securePassword -DurabilityLevel Silver -Verbose -VMImageSku 18.04-LTS
    ```
* Add parameter `IsPrimaryNodeType` to `Add-AzServiceFabricNodeType` to be able to create an additional primary node type,
    for the purpose of transitioning the primary node type to another one in the case of OS migration.
* `Add-AzServiceFabricNodeType` now correctly copies the LinuxDiagnostic extension. This was previously not working for Linux.
* `Add-AzServiceFabricNodeType` now correctly copies the RDP/SSH load balancer inbound NAT port mapping to the new node type.
* Add template for `UbuntuServer1804` for creating Ubuntu 18.04 clusters using `New-AzServiceFabricCluster`.
* `Remove-AzServiceFabricNodeType` was incorrectly blocking Bronze durability node types for removal, and this has been updated
    to only block when the Bronze durability level differs between the SF node type and VMSS setting.
* Added cmdlet `Update-AzServiceFabricVmImage` to update the delivered SF runtime package type. This must be changed when migrating from Ubuntu 16 to 18.
* Added cmdlet `Update-AzServiceFabricNodeType` to update the properties of a cluster node type. For now this is solely used to update whether the node type is primary via bool parameter `-IsPrimaryNodeType $false`.
* `Update-AzServiceFabricReliability` is now able to update reliability level when the cluster has more than one primary node type.
    To do this, the name of the node type is supplied via the new -NodeType parameter.

## Checklist

- [ x ] I have read the [_Submitting Changes_](../blob/master/CONTRIBUTING.md#submitting-changes) section of [`CONTRIBUTING.md`](../blob/master/CONTRIBUTING.md)
- [ x ] The title of the PR is clear and informative
- [ x ] The appropriate `ChangeLog.md` file(s) has been updated:
- [ x ] The PR does not introduce [breaking changes](../blob/master/documentation/breaking-changes/breaking-changes-definition.md)
- [ x ] If applicable, the changes made in the PR have proper test coverage
        - {Please put the link here}

### DIFF
--- a/src/ServiceFabric/ServiceFabric/Az.ServiceFabric.psd1
+++ b/src/ServiceFabric/ServiceFabric/Az.ServiceFabric.psd1
@@ -112,6 +112,7 @@ CmdletsToExport = 'Add-AzServiceFabricClientCertificate',
                'Add-AzServiceFabricManagedNodeTypeVMSecret', 
                'Remove-AzServiceFabricManagedNodeTypeVMExtension', 
                'Restart-AzServiceFabricManagedNodeType',
+               'Update-AzServiceFabricNodeType',
                'Update-AzServiceFabricVmImage'
 
 # Variables to export from this module
--- a/src/ServiceFabric/ServiceFabric/ChangeLog.md
+++ b/src/ServiceFabric/ServiceFabric/ChangeLog.md
@@ -37,6 +37,7 @@
 * `Remove-AzServiceFabricNodeType` was incorrectly blocking Bronze durability node types for removal, and this has been updated
     to only block when the Bronze durability level differs between the SF node type and VMSS setting.
 * Added cmdlet `Update-AzServiceFabricVmImage` to update the delivered SF runtime package type. This must be changed when migrating from Ubuntu 16 to 18.
+* Added cmdlet `Update-AzServiceFabricNodeType` to update the properties of a cluster node type. For now this is solely used to update whether the node type is primary via bool parameter `-IsPrimaryNodeType $false`.
 * `Update-AzServiceFabricReliability` is now able to update reliability level when the cluster has more than one primary node type.
     To do this, the name of the node type is supplied via the new -NodeType parameter.
 
@@ -63,7 +64,7 @@
     - `Add-AzServiceFabricManagedNodeTypeVMExtension`
     - `Add-AzServiceFabricManagedNodeTypeVMSecret`
     - `Remove-AzServiceFabricManagedNodeTypeVMExtension`
-    - `Restart-AzServiceFabricManagedNodeTyp`
+    - `Restart-AzServiceFabricManagedNodeType`
 * Upgraded Service Fabric SDK to version 1.2.0 which uses service fabric resource provider api-version 2020-03-01 for the current model and 2020-01-01-preview for managed clusters.
 
 ## Version 2.1.0


src/ServiceFabric/ServiceFabric/Commands/UpdateAzServiceFabricVmImage.cs
```diff
@@ -0,0 +1,74 @@
...
+        [Parameter(Mandatory = true, Position = 1, ValueFromPipelineByPropertyName = true,
+                   HelpMessage = "Specify the name of the cluster")]
+        [ValidateNotNullOrEmpty()]
+        [Alias("ClusterName")]
+        public override string Name { get; set; }
+
+        /// <summary>
+        /// VmImage sets the Service Fabric runtime package delivery hint for the cluster, which results in dropping the appropriately-mapped package for the VM image OS.
+        /// Accepted values: Windows, Linux, Ubuntu, Ubuntu18_04
+        /// Note: Linux & Ubuntu both map to Ubuntu 16.04
```
a-santamaria, line 41: why have both Linux and ubuntu if they are both 16.04?

src/ServiceFabric/ServiceFabric/Commands/UpdateAzServiceFabricNodeType.cs
```diff
@@ -0,0 +1,101 @@
...
+                }
+                else
+                {
+                    WriteVerbose(string.Format(ServiceFabricProperties.Resources.NodeTypeUpdateIsNoOp, this.NodeType));
+                    WriteObject(new PSCluster(cluster), true);
+                }
+            }
+        }
+
+        private bool PatchRequired(NodeTypeDescription existingNodeType)
```
a-santamaria, line 96: nit: suggest to rename to IsPatchRequired

src/ServiceFabric/ServiceFabric/Commands/UpdateAzServiceFabricNodeType.cs
```diff
@@ -0,0 +1,101 @@
...
+            if (ShouldProcess(target: this.NodeType, action: string.Format("Update nodetype {0}", this.NodeType)))
+            {
+                if (cluster.NodeTypes == null)
+                {
+                    throw new PSInvalidOperationException(ServiceFabricProperties.Resources.NodeTypesNotDefinedInCluster);
+                }
+
+                if (PatchRequired(existingNodeType))
+                {
+                    if (this.IsPrimaryNodeType.HasValue && this.IsPrimaryNodeType.Value != existingNodeType.IsPrimary)
```
a-santamaria, line 74: do you need to check this again?

src/ServiceFabric/ServiceFabric/Commands/UpdateAzServiceFabricNodeType.cs
```diff
@@ -0,0 +1,101 @@
...
+            }
+
+            if (ShouldProcess(target: this.NodeType, action: string.Format("Update nodetype {0}", this.NodeType)))
+            {
+                if (cluster.NodeTypes == null)
+                {
+                    throw new PSInvalidOperationException(ServiceFabricProperties.Resources.NodeTypesNotDefinedInCluster);
+                }
+
+                if (PatchRequired(existingNodeType))
```
a-santamaria, line 72: you might want to check that at least one primary node type is left if changing to false

src/ServiceFabric/ServiceFabric/Commands/UpdateAzServiceFabricVmImage.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// </summary>
+        [Parameter(Mandatory = true, ValueFromPipeline = true,
+                   HelpMessage = "VM Image hint for determining SF runtime package delivery: i.e. .cab/.deb")]
+        [ValidateNotNullOrEmpty()]
+        public VmImageKind VmImage { get; set; }
+
+        public override void ExecuteCmdlet()
+        {
+            var cluster = GetCurrentCluster();
+            var oldVmImage = GetVmImage(cluster);
```
a-santamaria, line 51: We shouldn't allow going from windows to Linux and vise versa right? should we add the validation here?

src/ServiceFabric/ServiceFabric/Commands/UpdateAzServiceFabricVmImage.cs
maburlik: Good question. This is how it is in SFRP, and some existing ARM templates may have either in the ARM template, but both are accepted.

src/ServiceFabric/ServiceFabric/Commands/UpdateAzServiceFabricNodeType.cs
maburlik: Good point. I was thinking of this like a staged operation - to check if any change was made at all, then apply changes per parameter (as update node type cmdlet is extended). This could be done another way.

src/ServiceFabric/ServiceFabric/Commands/UpdateAzServiceFabricNodeType.cs
maburlik: SFRP would reject this logic right? This is obviously something that should be validated out, but is business logic supposed to be handled in this step? I would say yes if modifying node type involved updating multiple ARM resources i.e. VMSS, SF, Networking components since Az PS is the actor that modifies multiple components so would assume managing staggered requests via validation & state. In this case the singular request is sent to SFRP.

src/ServiceFabric/ServiceFabric/Commands/UpdateAzServiceFabricVmImage.cs
maburlik: If there is investment in hybrid clusters this setting would expectedly get deprecated, or another "Hybrid" or similar setting would be introduced to allow such combinations. In theory a Linux to Windows switch would be fine if the running apps continue to work.

As long as extension does 'PUT node' with OS details and uses the bootstrap API for retrieving cluster package, this setting will only really used to deliver packages for upgrades, so this would be the earnest indicator that the user intends to switch the primary OS of the cluster to Windows and future incoming packages for upgrades will be based on the cab.

The scenario could be as such:
- User has an app that runs dotnet core which is assumed to be able to activate on Windows & Linux machines without repackaging.
- vmImage is updated from Ubuntu18_04 to Windows so upgrades will not work until all cluster nodes are on a homogeneous OS
- A new node type is created with a WindowsServer OS and the app is deployed to the nodes as they come up
- The old node type is decommissioned

wyunchi-ms:
`TestAddAzureRmServiceFabricNodeType` fails, please rerecord it.

maburlik:
> `TestAddAzureRmServiceFabricNodeType` fails, please rerecord it.

Working on this.

wyunchi-ms (approved):

PR #15285, opened 2021-06-16
Print default log of auto gen module to verbose stream

dolauli:
Part of fix to https://github.com/Azure/autorest.powershell/issues/689.

### DIFF
--- a/src/Accounts/Accounts/ChangeLog.md
+++ b/src/Accounts/Accounts/ChangeLog.md
@@ -20,6 +20,7 @@
 
 ## Upcoming Release
 * Printed auto generated modules' default logs to verbose stream
+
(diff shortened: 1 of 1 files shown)


src/Accounts/Accounts/ChangeLog.md
```diff
@@ -19,7 +19,7 @@
 -->
 
 ## Upcoming Release
-
+* Printed auto generated modules' default logs to verbose stream
```
BethanyZhou, line 22:
```suggestion
* Printed auto generated modules' default logs to verbose stream

```

BethanyZhou (changes requested):

BethanyZhou (approved):

PR #15333, opened 2021-06-22
Adding new parameters for CustomIpPrefix

gitlwh:
## Checklist

        - {Please put the link here}

### DIFF
--- a/src/Network/Network/BYOIP/CustomIpPrefix/UpdateAzureCustomIpPrefixCommand.cs
+++ b/src/Network/Network/BYOIP/CustomIpPrefix/UpdateAzureCustomIpPrefixCommand.cs
@@ -77,7 +77,7 @@ namespace Microsoft.Azure.Commands.Network
         public SwitchParameter Deprovision { get; set; }
 
         [Parameter(
-            Mandatory = true,
+            Mandatory = false,
             ValueFromPipelineByPropertyName = true,
             HelpMessage = "The customIpPrefix CIDR.")]
         [ValidateNotNullOrEmpty]


src/Network/Network/BYOIP/CustomIpPrefix/UpdateAzureCustomIpPrefixCommand.cs
```diff
@@ -67,7 +67,13 @@ public class UpdateAzureCustomIpPrefixCommand : CustomIpPrefixBaseCmdlet
         public SwitchParameter Commission { get; set; }
 
         [Parameter(Mandatory = false, HelpMessage = "Run cmdlet in the background")]
-        public SwitchParameter Decomission { get; set; }
+        public SwitchParameter Decommission { get; set; }
```
BethanyZhou, line 70: Hey, please add Decomission as the alias of Decommission to avoid breaking change.

src/Network/Network/BYOIP/CustomIpPrefix/UpdateAzureCustomIpPrefixCommand.cs
```diff
@@ -67,7 +67,13 @@ public class UpdateAzureCustomIpPrefixCommand : CustomIpPrefixBaseCmdlet
...
+
+        [Parameter(Mandatory = false, HelpMessage = "Run cmdlet in the background")]
+        public SwitchParameter Provision { get; set; }
```
BethanyZhou, line 73: - I recommend using bool as the type of Provision so that we needn't two parameters for one function.
- Please update the help message.

BethanyZhou (changes requested):
Please add change log for your change.

BethanyZhou:
Hi @gitlwh , How about Provision and Deprovision?

gitlwh (commented):
> Hi @gitlwh , How about Provision and Deprovision?

What do you mean by that? I think they should have same behavior as commission and decommission. Only one of them should exist.

BethanyZhou:
> Only one of the four should exist.

How to check this logic? It's in backend side?

>Type would change behavior of our command because we need to use like
'''-Provision $true''' right?

Yes, I prefer this behavior because we can avoid customer using Provision and Deprovision at same time

BethanyZhou:
> And if we change the behavior of command it would be a break change right?

Thank you for you quick response. Yes, It would be a breaking change. That's why I only discuss the design of Provision with you.

src/Network/Network/ChangeLog.md
```diff
@@ -48,6 +48,9 @@
     - `Update-AzVirtualHub`
 * Updated cmdlets to expose two read-only properties of client certificate.
     - `Get-AzApplicationGatewayTrustedClientCertificate`
+* Updated cmdlets to add properties for new BYOIP features.
+    - `New-AzCustomIpPrefix`
```
BethanyZhou, line 52: Please sync your code with latest network-march branch so that your change log msg can include in upcoming release section.
image

BethanyZhou (changes requested):

gitlwh:
> > And if we change the behavior of command it would be a break change right?
> 
> * Thank you for you quick response. Yes, It would be a breaking change if we change the behavior of Commission. Since Commission had been release, I want to discuss about the design of Provision with you.
> * I understand that you want to keep the behavior of Provision same with Commission. Your design looks acceptable but I'm wondering how to design is more friendly to customer. Will you feel better if you are customer and can set Provision $true, $null and $false by one parameter?

No, The original way is more simple for cx to use and we don't have a plan to change this bahavior.

gitlwh:
Added one more parameter for update operation CIDR

src/Network/Network/BYOIP/CustomIpPrefix/UpdateAzureCustomIpPrefixCommand.cs
```diff
@@ -63,11 +63,25 @@ public class UpdateAzureCustomIpPrefixCommand : CustomIpPrefixBaseCmdlet
...
+        public SwitchParameter Decommission { get; set; }
+
+        [Parameter(Mandatory = false, HelpMessage = "start provisioning process.")]
+        public SwitchParameter Provision { get; set; }
+
+        [Parameter(Mandatory = false, HelpMessage = "start deprovisioning process.")]
+        public SwitchParameter Deprovision { get; set; }
+
+        [Parameter(
+            Mandatory = true,
```
BethanyZhou, line 80: New required parameter will cause breaking change, which is not allowed in regular release, please wait for our major release

gitlwh: this is a mistake. just changed to false.

BethanyZhou (approved):

PR #15511, opened 2021-07-21
update Az.Compute to use latest Compute SDK

grizzlytheodore:
…ges from 'SecurityType" to "string"  end user experience stays the same.

## Checklist

        - {Please put the link here}


grizzlytheodore:
SecurityType parameter type changed from "SecurityType" to "string". 
because this enum property in Swagger went from modelAsString: false to true. 
shown here https://github.com/Azure/azure-rest-api-specs/commit/9c4f909644f4b0194bd17d02b87c3fd2cd6fb47d

but the end user experience should stay the same.

tools/StaticAnalysis/Exceptions/Az.Compute/BreakingChangeIssues.csv
```diff
@@ -0,0 +1,6 @@
...
+Az.Compute,Microsoft.Azure.Commands.Compute.SetAzureVMSecurityProfile,Set-AzVMSecurityProfile,0,2020,The cmdlet 'Set-AzVMSecurityProfile' no longer supports the type 'Microsoft.Azure.Management.Compute.Models.SecurityTypes' for parameter 'SecurityType'.,Change the type for parameter 'SecurityType' back to 'Microsoft.Azure.Management.Compute.Models.SecurityTypes'.
+Az.Compute,Microsoft.Azure.Commands.Compute.Automation.SetAzureRmVmssSecurityProfileCommand,Set-AzVmssSecurityProfile,0,2020,The cmdlet 'Set-AzVmssSecurityProfile' no longer supports the type 'Microsoft.Azure.Management.Compute.Models.SecurityTypes' for parameter 'SecurityType'.,Change the type for parameter 'SecurityType' back to 'Microsoft.Azure.Management.Compute.Models.SecurityTypes'.
```
isra-fel, line 6: Changing enum to string should not be a breaking change, but removal of properties may cause user's script to break, so they should not be suppressed.

isra-fel (changes requested):
`BreakingChangeIssues.csv` is in wrong format -- each cell should be double-quoted -- make sure you copied them from the csv files in the artifacts of a failed pipeline, not from the logs.

tools/StaticAnalysis/Exceptions/Az.Compute/BreakingChangeIssues.csv
grizzlytheodore: the first two ones are not exactly removed, but 'PublicIpAddressesSkuName' and 'PublicIPAddressSkuTier' are changed to 'name' and 'tier'.    But there are no cmdlets right now that use those models.

grizzlytheodore:
please review again !

tools/StaticAnalysis/Exceptions/Az.Compute/BreakingChangeIssues.csv
isra-fel: Okay in that case they don't count as breaking change

isra-fel (approved):

PR #15605, opened 2021-08-04
add params to New-Azvm 

grizzlytheodore:
adding '-SshKeyName' and 'GenerateSshKey' parameters to New-AzVm
and fixing small bug in New-AzSshKey

## Checklist

        - {Please put the link here}

### DIFF
--- a/src/Compute/Compute/ChangeLog.md
+++ b/src/Compute/Compute/ChangeLog.md
@@ -20,6 +20,7 @@
 
 -->
 ## Upcoming Release
+* Added new parameters `-SshKeyName` and `-GenerateSshKey` to `New-AzVM` to create a VM with SSH Key
 
 ## Version 4.16.0
 * Fixed the warning in `New-AzVM` cmdlet stating the sku of the VM is being defaulted even if a sku size is provided by the user. Now it only occurs when the user does not provide a sku size.
@@ -39,7 +40,6 @@
     - `New-AzVmssConfig`
     - `Update-AzVm`
     - `Update-AzVmss`
-* Added new parameters `-SshKeyName` and `-GenerateSshKey` to `New-AzVM` to create a VM with SSH Key
 
 ## Version 4.15.0
 * Added optional parameter `-OrchestrationMode` to `New-AzVmss` and `New-AzVmssConfig`


src/Compute/Compute/VirtualMachine/Operation/NewAzureVMCommand.cs
```diff
@@ -966,5 +1068,106 @@ private static string GetStorageAccountNameFromUriString(string uriStr)
...
+            {
+                throw new Exception("Please provide parameter '-SshKeyName' to be used with '-GenerateSshKey'");
+            }
+
+
+            // check if -SshKeyName exists in azure
+            Microsoft.Azure.Commands.Compute.Automation.NewAzureSshKey sshKeyClass = new Microsoft.Azure.Commands.Compute.Automation.NewAzureSshKey();
+            SshPublicKeyResource SshPublicKey;
+            string publicKey = "value";
+            try
```
VeryEarly, line 1085: logic in line 1085 to line 1139 should be: if(this.GenerateSshKey.IsPresent){ trygetkey() ? exception : generatenewkey } else { trygetkey() ? continue : exception }

VeryEarly (changes requested):

src/Compute/Compute/VirtualMachine/Operation/NewAzureVMCommand.cs
```diff
@@ -667,11 +735,44 @@ public void DefaultExecuteCmdlet()
                         }
                     }
 
-                    var result = this.VirtualMachineClient.CreateOrUpdateWithHttpMessagesAsync(
+                    Rest.Azure.AzureOperationResponse<VirtualMachine> result;
+
+                    if (this.IsParameterBound(c => c.SshKeyName) || this.GenerateSshKey.IsPresent)
```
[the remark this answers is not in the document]
grizzlytheodore: i still need this if statement to create/retrieve ssh key if the parameters are given

src/Compute/Compute/Usage/NewAzureSshKey.cs
```diff
@@ -42,28 +42,29 @@ public override void ExecuteCmdlet()
             base.ExecuteCmdlet();
             ExecuteClientAction(() =>
             {
-            string resourceGroupName = this.ResourceGroupName;
```
grizzlytheodore, line 45: indentations are fixed in this file

grizzlytheodore:
@VeryEarly thank you for all your comments, I really appreciate it looking at it thoroughly. 
I've addressed most of your comments except for the ones advising new parameter sets. 

So you want me to add two new parameter sets branch off of DefaultParameterSet and SimpleParameter set that has '-Linux', '-SshKeyName' and '-GenerateSshKey'? 

Could you elaborate on why that would be better? 

I am hesitant because wouldn't that add more room for error like: '-Linux' is given but '-Image' or OSRprofile is not Linux.
Also, if we were to add two new parameter sets for this occasion to such commonly used cmdlet with so many parameters some day we will end up with numerous parameter sets for this cmdlets. Should I consult a PM?

VeryEarly:
> @VeryEarly thank you for all your comments, I really appreciate it looking at it thoroughly.
> I've addressed most of your comments except for the ones advising new parameter sets.
> 
> So you want me to add two new parameter sets branch off of DefaultParameterSet and SimpleParameter set that has '-Linux', '-SshKeyName' and '-GenerateSshKey'?
> 
> Could you elaborate on why that would be better?
> 
> I am hesitant because
> 
> 1. Current way of checking the OS is sufficient and works fine.
> 2. wouldn't that add more room for error like: '-Linux' is given but '-Image' or OSRprofile is not Linux.
> 3. we were to add two new parameter sets for this occasion to such a commonly used cmdlet with so many parameters, someday we will end up with countless parameter sets for this cmdlet.
> 
> Should I consult a PM?

please correct me if I'm wrong, the exception thrown at line 452 (NewAzureVMCommand.cs), won't be executed until line 625 (NewAzureVMCommand.cs) was reached:
```
result = await client.RunAsync(client.SubscriptionId, parameters, asyncCmdlet);
```
At this point, some other resources should be already created (storage/network? need your confirmation).

This is too heavy for a parameter level behavior, that is to say, this exception can be discovered at the very beginning when cmdlet was invoked (using parameter set) while not to call any actual business logic.

Please let me know if you have any concern.

grizzlytheodore:
> > @VeryEarly thank you for all your comments, I really appreciate it looking at it thoroughly.
> > I've addressed most of your comments except for the ones advising new parameter sets.
> > So you want me to add two new parameter sets branch off of DefaultParameterSet and SimpleParameter set that has '-Linux', '-SshKeyName' and '-GenerateSshKey'?
> > Could you elaborate on why that would be better?
> > I am hesitant because
> > 
> > 1. Current way of checking the OS is sufficient and works fine.
> > 2. wouldn't that add more room for error like: '-Linux' is given but '-Image' or OSRprofile is not Linux.
> > 3. we were to add two new parameter sets for this occasion to such a commonly used cmdlet with so many parameters, someday we will end up with countless parameter sets for this cmdlet.
> > 
> > Should I consult a PM?
> 
> please correct me if I'm wrong, the exception thrown at line 452 (NewAzureVMCommand.cs), won't be executed until line 625 (NewAzureVMCommand.cs) was reached:
> 
> ```
> result = await client.RunAsync(client.SubscriptionId, parameters, asyncCmdlet);
> ```

this part is right 

> At this point, some other resources should be already created (storage/network? need your confirmation).

but other resources are not created. the PowerShell sends one payload, and storage, network, and VM are all created together in the backend.  Many of the storage and network properties will be defaulted to a value because the simple parameter set provides very little bit of that. 

> This is too heavy for a parameter level behavior, that is to say, this exception can be discovered at the very beginning when cmdlet was invoked (using parameter set) while not to call any actual business logic.
> 
> Please let me know if you have any concern.

VeryEarly:
> > > @VeryEarly thank you for all your comments, I really appreciate it looking at it thoroughly.
> > > I've addressed most of your comments except for the ones advising new parameter sets.
> > > So you want me to add two new parameter sets branch off of DefaultParameterSet and SimpleParameter set that has '-Linux', '-SshKeyName' and '-GenerateSshKey'?
> > > Could you elaborate on why that would be better?
> > > I am hesitant because
> > > 
> > > 1. Current way of checking the OS is sufficient and works fine.
> > > 2. wouldn't that add more room for error like: '-Linux' is given but '-Image' or OSRprofile is not Linux.
> > > 3. we were to add two new parameter sets for this occasion to such a commonly used cmdlet with so many parameters, someday we will end up with countless parameter sets for this cmdlet.
> > > 
> > > Should I consult a PM?
> > 
> > 
> > please correct me if I'm wrong, the exception thrown at line 452 (NewAzureVMCommand.cs), won't be executed until line 625 (NewAzureVMCommand.cs) was reached:
> > ```
> > result = await client.RunAsync(client.SubscriptionId, parameters, asyncCmdlet);
> > ```
> 
> this part is right
> 
> > At this point, some other resources should be already created (storage/network? need your confirmation).
> 
> but here no. Other resources are not created. the PowerShell sends one payload to Azure on this line 625, and in the backend, the storage resources, network resources, and VM are all created together. Many of the storage and network properties will default to a value in the backend because the simple parameter set provides very little bit of that.
> 
> > This is too heavy for a parameter level behavior, that is to say, this exception can be discovered at the very beginning when cmdlet was invoked (using parameter set) while not to call any actual business logic.
> > Please let me know if you have any concern.

Make sense. 

I suggest to

- Based on your implementation I'll say when user want to use ssh key, the parameter "sshkeyname" is required, so for if statement you are using like ```!String.IsNullOrEmpty(_cmdlet.SshKeyName) || _cmdlet.GenerateSshKey```, should all be changed to ```!String.IsNullOrEmpty(_cmdlet.SshKeyName)```
- create a method ```privatevoid Validate()``` for class Parameters to handle all these logics (check whether sshkeyname/generated are provided, if linux, and other new parameters in the future)
- put this method in the constructor of Parameters, remove all if blocks that validate parameters
- create another method called cleanup or whatever, to handle all the remove generated ssh key logic (and other clean up jobs in the future), invoke this method in catch exceptions whenever exceptions happened (in your case, if cloudexception, remove generated sshkey)
- make "addSshPublicKey" and "SshKeyForLinux" method of class Parameters

The reason I suggested to do so is because these business logics are not supposed to be blended with the concrete strategy, please let me know if you have any concerns.

grizzlytheodore:
> > > > @VeryEarly thank you for all your comments, I really appreciate it looking at it thoroughly.
> > > > I've addressed most of your comments except for the ones advising new parameter sets.
> > > > So you want me to add two new parameter sets branch off of DefaultParameterSet and SimpleParameter set that has '-Linux', '-SshKeyName' and '-GenerateSshKey'?
> > > > Could you elaborate on why that would be better?
> > > > I am hesitant because
> > > > 
> > > > 1. Current way of checking the OS is sufficient and works fine.
> > > > 2. wouldn't that add more room for error like: '-Linux' is given but '-Image' or OSRprofile is not Linux.
> > > > 3. we were to add two new parameter sets for this occasion to such a commonly used cmdlet with so many parameters, someday we will end up with countless parameter sets for this cmdlet.
> > > > 
> > > > Should I consult a PM?
> > > 
> > > 
> > > please correct me if I'm wrong, the exception thrown at line 452 (NewAzureVMCommand.cs), won't be executed until line 625 (NewAzureVMCommand.cs) was reached:
> > > ```
> > > result = await client.RunAsync(client.SubscriptionId, parameters, asyncCmdlet);
> > > ```
> > 
> > 
> > this part is right
> > > At this point, some other resources should be already created (storage/network? need your confirmation).
> > 
> > 
> > but here no. Other resources are not created. the PowerShell sends one payload to Azure on this line 625, and in the backend, the storage resources, network resources, and VM are all created together. Many of the storage and network properties will default to a value in the backend because the simple parameter set provides very little bit of that.
> > > This is too heavy for a parameter level behavior, that is to say, this exception can be discovered at the very beginning when cmdlet was invoked (using parameter set) while not to call any actual business logic.
> > > Please let me know if you have any concern.
> 
> Make sense.
> 
> I suggest to
> 
> * Based on your implementation I'll say when user want to use ssh key, the parameter "sshkeyname" is required, so for if statement you are using like `!String.IsNullOrEmpty(_cmdlet.SshKeyName) || _cmdlet.GenerateSshKey`, should all be changed to `!String.IsNullOrEmpty(_cmdlet.SshKeyName)`
> * create a method `privatevoid Validate()` for class Parameters to handle all these logics (check whether sshkeyname/generated are provided, if linux, and other new parameters in the future)
> * put this method in the constructor of Parameters, remove all if blocks that validate parameters
> * create another method called cleanup or whatever, to handle all the remove generated ssh key logic (and other clean up jobs in the future), invoke this method in catch exceptions whenever exceptions happened (in your case, if cloudexception, remove generated sshkey)
> * make "addSshPublicKey" and "SshKeyForLinux" method of class Parameters
> 
> The reason I suggested to do so is because these business logics are not supposed to be blended with the concrete strategy, please let me know if you have any concerns.


@VeryEarly 
Thank you for all the advice. I see how this will be a cleaner design. 

I do have a concern about the last bullet point. The default parameter set does not use the Parameters class, and I still need to use "addSshPublicKey" and "SshKeyForLinux" for it. Shown https://github.com/Azure/azure-powershell/blob/5a488b6e8b08c787eaf2c1361bbd55f3afc13bb4/src/Compute/Compute/VirtualMachine/Operation/NewAzureVMCommand.cs#L739   any other suggestion on how I could structure it better?

grizzlytheodore:
@VeryEarly  I've implemented all your suggestions except for the last bullet point. I've made validate() a method of the cmdlet class.

src/Compute/Compute/VirtualMachine/Operation/NewAzureVMCommand.cs
```diff
@@ -966,5 +1036,128 @@ private static string GetStorageAccountNameFromUriString(string uriStr)
...
+        }
+
+        private void validate()
+        {
+            if (this.GenerateSshKey.IsPresent && !this.IsParameterBound(c => c.SshKeyName))
+            {
+                throw new Exception("Please provide parameter '-SshKeyName' to be used with '-GenerateSshKey'");
+            }
+
+            if (this.ParameterSetName == "DefaultParameterSet" && !IsLinuxOs())
```
VeryEarly, line 1146: is this condition correct?

VeryEarly:
> @VeryEarly I've implemented all your suggestions except for the last bullet point. I've made validate() a method of the cmdlet class.

I see, it might need more refactor to this cmdlet then, I'm good with current implementation, thanks for your quick response.
Please check the new comments, it caused test case failures.

src/Compute/Compute/VirtualMachine/Operation/NewAzureVMCommand.cs
grizzlytheodore: it was faulty. fixed it. thank you for pointing that out

grizzlytheodore:
@veryearly pls review again!

VeryEarly (approved):

VeryEarly (approved):

grizzlytheodore:
@VeryEarly  pls merge

PR #15896, opened 2021-09-17
[Logz] Release version 2020-10-01-preview

LucasYao93:
## Description

+ Customization cmdlet
+ Add test cases
+ Add examples


LucasYao93:
Don’t merge it now because we are waiting service team’s response

src/Logz/custom/Get-AzLogzMonitoredResource.ps1
```diff
@@ -0,0 +1,130 @@
...
+.Description
+List the resources currently being monitored by the Logz monitor resource.
+.Example
+PS C:\> {{ Add code here }}
+
+{{ Add output here }}
+.Example
+PS C:\> {{ Add code here }}
+
+{{ Add output here }}
```
BethanyZhou, line 28: Example section should be removed here.

PR #16449, opened 2021-11-19
Remove secrets in PSAzurerRmAccount from display

msJinLei:
The PR is to fix https://github.com/Azure/azure-powershell/issues/15427

Service principal secret and service principal certificate are required for service principal user to access Azure resources.
Previously PowerShell can access the plain text values of service principal secret and service principal certificate password using 
```
- (Get-AzContext).Accounts.ExtendedProperties.ServicePrincipalSecret
- (Get-AzContext).Accounts.ExtendedProperties.CertificatePassword
```

To improve the security, the PR removes the these 2 attributes from `(Get-AzContext).Accounts`, whose type is `PSAzurerRmAccount`.

After the change, the context object retrieved by Get-AzContext no longer has ServicePrincipalSecret and CertificatePassword. The cmdlet Set-AzContext, Rename-AzContext accept context object as the input, which brings the risk that the context without  ServicePrincipalSecret and CertificatePassword will be added into context list and even be set as current context and then break the current environment.

To solve the issue, the PR also changes the logic when new context object is added to the context list. Firstly find whether there is an existing object with the same account id, tenant and subscription. If found, merge the attributes in the existing one but absent in the new one to the new one and then add the new one to the context list.

## Checklist

        - {Please put the link here}

### DIFF
--- a/src/Accounts/Accounts/ChangeLog.md
+++ b/src/Accounts/Accounts/ChangeLog.md
@@ -19,7 +19,7 @@
 -->
 
 ## Upcoming Release
-* Removed `ServicePrincipalSecret` and `CertificatePassword` in Account of Context from display [#15427]
+* Removed `ServicePrincipalSecret` and `CertificatePassword` in `PSAzureRmAccount` [#15427]
 * Added optional parameter `MicrosoftGraphAccessToken` to `Connect-AzAccount`
 * Added optional parameters `MicrosoftGraphEndpointResourceId`, `MicrosoftGraphUrl` to `Add-AzEnvironment` and `Set-AzEnvironment`
 * Added `-AccountId` property to `UserWithSubscriptionId` parameter set of `Connect-AzAccount` which allows a user name to be pre-selected for interactive logins
--- a/src/Accounts/Authentication.ResourceManager/AzureRmProfile.cs
+++ b/src/Accounts/Authentication.ResourceManager/AzureRmProfile.cs
@@ -502,7 +502,7 @@ namespace Microsoft.Azure.Commands.Common.Authentication.Models
             {
                 if (TryFindContext(context, out string oldName))
                 {
-                    var oldContext = Contexts[oldName];
+                    var oldContext = Contexts[oldName].DeepCopy();
                     oldContext.Update(context);
                     context = oldContext;
                 }


src/Accounts/Accounts/ChangeLog.md
```diff
@@ -19,6 +19,7 @@
 -->
 
 ## Upcoming Release
+* Removed `ServicePrincipalSecret` and `CertificatePassword` in Account of Context from display [#15427]
```
dingmeng-xue, line 22: from display or we have removed them from Azure Context psobject? I remember it is latter.

src/Accounts/Authentication.ResourceManager/AzureRmProfile.cs
```diff
@@ -488,11 +488,25 @@ public bool TryRenameContext(string sourceName, string targetName)
...
         public bool TrySetContext(string name, IAzureContext context)
         {
             bool result = false;
             if (Contexts != null)
             {
+                string oldName = null;
+                if (TryFindContext(context, out oldName))
+                {
+                    var oldContext = Contexts[oldName];
+                    oldContext.Update(context);
```
dingmeng-xue, line 507: `context` is from user's import. Is it possible this line will change `oldContext` in `contexts`?

src/Accounts/Authentication.ResourceManager/AzureRmProfile.cs
```diff
@@ -488,11 +488,25 @@ public bool TryRenameContext(string sourceName, string targetName)
...
+        /// </summary>
+        /// <param name="name">The specified new name of the context.</param>
+        /// <param name="context">The new context to set as default.</param>
         public bool TrySetContext(string name, IAzureContext context)
         {
             bool result = false;
             if (Contexts != null)
             {
+                string oldName = null;
+                if (TryFindContext(context, out oldName))
```
dingmeng-xue, line 504:
```suggestion
                if (TryFindContext(context, out string oldName))
```

src/Accounts/Authentication.ResourceManager/AzureRmProfile.cs
msJinLei: Should use deepCopy. Fixed

src/Accounts/Accounts/ChangeLog.md
msJinLei: > 

There are 2 classes that relate to Account of Context in Az.Accounts, AzureAccount is for internal use and PSAzureRmAccount is for powershell operation. The PR just removes `ServicePrincipalSecret` and `CertificatePassword` from PSAzureRmAccount.

dingmeng-xue (approved):

PR #16503, opened 2021-11-26
[ConnectedKubernetes] move connected kubernetes to main

lijinpei2008:
Move connected kubernetes to main.

Create new cmdlet: Get-AzConnectedKubernetesUserCredential.

And update the New-AzConnectedKubernetes to support helm from version 3.7+

### DIFF
--- a/src/ConnectedKubernetes/ChangeLog.md
+++ b/src/ConnectedKubernetes/ChangeLog.md
@@ -18,8 +18,6 @@
         - Additional information about change #1
 -->
 ## Upcoming Release
-
-## Version 0.3.0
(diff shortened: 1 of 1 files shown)


src/ConnectedKubernetes/ChangeLog.md
```diff
@@ -19,6 +19,11 @@
 -->
 ## Upcoming Release
 
+## Version 0.3.0
```
isra-fel, line 22: This line should be removed because versions will not be updated until we begin the release process.

isra-fel (approved):
LGTM

PR #16687, opened 2021-12-22
Fix update SignalR bugs and add tests

Y-Sindo:


src/SignalR/SignalR/Az.SignalR.psd1
```diff
@@ -12,7 +12,7 @@
 # RootModule = ''
 
 # Version number of this module.
-ModuleVersion = '1.4.0'
```
msJinLei, line 15: Please revert the change of the line. We will update the version using automatic tools

msJinLei (changes requested):

src/SignalR/SignalR/Az.SignalR.psd1
```diff
@@ -12,7 +12,7 @@
...
+ModuleVersion = '1.4.1'
```
Y-Sindo, line 15:
```suggestion
ModuleVersion = '1.4.0'
```

msJinLei (approved):

PR #17477, opened 2022-03-15
Update Az.Accounts for telemetry

dolauli:
Major changes are as below.

1. Use correlationId as the telemetryId
2. Add two delegates GetTelemetryId and Telemetry to serve auto gen cmdlets for telemetry
3. Use MetricHelper.SessionId as the telemetry SessionId


src/Accounts/Accounts/CommonModule/AzModule.cs
```diff
@@ -133,19 +134,40 @@ public async Task EventListener(string id, CancellationToken cancellationToken,
                     await OnProcessRecordAsyncEnd(id, cancellationToken, getEventData, signal, processRecordId);
```
dingmeng-xue, line 134: Shall `processRecordId` be changed?

dolauli: No need to change since in function OnProcessRecordAsyncEnd, the processRecordId is actually a redundant parameter without being used.

dingmeng-xue (approved):

PR #17799, opened 2022-04-14
Enabled public network access setting

BethanyZhou:
## Description
- Added a new property named PublicNetworkAccess for managed HSM and key vault.
- supported setting PublicNetworkAccess at the time of creating or updating
- added test cases for PublicNetworkAccess 

## Checklist

        - {Please put the link here}

### DIFF
--- a/src/KeyVault/KeyVault/Commands/NewAzureKeyVault.cs
+++ b/src/KeyVault/KeyVault/Commands/NewAzureKeyVault.cs
@@ -98,7 +98,7 @@ namespace Microsoft.Azure.Commands.KeyVault
         public int SoftDeleteRetentionInDays { get; set; }
 
         [Parameter(Mandatory = false,
-            HelpMessage = "Specifies whether the vault will accept traffic from public internet. If set to 'disabled' all traffic except private endpoint traffic and that that originates from trusted services will be blocked. This will override the set firewall rules, meaning that even if the firewall rules are present we will not honor the rules. By default, we will enable public network access.")]
+            HelpMessage = "Specifies whether the vault will accept traffic from public internet. If set to 'disabled' all traffic except private endpoint traffic and that originates from trusted services will be blocked. This will override the set firewall rules, meaning that even if the firewall rules are present we will not honor the rules. By default, we will enable public network access.")]
         [PSArgumentCompleter("Enabled", "Disabled")]
         public string PublicNetworkAccess { get; set; }
 
--- a/src/KeyVault/KeyVault/Commands/UpdateAzureKeyVault.cs
+++ b/src/KeyVault/KeyVault/Commands/UpdateAzureKeyVault.cs
@@ -55,7 +55,7 @@ namespace Microsoft.Azure.Commands.KeyVault
         public bool? EnableRbacAuthorization { get; set; }
 
         [Parameter(Mandatory = false,
-            HelpMessage = "Specifies whether the vault will accept traffic from public internet. If set to 'disabled' all traffic except private endpoint traffic and that that originates from trusted services will be blocked. This will override the set firewall rules, meaning that even if the firewall rules are present we will not honor the rules.")]
+            HelpMessage = "Specifies whether the vault will accept traffic from public internet. If set to 'disabled' all traffic except private endpoint traffic and that originates from trusted services will be blocked. This will override the set firewall rules, meaning that even if the firewall rules are present we will not honor the rules.")]
         [PSArgumentCompleter("Enabled", "Disabled")]
         public string PublicNetworkAccess { get; set; }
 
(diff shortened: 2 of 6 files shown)


src/KeyVault/KeyVault/Commands/NewAzureKeyVault.cs
```diff
@@ -97,6 +97,11 @@ public class NewAzureKeyVault : KeyVaultManagementCmdletBase
         [ValidateNotNullOrEmpty]
         public int SoftDeleteRetentionInDays { get; set; }
 
+        [Parameter(Mandatory = false,
+            HelpMessage = "Specifies whether the vault will accept traffic from public internet. If set to 'disabled' all traffic except private endpoint traffic and that that originates from trusted services will be blocked. This will override the set firewall rules, meaning that even if the firewall rules are present we will not honor the rules. By default, we will enable public network access.")]
```
isra-fel, line 101:
```suggestion
            HelpMessage = "Specifies whether the vault will accept traffic from public internet. If set to 'disabled' all traffic except private endpoint traffic and that originates from trusted services will be blocked. This will override the set firewall rules, meaning that even if the firewall rules are present we will not honor the rules. By default, we will enable public network access.")]
```

src/KeyVault/KeyVault/Commands/NewAzureKeyVault.cs
```diff
@@ -97,6 +97,11 @@ public class NewAzureKeyVault : KeyVaultManagementCmdletBase
...
+        [PSArgumentCompleter("Enabled", "Disabled")]
```
isra-fel, line 102: Should this be a ValidateSet instead of ArgumentCompleter? Also, would it be better to use just boolean or switch parameter?

src/KeyVault/KeyVault/Commands/UpdateAzureKeyVault.cs
```diff
@@ -54,6 +54,12 @@ public class UpdateTopLevelResourceCommand : KeyVaultManagementCmdletBase
         [Parameter(Mandatory = false, HelpMessage = "Enable or disable this key vault to authorize data actions by Role Based Access Control (RBAC).")]
         public bool? EnableRbacAuthorization { get; set; }
 
+        [Parameter(Mandatory = false,
+            HelpMessage = "Specifies whether the vault will accept traffic from public internet. If set to 'disabled' all traffic except private endpoint traffic and that that originates from trusted services will be blocked. This will override the set firewall rules, meaning that even if the firewall rules are present we will not honor the rules.")]
```
isra-fel, line 58:
```suggestion
            HelpMessage = "Specifies whether the vault will accept traffic from public internet. If set to 'disabled' all traffic except private endpoint traffic and that originates from trusted services will be blocked. This will override the set firewall rules, meaning that even if the firewall rules are present we will not honor the rules.")]
```

src/KeyVault/KeyVault/Commands/UpdateAzureManagedHsm.cs
```diff
@@ -53,6 +53,11 @@ public class UpdateAzureManagedHsm : KeyVaultManagementCmdletBase
             HelpMessage = "specifying whether protection against purge is enabled for this managed HSM pool. The setting is effective only if soft delete is also enabled. Enabling this functionality is irreversible.")]
         public SwitchParameter EnablePurgeProtection { get; set; }
 
+        [Parameter(Mandatory = false,
+            HelpMessage = "Controls permission for data plane traffic coming from public networks while private endpoint is enabled..")]
```
isra-fel, line 57:
```suggestion
            HelpMessage = "Controls permission for data plane traffic coming from public networks while private endpoint is enabled.")]
```

isra-fel (changes requested):

src/KeyVault/KeyVault/Commands/NewAzureKeyVault.cs
BethanyZhou: The reason why use string in this case is PublicNetworkAccess defines as string in swagger. I'm afraid that there may be other states for PublicNetworkAccess like LimitedEnabled or sth else. So I decided to make PublicNetworkAccess as string to align swagger. What's your thought? Refer to https://github.com/Azure/azure-rest-api-specs/blob/5575aeff7364faef03c26a3661e1e64a6bb693a7/specification/keyvault/resource-manager/Microsoft.KeyVault/stable/2021-10-01/keyvault.json#L1308

isra-fel: Very well

BethanyZhou: Wanna keep ArgumentCompleter to avoid modifying validate set if adding additional valid values.

isra-fel (approved):

PR #17848, opened 2022-04-19
[Smoketest]Add try 3 times with 30s sleeping time

lijinpei2008:
## Description
Add try 3 times with 30s sleeping time between each try.

## Checklist

        - {Please put the link here}

### DIFF
--- a/tools/Test/SmokeTest/RmCoreSmokeTests.ps1
+++ b/tools/Test/SmokeTest/RmCoreSmokeTests.ps1
@@ -52,7 +52,7 @@ function Retry-AzCommand {
     $loopLimit = 0
     do {
         try {
-            $script = if($loopLimit -gt 0) {"`$ErrorActionPreference='Continue' `n"} Else {"`$ErrorActionPreference='SilentlyContinue' `n"}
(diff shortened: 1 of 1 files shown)


isra-fel:
Have you tested it?

lijinpei2008:
> Have you tested it?

Yes, passed the test locally whether it's PowerShell or Windows Powershell.

isra-fel:
How about the negative case: you deliberately make a test case fail and check:
1. if the error message contains meaningful info (like exception message and stack trace)
2. if the case is retried

lijinpei2008:
> How about the negative case: you deliberately make a test case fail and check:
> 
> 1. if the error message contains meaningful info (like exception message and stack trace)
> 2. if the case is retried

Yes, change the value of the environment variable "$ErrorActionPreference" to "Continue" when the command runs incorrectly and retry.
If it succeeds, it skips. If the error continues, and the error message is printed on the screen. This behavior will continue 2 times.
For example, when I provide a wrong parameter, I can see an explicit error message telling me that the parameter "Names" does not exist.

isra-fel (approved):
LGTM

PR #18496, opened 2022-06-13
Update the data in the test cases

kceiw:
## Description

We use the PowerShell api to parse the cmdlet from the user input. It requires the corresponding module to be installed so that it can import it and use it to extract the parameter names from the user input. But that module (e.g. Az modules) are not installed in build machines. The test cases that need to extract the parameter names will fail. The solution here is to not to use Az cmdlets as the input for the test cases. We use the cmdlets from the built-in modules, which are copied to the test directory. Thus we can maintain minimal set up for the test machines and still be able to test the functionality in the product code.

## Checklist

        - {Please put the link here}


jjaguirre394 (approved):

mirdaki (approved):
LGTM

PR #18899, opened 2022-07-07
Reformatted exception message for cmdlets

arpit-gagneja:
## Description

As of today, the cmdlets just print the top level message. In our service case, that message is not that informative for the customers. For our HTTP responses, we includes details in the response as well which explains the message with more information.

For example, here is the testing snapshot after changes for one cmdlet response. The first error message for changed cmdlet response contains more details whereas the second error message for unchanged cmdlet is not informative.

Full model validation response

## Checklist

        - {Please put the link here}


BethanyZhou:
@dolauli , could you help review this PR? Is this correct way to reformat exception message?

src/DnsResolver/custom/csharp/RestErrorHandler.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.PowerShell.Cmdlets.DnsResolver.Models.Api20220701;
+
+namespace Microsoft.Azure.PowerShell.Cmdlets.DnsResolver.Cmdlets
+{
+    internal static class CmdletRestExtension
+    {
+        public static void WriteError(this Cmdlet cmdlet, HttpResponseMessage responseMessage, Task<ICloudError> errorResponseTask, ref Task<bool> returnNow)
+        {
+            var errorString = responseMessage.Content.ReadAsStringAsync().ConfigureAwait(false).GetAwaiter().GetResult();
```
dolauli, line 15: We should avoid to print the whole response. According to your ask in the teams channel, can we just print the details section in the response.

arpit-gagneja: @dolauli : we need to print the **message + details**.

For the customer request validation, the details section includes all the missing things in the request payload.
For the business rule validation, the message includes as to why the validation failed.

dolauli: @arpit-gagneja, if there is no details section, you may re-use the logic in ondefault function. Besides we assume Azure services should follow the similar schema to return errors. And actually we have implemented the logic to desterilizing details from the error response, please see .\generated\api\Models\xxx\ErrorResponse.cs for details. So you may access the details by using something like `(await response)?.Detail;`

arpit-gagneja: thanks @dolauli for the suggestion. one more question here: I have made below changes and want to know if there is a utility function to convert Detail to string.

image

arpit-gagneja: @dolauli : I have addressed the comment as suggested and tested the changes locally. The changes look fine.
Please review again at your earliest convenience.

src/DnsResolver/custom/csharp/RestErrorHandler.cs
```diff
@@ -0,0 +1,36 @@
...
+    internal static class CmdletRestExtension
+    {
+        public static void WriteError(this Cmdlet cmdlet, HttpResponseMessage responseMessage, Task<ICloudError> errorResponseTask, ref Task<bool> returnNow)
+        {
+            var errorResponse = errorResponseTask?.ConfigureAwait(false).GetAwaiter().GetResult();
+
+            if (errorResponse?.Detail != null)
+            {
+                var errorDetail = errorResponse.Detail;
+                var errorDetailString = JsonConvert.SerializeObject(errorDetail);
```
dolauli, line 21: Do not use newtonsoft, which is not included in autogen modules. Autorest.powershell actually has an embedded Json support. And [here ](https://github.com/Azure/azure-powershell/blob/main/src/Databricks/generated/api/Models/Api20210401Preview/ErrorInfo.json.cs#L108-L112)is an example about how to use it. Besides, you may loop the errorDetail to call errorDetail[i].ToJsonString(), and then combine them together. Actually, I think you may just need to combine errorDetail[i].Message, and do not need to return the target and the code.

PR #18903, opened 2022-07-08
Use "module" instead of "regex" when updating CreateMappings_rules.json

isra-fel:
## Description

We should avoid using "regex" in CreateMappings_rules.json because regex can easily conflict with each other.
On the other hand, "module" is always accurate.

## Checklist

        - {Please put the link here}

### DIFF
--- a/tools/Gen2Master/MoveFromGeneration2Master.ps1
+++ b/tools/Gen2Master/MoveFromGeneration2Master.ps1
@@ -361,7 +361,7 @@ Function Update-MappingJson {
         $MappingPath = Join-Path -Path (Join-Path -Path $PSScriptRoot -ChildPath '..') -ChildPath "CreateMappings_rules.json"
         $MappingObject = Get-Content -Path $MappingPath | ConvertFrom-Json
         Foreach ($Item in $MappingObject) {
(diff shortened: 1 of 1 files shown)


tools/Gen2Master/MoveFromGeneration2Master.ps1
```diff
@@ -361,11 +361,11 @@ Function Update-MappingJson {
         $MappingPath = Join-Path -Path (Join-Path -Path $PSScriptRoot -ChildPath '..') -ChildPath "CreateMappings_rules.json"
         $MappingObject = Get-Content -Path $MappingPath | ConvertFrom-Json
         Foreach ($Item in $MappingObject) {
-            If ($ModuleName -eq $Item.regex) {
+            If ($ModuleName -eq $Item.module) {
```
BethanyZhou, line 364: Do all generated modules have "module" property?

isra-fel: Thanks. I've updated the condition so that a new rule will be added only if there are no rules match module OR regex.

BethanyZhou (approved):
Looks good to me

PR #19059, opened 2022-07-26
[SQL] Rework confirmation prompt usage in sql base class

ostojicb:
## Description
before:
image

after:
image

Most of our cmdlets do explicitly check `ShouldProcess`, but the base cmdlet class also does the similar thing - so we end up prompting twice for the same thing. Base class's `ExecuteCmdlet` doesn't really support `ShouldContinue` logic, so here I'm switching it up a little bit to enable that (hopefully without breaking existing cmdlet logic)

Also, base `ExecuteCmdlet` would write output even if the action was not performed (i.e. prompt declined) - and would write the 'updated' model which IMO is quite confusing so that's also fixed. 

This PR though doesn't attempt to clean up the double-prompts (that can be done separately)
In addition to that, we could work on the docs to avoid further prompt duplication in sql module cmdlets

## Checklist

    * Create new module from the scratch
* **SHOULD NOT** change version of module in pull request


src/Sql/Sql/Properties/Resources.resx
```diff
@@ -669,8 +669,8 @@
...
     <value>Invalid format of the resource identifier. A long term retention backup resource Id should follow one of the following formats:
-/subscriptions/&lt;subscriptionId&gt;/providers/Microsoft.Sql/locations/&lt;location&gt;/longTermRetentionServers/&lt;serverName&gt;/longTermRetentionDatabases/&lt;databaseName&gt;/longTermRetentionBackups/&lt;backupName&gt;
-/subscriptions/&lt;subscriptionId&gt;/resourceGroups/&lt;resourceGroupName&gt;/providers/Microsoft.Sql/locations/&lt;location&gt;/longTermRetentionServers/&lt;serverName&gt;/longTermRetentionDatabases/&lt;databaseName&gt;/longTermRetentionBackups/&lt;backupName&gt;
+ /subscriptions/&lt;subscriptionId&gt;/providers/Microsoft.Sql/locations/&lt;location&gt;/longTermRetentionServers/&lt;serverName&gt;/longTermRetentionDatabases/&lt;databaseName&gt;/longTermRetentionBackups/&lt;backupName&gt;
+ /subscriptions/&lt;subscriptionId&gt;/resourceGroups/&lt;resourceGroupName&gt;/providers/Microsoft.Sql/locations/&lt;location&gt;/longTermRetentionServers/&lt;serverName&gt;/longTermRetentionDatabases/&lt;databaseName&gt;/longTermRetentionBackups/&lt;backupName&gt;
```
ostojicb, line 673: unrelated change - In 'Designer.cs' file these lines would start with triple `/` (`///`) which is confusing to our compiler, so I'm adding a whitespace - to avoid manually editing .cs file every time we change .resx

src/Sql/Sql/DistributedAvailabilityGroup/Cmdlet/NewAzureSqlManagedInstanceLink.cs
```diff
@@ -123,16 +150,8 @@ public override void ExecuteCmdlet()
                 default:
                     break;
             }
+            base.ExecuteCmdlet();
 
-            // messages describing behavior with -WhatIf and -Confirm flags
```
BethanyZhou, line 127: We should support should process for New-Az* cmdlet. Please revert these changes.

ostojicb: we're still supporting it - via [AzureSqlCmdletBase](https://github.com/Azure/azure-powershell/pull/19059/files#diff-8f33295ee2f2df46a431874d5cd802002e0d98732b0087738f0da851045b0209R149)

BethanyZhou: The problem is not all commands inherited from AzureSqlCmdletBase should support shouldprocess, like Get-Az* cmdlets, see https://cs.github.com/AuxMon/azure-powershell/blob/17a2a45099e51f9420f4705f820667b0378fd610/src/Sql/Sql/DataClassification/Cmdlet/GetAzSqlInstanceDatabaseSensitivityRecommendation.cs?q=%3A+AzureSqlCmdletBase%3C#L28

ostojicb: they don't support it magically; you still need to explicitly specify `SupportsShouldProcess = true`

isra-fel (changes requested):
It's generally not a good idea to call ShouldProcess() in the base class. Quoting the comments [from PowerShell's source code](https://github.com/PowerShell/PowerShell/blob/7dc4587014bfa22919c933607bf564f0ba53db2e/src/System.Management.Automation/engine/cmdlet.cs#L781):
> A Cmdlet should declare [Cmdlet( SupportsShouldProcess = true )] if-and-only-if it calls ShouldProcess before making changes.
However refactoring would take a lot of effort and the cmdlets are working fine, so I'm OK to continue using this pattern.

isra-fel:
Kindly ping @ostojicb

ostojicb:
> Kindly ping @ostojicb

putting this on hold for now

PR #19174, opened 2022-08-08
Added parameter ContentType in Import-AzKeyVaultCertificate to support importing pem via certificate string

BethanyZhou:
…t importing pem via certificate string

## Checklist

    * Create new module from the scratch
* **SHOULD NOT** change version of module in pull request

### DIFF
--- a/src/KeyVault/KeyVault/Commands/Certificate/ImportAzureKeyVaultCertificate.cs
+++ b/src/KeyVault/KeyVault/Commands/Certificate/ImportAzureKeyVaultCertificate.cs
@@ -89,6 +89,15 @@ namespace Microsoft.Azure.Commands.KeyVault
         public string CertificateString { get; set; }
 
         /// <summary>
+        /// Specifies type of the certificate to be imported.
+        /// </summary>
+        [Parameter(Mandatory = false,
+                   ParameterSetName = ImportWithPrivateKeyFromStringParameterSet,
+                   HelpMessage = "Specifies the type of the certificate to be imported. Regards certificate string as PFX format by default.")]
+        [PSArgumentCompleter(Constants.Pkcs12ContentType, Constants.PemContentType)]
+        public string ContentType { get; set; } = Constants.Pkcs12ContentType;
+
+        /// <summary>
         /// Password
         /// </summary>
         [Parameter(Mandatory = false,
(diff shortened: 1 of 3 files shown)


src/KeyVault/KeyVault/Commands/Certificate/ImportAzureKeyVaultCertificate.cs
```diff
@@ -88,6 +88,15 @@ public class ImportAzureKeyVaultCertificate : KeyVaultCmdletBase
...
         public string CertificateString { get; set; }
 
+        /// <summary>
+        /// Specifies type of the certificate to be imported.
+        /// </summary>
+        [Parameter(Mandatory = false,
+                   ParameterSetName = ImportWithPrivateKeyFromStringParameterSet,
+                   HelpMessage = "Specifies the type of the certificate to be imported. Regards certificate string as PFX format by default.")]
+        [PSArgumentCompleter(Constants.Pkcs12ContentType, Constants.PemContentType)]
+        public string ContentType { get; set; } = Constants.Pkcs12ContentType;
```
isra-fel, line 98: I'm thinking whether we should expose the raw content type like `-ContentType application/x-pkcs12` or an enum (or string) to represent the content type for example `-ContentType Pkcs12` and translate the enum to the actual content type before calling SDK. The latter syntax is definitely simpler to use.
What does CLI's interface look like?

isra-fel (commented):
Please also update the help doc.

src/KeyVault/KeyVault/Commands/Certificate/ImportAzureKeyVaultCertificate.cs
BethanyZhou: The reason why I didn't do this translation is I don't want to maintain a mapping list for this kind of content type.

CLI hasn't supported importing certificate from string directly. So they didn't expose this parameter. 

To align with Portal, I agree we can use wrapped value to represent the content type. 

image

isra-fel: I vote for "PKCS12" and "PEM" because they are much clearer to use.

isra-fel (approved):

src/KeyVault/KeyVault/Commands/Certificate/ImportAzureKeyVaultCertificate.cs
BethanyZhou: Has exposed these raw values to customer in New/Set-AzKeyVaultCertificatePolicy. Let us keep align with previous design.

PR #19500, opened 2022-09-15
Remove reference to unused object in New-AzRoleAssignment

dagoroz:
## Description

Removed reference to ResourceIdentifier object in the New-AzRoleAssignment workflow. It is not used at any point and it causes issues for tenants that do not have subscriptions yet

## Checklist

### DIFF
--- a/src/Resources/Resources/RoleAssignments/NewAzureRoleAssignmentCommand.cs
+++ b/src/Resources/Resources/RoleAssignments/NewAzureRoleAssignmentCommand.cs
@@ -318,14 +318,6 @@ namespace Microsoft.Azure.Commands.Resources
                     Id = ObjectId,
                     ObjectType = ObjectType,
                 },
-                ResourceIdentifier = new ResourceIdentifier()
-                {
-                    ParentResource = ParentResource,
-                    ResourceGroupName = ResourceGroupName,
-                    ResourceName = ResourceName,
-                    ResourceType = ResourceType,
-                    Subscription = DefaultProfile.DefaultContext.Subscription.Id,
-                },
                 CanDelegate = AllowDelegation.IsPresent ? true : false,
                 Description = Description,
                 Condition = Condition,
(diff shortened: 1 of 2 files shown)


src/Resources/Resources/RoleAssignments/NewAzureRoleAssignmentCommand.cs
```diff
@@ -318,14 +318,6 @@ public override void ExecuteCmdlet()
                     Id = ObjectId,
                     ObjectType = ObjectType,
                 },
-                ResourceIdentifier = new ResourceIdentifier()
```
RochelleChou, line 321: I don’t think it is a good choice to remove the whole object “ResourceIdentifier”. It will block NewAzureRoleAssignment other parameters(“ParentResource”, “ResourceGroupName” and so on). 
I think we should check the value of “DefaultProfile” is null? “DefaultProfile.DefaultContext” is null? “DefaultProfile.DefaultContext.Subscription” is null? If we have the value “DefaultProfile.DefaultContext.Subscription.Id” we can assign it to “ResourceIdentifier.Subscription”, else we didn’t set any value to “ResourceIdentifier.Subscription”.

isra-fel:
/azp run 
azure-powershell - powershell-core

src/Resources/Resources/RoleAssignments/NewAzureRoleAssignmentCommand.cs
dagoroz: The resource identifier object is never used during the flow of NewAzureRoleAssignmentCommand that is why it was removed instead of patched.

isra-fel: This `FilterRoleAssignmentsOptions` object is used by [`AuthorizationClient.CreateRoleAssignment`](https://github.com/Azure/azure-powershell/blob/38bb792a85ecd428cee2f58be02018958620898c/src/Resources/Resources/Models.Authorization/AuthorizationClient.cs#L134) and the `ResourceIdentifier` property is never referenced, so I agree that it's OK to be removed.

isra-fel (approved):

PR #19712, opened 2022-09-29
Fix aad graph endpoint for SQL server and managed instance APIs

amolagar5:
## Description

Move endpoint used in SQL Server and SQL Instance from AD Graph to MS Graph


strehan1993 (approved):

ericshape (dismissed):
LGTM

dingmeng-xue:
Do you expect this feature go into Ignite release?

tools/TestFx/TestEnvironmentFactory.cs
```diff
@@ -30,7 +30,7 @@ static TestEnvironmentFactory()
 
         internal static TestEnvironment BuildTestFxEnvironment()
         {
-            _environment.SetRecordedEnvironmentVariables();
+            _environment.SetEnvironmentVariables();
```
dingmeng-xue, line 33: why the files under TestFx need to be changed?

dingmeng-xue (changes requested):

tools/TestFx/TestEnvironmentFactory.cs
vidai-msft: For batch tests, the TestEnvironment object is initiated only once. However, the environment variables (set by SetEnvironmentVariables()) need to be re-populated for each test case. Previously, only Playback mode got it re-populated, so it caused SubscriptionId missing in json file during recording.

amolagar5: Thanks @vidai-msft for fixing this.

amolagar5:
> Do you expect this feature go into Ignite release?

Yes. sooner we can release this, better it is. 
We have 2 CRIs due to this so far already.

wyunchi-ms (approved):

PR #19904, opened 2022-10-21
Implementation of interface IPSCmdletAction for test coverage

vidai-msft:
## Description

Implementation of the test coverage to log test raw data
Added config in the pipeline to trigger the log action

## Checklist


src/Accounts/Authentication/Config/ConfigInitializer.cs
```diff
@@ -190,6 +190,12 @@ private void RegisterConfigs(IConfigManager configManager)
                 true,
                 AzurePSDataCollectionProfile.EnvironmentVariableName,
                 new[] { AppliesTo.Az }));
+            configManager.RegisterConfig(new SimpleTypedConfig<bool>(
```
isra-fel, line 193: We should also wrap this `RegisterConfig()` call with `#if`, because we don't want it exposed to customers.

isra-fel (changes requested):

tools/TestCoverageAnalysis/AnalyzeTestCoverage.ps1
```diff
@@ -0,0 +1,169 @@
+$repoRootDirectory = $PSScriptRoot | Split-Path | Split-Path
```
isra-fel, line 1: - Please add [license notice](https://github.com/Azure/azure-powershell/blob/main/tools/CheckAssemblies.ps1) at top of the file.
- Also it would be nice to add doc about how to use this script. [Here's an example](https://github.com/Azure/azure-powershell/blob/main/tools/StaticAnalysis/ExampleAnalyzer/Measure-MarkdownOrScript.ps1)
- We should consider calling this script from pipeline, so that we can validate its correctness.

src/Accounts/Accounts/Account/ConnectAzureRmAccount.cs
```diff
@@ -750,6 +750,7 @@ public void OnImport()
                 AzureSession.Instance.RegisterComponent(nameof(AzureCredentialFactory), () => new AzureCredentialFactory());
                 AzureSession.Instance.RegisterComponent(nameof(MsalAccessTokenAcquirerFactory), () => new MsalAccessTokenAcquirerFactory());
                 AzureSession.Instance.RegisterComponent<ISshCredentialFactory>(nameof(ISshCredentialFactory), () => new SshCredentialFactory());
+                AzureSession.Instance.RegisterComponent<ITestCoverage>(nameof(ITestCoverage), () => new TestCoverage());
```
isra-fel, line 753: To minimize the impact to customers, let's wrap this with `#if DEBUG || TESTCOVERAGE` as well.
Also, if we add it here, we can remove the `#if` in `TestCoverage.cs`

src/Accounts/Authentication/Config/ConfigInitializer.cs
```diff
@@ -190,6 +190,14 @@ private void RegisterConfigs(IConfigManager configManager)
                 true,
                 AzurePSDataCollectionProfile.EnvironmentVariableName,
                 new[] { AppliesTo.Az }));
+#if DEBUG || TESTCOVERAGE
+            configManager.RegisterConfig(new SimpleTypedConfig<bool>(
+                ConfigKeys.EnableTestCoverage,
+                "N/A",
```
isra-fel, line 196: Let's make the help message meaningful just in case.
```suggestion
                "When enabled, the test framework will calculate and output a test coverage report.",
```

isra-fel (changes requested):

tools/TestCoverageAnalysis/AnalyzeTestCoverage.ps1
vidai-msft: License and comments added
The script will be combined with the GitHub bot. It likely will be updated again.

src/Accounts/Authentication/Config/ConfigInitializer.cs
vidai-msft: Help message added.

vidai-msft:
Azure pipeline formatting code added so that it is easy for us to notice the errors during raw data generation

isra-fel:
LGTM. Pending new common build.

.azure-pipelines/powershell-core.yml
```diff
@@ -17,6 +17,7 @@ variables:
   AnalysisTimeoutInMinutes: 120
   TestTimeoutInMinutes: 180
   BuildAzPredictor: false
+  Azure_PS_TestCoverage: true
```
wyunchi-ms, line 20: I think it will be better if the name can align with other variables such as `EnableAzurePSTestCoverage` or just `EnableTestCoverage`

vidai-msft: This is an environment variable and actually the naming convention is from the config framework. I will update to match with the convention for the CI

Nickcandy (approved):

PR #20007, opened 2022-11-02
Updating NewAzurePublicIpPrefixCommand.cs to make location mandatory

brianlehr:
## Description

Updating NewAzurePublicIpPrefixCommand.cs to make location mandatory, as it is always required

## Checklist

- [x] Check this box to confirm: **I have read the [_Submitting Changes_](../blob/main/CONTRIBUTING.md#submitting-changes) section of [`CONTRIBUTING.md`](../blob/main/CONTRIBUTING.md) and reviewed the following information:**

* **SHOULD** select appropriate branch. Cmdlets from Autorest.PowerShell should go to [`generation`](https://github.com/Azure/azure-powershell/tree/generation) branch. 
* **SHOULD** make the title of PR clear and informative, and in the present imperative tense. 
* **SHOULD** update `ChangeLog.md` file(s) appropriately
    * For any service, the `ChangeLog.md` file can be found at `src/{{SERVICE}}/{{SERVICE}}/ChangeLog.md`
    * A snippet outlining the change(s) made in the PR should be written under the `## Upcoming Release` header in the past tense. Add changelog in description section if PR goes into [`generation`](https://github.com/Azure/azure-powershell/tree/generation) branch.
    * Should **not** change `ChangeLog.md` if no new release is required, such as fixing test case only.
* **SHOULD** have approved design review for the changes in [this repository](https://github.com/Azure/azure-powershell-cmdlet-review-pr) ([_Microsoft internal only_](../blob/main/CONTRIBUTING.md#onboarding)) with following situations
    * Create new module from scratch
    * Create new resource types which are not easy to conform to [Azure PowerShell Design Guidelines](../blob/main/documentation/development-docs/design-guidelines)
    * Create new resource type which name doesn't use module name as prefix
    * Have design question before implementation
* **SHOULD** regenerate markdown help files if there is cmdlet API change. [Instruction](../blob/main/documentation/development-docs/help-generation.md#updating-all-markdown-files-in-a-module)
* **SHOULD** have proper test coverage for changes in pull request.
* **SHOULD NOT** introduce [breaking changes](../blob/main/documentation/breaking-changes/breaking-changes-definition.md) in Az minor release except preview version.
* **SHOULD NOT** adjust version of module manually in pull request

PR #20008, opened 2022-11-02
Updating NewAzurePublicIpAddressCommand.cs to make location mandatory

brianlehr:
## Description

Updating NewAzurePublicIpAddressCommand.cs to make location mandatory, as it is always required

## Checklist

### DIFF
--- a/src/Network/Network/PublicIpAddress/NewAzurePublicIpAddressCommand.cs
+++ b/src/Network/Network/PublicIpAddress/NewAzurePublicIpAddressCommand.cs
@@ -51,7 +51,7 @@ namespace Microsoft.Azure.Commands.Network
         public virtual string ResourceGroupName { get; set; }
 
         [Parameter(
-            Mandatory = false,
+            Mandatory = true,
             ValueFromPipelineByPropertyName = true,
             HelpMessage = "The public IP address location.")]
         [LocationCompleter("Microsoft.Network/publicIPAddresses")]


Nickcandy:
This pr would cause a breaking change which is not allowed in regular releases.

Nickcandy:
@BethanyZhou  Hi Beisi, what do you think

brianlehr:
Note that this is already mandatory in practice, you can test it out yourself, I'm just trying to fix auto-generated documentation.

BethanyZhou:
Hi @brianlehr , we are verifying the case you mentioned. 

- From the swagger definition, it looks like it's a location-optional resource, https://github.com/Azure/azure-rest-api-specs/blob/00ea1cdee06eb16e8befcbb8ffa8abf35ba99f4d/specification/network/resource-manager/Microsoft.Network/stable/2022-01-01/network.json#L156
- But in fact, we get response "the location property is required for this definition" if we don's provide location. 

If this parameter is only required in some cases, and it's optional other cases. It still should be marked as optional. We need to check which behavior is by design. @wdehrich could you help confirm this case?

BTW,[ the help markdown of this command](https://github.com/Azure/azure-powershell/blob/main/src/Network/Network/help/New-AzPublicIpAddress.md) needs to be updated if true this parameter is required. And a change log message is required in [ChangeLog.md](https://github.com/Azure/azure-powershell/blob/main/src/Network/Network/ChangeLog.md).

brianlehr:
My understanding was that it was required in call cases - @wdehrich please let me know if there is a case where it is optional.

msJinLei:
duplicate with https://github.com/Azure/azure-powershell/pull/20007

msJinLei:
@brianlehr To switch an optional parameter to mandatory is a break change. We can accept it only if the cmdlet always fails without the parameter. Is that true?

>**Please response as soon as possible for we will freeze code next Monday. Thank you for your understanding and cooperation.**

brianlehr:
> @brianlehr To switch an optional parameter to mandatory is a break change. We can accept it only if the cmdlet always fails without the parameter. Is that true?
> 
> > **Could you response as soon as possible for we will freeze code next Monday? Thank you for your understanding and cooperation.**

@msJinLei This field is already mandatory, and yes it always fails without parameter.

msJinLei:
@brianlehr Please fix the static analysis error refer to the following example.

```
Suppress the breaking change errors in "tools/StaticAnalysis/Exceptions/${ModuleName}/BreakingChangeIssues.csv"
"AssemblyFileName","ClassName","Target","Severity","ProblemId","Description","Remediation"
"Microsoft.Azure.PowerShell.Cmdlets.Network.dll","Microsoft.Azure.Commands.Network.NewAzFirewallPolicyNatRuleCommand","New-AzFirewallPolicyNatRule","0","1050","The parameter set '__AllParameterSets' for cmdlet 'New-AzFirewallPolicyNatRule' has been removed.","Add parameter set '__AllParameterSets' back to cmdlet 'New-AzFirewallPolicyNatRule'."
```

brianlehr:
@msJinLei I'm sorry, I don't understand what the Firewall Policy Rules have to do with my ask?

wyunchi-ms:
This is fixed in https://github.com/Azure/azure-powershell/pull/22909

PR #20196, opened 2022-11-22
Assembly loading improvements

isra-fel:
## Description

* This PR disables assembly preloading because we already implemented lazy-loading.
* The definitions of dependency assemblies are centralized, and with more flexibility. You can specify a range of PowerShell versions to load a certain assembly, or a type of OS. Thanks to this the size of Az module is also reduced.
  * Duplicated `Azure.Identity.dll` is removed.
  * The following assemblies used to have both netframework and netstandard version in our codebase. This PR removed the netframework versions:
    * Microsoft.Bcl.AsyncInterfaces
    * Microsoft.IdentityModel.Abstractions
    * System.Memory.Data
    * System.Text.Json

## Design
The core implementation is a new project named `AssemblyLoading` (and a new assembly `Microsoft.Azure.PowerShell.AssemblyLoading`) in Az.Accounts module. The two assembly loaders `CustomeAssemblyLoader` and `AzAssemblyLoadContext` will depend on this project to provide the list of assemblies to load according to the environment.

## Checklist


isra-fel:
todo:
- https://cs.github.com/Azure/azure-powershell/blob/34f9661146728f308b064663bd4f0e94c1794ecc/tools/Tools.Common/Loaders/CmdletLoader.cs?q=repo%3AAzure%2Fazure-powershell+PreloadAssemblies#L45
- https://cs.github.com/Azure/azure-powershell/blob/34f9661146728f308b064663bd4f0e94c1794ecc/src/Storage/Storage/Storage.csproj?q=repo%3AAzure%2Fazure-powershell+PreloadAssemblies#L27
- https://cs.github.com/Azure/azure-powershell/blob/34f9661146728f308b064663bd4f0e94c1794ecc/src/lib/test.net472.config?q=repo%3AAzure%2Fazure-powershell+PreloadAssemblies#L12

src/Storage/Storage/Storage.csproj
```diff
@@ -23,11 +23,7 @@
     <PackageReference Include="Microsoft.Azure.Storage.File" Version="11.2.2" />
     <PackageReference Include="Microsoft.Azure.Storage.Queue" Version="11.2.2" />
   </ItemGroup>
-  <Target Name="CopyFiles" AfterTargets="Build">
-    <Copy SourceFiles="@(PreLoadAssemblies)" DestinationFolder="$(TargetDir)\PreloadAssemblies" />
-    <Copy SourceFiles="@(NetCoreAssemblies)" DestinationFolder="$(TargetDir)\NetCoreAssemblies" />
-  </Target>
```
isra-fel, line 29: These lines were added to support data plane track 2 SDK (#11520). They preload Azure.Core.dll and its dependencies. But now Azure.Core is handled by Az.Accounts so these are no longer needed.
cc @blueww

src/Accounts/AuthenticationAssemblyLoadContext/AzAssemblyLoadContextBase.cs
```diff
@@ -0,0 +1,99 @@
...
+    internal abstract class AzAssemblyLoadContextBase : AssemblyLoadContext
+    {
+        private ConcurrentDictionary<string, Assembly> AssemblyCache { get; set; } =
+            new ConcurrentDictionary<string, Assembly>(StringComparer.OrdinalIgnoreCase);
+
+        protected AzAssemblyLoadContextBase(string name) : base(name) { }
+
+        protected void CacheAssembly(string name, Assembly assembly)
+        {
+            AssemblyCache.TryAdd(name, assembly);
```
wyunchi-ms, line 34: Will it be better to write a debug message is try add fails? If no, why not use `AddOrUpdate` to instead?

src/Accounts/AuthenticationAssemblyLoadContext/AzSharedAssemblyLoadContext.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                var loadedAssembly = LoadFromAssemblyPath(assembly.Path);
+                var loadedAssemblyName = loadedAssembly.GetName();
+                if (IsAssemblyMatching(requestedAssemblyName, loadedAssemblyName))
+                {
+                    CacheAssembly(loadedAssemblyName.Name, loadedAssembly);
+                }
+                return loadedAssembly;
+            }
+            return null;
```
wyunchi-ms, line 54: Will it be better to add a warning message here if the assembly is not found for debug?

src/Accounts/AuthenticationAssemblyLoadContext/AzAssemblyLoadContextBase.cs
isra-fel: @wyunchi-ms thanks for reviewing :)

>  write a debug message is try add fails

You are right, it would be better. However it would take much effort to implement because (1) we don't have a cmdlet instance so it's tricky to write debug log here and (2) we wouldn't want the assembly loading debug message to be mixed with normal debug message because there will be plenty of them and they are not useful for debugging in most cases.

> why not use AddOrUpdate

It does not make sense to cache the same assembly twice, so we don't need `AddOrUpdate` here. `TryAdd` is enough.

wyunchi-ms (approved):

PR #20382, opened 2022-12-07
Fixed bugs for test coverage

vidai-msft:
This PR modified the followings
- Fixed bug of test coverage not generated on Linux and MacOS due to path symbol issue for both SDK and AutoRest
- Fixed bug of test coverage not generated on Windows due to wrong relationship/pattern between current location and test coverage data location
- Added support for test coverage to specify custom data location. If the location is not specified, it will fall back to .azure folder
- Fixed bug of test coverage calculation that cmdlets/parametersets were not counted in when no raw data file was generated for that module.
- Added properties StartDateTime and EndDateTime for test coverage raw data

## Description

Debugging test coverage

## Checklist

### DIFF
--- a/src/Accounts/Authentication/TestCoverage.cs
+++ b/src/Accounts/Authentication/TestCoverage.cs
@@ -12,6 +12,7 @@
 // limitations under the License.
 // ----------------------------------------------------------------------------------
 
+using Microsoft.Azure.Commands.Common.Authentication.Properties;
 using Microsoft.Azure.Commands.Shared.Config;
 using Microsoft.Azure.PowerShell.Common.Config;
 using Microsoft.WindowsAzure.Commands.Common;
@@ -57,7 +58,9 @@ namespace Microsoft.Azure.Commands.Common.Authentication
             }
             if (string.IsNullOrEmpty(testCoverageRootPath))
             {
-                testCoverageRootPath = GetFallbackLocation();
+                testCoverageRootPath = Path.Combine(
+                    Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),
+                    Resources.AzureDirectoryName);
             }
 
             s_testCoveragePath = Path.Combine(testCoverageRootPath, "TestCoverageAnalysis", "Raw");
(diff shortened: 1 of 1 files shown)


src/Accounts/Authentication/TestCoverage.cs
```diff
@@ -41,36 +44,37 @@ public class TestCoverage : ITestCoverage
...
 
-            string result = Directory.Exists(directoryPath) ? Path.GetFullPath(directoryPath) : null;
-            return result;
+            Console.WriteLine($"Test coverage data location: ${s_testCoveragePath}");
+        }
+
+        private static string GetFallbackLocation()
+        {
+            var profilePath = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
+            var azurePath = Path.Combine(profilePath, ".azure");
```
isra-fel, line 76: It's capitalized A
```suggestion
            var azurePath = Path.Combine(profilePath, ".Azure");
```

isra-fel (changes requested):

src/Accounts/Authentication/TestCoverage.cs
vidai-msft: @isra-fel in my local environments (both Windows and Linux WSL), they are all with lower cases. Do we have any official document mentioning that?

isra-fel: No documentation but we use this string for folder name. You should use it too.
https://github.com/Azure/azure-powershell/blob/2d4550c23745a3a9710ecbe5ff3c378f1d4cd300/src/Accounts/Authentication/Properties/Resources.resx#L124

Also I tested cloud shell, our metadata files are all in ".Azure"
image

isra-fel: BTW If you can reproduce how to create the lower case ".azure" folder, it may prove there's an issue in our code.

vidai-msft: Not able to reproduce. Let me update to upper case first.

isra-fel: OK. But I actually suggest using the resource string to avoid magic string, i.e. `Resources.AzureDirectoryName`.

vidai-msft: Changed to use resource string

isra-fel (approved):

Issue #20323, opened 2022-11-29
Get-AzRoleAssignment -scope "/" returns BadRequest, but Get-AzRoleAssignment | Where-Object {$_.scope -eq "/"} is fine


ChristopherGLewis:
### Description

`Get-AzRoleAssignment -scope "/"` returns a BadRequest error:

``` PowerShell
c:\> Get-AzRoleAssignment -scope "/"                       
Get-AzRoleAssignment: Operation returned an invalid status code 'BadRequest'
```
but this works OK with a Where-Object 

``` PowerShell
c:\> Get-AzRoleAssignment | Where-Object {$_.scope -eq "/"}

RoleAssignmentName : 8f0f3d8b-4536-47ce-95a0-edcb424c49db
RoleAssignmentId   : /providers/Microsoft.Authorization/roleAssignments/8f0f3d8b-4536-47ce-95a0-edcb424c49db
Scope              : /
DisplayName        : First Last
SignInName         : First.Last@domain.com
RoleDefinitionName : User Access Administrator
RoleDefinitionId   : 18d7d88d-d35e-4fb5-a5c3-7773c20a72d9
ObjectId           : 901c8c08-56c4-4df6-a651-2e4ba77c74f0
ObjectType         : User
CanDelegate        : False
Description        : 
ConditionVersion   : 
Condition          : 

RoleAssignmentName : 4dd06e22-14cf-4fab-9ff4-1845878a4d6f
RoleAssignmentId   : /providers/Microsoft.Authorization/roleAssignments/4dd06e22-14cf-4fab-9ff4-1845878a4d6f
Scope              : /
DisplayName        : First Last
SignInName         : First.Last@domain.com
RoleDefinitionName : Owner
RoleDefinitionId   : 8e3af657-a8ff-443c-a75c-2fe8c4bcb635
ObjectId           : 901c8c08-56c4-4df6-a651-2e4ba77c74f0
ObjectType         : User
CanDelegate        : False
Description        : 
ConditionVersion   : 
Condition          : 
'
```

Note this works perfectly fine using the AZ CLI:

``` sh
c:\>  az role assignment list --scope "/"
[
  {
    "canDelegate": null,
    "condition": null,
    "conditionVersion": null,
    "description": null,
    "id": "/providers/Microsoft.Authorization/roleAssignments/8f0f3d8b-4536-47ce-95a0-edcb424c49db",
    "name": "8f0f3d8b-4536-47ce-95a0-edcb424c49db",
[...]

PR #20433, opened 2022-12-13
Fixed `Get-AzRoleAssignment` BadRequest when scope is '/' [#20323]

rachelxj-ms:
## Description

reference: https://learn.microsoft.com/en-us/azure/role-based-access-control/role-assignments-list-rest
<html>
<body>

Filter | Description
-- | --
$filter=atScope() | Lists role assignments for only the specified scope, not including the role assignments at subscopes.
$filter=assignedTo('{objectId}') | Lists role assignments for a specified user or service principal.If the user is a member of a group that has a role assignment, that role assignment is also listed. This filter is transitive for groups which means that if the user is a member of a group and that group is a member of another group that has a role assignment, that role assignment is also listed.This filter only accepts an object ID for a user or a service principal. You cannot pass an object ID for a group.
$filter=atScope()+and+assignedTo('{objectId}') | Lists role assignments for the specified user or service principal and at the specified scope.
$filter=principalId+eq+'{objectId}' | Lists role assignments for a specified user, group, or service principal.

</body>
</html>

https://learn.microsoft.com/en-us/azure/role-based-access-control/elevate-access-global-admin#elevate-access-for-a-global-administrator-1
https://learn.microsoft.com/en-us/rest/api/authorization/role-assignments/list-for-scope?tabs=HTTP

check root scope "/" only to fix bad request.
            

## Checklist

- [*] Check this box to confirm: **I have read the [_Submitting Changes_](../blob/main/CONTRIBUTING.md#submitting-changes) section of [`CONTRIBUTING.md`](../blob/main/CONTRIBUTING.md) and reviewed the following information:**

### DIFF
--- a/src/Resources/Resources/Models.Authorization/AuthorizationClient.cs
+++ b/src/Resources/Resources/Models.Authorization/AuthorizationClient.cs
@@ -229,9 +229,11 @@ namespace Microsoft.Azure.Commands.Resources.Models.Authorization
                     principalId = string.IsNullOrEmpty(options.ADObjectFilter.Id) ? adObject.Id.ToString() : options.ADObjectFilter.Id;
                     if (isRootScope)
                     {
-                        // $filter = principalId + eq + '{objectId}' Lists role assignments for a specified user, group, or service principal.
-                        // If you use atScope() and principalId+eq + '{objectId}' together, it will throw exception because the API doesn't allow it.
-                        // objectId could represent a group, so can't use atScope() and assignedTo('{objectId}').
+                        /* $filter = principalId + eq + '{objectId}' Lists role assignments for a specified user, group, or service principal.
+                         * If you use atScope() and principalId+eq + '{objectId}' together, it will throw exception because the API doesn't allow it.
+                         * objectId could represent a group, so can't use atScope() and assignedTo('{objectId}') as alternative,
+                         * must filter after the results return from server.
+                         */
                         odataQuery = new ODataQuery<RoleAssignmentFilter>(f => f.AtScope());
                         needsFilterPrincipalId = true;
                     } 


src/Resources/Resources/Models.Authorization/AuthorizationClient.cs
```diff
@@ -187,7 +187,20 @@ public List<PSRoleAssignment> FilterRoleAssignments(FilterRoleAssignmentsOptions
...
+            // refer to https://learn.microsoft.com/en-us/azure/role-based-access-control/role-assignments-list-rest
+            // https://learn.microsoft.com/en-us/azure/role-based-access-control/elevate-access-global-admin#elevate-access-for-a-global-administrator-1
+            // scope is path variable in REST API. When scope is '/', $filter=atScope() is required, or else it will throw BadRequest.
+            // $filter=atScope() to return all role assignments at or above the scope. or else it returns all role assignments at, above or below the scope for the specific scope.
+            //
+            // Known issue:
+            // Before change: This command will return all role assignments at, above or below the scope for the specific scope.
+            // After change: This command will return all role assignments at the scope for the specific scope.
+            // But Azure cli az role assignment list --scope will only return role assignments at the specific scope, --include-inherited will return the assignments above the scope
+            // In order to implement this feature, it needs to add a new parameter IncludeInherited.
```
dagoroz, line 199: This block of comments is very hard to read.
Do you intend to return role assignments at scope for ALL scopes or only if no scope is provided? (i.e. intended scope is "/")

src/Resources/Resources/Models.Authorization/AuthorizationClient.cs
```diff
@@ -209,14 +222,35 @@ public List<PSRoleAssignment> FilterRoleAssignments(FilterRoleAssignmentsOptions
                     }
 
                     principalId = adObject.Id.ToString();
-                    odataQuery = new ODataQuery<RoleAssignmentFilter>(f => f.AssignedTo(principalId));
+                    if (!hasScope)
```
dagoroz, line 225: this conditional is hard to read. hasScope is already a negated statement (scope is NOT null or empty)
remove the ! and swap the code blocks. Makes the flow easier to follow.

src/Resources/Resources/Models.Authorization/AuthorizationClient.cs
```diff
@@ -209,14 +222,35 @@ public List<PSRoleAssignment> FilterRoleAssignments(FilterRoleAssignmentsOptions
...
+                    else
+                    {
+                        odataQuery = new ODataQuery<RoleAssignmentFilter>(f => (f.AtScope() && f.AssignedTo(principalId)));
+                    }
                 }
                 else
                 {
                     principalId = string.IsNullOrEmpty(options.ADObjectFilter.Id) ? adObject.Id.ToString() : options.ADObjectFilter.Id;
-                    odataQuery = new ODataQuery<RoleAssignmentFilter>(f => f.PrincipalId == principalId);
+                    if (!hasScope)
```
dagoroz, line 237: same comment about hard to read code

src/Resources/Resources/Models.Authorization/AuthorizationClient.cs
```diff
@@ -209,14 +222,35 @@ public List<PSRoleAssignment> FilterRoleAssignments(FilterRoleAssignmentsOptions
...
+                    {
+                        odataQuery = new ODataQuery<RoleAssignmentFilter>(f => f.PrincipalId == principalId);
+                    } 
+                    else
+                    {
+                        // atScope() can't use with pricipalId eq {objectId} expression, Principal id could be group id.
```
dagoroz, line 243: true but this is also true with other scopes, idk why you do this distinction here

src/Resources/Resources/Models.Authorization/AuthorizationClient.cs
```diff
@@ -209,14 +222,35 @@ public List<PSRoleAssignment> FilterRoleAssignments(FilterRoleAssignmentsOptions
...
+                        // Only pricipalId eq {objectId} can list role assignments for Group.
```
dagoroz, line 244: this might be poorly phrased, the way it's worded it looks like it contradicts your comment just one line before

dagoroz (commented):
Besides the comments I made in the code I don't understand the intention behind this PR.
The title makes it sound like this is about allowing requests when scope is '/' but the code as written today will change the behavior of the cmdlet taking away the scope and bellow capabilities.

We're planning on migrating to auto-generated cmdlets based on the swagger specs in the not too distant future, therefore we would prefer not to make major behavior changes before that.

src/Resources/Resources/Models.Authorization/AuthorizationClient.cs
rachelxj-ms: > This block of comments is very hard to read. Do you intend to return role assignments at scope for ALL scopes or only if no scope is provided? (i.e. intended scope is "/")

With -scope set in the command, the behavior in powershell is different from cli command.  If you check the document, this is a bug. If you think the change is too much, I can change it to check scope '/' only, but please be aware of this existing bug.
https://learn.microsoft.com/en-us/powershell/module/az.resources/get-azroleassignment?view=azps-9.1.0#-scope

src/Resources/Resources/Models.Authorization/AuthorizationClient.cs
rachelxj-ms: <html>
<body>


$filter=principalId+eq+'{objectId}' | Lists role assignments for a specified user, group, or service principal.
-- | --



</body>
</html>
if you use atScope() and principalId+eq+'{objectId}' together, it will throw exception because the API doesn't allow it. 
objectId could represent a group, so can't use atScope() and assignedTo('{objectId}').

src/Resources/Resources/Models.Authorization/AuthorizationClient.cs
rachelxj-ms: I have removed the extra fix for -scope, only check '/' to fix BadRequest here.

dagoroz: I have taken note of the bug and will address after we migrate to autogeneration.
Please do not change the behavior in this PR

dagoroz (approved):
LGTM in its current state.
Please double check tests don't break and are still testing appropriately

msJinLei (approved):

msJinLei (approved):

PR #20679, opened 2023-01-17
[Storage] Add warning message for AllowBlobPublicAccess default behavior change

yifanz0:
## Description

Add a warning message for the upcoming breaking change in New-AzStorageAccount.
The breaking change will be changing the default value of AllowBlowPublicAccess to False from True when creating a Storage account. 

## Checklist


blueww (dismissed):
@yifanz0 

Would you please give an example of the warning message?

And better add a change log for that.

yifanz0:
@blueww change log added. Following is an example of the warning message - 
image

src/Storage/Storage.Management/ChangeLog.md
```diff
@@ -18,7 +18,8 @@
         - Additional information about change #1
 -->
 ## Upcoming Release
-
+* Add a warning message for the upcoming breaking change when creating a Storage account 
+    - `New-AzStorageAccount`
```
blueww, line 22: Add an empty line after this line.

src/Storage/Storage.Management/StorageAccount/NewAzureStorageAccount.cs
```diff
@@ -22,9 +22,12 @@
 using Microsoft.Azure.Commands.ResourceManager.Common.ArgumentCompleters;
 using System;
 using System.Collections.Generic;
+using Microsoft.WindowsAzure.Commands.Common.CustomAttributes;
 
 namespace Microsoft.Azure.Commands.Management.Storage
 {
+    [GenericBreakingChange("Default behavior of AllowBlobPublicAccess will be changed", OldWay = "AllowBlobPublicAccess is set to True by defult.",
```
blueww, line 29: "Default behavior of AllowBlobPublicAccess will be changed" -> "Default behavior of AllowBlobPublicAccess will be changed in a future release."

And it's better to add the impact of the default value change, like after the change, with default value as false, user can't set the container public access, or the public access setting on container won't work (not allow anonymous access to containers and blobs)? Please test which way is correct and update it.

src/Storage/Storage.Management/ChangeLog.md
yifanz0: Done

src/Storage/Storage.Management/StorageAccount/NewAzureStorageAccount.cs
yifanz0: Done. Can you help review if the updated warning message detailed enough? 

image

src/Storage/Storage.Management/StorageAccount/NewAzureStorageAccount.cs
```diff
@@ -22,9 +22,15 @@
 using Microsoft.Azure.Commands.ResourceManager.Common.ArgumentCompleters;
 using System;
 using System.Collections.Generic;
+using Microsoft.WindowsAzure.Commands.Common.CustomAttributes;
 
 namespace Microsoft.Azure.Commands.Management.Storage
 {
+    [GenericBreakingChange("Default behavior of AllowBlobPublicAccess will be changed in a future release. The default value of the property will be " +
+        "changed from True to False. With AllowBlobPublicAccess set to False for the Storage account, a container's public access level cannot be set. " +
+        "If you attempt to set the public access level, it will error out indicating that the container's public access level cannot be set.",
```
blueww, line 31: I think we don't need  "If you attempt to set the public access level, it will error out indicating that the container's public access level cannot be set."

How about change it to:  "Default value of AllowBlobPublicAccess will be changed from True to False in a future release. When AllowBlobPublicAccess is False on a storage account, it is not permitted to configure container ACLs to allow anonymous access to blobs within the storage account. "

yifanz0: Yes this one looks good. Thanks!

blueww (approved):

isra-fel (approved):
LGTM

isra-fel (approved):

PR #21427, opened 2023-04-04
[Storage] Fix the issue of listing blobs with leading slashes 

yifanz0:
## Description

Fix issue https://dev.azure.com/msazure/One/_workitems/edit/17659911
This will only fix listing blobs. Getting a single blob will be fixed in the next major release

## Checklist


src/Storage/Storage/Blob/Cmdlet/GetAzureStorageBlob.cs
```diff
@@ -329,6 +329,7 @@ internal async Task ListBlobsByPrefix(long taskId, IStorageBlobManagement localC
                     {
                         if (blobFilter == null || blobFilter(item.Name))
                         {
+                            ClientOptions.TrimBlobNameSlashes = false;
```
blueww, line 332: Not sure if this really works. Have you tested the TrimBlobNameSlashes = false here works?

yifanz0: It works for the cases of listing blobs below - 
image

yifanz0: One case that is still not working is getting a single blob with a leading slash in the name, say, Get-AzStorageBlob -Container $containerName -Blob //aaa/test1 
As the fix only applies for listing blobs, this command will still result in blob not found

src/Storage/Storage/Common/Util.cs
```diff
@@ -419,6 +419,25 @@ public static BlobBaseClient GetTrack2BlobClient(BlobContainerClient track2conta
             return GetTrack2BlobClient(blobUriBuilder.ToUri(), context, options, blobType);
         }
 
+        public static BlobBaseClient GetTrack2BlobClientNoSlashTrim(BlobContainerClient track2container, string blobName, AzureStorageContext context, string versionId = null, bool? IsCurrentVersion = null, string snapshot = null, BlobClientOptions options = null, global::Azure.Storage.Blobs.Models.BlobType? blobType = null)
+        {
+            //Get Track2 Blob Client Uri
+            BlobUriBuilder blobUriBuilder = new BlobUriBuilder(track2container.Uri, false)
```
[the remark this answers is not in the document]
blueww: I have another idea, we might don't need a new function, but just add a new parameter like "TrimBlobNameSlashes = true" as the last parameter to the orignal function:
1. Don't need dup code
2. Other place call this function don't need change, since the default is still true
3. For the list blob, call the function by adding parameter "TrimBlobNameSlashes: true"

How do you think about it?

src/Storage/Storage/Blob/Cmdlet/GetAzureStorageBlob.cs
blueww: For single blob, there's no regression, might be OK.
However, if you can also fix single blob one, it should be better.
(Or if we consider this as breaking, might can add a breaking notice in 4/25 release, then add the single blob implemetation on May release.)

src/Storage/Storage/Common/Util.cs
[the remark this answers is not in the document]
yifanz0: Yes added.

src/Storage/Storage/Blob/Cmdlet/GetAzureStorageBlob.cs
yifanz0: Yes I'm thinking that for single blob, this might be a breaking change for the following scenario: 
If the customer has both "//aa/test1" and "aa/test1", then when getting a single blob with input "//aa/test1", currently the cmdlet will return "aa/test1" instead of "//aa/test1".

Will add the breaking notice in another PR

src/Storage/Storage/Blob/Cmdlet/GetAzureStorageBlob.cs
```diff
@@ -34,6 +34,9 @@ namespace Microsoft.WindowsAzure.Commands.Storage.Blob.Cmdlet
     /// <summary>
     /// list azure blobs in specified azure container
     /// </summary>
+    [GenericBreakingChange("Leading and trailing slashes on the blob name will not be trimmed when getting a single blob",
+       OldWay = "Leading and trailing slashes on the blob name are trimmed",
```
blueww, line 38: Could we don't add oldway/newway?
This change impact might be small, so don't want to make the warning too long.

yifanz0: Yes. Deleted.

blueww (dismissed):

wyunchi-ms (approved):

PR #21603, opened 2023-04-19
Inspection Limit Feature for Application Gateway Firewall Policy Settings

Shawnli222:
## Description

Allow customers to put limit on their inspection for application gateway firewall policy settings. This feature is built on the existing Firewall PolicySettings framework. A new field requestBodyInspectLimitInKB, requestBodyEnforcement, fileUploadEnforcement is added to provide inspection limit feature.

## Checklist

### DIFF
--- a/src/Network/Network/FirewallPolicy/PolicySettings/AzureApplicationGatewayFirewallPolicySetting.cs
+++ b/src/Network/Network/FirewallPolicy/PolicySettings/AzureApplicationGatewayFirewallPolicySetting.cs
@@ -40,7 +40,7 @@ namespace Microsoft.Azure.Commands.Network
 
         [Parameter(Mandatory = false, HelpMessage = "Max inspection limit in KB for request body inspection.")]
         [ValidateNotNullOrEmpty]
-        public int RequestBodyInspectLimitInKB { get; set; }
+        public int? RequestBodyInspectLimitInKB { get; set; }
 
         [Parameter(
             HelpMessage = "Disable Request Body check.")]
@@ -99,6 +99,11 @@ namespace Microsoft.Azure.Commands.Network
             if (!this.MyInvocation.BoundParameters.ContainsKey("CustomBlockResponseStatusCode"))
             {
                 this.CustomBlockResponseStatusCode = (int?)null;
+            }
+
+            if (!this.MyInvocation.BoundParameters.ContainsKey("RequestBodyInspectLimitInKB"))
+            {
+                this.RequestBodyInspectLimitInKB = (int?)null;
             }
 
             if (!this.MyInvocation.BoundParameters.ContainsKey("DisableFileUploadEnforcement"))
--- a/src/Network/Network/Models/PSApplicationGatewayFirewallPolicySettings.cs
+++ b/src/Network/Network/Models/PSApplicationGatewayFirewallPolicySettings.cs
@@ -32,7 +32,7 @@ namespace Microsoft.Azure.Commands.Network.Models
         public bool? RequestBodyEnforcement { get; set; }
 
         [Ps1Xml(Target = ViewControl.Table)]
-        public int RequestBodyInspectLimitInKB { get; set; }
+        public int? RequestBodyInspectLimitInKB { get; set; }
 
         [Ps1Xml(Target = ViewControl.Table)]
         public bool RequestBodyCheck { get; set; }
--- a/src/Network/Network/help/New-AzApplicationGatewayFirewallPolicySetting.md
+++ b/src/Network/Network/help/New-AzApplicationGatewayFirewallPolicySetting.md
@@ -67,7 +67,7 @@ Accept wildcard characters: False
 Max inspection limit in KB for request body inspection.
 
 ```yaml
-Type: System.Int32
+Type: System.Nullable`1[System.Int32]
 Parameter Sets: (All)
 Aliases:
 
--- a/src/Network/Network.Test/SessionRecords/Commands.Network.Test.ScenarioTests.ApplicationGatewayTests/TestApplicationGatewayFirewallPolicyWithInspectionLimit.json
+++ b/src/Network/Network.Test/SessionRecords/Commands.Network.Test.ScenarioTests.ApplicationGatewayTests/TestApplicationGatewayFirewallPolicyWithInspectionLimit.json
@@ -6,7 +6,7 @@
       "RequestMethod": "GET",
       "RequestHeaders": {
         "x-ms-client-request-id": [
-          "204026e1-c4dc-4ad9-9d8b-9e5e6e0437ad"
+          "7e0abb65-4863-4534-8ef0-c220c09f6972"
         ],
         "Accept-Language": [
           "en-US"
@@ -30,13 +30,13 @@
           "11999"
         ],
         "x-ms-request-id": [
-          "1ab8890a-658a-49b0-a4ab-8b1ecc059cf2"
+          "12df3bc6-b379-4c8d-9f30-58720b1a6732"
         ],
         "x-ms-correlation-request-id": [
-          "1ab8890a-658a-49b0-a4ab-8b1ecc059cf2"
+          "12df3bc6-b379-4c8d-9f30-58720b1a6732"
         ],
         "x-ms-routing-request-id": [
-          "WESTUS:20230502T182555Z:1ab8890a-658a-49b0-a4ab-8b1ecc059cf2"
+          "WESTUS:20230503T045755Z:12df3bc6-b379-4c8d-9f30-58720b1a6732"
         ],
         "Strict-Transport-Security": [
           "max-age=31536000; includeSubDomains"
@@ -45,7 +45,7 @@
           "nosniff"
         ],
         "Date": [
-          "Tue, 02 May 2023 18:25:55 GMT"
+          "Wed, 03 May 2023 04:57:55 GMT"
         ],
         "Content-Type": [
           "application/json; charset=utf-8"
@@ -61,12 +61,12 @@
       "StatusCode": 200
     },
     {
-      "RequestUri": "/subscriptions/1eca0587-27cc-4d6e-97ce-0b4935c307f4/resourcegroups/ps6301?api-version=2016-09-01",
-      "EncodedRequestUri": "L3N1YnNjcmlwdGlvbnMvMWVjYTA1ODctMjdjYy00ZDZlLTk3Y2UtMGI0OTM1YzMwN2Y0L3Jlc291cmNlZ3JvdXBzL3BzNjMwMT9hcGktdmVyc2lvbj0yMDE2LTA5LTAx",
+      "RequestUri": "/subscriptions/1eca0587-27cc-4d6e-97ce-0b4935c307f4/resourcegroups/ps2886?api-version=2016-09-01",
+      "EncodedRequestUri": "L3N1YnNjcmlwdGlvbnMvMWVjYTA1ODctMjdjYy00ZDZlLTk3Y2UtMGI0OTM1YzMwN2Y0L3Jlc291cmNlZ3JvdXBzL3BzMjg4Nj9hcGktdmVyc2lvbj0yMDE2LTA5LTAx",
       "RequestMethod": "PUT",
       "RequestHeaders": {
         "x-ms-client-request-id": [
-          "7306118a-d5d0-4c0f-88b3-aa892f40d224"
+          "ef876282-bae7-411e-bfa4-9a71cddd6412"
         ],
         "Accept-Language": [
           "en-US"
@@ -96,13 +96,13 @@
           "1199"
         ],
         "x-ms-request-id": [
-          "ec316b6f-2c98-4486-82e3-9c691eb41b5f"
+          "4b1a4d7b-1a9e-4e59-b4fc-dd60f92d0885"
         ],
         "x-ms-correlation-request-id": [
-          "ec316b6f-2c98-4486-82e3-9c691eb41b5f"
+          "4b1a4d7b-1a9e-4e59-b4fc-dd60f92d0885"
         ],
         "x-ms-routing-request-id": [
-          "WESTUS:20230502T182556Z:ec316b6f-2c98-4486-82e3-9c691eb41b5f"
+          "WESTUS:20230503T045757Z:4b1a4d7b-1a9e-4e59-b4fc-dd60f92d0885"
         ],
         "Strict-Transport-Security": [
           "max-age=31536000; includeSubDomains"
@@ -111,7 +111,7 @@
           "nosniff"
         ],
         "Date": [
-          "Tue, 02 May 2023 18:25:55 GMT"
+          "Wed, 03 May 2023 04:57:56 GMT"
         ],
         "Content-Length": [
           "238"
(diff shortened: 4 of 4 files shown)


src/Network/Network/FirewallPolicy/PolicySettings/AzureApplicationGatewayFirewallPolicySetting.cs
```diff
@@ -43,6 +51,10 @@ public class AzureApplicationGatewayFirewallPolicySetting : NetworkBaseCmdlet
         [ValidateNotNullOrEmpty]
         public int MaxRequestBodySizeInKb { get; set; }
 
+        [Parameter(Mandatory = false, HelpMessage = "Whether allow WAF to enforce file upload limits.")]
```
tejasshah7, line 54: Change to "Allow WAF to enforce file upload limit". Do we not specific default value here?

src/Network/Network/FirewallPolicy/PolicySettings/AzureApplicationGatewayFirewallPolicySetting.cs
```diff
@@ -33,7 +33,15 @@ public class AzureApplicationGatewayFirewallPolicySetting : NetworkBaseCmdlet
         [ValidateSet("Disabled", "Enabled", IgnoreCase = true)]
         [ValidateNotNullOrEmpty]
         public string State { get; set; }
-        
+
+        [Parameter(Mandatory = false, HelpMessage = "Whether allow WAF to enforce request body limits.")]
```
tejasshah7, line 37: Change to "Configure WAF to enforce request body limit". Do we not specific default value here?

src/Network/Network/FirewallPolicy/PolicySettings/AzureApplicationGatewayFirewallPolicySetting.cs
```diff
@@ -33,7 +33,15 @@ public class AzureApplicationGatewayFirewallPolicySetting : NetworkBaseCmdlet
...
+        [ValidateNotNullOrEmpty]
+        public bool RequestBodyEnforcement { get; set; }
```
tejasshah7, line 39: Should we rather change to EnableRequestBodyEnforcement to make it more clear?

src/Network/Network/FirewallPolicy/PolicySettings/AzureApplicationGatewayFirewallPolicySetting.cs
```diff
@@ -43,6 +51,10 @@ public class AzureApplicationGatewayFirewallPolicySetting : NetworkBaseCmdlet
         [ValidateNotNullOrEmpty]
         public int MaxRequestBodySizeInKb { get; set; }
 
+        [Parameter(Mandatory = false, HelpMessage = "Whether allow WAF to enforce file upload limits.")]
+        [ValidateNotNullOrEmpty]
+        public bool FileUploadEnforcement { get; set; }
```
tejasshah7, line 56: Should we rather change to EnableFileUploadEnforcement?

src/Network/Network/FirewallPolicy/PolicySettings/AzureApplicationGatewayFirewallPolicySetting.cs
Shawnli222: These are description from swagger PR

src/Network/Network/FirewallPolicy/PolicySettings/AzureApplicationGatewayFirewallPolicySetting.cs
Shawnli222: This name is from swagger

src/Network/Network/FirewallPolicy/PolicySettings/AzureApplicationGatewayFirewallPolicySetting.cs
Shawnli222: same as above

tejasshah7: We should have changed it there :( I wonder why swagger team didn't raise it

src/Network/Network/FirewallPolicy/PolicySettings/AzureApplicationGatewayFirewallPolicySetting.cs
Shawnli222: Actually we cannot change the name of the variable because we need it to match the one in the swagger network sdk, else the test would fail.

src/Network/Network.Test/ScenarioTests/ApplicationGatewayTests.ps1
```diff
@@ -4352,6 +4352,60 @@ function Test-ApplicationGatewayFirewallPolicyWithUppercaseTransform
...
+
+	try
+	{
+		$resourceGroup = New-AzResourceGroup -Name $rgname -Location $location -Tags @{ testtag = "APPGw tag"}
+
+		# WAF Policy and Custom Rule
+		$variable = New-AzApplicationGatewayFirewallMatchVariable -VariableName RequestHeaders -Selector Content-Length
+		$condition =  New-AzApplicationGatewayFirewallCondition -MatchVariable $variable -Operator GreaterThan -MatchValue 1000 -Transform Uppercase -NegationCondition $False
+		$rule = New-AzApplicationGatewayFirewallCustomRule -Name example -Priority 2 -RuleType MatchRule -MatchCondition $condition -Action Block
+		$policySettings = New-AzApplicationGatewayFirewallPolicySetting -Mode Prevention -State Enabled -RequestBodyEnforcement $True -RequestBodyInspectLimitInKB 2000 -MaxFileUploadInMb 70 -FileUploadEnforcement $True -MaxRequestBodySizeInKb 70
```
karanbazaz, line 4371: Can we put non default values "false" for the new boolean fields so that we can make sure that they are set. Default value ("true") will be set any ways and so might not be a true test.

Shawnli222: I think in the swagger the default value is set to true

Shawnli222: "requestBodyEnforcement": {
          "type": "boolean",
          "default": true,
          "description": "Whether allow WAF to enforce request body limits."
        },
        "maxRequestBodySizeInKb": {
          "type": "integer",
          "format": "int32",
          "minimum": 8,
          "exclusiveMinimum": false,
          "description": "Maximum request body size in Kb for WAF."
        },
        "fileUploadEnforcement": {
          "type": "boolean",
          "default": true,
          "description": "Whether allow WAF to enforce file upload limits."
        },

karanbazaz: @Shawnli222 agree with what you are saying. What I am asking is that when are trying to put a test case which will set a value, we should try to set the non-default value since default value will anyways be set.

Shawnli222: oh I see, will make this change

src/Network/Network/FirewallPolicy/PolicySettings/AzureApplicationGatewayFirewallPolicySetting.cs
```diff
@@ -74,6 +86,16 @@ public override void ExecuteCmdlet()
                 this.State = "Enabled";
             }
 
+            if (!this.MyInvocation.BoundParameters.ContainsKey("RequestBodyEnforcement"))
+            {
+                this.RequestBodyEnforcement = true;
```
tejasshah7, line 91: What will happen if a user is doing a Get where this field is not set and then configures this and Request Body Check is disabled. Shouldn't this be False?

Same comment for below variable.

@karanbazaz thoughts?

src/Network/Network/help/New-AzApplicationGatewayFirewallPolicySetting.md
```diff
@@ -39,8 +39,60 @@ $condition = New-AzApplicationGatewayFirewallPolicySetting -State $enabledState
...
+Enable/Disable request body enforcement limits for WAF.
+
+```yaml
+Type: System.Boolean
+Parameter Sets: (All)
+Aliases:
+
+Required: False
+Position: Named
+Default value: None
```
tejasshah7, line 61: Is this really true? We are setting default as True. May be actually we SHOULD NOT set it and then this would be correct.

src/Network/Network/help/New-AzApplicationGatewayFirewallPolicySetting.md
```diff
@@ -39,8 +39,60 @@ $condition = New-AzApplicationGatewayFirewallPolicySetting -State $enabledState
...
+Enable/Disable file upload enforcement limits for WAF.
+
+```yaml
+Type: System.Boolean
+Parameter Sets: (All)
+Aliases:
+
+Required: False
+Position: Named
+Default value: None
```
tejasshah7, line 91: Is this really true? We are setting default as True. May be actually we SHOULD NOT set it and then this would be correct.

src/Network/Network/FirewallPolicy/PolicySettings/AzureApplicationGatewayFirewallPolicySetting.cs
```diff
@@ -74,6 +86,16 @@ public override void ExecuteCmdlet()
                 this.State = "Enabled";
             }
 
+            if (!this.MyInvocation.BoundParameters.ContainsKey("RequestBodyEnforcement"))
```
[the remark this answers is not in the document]
tejasshah7: Yeah I have same comment above. Lets discuss this offline.

src/Network/Network/FirewallPolicy/PolicySettings/AzureApplicationGatewayFirewallPolicySetting.cs
```diff
@@ -33,7 +33,15 @@ public class AzureApplicationGatewayFirewallPolicySetting : NetworkBaseCmdlet
         [ValidateSet("Disabled", "Enabled", IgnoreCase = true)]
         [ValidateNotNullOrEmpty]
         public string State { get; set; }
-        
+
+        [Parameter(Mandatory = false, HelpMessage = "Enable/Disable request body enforcement limits for WAF.")]
+        [ValidateNotNullOrEmpty]
+        public bool RequestBodyEnforcement { get; set; }
+
+        [Parameter(Mandatory = false, HelpMessage = "Max inspection limit in KB for request body inspection for WAF.")]
```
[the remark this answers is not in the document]
karanbazaz: @baoqihuang0326 I think the current comment is good. Inspection limit might need to be explained. We can just skip the "for WAF" at the end to maintain consistency with the MaxRequestBodySizeInKb HelpMessage

Shawnli222:
@msJinLei hi Jin Lei can you please approve and merge this when you come from vacation? Addressed all comments from reviewers.

src/Network/Network/FirewallPolicy/PolicySettings/AzureApplicationGatewayFirewallPolicySetting.cs
```diff
@@ -106,8 +128,11 @@ protected PSApplicationGatewayFirewallPolicySettings NewObject()
             {
                 Mode = this.Mode,
                 State = this.State,
+                RequestBodyEnforcement = this.DisableRequestBodyEnforcement != null && (this.DisableRequestBodyEnforcement == true ? false : true),
```
tejasshah7, line 131: Too complicated, lets move this out and set this field below.

Shawnli222: I think in this PSApplicationGatewayFirewallPolicySettings we cant have fields below since RequestBodyEnforcement is only inside PSApplicationGatewayFirewallPolicySettings

tejasshah7: You don't need to return right away. You can create a local object, do whatever you want and then return that object.

Shawnli222: move the condition outside of the return object

tejasshah7 (approved):

src/Network/Network/FirewallPolicy/PolicySettings/AzureApplicationGatewayFirewallPolicySetting.cs
```diff
@@ -33,7 +33,15 @@ public class AzureApplicationGatewayFirewallPolicySetting : NetworkBaseCmdlet
...
         public string State { get; set; }
-        
+
+        [Parameter(Mandatory = false, HelpMessage = "Disable request body enforcement limits for WAF.")]
+        [ValidateNotNullOrEmpty]
+        public bool? DisableRequestBodyEnforcement { get; set; }
+
+        [Parameter(Mandatory = false, HelpMessage = "Max inspection limit in KB for request body inspection.")]
+        [ValidateNotNullOrEmpty]
+        public int RequestBodyInspectLimitInKB { get; set; }
```
karanbazaz, line 43: this should also be int? like bool? since its a nullable field.

src/Network/Network/Models/PSApplicationGatewayFirewallPolicySettings.cs
```diff
@@ -28,12 +28,21 @@ public partial class PSApplicationGatewayFirewallPolicySettings
         [Ps1Xml(Target = ViewControl.Table)]
         public string Mode { get; set; }
 
+        [Ps1Xml(Target = ViewControl.Table)]
+        public bool? RequestBodyEnforcement { get; set; }
+
+        [Ps1Xml(Target = ViewControl.Table)]
+        public int RequestBodyInspectLimitInKB { get; set; }
```
karanbazaz, line 35: same here should be int? @tejasshah7 to confirm?

src/Network/Network/FirewallPolicy/PolicySettings/AzureApplicationGatewayFirewallPolicySetting.cs
```diff
@@ -89,6 +101,16 @@ public override void ExecuteCmdlet()
                 this.CustomBlockResponseStatusCode = (int?)null;
             }
 
+            if (!this.MyInvocation.BoundParameters.ContainsKey("DisableFileUploadEnforcement"))
+            {
+                this.DisableFileUploadEnforcement = (bool?)null;
+            }
+
+            if (!this.MyInvocation.BoundParameters.ContainsKey("DisableRequestBodyEnforcement"))
```
karanbazaz, line 109: Missing the same if condition for RequestBodyInspectLimitInKB.

src/Network/Network/help/New-AzApplicationGatewayFirewallPolicySetting.md
```diff
@@ -39,8 +39,60 @@ $condition = New-AzApplicationGatewayFirewallPolicySetting -State $enabledState
...
+Accept pipeline input: False
+Accept wildcard characters: False
+```
+
+### -RequestBodyInspectLimitInKB  
+Max inspection limit in KB for request body inspection.
+
+```yaml
+Type: System.Int32
```
karanbazaz, line 70: The type should be Type: System.Nullable`1[System.Int32] since the field is also nullable.

karanbazaz (approved):

msJinLei:
@Shawnli222 when the sdk ver 26.0.0 is going to be released?

Shawnli222:
@msJinLei I think it’s 5/12

baoqihuang0326 (approved):

tejasshah7 (approved):

msJinLei:
> @msJinLei I think it’s 5/12

Please remember to remove package from local feed after the SDK is released.

msJinLei (approved):

PR #21717, opened 2023-05-08
[Storage] Support setting cors rules in mgmt plane cmdlets

yifanz7:
## Checklist


src/Storage/Storage.Management/Models/PSBlobServiceProperties.cs
```diff
@@ -292,15 +295,37 @@ public CorsRule ParseCorsRule()
         /// <returns>String Array</returns>
         private static string[] ListToArray(IList<string> stringList)
         {
+            string[] stringArray = new string[stringList.Count];
```
blueww, line 298: It looks this line will throw null reference exception if stringList is null.

src/Storage/Storage.Management/Models/PSBlobServiceProperties.cs
```diff
@@ -292,15 +295,37 @@ public CorsRule ParseCorsRule()
...
+        {
+            List<string> methods = new List<string>();
+            foreach(string method in allowedMethods)
+            {
+                if (AllowedMethodList.Contains(method.ToUpper())) {
+                    methods.Add(method.ToUpper());
+                } 
+                else
+                {
+                    throw new InvalidOperationException(string.Format("{0} is an invalid HTTP method.", method));
```
blueww, line 323: I am doubt should we return error here. 
It might block the API when some new method added in the future.
When the method not exist, we might can just add the orignal value.

src/Storage/Storage.Management/help/Update-AzStorageBlobServiceProperty.md
```diff
@@ -103,6 +105,41 @@ IsVersioningEnabled           : True
 
 This command enables Versioning on Blob service of a Storage account
 
+### Example 4: Update CORS rules
```
blueww, line 108: Please also give the sample to clear the CORS rule.
To match the usage of each dataplane CORS rule cmdlets.

src/Storage/Storage.Management.Test/ScenarioTests/StorageBlobTests.ps1
```diff
@@ -594,6 +605,34 @@ function Test-StorageBlobServiceProperties
 		Assert-AreEqual $false $property.DeleteRetentionPolicy.Enabled
 		Assert-AreEqual $null $property.DeleteRetentionPolicy.Days
 
+		$property = Update-AzStorageBlobServiceProperty -ResourceGroupName $rgname -StorageAccountName $stoname -CorsRule $CorsRules
```
blueww, line 608: Please also add test to clear CORS rule, for both file/blob.

src/Storage/Storage.Management/Models/PSBlobServiceProperties.cs
yifanz7: I've added the check here just to make sure we can throw the exception at client side just as what dataplane cmdlets are doing. But it's true that whenever a new value is added in the future, the value will be added to the list here too.
Regarding "When the method not exist, we might can just add the original value.", I think in this case we just don't need to check if the method exists, but add the value to the list anyways

yifanz7: Thus I guess the decision here should be - do we want to align with the dataplane cmdlets to check for it on the client side? Or do we just want the server to throw the exception?

src/Storage/Storage.Management/Models/PSBlobServiceProperties.cs
yifanz7: Checked in the changes by mistake. Reverted it.

src/Storage/Storage.Management/help/Update-AzStorageBlobServiceProperty.md
yifanz7: Added

isra-fel (approved):

isra-fel (approved):

isra-fel (approved):

isra-fel (approved):

PR #21921, opened 2023-05-24
[Compute] Automapper fix for bug in Update-AzDisk and Update-AzSnapshot 

haagha:
## Description

Bug in Update-AzDisk and Update-AzSnapshot causes null Tags property to be mapped as empty dictionaries. When passed to the PATCH call, it results in clearing user's Tags.

This bug fix will add a condition for the mapping to only map when Tag's property is not null.

## Checklist

### DIFF
--- a/src/Compute/Compute/ChangeLog.md
+++ b/src/Compute/Compute/ChangeLog.md
@@ -26,6 +26,7 @@
 * Added `-NetworkAccessPolicy` parameter to `New-AzSnapshotUpdateConfig`.
 * Added `-SharedGalleryImageId` parameter to `New-AzVM`, `New-AzVmConfig`, `New-AzVmss`, `New-AzVmssConfig`, `Update-AzVmss`, and `Set-AzVmssStorageProfile`.
 * Updated `Set-AzVMDiagnosticsExtension` to correct capitalization by passing "StorageAccount" as configuration property instead of "storageAccount".
+* Added condition in Automapper configurations to check for null Tag values for `PSDiskUpdate` and `PSSnapshotUpdate` to fix bug in Update-AzDisk and Update-AzSnapshot.
 
 ## Version 5.5.0
 * Added breaking change message for `New-AzVmss`.


isra-fel:
+ @NoriZC + @wyunchi-ms  to review as well.
BTW @haagha please update the changelog.

wyunchi-ms:
Is there any other cmdlets act like this way?

src/Compute/Compute/Generated/Models/ComputeAutoMapperProfile.cs
```diff
@@ -100,12 +100,16 @@ private static void Initialize()
...
+                cfg.CreateMap<TO.PSDiskUpdate, FROM.DiskUpdate>()
+                    .ForMember(c => c.Tags, o => o.Condition(r => (r.Tags != null)));
                 cfg.CreateMap<FROM.Snapshot, FROM.SnapshotUpdate>();
                 cfg.CreateMap<FROM.SnapshotUpdate, FROM.Snapshot>();
-                cfg.CreateMap<FROM.SnapshotUpdate, TO.PSSnapshotUpdate>();
-                cfg.CreateMap<TO.PSSnapshotUpdate, FROM.SnapshotUpdate>();
+                cfg.CreateMap<FROM.SnapshotUpdate, TO.PSSnapshotUpdate>()
+                    .ForMember(c => c.Tags, o => o.Condition(r => (r.Tags != null)));
+                cfg.CreateMap<TO.PSSnapshotUpdate, FROM.SnapshotUpdate>()
+                    .ForMember(c => c.Tags, o => o.Condition(r => (r.Tags != null)));
```
Sandido, line 112: I'm a bit confused as to why this will fix the issue. The issue is that customers have Tag values that are being lost, so their Tags are filled and are being mapped to an empty tag. 
Now this condition checks if the customer Tag has values in it and maps it only if there are tag values? Isn't that the same behavior as before?

src/Compute/Compute/ChangeLog.md
```diff
@@ -26,6 +26,7 @@
 * Added `-NetworkAccessPolicy` parameter to `New-AzSnapshotUpdateConfig`.
 * Added `-SharedGalleryImageId` parameter to `New-AzVM`, `New-AzVmConfig`, `New-AzVmss`, `New-AzVmssConfig`, `Update-AzVmss`, and `Set-AzVmssStorageProfile`.
 * Updated `Set-AzVMDiagnosticsExtension` to correct capitalization by passing "StorageAccount" as configuration property instead of "storageAccount".
+* Added condition in Automapper configurations to check for null Tag values for `PSDiskUpdate` and `PSSnapshotUpdate` to fix bug in Update-AzDisk and Update-AzSnapshot.
```
Sandido, line 29: Add a test for this scenario.

Sandido (dismissed):

haagha:
@wyunchi-ms Thankfully we don't use Automapper for any other Update Objects. These are the only two.

src/Compute/Compute/Generated/Models/ComputeAutoMapperProfile.cs
haagha: @Sandido the issue is that in a PATCH call you only send items that need to be updated. If the tags are null, changing them to empty will be interpreted as Clearing out all tags. These conditions check to not map tags if they are null.

haagha:
Adding a few other changes requested by CXP before merging. Please hold on merge, Thanks.

src/Compute/Compute/Generated/Models/ComputeAutoMapperProfile.cs
vidai-msft: @haagha if the end users do want to clear the tags, would they fail to achieve that?

haagha: @vidai-msft The users can clear tags like normally before by passing an empty dictionary. An empty dictionary would not meet the condition.

vidai-msft: @haagha OK. got it. The expected behavior is empty dictionary will clear the tag while null should do nothing.

vidai-msft (approved):

vidai-msft:
@Sandido can you please approve this PR if you don't have any other question or concern?

vidai-msft:
Gently ping @Sandido

haagha:
@vidai-msft I've explained to @Sandido the behavior that will be because of this change. Please merge this.

PR #22007, opened 2023-06-06
Add dynamic message to user when they use soon to be deprecated Image Alias values

Sandido:
## Description

As per previous work and a future breaking change in New-AzVmss and New-AzVM, this is a message to tell users to stop using the soon to be deprecated image alias values (Ignite 2023). Instead use the newer versioned values as seen in the provided link in the message.

## Checklist


src/Compute/Compute/ChangeLog.md
```diff
@@ -22,6 +22,8 @@
 ## Upcoming Release
 * Added useful examples to the `New-AzVMConfig` help doc.
 * Added new `ResourceId` parameter to the `Get-AzVmss` cmdlet. 
+* Added a message to the user when they provid an outdated image alias to `New-AzVM` via the `-Image` parameter or to `New-AzVmss` via the `-ImageName` parameter.
```
vidai-msft, line 25: `provid` should be `provide`

src/Compute/Compute/Manual/VirtualMachineScaleSetCreateOrUpdateMethod.cs
```diff
@@ -274,6 +274,16 @@ public async Task<ResourceConfig<VirtualMachineScaleSet>> CreateConfigAsync()
 
             private async Task<ResourceConfig<VirtualMachineScaleSet>> SimpleParameterSetNormalMode()
             {
+                // Temporary message until after the Ignite 2023 release that should remove these outdated image aliases. 
+                if ((_cmdlet.ImageName != "")
```
vidai-msft, line 278: `(_cmdlet.ImageName != "")` is not necessary unless most of the `ImageName` is empty.

src/Compute/Compute/Manual/VirtualMachineScaleSetCreateOrUpdateMethod.cs
```diff
@@ -453,6 +463,16 @@ private async Task<ResourceConfig<VirtualMachineScaleSet>> SimpleParameterSetOrc
                 int platformFaultDomainCountFlexibleDefault = 1;
                 SwitchParameter singlePlacementGroupFlexibleDefault = false;
 
+                // Temporary message until after the Ignite 2023 release that should remove these outdated image aliases. 
+                if ((_cmdlet.ImageName != "")
```
vidai-msft, line 467: Same as above.

src/Compute/Compute/VirtualMachine/Operation/NewAzureVMCommand.cs
```diff
@@ -511,6 +511,16 @@ public async Task<ResourceConfig<VirtualMachine>> CreateConfigAsync()
 
                 if (_cmdlet.DiskFile == null)
                 {
+                    // Temporary message until after the Ignite 2023 release that should remove these outdated image aliases. 
+                    if ((_cmdlet.Image != "")
```
vidai-msft, line 515: Same as above.

vidai-msft (changes requested):

vidai-msft:
@Sandido please make it a non-draft PR when it is ready for review. Thanks!

src/Compute/Compute/Manual/VirtualMachineScaleSetCreateOrUpdateMethod.cs
Sandido: It is possible that the ImageName is empty because it is an optional parameter. This check should be here I think.

vidai-msft: @Sandido if the `ImageName` is empty, then it won't equal to "CentOS" | "Debian" | "RHEL" | "UbuntuLTS". So if most of the cmdlets would receive an empty string for this parameter, then it is fair enough to verify it at the beginning to save time. Otherwise, it is not that necessary.

vidai-msft: @Sandido any update?

vidai-msft: @Sandido What do you think of this comment?

Sandido: @vidai-msft , In the past 90 days in kusto, 
about 30% of New-AzVM calls that can have -Image (SimpleParameterSet) do not use it so it would be empty.
About 68% of New-AzVmss calls that can have -ImageName do not use it so it would be empty.

vidai-msft (approved):

PR #22208, opened 2023-06-30
Removing length validation for DNS TXT record to ensure consistency

avadhanulalavanya:
## Description

This PR contains removing length validation for DNS TXT records. 

Context : Total length applies at record set level. We currently have length validation in powershell only when a single TXT record is added to a record set, while we dont have any validation while adding multiple records within record set.

Current Error returned - New-AzDnsRecordConfig : "Cannot validate argument on parameter 'Value'. The character length of the 1979 argument is too long. Shorten the character length of the argument so it is fewer than or equal to "1024" characters, and then try the command again."
Error returned in all other scenarios is the error returned directly from DNS API , which is different from above error message. We also do not have length validations on CLI and Azure portal.

Removing validation for single TXT record in Powershell to make it consistent with portal/CLI and to make error returned consistent.

Changes(For both public and private DNS):
1.Removed length validation
2.Removed Min and Max TXTRecordLength variables
3.Removed Scenario tests for length validation on Powershell.

## Checklist

### DIFF
--- a/src/PrivateDns/PrivateDns/ChangeLog.md
+++ b/src/PrivateDns/PrivateDns/ChangeLog.md
@@ -18,7 +18,7 @@
         - Additional information about change #1
 -->
 ## Upcoming Release
-* Removing length validation for DNS TXT record to make it consistent with Azure CLI and Azure portal.
+* Removed length validation for DNS TXT record to make it consistent with Azure CLI and Azure portal.
 
 ## Version 1.0.3
 * Corrected verbose output string formatting for Remove-AzPrivateDnsRecordSet


src/Dns/Dns/ChangeLog.md
```diff
@@ -18,6 +18,7 @@
         - Additional information about change #1
 -->
 ## Upcoming Release
+* Removing length validation for DNS TXT record to make it consistent with Azure CLI and Azure portal.
```
vidai-msft, line 21:
```suggestion
* Removed length validation for DNS TXT record to make it consistent with Azure CLI and Azure portal.
```

src/PrivateDns/PrivateDns/ChangeLog.md
```diff
@@ -18,6 +18,7 @@
         - Additional information about change #1
 -->
 ## Upcoming Release
+* Removing length validation for DNS TXT record to make it consistent with Azure CLI and Azure portal.
```
vidai-msft, line 21:
```suggestion
* Removed length validation for DNS TXT record to make it consistent with Azure CLI and Azure portal.
```

vidai-msft (changes requested):

vidai-msft (approved):
LGTM

Issue #22453, opened 2023-08-01
Invoke-AzContainerInstanceCommand -PassThru returns nothing when called from ARM deployment script


andreygoran:
### Description

Invoke-AzContainerInstanceCommand -PassThru is supposed to return the output of the remotely executed command, but this does not happen when it's called from an ARM deployment script. NULL is returned instead of the command's output.

### Steps to reproduce

1. Create resource group
1. Create managed identity and assign it access to resource group
1. Create container instance in resource group
1. Run ARM deployment script that executes a command in container instance using managed identity

Save ARM template file to template.json:
```
{
  "$schema": "https://schema.management.azure.com/schemas/2019-08-01/deploymentTemplate.json#",
  "contentVersion": "1.0.0.0",
  "parameters": {
    "identity": {
      "type": "string"
    },
    "utcValue": {
      "type": "string",
      "defaultValue": "[utcNow()]"
    }
  },
  "resources": [
    {
      "type": "Microsoft.Resources/deploymentScripts",
      "apiVersion": "2020-10-01",
      "name": "script",
      "location": "[resourceGroup().location]",
      "kind": "AzurePowerShell",
      "identity": {
        "type": "userAssigned",
        "userAssignedIdentities": {
          "[parameters('identity')]": {}
        }
      },
      "properties": {
            "forceUpdateTag": "[parameters('utcValue')]",
            "scriptContent":"
                $DebugPreference='Continue'

                Connect-AzAccount -Identity

                $output_name = (Get-AzContainerGroup -Name myapp -ResourceGroupName MyResourceGroup).Name
                $output_invokecommand = (Invoke-AzContainerInstanceCommand -ContainerGroupName myapp -ContainerName myapp -ResourceGroupName MyResourceGroup -PassThru -Command hostname)

                $DeploymentScriptOutputs = @{}
                $DeploymentScriptOutputs['psversion'] = $PSVersionTable
                $DeploymentScriptOutputs['azmodule'] = (Get-Module Az*)
                $DeploymentScriptOutputs['name'] = [string]::IsNullOrEmpty($output_name) ? 'None' : $output_name
                $DeploymentScriptOutputs['invokecommand'] = [string]::IsNullOrEmpty($output_invokecommand) ? 'None' : $output_invokecommand
            ",
            "azPowerShellVersion":"10.1.0",
            "retentionInterval": "P1D"
      }
    }
  ],
  "outputs": {
    "psversion": {
[...]

Issue #19247, opened 2022-08-16
`Invoke-AzContainerInstanceCommand` doesn't wait for executed command completion in Azure DevOps pipeline Azure PowerShell task


andysbolton:
### Description

When I run this command in an Azure Devops pipeline with an ubuntu-20.04 image and the [Azure Powershell Task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/deploy/azure-powershell?view=azure-devops), I notice that it 
it returns immediately and continues running the  pipeline script. If I run locally, it blocks for 10 seconds before returning.
``` 
Invoke-AzContainerInstanceCommand `
        -ContainerGroupName $Name `
        -ContainerName $Name `
        -ResourceGroupName $ResourceGroup `
        -Command "sleep 10"
```
Because of this, I'm not able to determine in the pipeline whether the executed command has succeeded or failed.

(FYI I also created an issue in the azure-pipelines-task repo: https://github.com/microsoft/azure-pipelines-tasks/issues/16590)

I'm not sure if this is because the pipeline task is run in NonInteractive mode (although the above command works in NonInteractive mode locally).

If I run execute a command that writes to standard out and then put in a `Start-Sleep` command in the pipeline task, I receive all of the standard out, so it seems the websockets are working correctly, but it is not waiting for the executed command to finish.

Should this command be expected to block in the pipeline or is there some limitation in the pipeline/task or the PowerShell library that is preventing that from happening?

### Issue script & Debug output

```PowerShell
(Apologies for the garbled output, it's due to the pipeline obfuscating secrets)

$DebugPreference='Continue'
Invoke-AzContainerInstanceCommand `
        -ContainerGroupName $Name `
        -ContainerName $Name `
        -ResourceGroupName $ResourceGroup `
        -Command "sleep 10"

2022-08-16T15:33:18.2963070Z [***;1mDEBUG: 15:***:18 - GetAzureRMContextCommand end processing.[***m
2022-08-16T15:33:18.3262710Z [***;1mDEBUG: [CmdletBeginProcessing]: Starting command[***m
2022-08-16T15:33:18.3314054Z [***;1mDEBUG: CmdletBeginProcessing: [***m
2022-08-16T15:33:18.3316538Z [***;1mDEBUG: CmdletProcessRecordStart: [***m
2022-08-16T15:33:18.3318778Z [***;1mDEBUG: CmdletGetPipeline: [***m
[...]

Issue #15754, opened 2021-08-26
Invoke-AzContainerInstanceCommand output


janegilring:
## Description

Scenario: Execute a command within an Azure Container Instances container.

By using Azure CLI, this works as expected:

```
PS /> az container exec --resource-group test-rg --name test-aci-01 --container-name container-01 --exec-command "date"
Thu Aug 26 04:14:18 UTC 2021
```

When trying to accomplish the same using Azure PowerShell, the output is not the command output - but rather a password and a websocket URI as shown below. Could it be that the Azure CLI uses these credentials to actually gather the command output, but the PowerShell command does not? If so, it would be useful if this behaviour could be change similarly to Azure CLI.

Also, it is unclear why we in the PowerShell command need to specify -TerminalSizeCol and -TerminalSizeRow, while in the Azure CLI option we do not.

## Steps to reproduce

```powershell
PS /> Invoke-AzContainerInstanceCommand -ResourceGroupName test-rg -ContainerGroupName test-aci-01 -ContainerName container-01 -Command "date" -TerminalSizeCol 12 -TerminalSizeRow 12

Password                         WebSocketUri
--------                         ------------
random-string wss://abc123.norwayeast.atlas.cloudapp.azure.com:19390/execstream/sessionId/80035cbc-85e8-4fec-a0af-
```

## Environment data
```
PS /> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      7.1.3
PSEdition                      Core
GitCommitId                    7.1.3
OS                             Linux 5.10.16.3-microsoft-standard-WSL2 #1 SMP Fri Apr 2 22:23:49 UTC 2021
Platform                       Unix
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

## Module versions

```powershell
ModuleType Version    PreRelease Name                                ExportedCommands
---------- -------    ---------- ----                                ----------------
Script     2.5.2                 Az.Accounts                         {Add-AzEnvironment, Clear-AzContext, Clear-AzDefault, Connect-AzAccount…}
Script     2.1.0                 Az.ContainerInstance                {Add-AzContainerInstanceOutput, Get-AzContainerGroup, Get-AzContainerInstan…
```

## Debug output
```
DEBUG: 06:20:55 - GetAzureRMContextCommand end processing.
[...]

PR #22515, opened 2023-08-07
Fixed websocket related tasks in InvokeAzContainerInstanceCommand_ExecuteExpanded.cs

andreygoran:
## Description

This PR fixes the problem with Invoke-AzContainerInstanceCommand not returning any result when it's run under certain conditions, and closes https://github.com/Azure/azure-powershell/issues/22453

The problem occurres when powershell is launched with redirection of input stream (and apparently this is how it's being run in Azure ARM Powershell Deployment Scripts and in Azure Powershell Functions).

`Test.ps1` code used to reproduce the problem:

```powershell
$HostName = Invoke-AzContainerInstanceCommand -ContainerGroupName mypowershellapp -ContainerName mypowershellapp -ResourceGroupName MyResourceGroup -PassThru -Command 'hostname'
("Host: " + $HostName) | Out-Host
```

Results before the fix:
```
$ pwsh Test.ps1
SandboxHost-638266427706169802
Host: SandboxHost-638266427706169802
$ echo | pwsh Test.ps1
Host: 
```

Results after the fix:
```
$ pwsh Test.ps1
SandboxHost-638266427706169802
Host: SandboxHost-638266427706169802
$ echo | pwsh Test.ps1
SandboxHost-638266427706169802
Host: SandboxHost-638266427706169802
```

### Related Issues
Fixed: https://github.com/Azure/azure-powershell/issues/22453
Probably related: https://github.com/Azure/azure-powershell/issues/19247
Probably related: https://github.com/Azure/azure-powershell/issues/15754#issuecomment-1097655137

### Root Cause Analysis

The root cause of the bug was using a `return Task.Factory.StartNew(async () => {...})` construction to handle working with websockets doing actual communication with container instance. This code pattern is dangerous and should not be used, see for details [https://blog.stephencleary.com/2013/08/startnew-is-dangerous.html](https://blog.stephencleary.com/2013/08/startnew-is-dangerous.html)

Another problem was using `Console.KeyAvailable` which throws an exception when input redirection is used. This was fixed by adding `Console.IsInputRedirected` validation.

## Checklist

### DIFF
--- a/src/ContainerInstance/custom/InvokeAzContainerInstanceCommand_ExecuteExpanded.cs
+++ b/src/ContainerInstance/custom/InvokeAzContainerInstanceCommand_ExecuteExpanded.cs
@@ -48,7 +48,7 @@ namespace Microsoft.Azure.PowerShell.Cmdlets.ContainerInstance.Cmdlets
 
         private Task PullResponse()
         {
-            return Task.Run(async () =>
+            return Task.Factory.StartNew(async () =>
             {
                 string result = string.Empty;
                 var allBytes = new List<byte>();
@@ -71,12 +71,12 @@ namespace Microsoft.Azure.PowerShell.Cmdlets.ContainerInstance.Cmdlets
 
                 if (!string.IsNullOrEmpty(result) && PassThru.ToBool()) { WriteObject(result.TrimEnd()); }
 
-            }, this._cancellationTokenSource.Token);
+            }, this._cancellationTokenSource.Token).Unwrap();
         }
 
         private Task PushCommand()
         {
-            return Task.Run(async () =>
+            return Task.Factory.StartNew(async () =>
             {
                 if (Console.IsInputRedirected) return;
                 
@@ -97,7 +97,7 @@ namespace Microsoft.Azure.PowerShell.Cmdlets.ContainerInstance.Cmdlets
                     System.Threading.Thread.Sleep(250); 
                 }
 
-            }, this._cancellationTokenSource.Token);
+            }, this._cancellationTokenSource.Token).Unwrap();
 
         }
     }


src/ContainerInstance/custom/InvokeAzContainerInstanceCommand_ExecuteExpanded.cs
```diff
@@ -48,7 +48,7 @@ public partial class InvokeAzContainerInstanceCommand_ExecuteExpanded : global::
 
         private Task PullResponse()
         {
-            return Task.Factory.StartNew(async () =>
+            return Task.Run(async () =>
```
BethanyZhou, line 51: https://code-maze.com/csharp-task-run-vs-task-factory-startnew/
Task.Factory.StartNew() vs Task.Run()
1. TaskCreationOptions.AttachedToParent vs TaskCreationOptions.DenyChildAttach? TaskCreationOptions.AttachedToParent as WriteObject() in PullResponse() depends on parent task.
2. Current vs Default TaskScheduler? Default TaskScheduler. No sure is current TaskScheduler same with default scheduler in ARM deployment environment?
3. Task<Task<string>> vs Task<string>?  No difference in this case due to no return.
4.  Object State? No difference in this case as no shared variable is scoped outside of the task block except socket.

BethanyZhou: Maybe should use `Task.Factory.StartNew(action,  this._cancellationTokenSource.Token, TaskCreationOptions.None, TaskScheduler.Default);`? @andreygoran what's your thought?

andreygoran: @BethanyZhou,

3. `Task<Task<void>>` vs `Task<void>` I think it's actually crucial in this case. We use `Task.WaitAll` later in code https://github.com/Azure/azure-powershell/blob/d6e1dc5e5759ae025a6509886ac5d7580113f5a6/src/ContainerInstance/custom/InvokeAzContainerInstanceCommand_ExecuteExpanded.cs#L45C13-L45C25 and `Task<Task<void>>` means that the waiting is over when the inner `Task` is returned, but this does not mean that the inner `Task` is actually completed by that time. However `Task.Run` does understand the semantics of `Task<Task<void>>` and actually waits until the inner task is completed.
Another solution would be to use `.Unwrap()` instead of switching `Task.Factory.StartNew` to `Task.Run`:
```
            return Task.Factory.StartNew(async () =>
            {
...
            }, this._cancellationTokenSource.Token).Unwrap();
```

> Maybe should use `Task.Factory.StartNew/Task.Run(action, this._cancellationTokenSource.Token, TaskCreationOptions.None, TaskScheduler.Default);`?

I randomly checked a few usages of `Task.Run` in the project, and I do not see any additional parameters used anywhere. So probably it's fine the way it is.

BethanyZhou: Thanks for your opinions. 

I prefer to  
```
return Task.Factory.StartNew(async () =>
            {
...
            }, this._cancellationTokenSource.Token).Unwrap();
```

Because we need TaskCreationOptions.AttachedToParent but Task.Run() uses TaskCreationOptions.DenyChildAttach

andreygoran: @BethanyZhou I added new commit that restores Task.Factory.StartNew and adds Unwrap()

BethanyZhou (approved):
LGTM

PR #22534, opened 2023-08-09
Add SecurityType of Standard to VM and VMSS cmdlets for Trusted Launch, includes breaking changes

Sandido:
## Description
https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/1270 

This PR is preliminary work for the Trusted Launch feature where VMs need to have a way for customers to opt out of the TL feature once TL is the default in the near future. 
The plan is to allow the user to supply the value Standard to SecurityType. This value does not actually exist on the backend RP or in ARM yet. The PS and CLI clients will take in Standard and just ensure that the SecurityType is not defaulted to TL. 

This PR includes breaking changes and suppressions for them. The 2 breaking changes are the same type in 2 different cmdlets. 
1. Set-AzVMSecurityProfile
2. Set-AzVmssSecurityProfile
Where the change was adding a ValidateSet parameter group to the SecurityType parameter. I think this change should be allowed as the SecurityType parameter has always been restricted on the backend to only ConfidentialVM and TrustedLaunch. In the future, this will be expanded to Standard, hence my work here. 
I don't think this is a breaking change in practice since these values were always restricted to certain strings. 
If this is actually a breaking change that could harm our customers, please let me know and I will revert this. 
UPDATE: I removed these validations for now.

## Checklist


src/Compute/Compute/Generated/VirtualMachineScaleSet/Config/SetAzureRmVmssSecurityProfileCommand.cs
```diff
@@ -47,7 +48,8 @@ public partial class SetAzureRmVmssSecurityProfileCommand : Microsoft.Azure.Comm
             Mandatory = false,
             Position = 1,
             ValueFromPipelineByPropertyName = true)]
-        [PSArgumentCompleter("TrustedLaunch", "ConfidentialVM")]
+        [ValidateSet(ValidateSetValues.TrustedLaunch, ValidateSetValues.ConfidentialVM, ValidateSetValues.Standard, IgnoreCase = true)]
```
YanaXu, line 51: It's a breaking change. Can you hold this PR to next major release?

Sandido: @YanaXu , I'll just remove this and save it for later.

YanaXu (approved):

PR #22910, opened 2023-10-06
[Network] - Add DefaultOutboundAccess parameter on subnet

phrazfipho:
## Description

- Customers can create a new subnet that disables any default outbound access 
- WI: https://msazure.visualstudio.com/One/_workitems/edit/25007155/

## Checklist


VeryEarly (dismissed):

src/Network/Network/VirtualNetwork/Subnet/AzureVirtualNetworkSubnetConfigBase.cs
```diff
@@ -126,5 +126,11 @@ public class AzureVirtualNetworkSubnetConfigBase : NetworkBaseCmdlet
             ValueFromPipelineByPropertyName = true,
             HelpMessage = "IpAllocation")]
         public PSIpAllocation[] IpAllocation { get; set; }
+
+        [Parameter(
+            Mandatory = false,
+            ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Default outbound connectivity for all VMs in the subnet")]
+        public bool? DefaultOutboundAccess { get; set; }
```
VeryEarly, line 134: does this value has 3 states? null/true/false, please consider use switch parameter if it only represent 2 states.

phrazfipho: it has 3 states. nullable boolean

VeryEarly: For PUT API, this value could be either with or without, can you explain why it need null value?

phrazfipho: In NRP server side it is nullable, so in PS it should be nullable too.
(It is default to true in server side if no value provided from client, so if here it is non-nullable boolean, it will pass false as default to the server side)

VeryEarly (approved):

PR #23017, opened 2023-10-18
Move Azure Policy CRUD cmdlets to autorest.powershell generation

mentat9:
## Description
This PR is to reimplement the Azure Policy CRUD cmdlets using autorest.powershell generation.

## Checklist

- [X] Check this box to confirm: **I have read the [_Submitting Changes_](../blob/main/CONTRIBUTING.md#submitting-changes) section of [`CONTRIBUTING.md`](../blob/main/CONTRIBUTING.md) and reviewed the following information:**

### DIFF
--- a/src/DataMigration/DataMigration.Autorest/custom/Cmdlets/New-AzDataMigrationSqlServerSchema.ps1
+++ b/src/DataMigration/DataMigration.Autorest/custom/Cmdlets/New-AzDataMigrationSqlServerSchema.ps1
@@ -87,22 +87,59 @@ function New-AzDataMigrationSqlServerSchema
             #Testing Whether Console App is downloaded or not
             $TestExePath =  Test-Path -Path $ExePath;
 
-            #Downloading and extracting SchemaMigration Zip file
+            #Console app download address
+            $ZipSource = "https://migrationapps.blob.core.windows.net/schemamigration/SqlSchemaMigration.zip";
+            $ZipDestination = Join-Path -Path $BaseFolder -ChildPath "SqlSchemaMigration.zip";
+            
             if(-Not $TestExePath)
             {
-                $ZipSource = "https://migrationapps.blob.core.windows.net/schemamigration/SqlSchemaMigration.zip";
-                $ZipDestination = Join-Path -Path $BaseFolder -ChildPath "SqlSchemaMigration.zip";
+                Write-Host "Downloading and extracting latest SchemaMigration Zip file..."
+                
+                #Downloading and extracting SchemaMigration Zip file
                 Invoke-RestMethod -Uri $ZipSource -OutFile $ZipDestination;
-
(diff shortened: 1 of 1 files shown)


mentat9:
This PR is the initial submission of CRUD cmdlets generated by autorest from Azure Policy OpenAPI specs. These cover all of the Azure Policy artifacts/resource types:

- Microsoft.Authorization/policyDefinitions
- Microsoft.Authorization/policySetDefinitions
- Microsoft.Authorization/policyAssignments
- Microsoft.Authorization/policyExemptions

Notes on this PR:
1. The existing cmdlets being replaced here were written in custom C# years ago and have many special features with high customer adoption.
2. Those existing cmdlets are packaged in the Az.Resources module along with many cmdlets owned by other teams and targeting other namespaces. These new cmdlets need to ship in their own module allowing customers to run both sets of cmdlets side-by-side (to help ease the task of moving to the new module), while also allowing us to ship updates to the new cmdlets independently of everything else currently bundled in the Az.Resources module.
3. Moving to autorest required heavy customization to minimize customer impact from required breaking changes.
4. The new cmdlets currently offer a -BackwardCompatible switch that turns on more complete matching of the behavior of existing cmdlets. This switch may be deprecated in the future after customers have had time to update their code to accommodate the breaking changes.
5. All previous tests were ported to Pester with almost entirely unchanged behavior. All tests are passing with a small number of exceptions (5 tests are set to -Skip). These tests are all in the <root>\test\Backcompat folder. New tests unrelated to backward compatibility will be added under <root>\test.
6. There is work remaining before these cmdlets can ship. I will be working on this in parallel with PR review and applying feedback: new tests, docs, examples.

Details on the 5 skipped tests:
1. Two tests are skipped because autorest generated cmdlets running under Pester don't support dynamic parameters. These two tests actually work when the code is run interactively in a Powershll shell. It's possible these tests will work in Pester 5.x, but the repo is currently on 4.10.1.
2. One test of a piping scenario is skipped because the generated cmdlet crashes due to what appears to be a generation bug. That test is duplicated in the Repro.Tests.ps1 file (also skipped so tests aren't broken). If anyone on the Powershell team wants to investigate, please use the Repro.Tests.ps1 file for that.
3. One other test is skipped because it isn't implemented yet.

VeryEarly:
Hi @mentat9 ,

You need to raise another PR to delete related cmdlet(tests, models) from main branch

mentat9:
> Hi @mentat9 ,
> 
> You need to raise another PR to delete related cmdlet(tests, models) from main branch

OK. But I'm a little concerned with removing the old cmdlets entirely. What if we need to release a maintenance fix on the old cmdlets?

mentat9:
Here's the PR for removing the existing cmdlets from Az.Resources module: https://github.com/Azure/azure-powershell/pull/23031.

PR #22974, opened 2023-10-13
Make TrustedLaunch default in for VM and VMSS creation

Sandido:
## Description
I still have a lot of cleaning up to do, and the new default value requires me to re-record many PS tests, but the functionality is all there. 
When no SecurityType is provided, TrustedLaunch and its requirements will be defaulted in. 
If no Image is provided, the 2022AzureEdition image, compatible with TL, will be defaulted in. 
Resources made from a SharedGalleryImage are not considered for this feature. 
Resources made from an Image that is Gen1 (HyperVGeneration V1) are not considered for this feature. 
PM for this feature is ajay.kundnani@microsoft.com.

I have had issues testing the managed disk scenario (3) as the request always times out (VMAgentStatus and os health is never good), but it seems to probably be working. 

This feature consists of 4 different feature requests:

1. https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/1240 
For Vm and Vmss creation, with minimal inputs, the default will be SecurityType = TrustedLaunch, with UEFI settings also set to true for TL if the user had not manually set the securitytype.

2. https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/1248
For a new vm disk created from an Image, I query for that image and determine if it is HyperVGeneration of V2. If yes, then default to trusted launch turned on if the user had not manually set the securitytype.

3. https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/1243
For VMs created from a managed disk for its os disk, it checks to see if the disk supports trusted launch, and if yes it will default the VM to trusted launch if the related values were not manually set by the user. 

4. https://github.com/Azure/azure-powershell-cmdlet-review-pr/issues/1241
This impacts the Config parameter sets for VM and Vmss creation because it requires the user to set the ImageReference values, specifically not the Id as marketplace images like 2022-datacenter-azure-edition do not have an Id. This defaulting only occurs if the user did not manually set the TL related values. 

As this feature changes the default behavior for VM and Vmss creation when SecurityType is null (which is typical) then I had to update many tests to set SecurityType to "Standard". This is seen in the client logic and prevents TrustedLaunch defaulting logic, but since the API does not support Standard it is nulled out before the .Net SDK call is made to create the VM or VMSS. 

I ended up re-recording many tests without this Standard value to ensure previous behavior with this new defaulting is not broken, hence the many new json files. 

## Checklist

[...]

PR #23338, opened 2023-11-09
Remove GuestAttestation logic

Sandido:
## Description

Trusted Launch is now turned on by default: https://github.com/Azure/azure-powershell/pull/22974 
As part of this, the GuestAttestation extension is installed on the VM and Vmss by default and sets SystemAssignedIdentity to True. This feature specifically only occurred when the user explicitly set TrustedLaunch. 
The GuestAttestation extension was removed from other clients due to perf concerns. It adds to our perf by increasing it by about 10%. Considering this needs to be done in a breaking change release, we would like to remove it now. 

This PR also reverts logic that was [added ](https://github.com/Azure/azure-powershell/pull/23297)into the Vmss creation code to set SystemAssignedIdentity from True back to False if the vmss is a Flexible orchestration mode since SystemAssigned is not allowed and will cause the Vmss creation to fail (by default since Flex is also now turned on by default. This hotfix logic is not needed if the entire GuestAttestation logic is removed.

While this is an unannounced breaking change, 
1. the perf impact to our customers, 
2. as well as interfering with the new defaults of Vmss Flex, 
3. and to bring us in parity with other clients where this was removed, 
is worth the break. 

## Checklist


grizzlytheodore:
let's add a test creating a VMSS defaulting in both TL and Flex, that was failing before, should pass here. 

Also this seems to contain reverting https://github.com/Azure/azure-powershell/pull/23297

grizzlytheodore:
LGTM thank you adam

PR #23486, opened 2023-11-20
Update sync-aliases CI to fix issue caused by PowerShell upgrade

vidai-msft:
## Description

Fix the PowerShell to 7.4.x and .NET version to 8.0.x.

The reason I chose to use 7.4 is that we don't need to update the version for 2 years.

## Checklist

### DIFF
--- a/.azure-pipelines/sync-aliases.yml
+++ b/.azure-pipelines/sync-aliases.yml
@@ -13,27 +13,24 @@ jobs:
 
   steps:
   - task: UseDotNet@2
-    displayName: Install .NET 7 SDK
+    displayName: Install .NET 8 SDK
     inputs:
       packageType: sdk
-      version: 7.0.x
+      version: 8.0.x
   - template: util/get-github-pat-steps.yml
 
   - pwsh: |
       dotnet --version
       dotnet new tool-manifest --force
-      dotnet tool install powershell
-    displayName: Install Latest PowerShell (7.3+)
+      dotnet tool install powershell --version 7.4.*
+    displayName: Install PowerShell 7.4.x
 
-  - task: AzurePowerShell@5
-    inputs:
-      azureSubscription: 'Azure SDK Infrastructure'
-      ScriptType: 'InlineScript'
-      Inline: |
-        $ADOTokenValue = Get-AzKeyVaultSecret -VaultName ${env:KEYVAULTNAME} -Name ${env:ADOTOKENNAME} -AsPlainText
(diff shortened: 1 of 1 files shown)


NoriZC:
@vidai-msft Could you explain more about why we don't need to update the version for 2 years? Is it because powershell will not have new version?

vidai-msft:
> @vidai-msft Vincent Dai FTE Could you explain more about why we don't need to update the version for 2 years? Is it because powershell will not have new version?

@NoriZC Due to PowerShell support lifecycle document, PowerShell 7.4 will be EOL on Nov 2025 while PowerShell 7.3 will be EOL on May 2024. So If we choose 7.3, then we will update the CI again in 6 months. That is the reason why I chose 7.4. Hope that makes sense.

NoriZC:
> > @vidai-msft Vincent Dai FTE Could you explain more about why we don't need to update the version for 2 years? Is it because powershell will not have new version?
> 
> @NoriZC Due to PowerShell support lifecycle document, PowerShell 7.4 will be EOL on Nov 2025 while PowerShell 7.3 will be EOL on May 2024. So If we choose 7.3, then we will update the CI again in 6 months. That is the reason why I chose 7.4. Hope that makes sense.

I understand that 7.4 would be better than 7.3. But the original code is using latest version. Why are we hard coding it? Is it because `dotnet tool install powershell` will not use the latest version by default any more?

vidai-msft:
> > > @vidai-msft
> > >             
> > >             Vincent Dai
> > >             FTE Vincent Dai FTE Could you explain more about why we don't need to update the version for 2 years? Is it because powershell will not have new version?
> > 
> > 
> > @NoriZC
> >             
> >             Nori Zhang
> >             FTE Due to PowerShell support lifecycle document, PowerShell 7.4 will be EOL on Nov 2025 while PowerShell 7.3 will be EOL on May 2024. So If we choose 7.3, then we will update the CI again in 6 months. That is the reason why I chose 7.4. Hope that makes sense.
> 
> I understand that 7.4 would be better than 7.3. But the original code is using latest version. Why are we hard coding it? Is it because `dotnet tool install powershell` will not intsall the latest version by default any more?

Because the "latest" is too changeable and in addition, PowerShell version is built based on a specific .NET version. Since we have to pin the version of .NET, then we must pin the version of PowerShell as well.

NoriZC (approved):
Hopefully this hard code could help save efforts and last long

PR #23624, opened 2023-11-30
Removing ispresent from vm and vmss creation cmdlets

Sandido:
## Description

Switch parameters have historically been used assuming that if they are present, they must be true. If they are not present, the .IsPresent built in method will return false and the parameter is null. 
However, the switch parameter can be false if the user passes it in like this, -Param:$false. 
Additionally, the code sometimes checks if the switch parameter .IsPresent and passes that value directly to the API call. This PR changes this functionality to instead check if the Switch param is true then passes true, if not then passes in null. 
This may break some API calls depending on what the API expects (passing in null instead of false) so testing is needed to verify if the API can accept that. Might go back to passing in false instead. This is being tested because some features are behind feature flags, so passing in null will avoid errors about trying to pass a value to a feature and property the user does not have access to (causing a hard error).  

## Checklist

### DIFF
--- a/src/Compute/Compute.Test/ScenarioTests/VirtualMachineScaleSetTests.ps1
+++ b/src/Compute/Compute.Test/ScenarioTests/VirtualMachineScaleSetTests.ps1
@@ -5298,8 +5298,8 @@ function Test-VirtualMachineScaleSetIsPresentRemovedTests
     try
     {
         # Common
-        #$rgname = "adsandtest14";
-        #$loc = "eastus";
+        $rgname = "adsandtest16";
+        $loc = "eastus";
         New-AzResourceGroup -Name $rgname -Location $loc -Force;
 
         # New VMSS Parameters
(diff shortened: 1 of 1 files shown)


src/Compute/Compute/Manual/VirtualMachineScaleSetCreateOrUpdateMethod.cs
```diff
@@ -569,7 +569,7 @@ private async Task<ResourceConfig<VirtualMachineScaleSet>> SimpleParameterSetOrc
                     instanceCount: _cmdlet.InstanceCount,
                     dataDisks: _cmdlet.DataDiskSizeInGb,
                     zones: _cmdlet.Zone,
-                    ultraSSDEnabled: _cmdlet.EnableUltraSSD.IsPresent,
+                    ultraSSDEnabled: (_cmdlet.EnableUltraSSD == true) ? true : (bool?)null,
```
Sandido, line 572: Can New-AzVmss be used to update a vmss? If so, might want to allow users to update the value to false.

Sandido, line 572: Maybe just pass the value? If the user doesn't have access to that feature, then they shouldn't be able to access the feature related parameter, right? Check that. If that is true, then can just pass in the value as opposed to explicitly null. 
Wait, no the switch param defaults to False if not set, right? So cannot pass that into the property without worry for feature flag restriction.

haagha:
I think we should use SwitchParameter.ToBool() to get the correct boolean value for the SwitchParameter.

PR #23973, opened 2024-01-17
[SqlLoginMigration] Upgrade console app automatically

Bhavikshah123:
Description
Background:
Currently, SQL Login migration PS backend is uses SqlLoginMigration console app, which is stored in storage account blob. When customers run the PS first time, the SqlLoginMigration console app will be downloaded from the storage account. But if newer console app is released (uploaded to the storage account blob), it cannot be downloaded automatically when customers run login migration PS. If customers want to use the latest console app, they have to delete the downloaded one manually.

Fix:
Upload a version file that keeps the same version number to the latest version of released console app to the storage account blob. Every time, release a new version console app, update the version number in the file to make it same to the released console app. When running the PS, if the console app already exists, check the Login.Console.exe version first and compare it to the version file. If existing one is old, remove it and download the latest one from the storage account blob.

Local testing

image

### DIFF
--- a/src/DataMigration/DataMigration.Autorest/custom/Cmdlets/New-AzDataMigrationLoginsMigration.ps1
+++ b/src/DataMigration/DataMigration.Autorest/custom/Cmdlets/New-AzDataMigrationLoginsMigration.ps1
@@ -91,14 +91,58 @@ function New-AzDataMigrationLoginsMigration
             #Testing Whether Console App is downloaded or not
             $TestExePath =  Test-Path -Path $ExePath;
 
+            #Console app download address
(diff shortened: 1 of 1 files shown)


siyao-Siyang:
You need to check out the generation branch and work on that branch.

siyao-Siyang:
The changes look good to me. Please follow the wiki I shared.

wyunchi-ms:
Please follow siyao's guide. Changes should be commited into generation branch first.

Bhavikshah123:
Closing this PR and have created the new PR following the wiki.

PR #24378, opened 2024-03-15
Add security utilities detections.

michaelcfanning:
## Description

Update secrets detection working branch to use the `microsoft-security-utilities` OSS library for high confidence secret detection and redaction.

## Mandatory Checklist

- Please choose the target release of Azure PowerShell
  - [X ] N/A

- [ X] Check this box to confirm: **I have read the [_Submitting Changes_](../blob/main/CONTRIBUTING.md#submitting-changes) section of [`CONTRIBUTING.md`](../blob/main/CONTRIBUTING.md) and reviewed the following information:**

### DIFF
--- a/src/Accounts/Authentication/Sanitizer/Services/DefaultSanitizerService.cs
+++ b/src/Accounts/Authentication/Sanitizer/Services/DefaultSanitizerService.cs
@@ -25,11 +25,7 @@ namespace Microsoft.Azure.Commands.Common.Authentication.Sanitizer.Services
             new SecretMasker(WellKnownRegexPatterns.HighConfidenceMicrosoftSecurityModels,
                              generateCorrelatingIds: true);
 
-        public string SanitizedValue => "******";
-
-        public Dictionary<string, IEnumerable<string>> IgnoredProperties => BuildIgnoredProperties();
-
-        private Dictionary<string, IEnumerable<string>> BuildIgnoredProperties()
+        public Dictionary<string, IEnumerable<string>> IgnoredProperties => new Dictionary<string, IEnumerable<string>>()
         {
             /*
              * This dictionary is used to store the properties that should be ignored during sanitization.
--- a/src/Accounts/Authentication/Sanitizer/Providers/SanitizerProviderBase.cs
+++ b/src/Accounts/Authentication/Sanitizer/Providers/SanitizerProviderBase.cs
@@ -20,7 +20,7 @@ namespace Microsoft.Azure.Commands.Common.Authentication.Sanitizer.Providers
 {
     public abstract class SanitizerProviderBase
     {
-        private const int MaxDepth = 10;
+        private const int MaxDepth = 5;
 
         protected ISanitizerService Service { get; private set; }
 


src/Accounts/Authentication/Sanitizer/Services/DefaultSanitizerService.cs
```diff
@@ -12,13 +12,19 @@
...
 namespace Microsoft.Azure.Commands.Common.Authentication.Sanitizer.Services
 {
+
     internal class DefaultSanitizerService : ISanitizerService
     {
+        private SecretMasker secretMasker =
+            new SecretMasker(WellKnownRegexPatterns.HighConfidenceMicrosoftSecurityModels,
+                             generateSha256Hashes: true);
+
         public string SanitizedValue => "******";
```
michaelcfanning, line 28: >"******"; [](http://example.com/codeflow?start=40&length=9)

Vincent, currently this argument is not honored by the OSS masker, which currently injects `+++` for every redaction find. The next update of the library will actually inject a hash of the redacted find that can be used to correlate the redacted secret with other telemetry logs. We probably want this option to be configurable and we probably want the library to be configurable to update the default sanitized value. I'll look into this for 1.4.4.

src/Accounts/Authentication.Test/AuthenticatorsTest/DefaultSanitizerServiceTests.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using Xunit;
+
+namespace Microsoft.Azure.Commands.Common.Authentication.Test.AuthenticatorsTest
+{
+    public class DefaultSanitizerServiceTests
+    {
+        [Fact]
+        [Trait(Category.AcceptanceType, Category.CheckIn)]
+        public void DefaultSanitizerService_HighConfidenceSecurityModels()
```
michaelcfanning, line 28: >DefaultSanitizerService_HighConfidenceSecurityModels [](http://example.com/codeflow?start=20&length=52)

I know that you've added a higher level test insuring this functionality is working/integrated in actual command utilization. I thought we could use these focused tests to make sure things are working at the server object level.

src/Accounts/Authentication/Sanitizer/Services/DefaultSanitizerService.cs
```diff
@@ -39,66 +45,12 @@ internal class DefaultSanitizerService : ISanitizerService
...
             {
-                foreach (var pattern in SensitiveDataPatterns)
-                {
-                    if (Regex.IsMatch(data, pattern))
-                    {
-                        sanitizedData = Regex.Replace(data, pattern, SanitizedValue);
-                        return true;
-                    }
-                }
+                sanitizedData = this.secretMasker.MaskSecrets(data);
```
vidai-msft, line 52: @michaelcfanning Currently we just need to detect secrets instead of masking it. Could we replace with `this.secretMasker.DetectSecrets` ?

vidai-msft: @michaelcfanning The function has an out variable for sanitized data. However, in the current phase, this value won't be used and we only use this function to identify if the data has secrets. We will implement the real redaction in the next phase. Since the MaskSecrets has the logging system, I believe it is OK to continue using this function.
In addition, could you please help take a look at the test issue and upgrade the package version to a better one?

PR #24675, opened 2024-04-22
Replaced parameter `EnableRbacAuthorization` by `DisableRbacAuthorization` in `New-AzKeyVault` and `Update-AzKeyVault`

BethanyZhou:
## Mandatory Checklist

- Please choose the target release of Azure PowerShell
  - [ ] N/A


YanaXu (approved):

MattGal:
@YanaXu / @BethanyZhou ; can you please comment as to why did you need to remove the `-EnableRbacAuthorization` parameter instead of making it a no-op? 

 My team has release branches that contain usage of this flag, and the version of the Az PS Cmdlets is necessarily updated on build agents as a security measure, so we've had to make code changes to work around this.  

Leaving it in seems like a harmless thing to do, but removing it breaks users, so I'd like to understand the motivation.

BethanyZhou:
Hi @MattGal , thanks for asking.

Azure Key Vault is migrating from vault access policy to an Azure role-based access control (Azure RBAC) permission model. Azure RBAC is the default and recommended authorization system for Azure Key Vault. See [Migrate from vault access policy to an Azure role-based access control permission model](https://learn.microsoft.com/en-us/azure/key-vault/general/rbac-migration).

User had to specify EnableRbacAuthorization parameter explicitly before, which implies access policy is the default authorization systems. That's why we removed parameter EnableRbacAuthorization. Making it a no-op is an option but it's not a good practice for a new user. We preannounced the upcoming breaking change two sprints (almost two months) in advance to let user be aware of this and take actions when major version comes.

MattGal:
Counterpoint: the actual text of the commit you made includes you suppressing this warning:
`"The cmdlet 'New-AzKeyVault' no longer supports the parameter 'EnableRbacAuthorization' and no alias was found for the original parameter name.", "Add the parameter 'EnableRbacAuthorization' back to the cmdlet 'New-AzKeyVault', or add an alias to the original parameter name."`

... you never provided any period of time where users who don't actively read your preannouncements (nobody in my team, here at Microsoft, knew this was coming) got a warning, you simply removed the parameter and when the Az cmdlets moved forward, the change broke our official and release builds without so much as a warning.

Can you point to where this "make breaking changes but only communicate them externally to the product" principal exists in your documents?  I.e. the existence of `tools/StaticAnalysis/Exceptions/Az.KeyVault/BreakingChangeIssues.csv` implies that there's a process you purposefully circumvented here, and I'd like to represent the pain points of the customers surprised by this to whomever granted you that exception.

PR #24795, opened 2024-05-06
Deployment Stacks GA cmdlet changes.

dantedallag:
Parameter adjustments and new autorest sdk generation to prepare for deployment stacks GA.

## Description

This is the updates required for GA release of deployments stacks. In addition to updating the generated sdk, cmndlet parameters have been adjusted and a couple of other relevant changes have been made. Scenario tests have also been run in live mode.

Though this is "breaking", deployment stacks are still in preview, so I don't think it needs the tag on the PR? Let me know if that is not the case. 

## Mandatory Checklist

  - [ x] [General release](../blob/main/CONTRIBUTING.md#target-release-types)
  - [ ] No need for a release

- [ x] Check this box to confirm: **I have read the [_Submitting Changes_](../blob/main/CONTRIBUTING.md#submitting-changes) section of [`CONTRIBUTING.md`](../blob/main/CONTRIBUTING.md) and reviewed the following information:**

### DIFF
--- a/src/Resources/Resources/ChangeLog.md
+++ b/src/Resources/Resources/ChangeLog.md
@@ -19,9 +19,9 @@
 -->
 
 ## Upcoming Release
+* Deployment Stack cmdlets GA release/updates.
 * [Breaking Change] Redesigned CRUD cmdlets for `PolicyAssignment`, `PolicyDefinition`, `PolicyExemption`, `PolicySetDefinition`. Please see Az 12 migration guide https://learn.microsoft.com/en-us/powershell/azure/migrate-az-12.0.0 for more detail.
 * Added null check to the permissions object in the ToPSRoleDefinition method to return if the whole permissions object array is null.
-* Deployment Stack cmdlets GA release/updates.
 
 ## Version 6.16.2
 * Introduced secrets detection feature to safeguard sensitive data.


src/Resources/Resources/ChangeLog.md
```diff
@@ -24,6 +24,7 @@
 * Introduced secrets detection feature to safeguard sensitive data.
 * Migrated SDK generation from autorest csharp to autorest powershell.
 * Fixed deployment and deployment stack New/Set cmdlets to fail if template/parameter uri fails to downloads.
+* Adjusted Depoyment Stacks cmdlets to prepare for GA.
```
NoriZC, line 28: Do you have any reasons of adding changelogs under those 2 versions instead of under Upcoming Release?

src/Resources/Resources/ChangeLog.md
```diff
@@ -36,6 +37,7 @@
 * Fixed bug with custom types and deployments whatif. [#13245]
 * Fixed bug with nullable array parameters & outputs.
 * Fixed bug with TemplateParameterUri not downloading parameters correctly.
+* Add dynmaic parameters to stack New/Set cmdlets.
```
NoriZC, line 40:
```suggestion
* Added dynmaic parameters to stack New/Set cmdlets.
```
And the position is wired, we should not add changelogs under previous version

NoriZC:
> I am also unsure about the help file generation as I am seeing an unknown parameter "ProgressAction" being added to all of the cmdlets and it seems like there are PRs to remove this from other cmdlet's help files. Should I generate and manually remove references to "ProgressAction"?

You don't need to do so, we will delete them at once right before release.

NoriZC:
For the breaking changes: Yes as the DeploymentStack is under preview version, it's 

> @NoriZC I have updated help files and the changelog. Thanks for reviewing!
> 
> Let me know if you have thoughts on the other question I had about the change to the Client Factory.

Will discuss this issue with team and then get back to you :)

src/Accounts/Authentication/Factories/ClientFactory.cs
```diff
@@ -67,7 +67,15 @@ public ClientFactory()
             var subscriptionId = typeof(TClient).GetProperty("SubscriptionId");
             if (subscriptionId != null && context.Subscription != null)
             {
-                subscriptionId.SetValue(client, context.Subscription.Id.ToString());
+                // TODO: Once subscription type is updated to Guid everywhere, will no longer need this check.
```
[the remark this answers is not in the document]
NoriZC: We just discussed this modification. We understand that this is caused by a type change in common type v5 of swagger, and need to be resolved because other modules will also face same issues. But as it's a change in a base class, we need to investigate whether this would have deeper impact.

The investigation would cost some time, and this PR may not catch current release. If you are in a hurry, we might consider update the common type version in swagger using directive first.

NoriZC:
> I have completed requests and removed preview tags from the cmdlets. Let me know if there is anything else that needs to be done for GA release.

Please obviously announce Deployment Stack cmdlets GA in the changelog. Now the changelog is just saying that "prepare for GA"

src/Resources/Resources.Management.Sdk/Resources.Management.Sdk.sln
NoriZC, line 1: Please do not check in extra sln files.

NoriZC (approved):

NoriZC (approved):

PR #25027, opened 2024-05-29
Fixed a PrivateRange bug for classic firewall when trying to use /32

ajtms:
## Description

Fixed a PrivateRange bug for classic firewall when trying to use /32 in a private range. Modified the firewall/policy validation on PrivateRange to give a recommended masked address if the supplied one is improperly masked. Also adds and re-records (most of the line changes are from recording the session records) SessionTests/Records for the failing /32 address.

# Problem
When a user tries to add 255.255.255.255/32 to the private ranges of a classic firewall (no firewall policy), the cmdlet would throw an error.

# Solution
A bit-shift corner case was performing full shifts of an unsigned int (32 bits) which returns the original input (this is part of the C# language specification) and not 0. Similar code in the firewall policy module covers this case, but both situations have been updated to use a shared module for doing simple subnet and IP address validation.

# Features
- When a user tries to add a private range to classic firewall or to a firewall policy, if the address is not correctly masked by the host identifier/network length prefix, the error message now includes a suggested private range.
  - Ex. "192.168.255.0/20 is invalid" -> "Try using 192.168.240.0/20 instead"

# Additional changes
- Added 2 extra validations as part of the PrivateRangeCRUD tests for firewall and firewall policy.
- Updated the SessionRecords for those tests as well

## Mandatory Checklist

  - [ ] No need for a release

### DIFF
--- a/src/Network/Network/Common/Utils.cs
+++ b/src/Network/Network/Common/Utils.cs
@@ -86,7 +86,7 @@ namespace Microsoft.Azure.Commands.Network
         }
     }
 
-    public class NetworkValidationUtils
+    public static class NetworkValidationUtils
     {
         public static void ValidateIpAddress(string ipAddress)
         {


src/Network/Network/Common/Utils.cs
```diff
@@ -83,4 +85,180 @@ public static bool IsRegionRestrictedForBasicFirewall(string region)
             return false;
         }
     }
+
+    namespace Utilities.CIDR.IPv4
```
NoriZC, line 89: Could you explain why writing a nested namespace here rather than creating a new file? And the naming may also cause confusion.

From the [Namespace Naming Guidelines](http://msdn.microsoft.com/en-us/library/893ke618(vs.71).aspx):

The general rule for naming namespaces is to use the company name followed by the technology name and optionally the feature and design as follows.

CompanyName.TechnologyName[.Feature][.Design]

NoriZC (changes requested):

src/Network/Network/Common/Utils.cs
ajtms: Good call. The utility functions don't really need more namespaces, so I've migrated them back into a static class similar to the existing ones in that file.

NoriZC (approved):

PR #25457, opened 2024-07-04
Revert 23H2 Registration blocker 

priyjain358:
## Mandatory Checklist

  - [ ] No need for a release

### DIFF
--- a/src/StackHCI/StackHCI.Autorest/README.md
+++ b/src/StackHCI/StackHCI.Autorest/README.md
@@ -3,9 +3,6 @@
 This directory contains the PowerShell module for the StackHci service.
 
 ---
-## Status
-[![Az.StackHCI](https://img.shields.io/powershellgallery/v/Az.StackHCI.svg?style=flat-square&label=Az.StackHCI "Az.StackHCI")](https://www.powershellgallery.com/packages/Az.StackHCI/)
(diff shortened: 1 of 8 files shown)


VeryEarly:
This PR looks like a breaking change, any reason this feature needs to be reverted?

bniranjanbhat:
This change is related to HCI Device deliveries.
Earlier it was decided that the customers would not be allowed to use certain features if they are on newer version of OS. But later because of engineering constraints we had to allow the allow the customers use these features on the newer versions of OS as well.
Hence this change.
I would not call this as a breaking change, since we are loosing down the validation at this point.

VeryEarly:
> This change is related to HCI Device deliveries. Earlier it was decided that the customers would not be allowed to use certain features if they are on newer version of OS. But later because of engineering constraints we had to allow the allow the customers use these features on the newer versions of OS as well. Hence this change. I would not call this as a breaking change, since we are loosing down the validation at this point.

Thanks for the explanation, please provide release note for this change in the description or comments. Something like Allow newer version ... of ... for ...

bniranjanbhat:
Release Notes:  Allow registration for 23H2 and above versions of the device.

VeryEarly (approved):

PR #25935, opened 2024-08-26
Integrate security library with secrets detection

vidai-msft:
## Description

This PR is to integrate a common secret detection library to expand the scope of the patterns of secrets.

## Mandatory Checklist

  - [ ] No need for a release

### DIFF
--- a/src/Accounts/Authentication/Authentication.csproj
+++ b/src/Accounts/Authentication/Authentication.csproj
@@ -17,7 +17,7 @@
     <PackageReference Include="Microsoft.Identity.Client" Version="4.61.3" />
     <PackageReference Include="Microsoft.Identity.Client.Extensions.Msal" Version="4.61.3" />
     <PackageReference Include="Microsoft.Identity.Client.Broker" Version="4.61.3" />
-    <PackageReference Include="Microsoft.Security.Utilities.Core" Version="1.6.0" />
+    <PackageReference Include="Microsoft.Security.Utilities.Core" Version="1.8.0" />
   </ItemGroup>
 
   <ItemGroup>


vidai-msft:
@michaelcfanning if we only need to detect secrets instead of redacting them, is `SecretMasker` the correct object?

msJinLei (dismissed):

src/Accounts/Authentication/Sanitizer/Services/DefaultSanitizerService.cs
```diff
@@ -44,67 +45,16 @@ internal class DefaultSanitizerService : ISanitizerService
...
-            @"\b" // pre-match
-          + @"oy2[a-p][0-9a-z]{15}[aq][0-9a-z]{11}[eu][bdfhjlnprtvxz357][a-p][0-9a-z]{11}[aeimquy4]" // match
-          + @"\b", // post-match
-            
-            // NPM author keys.
-            @"\b" // pre-match
-          + @"npm_[0-9A-Za-z]{36}" // match
-          + @"\b", // post-match
-        };
+        private readonly SecretMasker _secretMaster = new SecretMasker(WellKnownRegexPatterns.HighConfidenceMicrosoftSecurityModels, generateCorrelatingIds: true);
```
michaelcfanning, line 48: >_secretMaster [](http://example.com/codeflow?start=38&length=13)

maybe name this `secretMasker`?

src/Accounts/Authentication/Sanitizer/Services/DefaultSanitizerService.cs
```diff
@@ -44,67 +45,16 @@ internal class DefaultSanitizerService : ISanitizerService
...
             if (!string.IsNullOrWhiteSpace(data))
             {
-                foreach (var pattern in SensitiveDataPatterns)
-                {
-                    if (Regex.IsMatch(data, pattern))
-                    {
-                        return true;
-                    }
-                }
+                var detections = _secretMaster.DetectSecrets(data);
```
michaelcfanning, line 56: >DetectSecrets [](http://example.com/codeflow?start=47&length=13)

so what you are doing here is only running the detections logic. and then reporting if you found at least one problem.

is that what you want? 

the `MaskSecrets(string) : string` API will accept an input and return a string with all detected secrets redacted from it, if that's what you need.

vidai-msft: > > DetectSecrets
> 
> so what you are doing here is only running the detections logic. and then reporting if you found at least one problem.
> 
> is that what you want?
> 
> the `MaskSecrets(string) : string` API will accept an input and return a string with all detected secrets redacted from it, if that's what you need.

Yes, I just need the detection logic instead of masking them.

michaelcfanning: ok, so you return a bool that basically tells the caller 'there is one or more secrets in the string that you provided me. I note that you aren't doing anything useful with the `sanitizedData` argument.

do you plan to do something with it in the future? the name of this method strongly suggests that sanitization should have occurred. :)

looking at the callers to this method, I see that you produce telemetry on detecting a secret in various data. that's good! your implementation would be improved if you flowed more detection data back to the caller. specifically, the kind of secret and its correlating id would be helpful. maybe for a future change?

vidai-msft: Yes, here the redaction is a placeholder for future development. I will log the secrets pattern in the telemetry. What data should I log? The correlation id and/or pattern id (SECXXX/XXX) or something else?

msJinLei (approved):

PR #26608, opened 2024-11-06
Update resourceManagement.yml

wanyang7:
Update resourceManagement.yml for Synapse, wonner is no longer used by Wan Yang, use wanyang7 instead.


vidai-msft:
@wanyang7 Please reach out to our PM for the change. It is not allowed to update it directly.

PR #26685, opened 2024-11-14
[AppServices] RestoreAzureWebAppSnapshot fetch the data from Microsoft.Web RP similar to other commands

vkarasala:
[AppServices] RestoreAzureWebAppSnapshot fetch the data from Microsoft.Web RP similar to other commands

## Description

Currently to fetch the source web app information we rely on Azure Resources API(https://learn.microsoft.com/en-us/rest/api/resources/resources/list?view=rest-resources-2021-04-01). 

The problem with above API is that the response is paginated, and filter is applied at backend on the paginated block. So, for example, if customer has 1000 sites, in first batch ARM fetches 100 from RP and applies the filter provided. If no results found, an empty collection is returned with nextLink to further query. If the customers have 1000's of resources, we end up making a lot of calls and get no data. The changes are to simplify that process.

For background, we changed to use ARM cache 5 years ago. This is because during an outage our RP is not functional, but ARM cache can help with required data for restore.

So, the changes still retain such handling but improve by calling to WEB RP to get the required data if it functional. Fallback will be made to ARM cache if WebRP is not functional.

Tested locally

image
image

## Mandatory Checklist

  - [ ] No need for a release


VeryEarly:
please update changelog: https://github.com/vkarasala/azure-powershell/blob/user/vkarasala/restore-snapshot-to-user-web-rp/src/Websites/Websites/ChangeLog.md

vkarasala:
> please update changelog: https://github.com/vkarasala/azure-powershell/blob/user/vkarasala/restore-snapshot-to-user-web-rp/src/Websites/Websites/ChangeLog.md

Updated.

jsharmaMSFT:
LGTM. No mocks/tests to add here?

vkarasala:
> LGTM. No mocks/tests to add here?

There are none defined, also it is understandable as the Snapshots are only available after an hour, tests are hard to define for such.

src/Websites/Websites/ChangeLog.md
```diff
@@ -21,6 +21,7 @@
 
 ## Version 3.2.2
 * Fix bug where parameters could not be set to false for `Publish-AzWebApp`
+* Fix the source app retrival from Microsoft.Web RP instead of ARM cache for `RestoreAzureWebAppSnapshot`
```
VeryEarly, line 24: please move this line under line 20

VeryEarly (approved):

PR #26900, opened 2024-12-18
Try removing track 1 deps

isra-fel:


isra-fel:
CI passed after the track 1 sdks are removed. We can safely assume that all sdk-based test projects don't have to depend on any track 1 management plain sdks.
For data plain it's another story, because we can't replace the current solution (test runner creates instance of sdk client and replaces real client) with the better solution (test runner adds http interceptor to the real client without referencing the concrete type), and that's because the creation of data plain sdks may not call client factory, so we may not have the chance to append http interceptor.

PR #26950, opened 2024-12-31
Remove ResourceManager track1 SDK dependency from testfx

VeryEarly:
## Description

We are removing all track 1 SDK dependencies from Azure-PowerShell. Removing "Microsoft.Azure.Management.ResourceManager" Version="3.17.3-preview" from testfx

## Mandatory Checklist

  - [ ] No need for a release

### DIFF
--- a/src/Resources/Resources.Test/PageExtensions.cs
+++ b/src/Resources/Resources.Test/PageExtensions.cs
@@ -1,0 +1,29 @@
+﻿// ----------------------------------------------------------------------------------
+//
+// Copyright Microsoft Corporation
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


isra-fel:
Adding some context to this PR: we were trying to migrate the arm sdk inside Resources module to sdk-autorest.powershell, but after that we found an error in the test projects of Resources saying there were ambiguous type references. Turned out test projects all depend on an ARM sdk inherited forom `Test.props`, so by migrating to sdk-ap we essentially introduced two sets of types with the same name space + type name but from different source, hence confusing the compiler.

This PR demonstrates by removing the reference to arm sdk in test.props the problem is fixed.

isra-fel (approved):
Please explain why removing the package reference didn't cause errors, cause it contradicts to what we discussed before.

VeryEarly:
> Please explain why removing the package reference didn't cause errors, cause it contradicts to what we discussed before.

Besides removing the track 1 SDK reference, this PR also move the class which depends on RM track1 sdk in testfx to resources.test. Hence resource.test depends on checked in RM sdk already, this class turn to depends on the checked in sdk instead.

PR #26974, opened 2025-01-07
Support pagination for enumerating deleted items

amrita-hegde:
## Description

Cosmos customers currently utilize the Azure PowerShell for restoring items. However, a limitation exists in the tool: it lacks the capability to retrieve the items by enumerating them page-wise. This change provides the capability to pass a continuation token to the Get-AzDataLakeStoreDeletedItem cmdlet so that, users can retrieve the next page using the continuation token. Maximum allowed value for this numResults(pagesize) is 4000. The number of returned entries could be more or less than numResults.

###Changes

**Azure PowerShell**

To enable subsequent pages to be fetched from PowerShell,

Define the optional parameter Continuation Token in Azure PowerShell for GetAzureRmDataLakeStoreDeletedItem command.

Use the continuation token to fetch subsequent pages when calling the API.

**.NET SDK**

https://github.com/Azure/azure-data-lake-store-net/pull/69
Change is released in https://www.nuget.org/packages/Microsoft.Azure.DataLake.Store/2.0.2 

This overloaded method will be called from Azure PowerShell by consuming the latest SDK version 2.0.2.

### Testing

Screenshot 2025-04-21 184949

#### Response from GetAzureRmDataLakeStoreDeletedItem cmd
image

#### Test execution
Screenshot 2025-03-13 165029

#### When expired/incorrect continuation token is passed
Screenshot 2025-03-13 114340

#### Local build
image

## Mandatory Checklist

  - [ ] No need for a release

### DIFF
--- a/src/DataLakeStore/DataLakeStore/help/Get-AzDataLakeStoreDeletedItemsWithToken.md
+++ b/src/DataLakeStore/DataLakeStore/help/Get-AzDataLakeStoreDeletedItemsWithToken.md
@@ -39,7 +39,7 @@ This operation may take considerable time when searching through millions of del
 
 ### Example 1: Search for deleted items.
 ```powershell
-PS C:\> Get-AzDataLakeStoreDeletedItemsWithToken -Account "ml1ptrashtest" -Filter "test0/file_123"
+Get-AzDataLakeStoreDeletedItemsWithToken -Account "ml1ptrashtest" -Filter "test0/file_123"
 ```
 
 ```output
@@ -55,7 +55,7 @@ This example searches for deleted items matching the pattern "test0/file_123".
 
 ### Example 2: Search for deleted items with pagination.
 ```powershell
-PS C:\> Get-AzDataLakeStoreDeletedItemsWithToken -Account "ml1ptrashtest" -Filter "test0/file_123" -ListAfter "133862976000000000/co3aa1020309024/"
+Get-AzDataLakeStoreDeletedItemsWithToken -Account "ml1ptrashtest" -Filter "test0/file_123" -ListAfter "133862976000000000/co3aa1020309024/"
 ```
 
 ```output


amrita-hegde:
@microsoft-github-policy-service agree company="Microsoft"


Regards

Amrita

src/DataLakeStore/DataLakeStore/DataPlaneModels/DataLakeStoreFileSystemClient.cs
```diff
@@ -56,7 +56,7 @@ public class DataLakeStoreFileSystemClient
         static DataLakeStoreFileSystemClient()
         {
             // Registering the custom target class
-            Target.Register<AdlsLoggerTarget>("AdlsLogger"); //generic
+            LogManager.Setup().SetupExtensions(ext => ext.RegisterTarget<AdlsLoggerTarget>("AdlsLogger"));
```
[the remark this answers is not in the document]
amrita-hegde: Target.Register<T>(string) has been marked as obsolete. LogManager.Setup().SetupExtensions() method is recommended. This change was introduced in NLog version 5.2 which is a dependency for new adls sdk.

src/Accounts/Accounts/Accounts.csproj
```diff
@@ -19,6 +19,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Identity.Client" Version="4.70.0" />
```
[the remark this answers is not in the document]
amrita-hegde: Yes, this was added unintentionally during testing. Removed it.

src/DataLakeStore/DataLakeStore/DataPlaneModels/AdlsClientFactory.cs
```diff
@@ -6,6 +6,11 @@
 using Microsoft.Rest;
 using Microsoft.WindowsAzure.Commands.Common;
 using System.Net.Http;
+using Azure.Core;
+using Azure.Identity;
```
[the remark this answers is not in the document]
amrita-hegde: Azure.Core and Azure.Identity are required because the TokenCredential and DefaultAzureCredential classes belong to these namespaces, respectively. The remaining unused namespaces have been removed from the code.

akharit (approved):

tools/Common.Netcore.Dependencies.targets
```diff
@@ -22,7 +22,7 @@
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Azure.Core" Version="1.44.1"/>
+    <PackageReference Include="Azure.Core" Version="1.45.0"/>
```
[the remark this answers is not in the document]
amrita-hegde: Thanks. Updated the binary

msJinLei:
Still 7 failed test cases

msJinLei (changes requested):

msJinLei:
The current issues of the PR
-	Request to upgrading Azure.Core to 1.45.0. My question is It is required? Can you use the 1.44.1?
-	Need to depend on DefaultAzureCredential from Azure.Identity. Please don’t depend on Azure.Identity’s authentication interface directly.
Could you describe your authentication requirement and we can see how to implement your requirement.
-	Failed test cases (can check it after the problems above are resolved)

src/DataLakeStore/DataLakeStore/DataPlaneModels/AdlsClientFactory.cs
```diff
@@ -47,17 +51,17 @@ internal static AdlsClient GetAdlsClient(string accntNm, IAzureContext context)
...
                 if (IsTest)
                 {
-                    client = AdlsClient.CreateClient(accntNm, MockCredentials, DataLakeStoreFileSystemClient.ImportExportMaxThreads, CustomDelegatingHAndler);
+                    client = AdlsClient.CreateClient(accntNm, MockCredentials, new string[] { });
                 }
                 else
                 {
-                    client = AdlsClient.CreateClient(accntNm, AzureSession.Instance.AuthenticationFactory.GetServiceClientCredentials(context,
-                        AzureEnvironment.Endpoint.AzureDataLakeStoreFileSystemEndpointSuffix), DataLakeStoreFileSystemClient.ImportExportMaxThreads);
+                    var tokenCredential = new DefaultAzureCredential();
```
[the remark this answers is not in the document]
msJinLei: Please create an class that derives from TokenCredential and use AuthenticationFactory.Authenticate to acquire token.
Here is the example https://github.com/Azure/azure-powershell/blob/0859c268394fce352bf6e8e7ce992a1a3200e409/src/Compute/Compute/Common/ComputeTokenCredential.cs

msJinLei:
> The current issues of the PR
> 
> * Request to upgrading Azure.Core to 1.45.0. My question is It is required? Can you use the 1.44.1?
- If so we are going to upgrading Azure.Core version
> * Need to depend on DefaultAzureCredential from Azure.Identity. Please don’t depend on Azure.Identity’s authentication interface directly.
- Please refer to https://github.com/Azure/azure-powershell/blob/0859c268394fce352bf6e8e7ce992a1a3200e409/src/Compute/Compute/Common/ComputeTokenCredential.cs to implement the token acquisition 
> * Failed test cases (can check it after the problems above are resolved)

tools/Common.Netcore.Dependencies.targets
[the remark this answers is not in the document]
amrita-hegde: Yes, required version is 1.45.0

amrita-hegde:
> The current issues of the PR
> 
> * Request to upgrading Azure.Core to 1.45.0. My question is It is required? Can you use the 1.44.1?
>>  We can't use Azure.Core version 1.44.1 because this PR requires the latest .NET SDK (2.0.2), which depends on Azure.Core 1.45.0 or higher. You can see the dependency here: https://www.nuget.org/packages/Microsoft.Azure.DataLake.Store/2.0.2#dependencies-body-tab So, updating the version to 1.45.0 is required.

> * Need to depend on DefaultAzureCredential from Azure.Identity. Please don’t depend on Azure.Identity’s authentication interface directly.
>   Could you describe your authentication requirement and we can see how to implement your requirement.
>> Currently, Azure PowerShell uses .NET SDK version 1.2.4-alpha, which authenticates the ADLS client using the [ServiceClientCredentials](https://learn.microsoft.com/en-us/dotnet/api/microsoft.rest.serviceclientcredentials?view=azure-dotnet) from Microsoft.Rest library. However, this authentication mechanism is no longer supported in the latest SDK version (2.0.2).
The new SDK requires the ADLS client to be instantiated using a TokenCredential from the Azure.Core library. This includes several authentication types (e.g., DefaultAzureCredential, ClientSecretCredential, etc.) provided by the Azure.Identity library. Among these, DefaultAzureCredential is the recommended option. https://learn.microsoft.com/en-us/dotnet/api/azure.core.tokencredential?view=azure-dotnet
> * Failed test cases (can check it after the problems above are resolved)

msJinLei:
@amrita-hegde The Azure.Core is upgraded to 1.45.0. Please pull the latest changes of main branch to your developing branch.

msJinLei (dismissed):

msJinLei:
@amrita-hegde 
Please refer to the logs https://dev.azure.com/azclitools/public/_build/results?buildId=247530&view=logs&j=b17395f6-68a3-5682-0476-d3f6f1043109&t=f098cad5-2fd2-534b-df25-06e0723952ab and fix the failed test cases

src/DataLakeStore/DataLakeStore.Test/ScenarioTests/DataLakeStoreTestRunner.cs
```diff
@@ -57,18 +64,42 @@ protected DataLakeStoreTestRunner(ITestOutputHelper output)
...
+                    var azureContext = new AzureContext
                     {
-                        AdlsClientFactory.IsTest = true;
-                        var creds = context.GetClientCredentials(AzureEnvironment.Endpoint.AzureDataLakeStoreFileSystemEndpointSuffix);
-                        AdlsClientFactory.CustomDelegatingHAndler = context.AddHandlers(creds, new AdlMockDelegatingHandler());
-                        AdlsClientFactory.MockCredentials = creds;
-                        return new object();
-                    }
-                )
+                        Account = AzureRmProfileProvider.Instance.Profile.DefaultContext.Account,
```
msJinLei, line 71: I have 2 modes: recording and playback. In playback mode, you should use MockCredential. Specifically in you case, you should set `AdlsClientFactory.IsTest = true`.  In pipeline, the test cases are all run in playback mode, which no access to EntraID service.

msJinLei (changes requested):

src/DataLakeStore/DataLakeStore.Test/ScenarioTests/DataLakeStoreTestRunner.cs
amrita-hegde: I have updated the code to use MockCredentials for test. Please take a look.

amrita-hegde (commented):
The CustomDelegatingHAndler is not passed to the AdlsClient. This method is removed from the 2.0.2 SDK. By passing the custom delegating handler, it will inject custom handler into the normal HttpClient. So marked 7 tests as LiveOnly mode since they were trying to call azure environment in playback mode.

src/DataLakeStore/DataLakeStore/DataPlaneCommands/GetAzureRmDataLakeStoreDeletedItem.cs
```diff
@@ -47,15 +47,27 @@ public class GetAzureDataLakeStoreDeletedItem : DataLakeStoreFileSystemCmdletBas
             HelpMessage = "Minimum number of entries to search for")]
         public int Count { get; set; } = 100;
 
+        [Parameter(ValueFromPipelineByPropertyName = true,
+            Mandatory = false,
+            ParameterSetName = DefaultParameterSet,
+            HelpMessage = "Token returned by system in the previous invocation")]
+        public string listAfter { get; set; }
```
vidai-msft: Per .NET and PowerShell naming convention, the parameter name should be ListAfter (capital 'L')

src/DataLakeStore/DataLakeStore/DataPlaneCommands/GetAzureRmDataLakeStoreDeletedItem.cs
```diff
@@ -47,15 +47,27 @@ public class GetAzureDataLakeStoreDeletedItem : DataLakeStoreFileSystemCmdletBas
...
-            var toReturn = DataLakeStoreFileSystemClient.EnumerateDeletedItems(Account, Filter, Count, this, CmdletCancellationToken).Select(entry => new DataLakeStoreDeletedItem(entry)).ToList();
-            WriteObject(toReturn);
+            var (deletedItems, continuationToken) = DataLakeStoreFileSystemClient.EnumerateDeletedItems(Account, Filter, Count, listAfter, this, CmdletCancellationToken);
+            var toReturn = deletedItems.Select(entry => new DataLakeStoreDeletedItem(entry)).ToList();
+
+            // Write the list of deleted items.
+            WriteObject(toReturn, enumerateCollection: true);
+
+            // Write the continuation token.
+            WriteObject($"Continuation Token: {continuationToken}");
```
vidai-msft: The WriteObject is kind of special in PowerShell SDK. It is used to output the returned object instead of a result message. Any reason why the token should be piped out?

src/DataLakeStore/DataLakeStore/DataPlaneModels/AdlsClientFactory.cs
```diff
@@ -47,17 +49,17 @@ internal static AdlsClient GetAdlsClient(string accntNm, IAzureContext context)
                 AdlsClient client = null;
                 if (IsTest)
                 {
-                    client = AdlsClient.CreateClient(accntNm, MockCredentials, DataLakeStoreFileSystemClient.ImportExportMaxThreads, CustomDelegatingHAndler);
+                    client = AdlsClient.CreateClient(accntNm, MockCredentials, new string[] { });
```
vidai-msft: The IsTest looks like has no use here. Can you please refactor the code here to avoid any confuse in the future?

src/DataLakeStore/DataLakeStore.Test/ScenarioTests/DataLakeStoreTestRunner.cs
```diff
@@ -57,18 +63,42 @@ protected DataLakeStoreTestRunner(ITestOutputHelper output)
...
-                        var creds = context.GetClientCredentials(AzureEnvironment.Endpoint.AzureDataLakeStoreFileSystemEndpointSuffix);
-                        AdlsClientFactory.CustomDelegatingHAndler = context.AddHandlers(creds, new AdlMockDelegatingHandler());
-                        AdlsClientFactory.MockCredentials = creds;
-                        return new object();
-                    }
-                )
+                {
+                    AdlsClientFactory.IsTest = true;
+
+                    var serviceClientCredentials = context.GetClientCredentials("https://datalake.azure.net");
```
vidai-msft: We have a class to predefine all the well-known URLs. datalake.azure.net could be referenced by AzureEnvironment.Endpoint.DataLakeEndpointResourceId. Can consider replacing with this variable.

src/DataLakeStore/DataLakeStore/Models/DataLakeStoreTokenCredential.cs
```diff
@@ -0,0 +1,46 @@
...
+            }
+
+            _accessToken = AzureSession.Instance.AuthenticationFactory.Authenticate(
+                context.Account,
+                context.Environment,
+                context.Tenant.Id,
+                null,
+                ShowDialog.Never,
+                null,
+                "https://datalake.azure.net/");
```
vidai-msft: Better use AzureEnvironment.Endpoint.DataLakeEndpointResourceId to reference the endpoint here.

src/DataLakeStore/DataLakeStore/help/Get-AzDataLakeStoreDeletedItem.md
```diff
@@ -14,8 +14,8 @@ Searches for deleted entries in trash which match the filter.
 ## SYNTAX
 
 ```
-Get-AzDataLakeStoreDeletedItem [-Account] <String> [-Filter] <String> [-Count <Int32>] [-AsJob]
- [-DefaultProfile <IAzureContextContainer>] [<CommonParameters>]
+Get-AzDataLakeStoreDeletedItem [-Account] <String> [-Filter] <String> [-Count <Int32>] [-listAfter <String>]
```
vidai-msft: Please update the listAfter to ListAfter and re-generate the help docs.

vidai-msft (changes requested):

src/DataLakeStore/DataLakeStore/DataPlaneCommands/GetAzureRmDataLakeStoreDeletedItem.cs
amrita-hegde: Continuation token will be the input to the next request. So user has to know the continuation token to fetch the next page.

vidai-msft: WriteObject will output multiple objects with different types which is not recommended. It may break the existing automation script and looks like a breaking change. Could you please write a test case to verify? Still I don't quite understand why the token is necessary here. Could it be wrapped into the object model?

src/DataLakeStore/DataLakeStore/DataPlaneModels/AdlsClientFactory.cs
amrita-hegde: removed

src/DataLakeStore/DataLakeStore/DataPlaneCommands/GetAzureRmDataLakeStoreDeletedItem.cs
amrita-hegde: We were using WriteObject to output the trash entries earlier. Only the continuation token is the addition here. The whole purpose of this change is to give back a continuation token to the user so that they can pass it in the next command to get the subsequent pages of the deleted item entries.  If it has to be wrapped into an object, it needs multiple change in dotnet sdk.

src/DataLakeStore/DataLakeStore/help/Get-AzDataLakeStoreDeletedItem.md
amrita-hegde: done

src/DataLakeStore/DataLakeStore/Models/DataLakeStoreTokenCredential.cs
amrita-hegde: done

src/DataLakeStore/DataLakeStore.Test/ScenarioTests/DataLakeStoreTestRunner.cs
amrita-hegde: This class is not needed anymore. removed

src/DataLakeStore/DataLakeStore/DataPlaneCommands/GetAzureRmDataLakeStoreDeletedItem.cs
amrita-hegde: Updated

src/DataLakeStore/DataLakeStore.Test/ScenarioTests/AdlsAliasTests.cs
```diff
@@ -70,7 +70,7 @@ public void TestAdlsTrustedIdProvider()
         }
 
         [Fact]
-        [Trait(Category.AcceptanceType, Category.CheckIn)]
+        [Trait(Category.AcceptanceType, Category.LiveOnly)]
```
YanaXu, line 73: Any reason to remove the test cases?

src/DataLakeStore/DataLakeStore/DataPlaneCommands/GetAzureRmDataLakeStoreDeletedItem.cs
YanaXu: I think the comment of @vidai-msft does make sense. WriteObject 2 times should not supported in Azure PowerShell. Could you prepare a cmdlet test example with screenshots to describes your case?

src/DataLakeStore/DataLakeStore.Test/ScenarioTests/AdlsAliasTests.cs
amrita-hegde: https://github.com/Azure/azure-powershell/pull/26974#pullrequestreview-2937930703 As mentioned here, passing the CustomDelegatingHAndler to adls client is not supported in the new .net sdk. Discussed this with @vidai-msft  and marked the testcases as LiveOnly.

amrita-hegde:
@YanaXu added a new cmdlet as suggested by @vidai-msft Requesting your review.

PR #27258, opened 2025-03-03
Fix AzDev module after redesign generation

isra-fel:
## Description

The PR mainly fixes the issue that autorest-based projects are missing from the results of `AzDev` module i.e. `Get-DevProject` and `Get-DevModule`. It also implements a logging mechanism that provides extra verbose and debug logging.

## Mandatory Checklist

  - [x] No need for a release


tools/AzDev/src/Services/ILogger.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// Logs a warning message.
+        /// </summary>
+        void Warning(string message);
+
+        /// <summary>
+        /// Logs an informational message.
+        /// </summary>
+        void Information(string message);
+    }
+}
```
isra-fel, line 42:
```suggestion
}

```

tools/AzDev/src/Services/PSCmdletLogger.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            _cmdlet.WriteWarning(message);
+        }
+
+        public void Information(string message)
+        {
+            _cmdlet.WriteInformation(new InformationRecord(message, "AzDev"));
+        }
+    }
+}
```
isra-fel, line 52:
```suggestion
}

```

tools/AzDev/src/Services/Conventions.cs
```diff
@@ -107,11 +107,13 @@ internal static bool IsExcludedProjectDirectory(IFileSystem fs, string path, out
                 return true;
             }
 
-            if (!TryGetOnlyCsprojPath(fs, path, out var _, out var cannotFindCsproj))
+            if (!TryGetOnlyCsprojPath(fs, path, out var _, out var cannotFindCsproj)
+                && !IsAutorestBasedProject(path, out var notAutorestBased))
```
isra-fel, line 111: Change to core business logic is here. Autorest-based projects as well as sdk-based ones both have a .csproj file that can be used to identify it is a project. This no longer holds for autorest-based projects because their csprojs are moved to another folder.

tools/AzDev/src/Services/DefaultCodebaseProvider.cs
```diff
@@ -12,34 +12,47 @@
...
 using System.IO.Abstractions;
 using AzDev.Models.Inventory;
 
 namespace AzDev.Services
 {
     internal class DefaultCodebaseProvider : ICodebaseProvider
     {
-        private IContextProvider _contextProvider;
-        private IFileSystem _fs;
+        private ILogger _logger;
```
VeryEarly, line 23: default logger or any safeguard to guarantee _logger was initialized before being used?

VeryEarly (commented):
NIT: any particular reason passing logger to models? instead of logging models in service which looks bit more natural to me. For example, define logging/print method in provider base (Ifilesystembasemodel, ILogger), and log model by this method in each service impl or base class

tools/AzDev/src/Services/DefaultCodebaseProvider.cs
isra-fel: Makes sense. Wouldn't want to blocker the main logic if logger is not ready. I'll assign it a default value.
Note that this is not thread safe

VeryEarly (approved):

PR #27908, opened 2025-06-10
[Storage] Support Storage Account saspolicy new properties SasExpirationAction

blueww:
## Mandatory Checklist

  - [ ] No need for a release

### DIFF
--- a/src/Storage/Storage.Management/StorageAccount/NewAzureStorageAccount.cs
+++ b/src/Storage/Storage.Management/StorageAccount/NewAzureStorageAccount.cs
@@ -882,11 +882,11 @@ namespace Microsoft.Azure.Commands.Management.Storage
                 }
                 else
                 {
-                    if (SasExpirationAction.ToLower() == ExpirationAction.Log.ToLower())
+                    if (String.Equals(SasExpirationAction, ExpirationAction.Log, StringComparison.OrdinalIgnoreCase))
                     {
                         SasExpirationAction = ExpirationAction.Log;
                     }
-                    else if (SasExpirationAction.ToLower() == ExpirationAction.Block.ToLower())
+                    else if (String.Equals(SasExpirationAction, ExpirationAction.Block, StringComparison.OrdinalIgnoreCase)) 
                     {
                         SasExpirationAction = ExpirationAction.Block;
                     }
--- a/src/Storage/Storage.Management/StorageAccount/SetAzureStorageAccount.cs
+++ b/src/Storage/Storage.Management/StorageAccount/SetAzureStorageAccount.cs
@@ -904,11 +904,11 @@ namespace Microsoft.Azure.Commands.Management.Storage
                         }
                         else
                         {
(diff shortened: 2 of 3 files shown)


blueww:
@vidai-msft

Would you please help to look at the only test case failure？
https://dev.azure.com/azclitools/public/_build/results?buildId=249688&view=logs&jobId=e44bd064-d8d9-5feb-6343-09b8c49ab1a8&j=e44bd064-d8d9-5feb-6343-09b8c49ab1a8&t=c414cfc4-3e2d-5217-f499-961bb73fce42

The failed test case "Microsoft.Azure.Commands.Management.Storage.Test.ScenarioTests.StorageAccountTests.TestNewSetAzureStorageAccountTLSveresionBlobPublicAccess" is no changed in this PR, and the test only fail on MacOS. I also can't repro this failure on my local machine (win11).

vidai-msft:
The function Retry-IfException looks like has a subtle issue. It relies on the global variable $Error which could be set from outside of the function scope. However, I did not see it was reset at the beginning of the function call.

blueww:
Hi @vidai-msft,

Thanks for looking at the test failure!

As the test pass after rerun, would you like to review and merge this PR if not other issue?

src/Storage/Storage.Management/StorageAccount/NewAzureStorageAccount.cs
```diff
@@ -863,10 +869,31 @@ public override void ExecuteCmdlet()
...
-            if (sasExpirationPeriod != null)
+            if (sasExpirationPeriod != null || SasExpirationAction != null)
             {
-                createParameters.SasPolicy = new SasPolicy(sasExpirationPeriod.Value.ToString(@"d\.hh\:mm\:ss"), "Log");
+                if (sasExpirationPeriod == null && SasExpirationAction != null)
+                {
+                    throw new ArgumentException("-SasExpirationAction can only be specified together with -SasExpirationPeriod.", "SasExpirationAction");
+                }
+                // Set the default action to Log to be aligned as before PSH release.
+                if (SasExpirationAction == null)
```
vidai-msft, line 879: This variable has a default value. Is it necessary to have null check here?

blueww: I believe when you don't input "SasExpirationAction", it will be null. (see below pic)
So we need give it a default value as Log when it's null.
e.g.: 
```
New-AzStorageAccount -ResourceGroupName weitry -Name weisaspolicy321 -SkuName Standard_LRS -Kind StorageV2 -Location eastus2euap -AllowBlobPublicAccess $false
```
image

It looks `[PSDefaultValue(Help = "Log", Value = StorageModels.ExpirationAction.Log)]` won't impact the value in PSH code, per my understanding it only update help, not sure if it's correct? 
Should I just remove this line?

vidai-msft: Yes, it is for doc only. Can keep it there.

PR #28159, opened 2025-07-09
Implement new parameter ClaimsChallenge to mitigate MFA enforcement issues

vidai-msft:
## Mandatory Checklist

  - [ ] No need for a release

### DIFF
--- a/src/Accounts/Accounts/Account/ConnectAzureRmAccount.cs
+++ b/src/Accounts/Accounts/Account/ConnectAzureRmAccount.cs
@@ -376,7 +376,7 @@ namespace Microsoft.Azure.Commands.Profile
             string claimsChallenge = null;
             if (this.IsParameterBound(c => c.ClaimsChallenge))
             {
-                if (!TryParseClaimsChallenge(ClaimsChallenge, out claimsChallenge))
+                if (!ClaimsChallengeUtilities.TryParseClaimsChallenge(ClaimsChallenge, out claimsChallenge))
                 {
                     throw new PSArgumentException(Resources.InvalidClaimsChallenge, nameof(ClaimsChallenge));
                 }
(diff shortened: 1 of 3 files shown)


src/Accounts/Accounts/help/Connect-AzAccount.md
isra-fel, line 1: Should we have an example to demonstrate how to opt-in MFA?

src/Accounts/Accounts/Account/ConnectAzureRmAccount.cs
```diff
@@ -598,6 +608,27 @@ public override void ExecuteCmdlet()
             }
         }
 
+        public static bool TryParseClaimsChallenge(string base64Input, out string claimsChallenge)
```
isra-fel: Minor: consider making it private or move to a Utility class
```suggestion
        private static bool TryParseClaimsChallenge(string base64Input, out string claimsChallenge)
```

src/Accounts/Accounts/Models/RMProfileClient.cs
```diff
@@ -130,21 +129,22 @@ public AzureRmProfile Login(
             bool skipValidation,
             IOpenIDConfiguration openIDConfigDoc,
             Action<string> promptAction,
+            string claimsChallenge = null,
```
isra-fel, line 132: Login() has no other usage except for in the tests. So this is fine

src/Accounts/Accounts/help/Connect-AzAccount.md
vidai-msft: In the error message, does it show the exact claims challenge that the end users could use?

src/Accounts/Accounts/Models/RMProfileClient.cs
vidai-msft: Yes. If don't assign with a default value, tests will fail.

isra-fel (approved):

PR #28259, opened 2025-07-23
[AzRest] added paging functionality to the invoke-azrest command

teresaritorto:
## Description
Introduce optional pagination support in Invoke-AzRest command to:
- Enable users to retrieve full datasets from paginated endpoints via server-driven pagination without the need to write their own custom pagination logic.
- Avoid breaking or unexpected changes to existing customers.

## Mandatory Checklist

  - [ ] No need for a release


src/Accounts/Accounts/Models/PSHttpResponseExtensions.cs
```diff
@@ -0,0 +1,150 @@
...
+            var nextLink = GetNextLink(nextLinkName, response);
+            
+            while (!string.IsNullOrEmpty(nextLink) && pageSize < maxPageSize)
+            {
+                response = await serviceClient
+                    .Operations
+                    .BeginHttpGetMessagesAsyncWithFullResponse(nextLink, null)
+                    .ConfigureAwait(false);
+
+                if (response.Response.StatusCode == (HttpStatusCode)429 && retry <= MaxRetryAttempts)
```
teresaritorto, line 89: May be useful to add verbose tracing here?

teresaritorto:
@microsoft-github-policy-service agree company="Microsoft"

src/Accounts/Accounts/Rest/InvokeAzRestMethodCommand.cs
```diff
@@ -109,6 +107,20 @@ public class InvokeAzRestMethodCommand : AzureRMCmdlet
         [ArgumentCompleter(typeof(FinalResultFromCompleter))]
         public string FinalResultFrom { get; set; }
 
+        [Parameter(Mandatory = false, HelpMessage = "Specifies to fetch all data for GET requests which will provide the collection of pageable items in the response.")]
+        public SwitchParameter FollowNextLink { get; set; }
+
+        [Parameter(Mandatory = false, HelpMessage = "Specifies the name of the property that provides the paginated next link. The default is nextLink")]
```
VeryEarly, line 113: have some validation for nextlinkname, pageableitemname, maxpagesize would be nice, since they are only useful with follownextlink

src/Accounts/Accounts/Rest/InvokeAzRestMethodCommand.cs
```diff
@@ -109,6 +107,20 @@ public class InvokeAzRestMethodCommand : AzureRMCmdlet
         [ArgumentCompleter(typeof(FinalResultFromCompleter))]
         public string FinalResultFrom { get; set; }
 
+        [Parameter(Mandatory = false, HelpMessage = "Specifies to fetch all data for GET requests which will provide the collection of pageable items in the response.")]
+        public SwitchParameter FollowNextLink { get; set; }
```
VeryEarly: NIT: haven't realize this before, although the description is precise enough, 'FollowNextLink' might not be the most intuitive name to hint this is a pagination flag. Maybe we can come up with a more meaning for alias for it?

src/Accounts/Accounts/Rest/InvokeAzRestMethodCommand.cs
VeryEarly: another idea is to add a new parameter set which has follownextlink mandatory (the validation of method == 'GET' is still necessary). And nextlinkname, pageableitemname, maxpagesize can be optional in this new parameter set without specific validation

src/Accounts/Accounts/ChangeLog.md
```diff
@@ -21,6 +21,7 @@
 ## Upcoming Release
 * Added new parameter `-ClaimsChallenge` to `Connect-AzAccount` to support claims challenge authentication for MFA.
 * Refined the error message when a cmdlet fails because of policy violations about Multi-Factor Authentication (MFA) to provide more actionable guidance.
+* Added Server-Side Pagination Support for `Invoke-AzRestMethod` command via `-FollowNextLink` parameter.
```
VeryEarly: Github sometimes incorrectly merge conflicts silently for Changelog.md, a good practice to avoid that is to always put new added release notes right under "## Upcoming Release" (line 22), thus conflicts will be guaranteed detected

src/Accounts/Accounts/Rest/InvokeAzRestMethodCommand.cs
teresaritorto: Thanks for the suggestion - I discussed with @Alex-AZPS and decided we will use 'Paginate' as the switch parameter as it is more intuitive to what it is actually doing.

src/Accounts/Accounts/Rest/InvokeAzRestMethodCommand.cs
teresaritorto: I've added validations in `ValidateParameters()` to enforce conditional parameter dependencies for `nextlinkname`, `pageableitemname`, `maxpagesize` only if `paginate` switch parameter is present

PR #28314, opened 2025-07-31
[Do not merge] [Test] Tsp gen SRP swagger migration test

blueww:
Please don't review it.
This is just for a test, to validate the SRP swagger -> TSP migration in https://github.com/Azure/azure-rest-api-specs/pull/35403

Several issues :

1. Swagger need update by adding `"x-ms-external": true,` to swagger "CloudError" and "CloudErrorBody", or PSH build will fail with "'CloudError' is an ambiguous reference between 'Microsoft.Azure.Management.Storage.Models.CloudError' and 'Microsoft.Rest.Azure.CloudError'". Temp workaround by update TSP gen swagger (not update TSP). Track official fix in https://github.com/Azure/azure-rest-api-specs/pull/35403#discussion_r2250853022 
2. Test case "TestAzureStorageAccountHierarchicalNamespaceUpgrade" failed since the resourcegroups in request Uri case change in swagger: [old swagger](https://github.com/Azure/azure-rest-api-specs/blob/5d8e7affbce1347ba7806a98a6af33a92d479ddd/specification/storage/resource-manager/Microsoft.Storage/stable/2025-01-01/storage.json#L824) has "resourcegroups", [new swagger](https://github.com/Azure/azure-rest-api-specs/blob/28f48ca0f88126a3d00f94fb7750bfeb5e4aa8fa/specification/storage/resource-manager/Microsoft.Storage/stable/2025-01-01/openapi.json#L3263) has "resourceGroups".  This looks old swagger issue. Fixed by update test record file Uri from "resourcegroups" to "resourceGroups".
3. Test case "TestShareSnapshot" failed since the request Uri query parameter sequence change from "$expand=snapshot&sapi-version=2025-01-01" to "api-version=2025-01-01&$expand=snapshots". This is caused by "ApiVersionParameter" in old swagger is the last two, in new swagger is the first. Track by https://github.com/Azure/azure-rest-api-specs/pull/35403#discussion_r2250825712. Temprate fix test case by update test record file.

Besides that, has run SRP regression test, all cases pass excepted keyvault related cases failed which is expected since roleassignment fail recently.


isra-fel:
Hi I'm going to convert this to a draft and please close it if won't be actively working on it. Thanks

Issue #28765, opened 2025-10-28
Invoke-AzSpotPlacementScore does not handle throttling situation


dohughes-msft:
### Description

When Invoke-AzSpotPlacementScore bumps up against API limits and gets throttled it does not handle this situation gracefully. It fails to parse the API response properly and ends with:

`Invoke-AzSpotPlacementScore_PostExpanded: Expected '{' or '['. Was String: You.`

### Issue script & Debug output

```PowerShell
PS C:\Users\dohughes\OneDrive - Microsoft\Documents\PowerShell> $DebugPreference='Continue'
PS C:\Users\dohughes\OneDrive - Microsoft\Documents\PowerShell> $resourceSku1 = @{sku = "Standard_D2_v3"}
PS C:\Users\dohughes\OneDrive - Microsoft\Documents\PowerShell> $resourceSku2 = @{sku = "Standard_D2_v2"}
PS C:\Users\dohughes\OneDrive - Microsoft\Documents\PowerShell> $resourceSku3 = @{sku = "Standard_D4_v3"}
PS C:\Users\dohughes\OneDrive - Microsoft\Documents\PowerShell> $desiredSizes = $resourceSku1,$resourceSku2,$resourceSku3
PS C:\Users\dohughes\OneDrive - Microsoft\Documents\PowerShell> $desiredLocations = 'japaneast','southcentralus','centralus'
PS C:\Users\dohughes\OneDrive - Microsoft\Documents\PowerShell> Invoke-AzSpotPlacementScore -Location eastus -DesiredCount 1 -DesiredLocation $desiredLocations -DesiredSize $desiredSizes -Debug
DEBUG: 17:04:51 - [ConfigManager] Got nothing from [DisplaySecretsWarning], Module = [], Cmdlet = []. Returning default value [True].
DEBUG: 17:04:51 - [ConfigManager] Got nothing from [DisplaySecretsWarning], Module = [], Cmdlet = []. Returning default value [True].
DEBUG: 17:04:51 - GetAzureRMContextCommand begin processing with ParameterSet 'GetSingleContext'.
DEBUG: 17:04:51 - [ConfigManager] Got [False] from [DisplayBreakingChangeWarning], Module = [Az.Accounts], Cmdlet = [Get-AzContext].
DEBUG: 17:04:51 - [ConfigManager] Got nothing from [DisplaySecretsWarning], Module = [], Cmdlet = []. Returning default value [True].
[...]

PR #29388, opened 2026-04-10
fix: handle plain-text throttle responses in Invoke-AzSpotPlacementScore

khang-11:
## Summary

- When `Invoke-AzSpotPlacementScore` receives a throttled HTTP 429 response with a plain-text body (e.g. `You are being throttled`), the autorest-generated client crashes with `Expected '{' or '['. Was String: You.` because `JsonNode.Parse()` is called unconditionally on the response body regardless of content type.
- Wrap all `FromJson(JsonNode.Parse(...))` call sites in `Compute.cs` with `try/catch` so that unparseable bodies return `null` instead of throwing. The existing `onDefault` handler already falls back to a `RestException` using the raw response body as the message when `code`/`message` are null, so the user now sees the actual throttle message rather than a parse exception.
- Add playback test cases for `Invoke-AzSpotPlacementScore` covering 429 with a `Retry-After` header and 429 without, asserting that no JSON parse exception is surfaced.

## Test coverage

| Test | What it covers |
|---|---|
| `ThrottledWithRetryAfter` | 429 + plain-text body + `Retry-After` header → no parse crash |
| `ThrottledWithoutRetryAfter` | 429 + plain-text body + no `Retry-After` header → no parse crash |

## Notes

- The fix is applied to all `_Call` methods in `Compute.cs` (not just SpotPlacement) since the same parse pattern is repeated for every API in the generated file and any of them could receive a plain-text throttle body.
- `ISendAsync.cs` retry middleware (`Retry429`, `RetryError`) is unchanged.

### DIFF
--- a/generated/Compute/Compute.Autorest/generated/api/Compute.cs
+++ b/generated/Compute/Compute.Autorest/generated/api/Compute.cs
@@ -450,13 +450,13 @@ namespace Microsoft.Azure.PowerShell.Cmdlets.Compute
                         case global::System.Net.HttpStatusCode.OK:
                         {
                             await eventListener.Signal(Microsoft.Azure.PowerShell.Cmdlets.Compute.Runtime.Events.BeforeResponseDispatch, _response); if( eventListener.Token.IsCancellationRequested ) { return null; }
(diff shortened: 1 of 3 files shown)


khang-11:
Fixes: #28765

VeryEarly:
- please read: https://github.com/Azure/azure-powershell/blob/main/CONTRIBUTING.md#submitting-changes
- please do not touch files under /generated
- please do not update test recording manually

PR #29393, opened 2026-04-10
Add App Config to Azure Environment

mrm9084:
## Description

Make it so you can set the App Config environment. Required for new clouds.

## Mandatory Checklist

  - [X] [General release](../blob/main/CONTRIBUTING.md#target-release-types)
  - [ ] No need for a release

- [X] Check this box to confirm: **I have read the [_Submitting Changes_](../blob/main/CONTRIBUTING.md#submitting-changes) section of [`CONTRIBUTING.md`](../blob/main/CONTRIBUTING.md) and reviewed the following information:**

### DIFF
--- a/src/Accounts/Accounts/Environment/SetAzureRMEnvironment.cs
+++ b/src/Accounts/Accounts/Environment/SetAzureRMEnvironment.cs
@@ -175,6 +175,14 @@ namespace Microsoft.Azure.Commands.Profile
         public string AzureAttestationServiceEndpointResourceId { get; set; }
 
         [Parameter(Mandatory = false, ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Dns suffix of Azure App Configuration.")]
+        public string AzureAppConfigurationEndpointSuffix { get; set; }
+
+        [Parameter(Mandatory = false, ValueFromPipelineByPropertyName = true,
+            HelpMessage = "The resource identifier of the Azure App Configuration service that is the recipient of the requested token.")]
+        public string AzureAppConfigurationEndpointResourceId { get; set; }
+
+        [Parameter(Mandatory = false, ValueFromPipelineByPropertyName = true,
             HelpMessage = "Dns suffix of Azure Synapse Analytics.")]
         public string AzureSynapseAnalyticsEndpointSuffix { get; set; }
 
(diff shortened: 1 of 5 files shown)


isra-fel (changes requested):
Please remove the parameter from `MetadataParameterSet` because the parameter set handles only a few endpoints returned by the metadata API.

Besides, please update the change log. Thank you.

isra-fel (changes requested):
Code change looks good. One last thing - could you refresh the help documents by following the steps in https://github.com/Azure/azure-powershell/blob/main/documentation/development-docs/help-generation.md

The reference docs for the Add and Set cmdlets are expected to be updated.
Thanks!

mrm9084:
> Code change looks good. One last thing - could you refresh the help documents by following the steps in https://github.com/Azure/azure-powershell/blob/main/documentation/development-docs/help-generation.md
> 
> The reference docs for the Add and Set cmdlets are expected to be updated. Thanks!

@isra-fel I'm unsure what I'm doing wrong when I generate I only get ProgressAction changes, nothing to do with what I did.

src/Accounts/Accounts/Environment/SetAzureRMEnvironment.cs
```diff
@@ -174,6 +174,14 @@ public string DataLakeAudience
             HelpMessage = "The resource identifier of the Azure Attestation service that is the recipient of the requested token.")]
         public string AzureAttestationServiceEndpointResourceId { get; set; }
 
+        [Parameter(Mandatory = false, ValueFromPipelineByPropertyName = true,
+            HelpMessage = "Dns suffix of Azure App Configuration.")]
+        public string AzureAppConfigurationEndpointSuffix { get; set; }
```
jimmyca15, line 179: We shouldn't be including endpoint suffix. It's subject to change, and clients should query endpoint off an app configuration resource rather than build endpoints.

jimmyca15: Given this is the established pattern, this looks like the right step to unblock usage in sovereign clouds. However, we should pursue a solution to where App Config data plane commands always use the configured "endpoint resource id" if configured to ensure that things continue to work if the app configuration endpoint suffix changes.

jimmyca15 (approved):

isra-fel (approved):
LGTM

PR #29731, opened 2026-06-15
powershell cmdlet for what if 

anamikapan11:
## Mandatory Checklist

  - [ ] No need for a release


src/Resources/ResourceManager/SdkClient/DeploymentStacksWhatIfSdkClient.cs
```diff
@@ -0,0 +1,960 @@
...
+                applyToChildScopes: denySettingsApplyToChildScopes);
+
+            // Use user-provided stackResourceId if available, else construct from name + scope
+            string resolvedStackResourceId = stackResourceId;
+            if (string.IsNullOrEmpty(resolvedStackResourceId))
+            {
+                if (!string.IsNullOrEmpty(resourceGroupName))
+                {
+                    resolvedStackResourceId = $"/subscriptions/{azureContext.Subscription.Id}/resourceGroups/{resourceGroupName}/providers/Microsoft.Resources/deploymentStacks/{deploymentStackName}";
+                }
```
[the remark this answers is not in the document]
anamikapan11: CreateDeploymentStacksWhatIfResult builds a default stack resource ID using azureContext.Subscription.Id when StackResourceId is not provided. But azureContext is only set when DeploymentStacksWhatIfSdkClient is created with the IAzureContext constructor. If the client is created with the IDeploymentStacksClient constructor, azureContext can be null, which could cause a NullReferenceException.
This preserves both constructor paths and keeps the IDeploymentStacksClient constructor usable for tests or injected clients. It only requires IAzureContext when the code actually needs to infer a subscription-scoped stack resource ID. If the caller provides StackResourceId, or if the scope is management group, no subscription context is needed. This avoids a confusing NullReferenceException and gives an actionable error message.

src/Resources/ResourceManager/SdkClient/DeploymentStacksWhatIfSdkClient.cs
```diff
@@ -0,0 +1,960 @@
...
+                
+                while (finalResult.Properties?.ProvisioningState == "Running" && attempts < maxPollingAttempts)
+                {
+                    WriteVerbose($"Polling What-If result... attempt {attempts + 1}");
+                    System.Threading.Thread.Sleep(pollingInterval * 1000);
+                    finalResult = DeploymentStacksClient.DeploymentStacksWhatIfResultsAtResourceGroup.Get(
+                        resourceGroupName,
+                        deploymentStackName);
+                    attempts++;
+                }
```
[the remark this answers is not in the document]
anamikapan11: The WhatIfResult polling loop uses Thread.Sleep, which blocks the PowerShell pipeline thread and may not respond quickly to stop/cancel. Should we keep the synchronous polling model but make the sleep cancellation-aware/cooperative, or should we consider a larger async/cancellation-token-based polling change?

There are two possible fixes i can think of -
1.	 keep sync polling but sleep in smaller chunks and check for stop/cancellation between chunks.
2.	add capped exponential backoff to reduce ARM GET pressure.

src/Resources/ResourceManager/SdkModels/DeploymentStackWhatIf/PSDeploymentStackWhatIfResult.cs
```diff
@@ -0,0 +1,34 @@
+// ----------------------------------------------------------------------------------
+// Copyright Microsoft Corporation
+// Licensed under the Apache License, Version 2.0 (the "License");
+// ----------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
dantedallag: Did this file get edited after being generated? Wondering why the license wasn't part of the generation.

src/Resources/ResourceManager/SdkModels/Deployments/PSDeploymentWhatIfCmdletParameters.cs
```diff
@@ -102,11 +102,17 @@ public DeploymentWhatIf ToDeploymentWhatIf()
             // Populate template properties.
             if (!string.IsNullOrEmpty(this.TemplateSpecId))
             {
-                properties.TemplateLink = new TemplateLink(id: this.TemplateSpecId);
+                properties.TemplateLink = new Microsoft.Azure.Management.Resources.Models.TemplateLink
```
dantedallag: I don't think that this needs to change. Could we revert this file?

src/Resources/Resources/help/Get-AzManagementGroupDeploymentStackWhatIfResult.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+## DESCRIPTION
+Gets one or more management group scoped deployment stack WhatIf result resources.
+
+## EXAMPLES
+
+### Example 1: Use Get-AzManagementGroupDeploymentStackWhatIfResult
+```powershell
+Get-AzManagementGroupDeploymentStackWhatIfResult -ManagementGroupId MyManagementGroup
```
dantedallag, line 40: Normally we have more than one example in the help docs. Check the stack cmdlets for the pattern.

src/Resources/Resources/Resources.format.ps1xml
```diff
@@ -706,5 +706,22 @@
         </ListEntries>
       </ListControl>
     </View>
+    <View>
+      <Name>Microsoft.Azure.Commands.ResourceManager.Cmdlets.SdkModels.DeploymentStackWhatIf.PSDeploymentStackWhatIfResult</Name>
+      <ViewSelectedBy>
+        <TypeName>Microsoft.Azure.Commands.ResourceManager.Cmdlets.SdkModels.DeploymentStackWhatIf.PSDeploymentStackWhatIfResult</TypeName>
```
dantedallag, line 712: Just double checking...do we need any more custom formatters? Can we confirm that all nested properties of the what-if result print out nicely when accessed via powershell?

src/Resources/ResourceManager/SdkClient/DeploymentStacksSdkClient.cs
```diff
@@ -525,7 +526,7 @@ string resourcesWithoutDeleteSupport
         )
         {
             var deleteResponse = DeploymentStacksClient.DeploymentStacks
-                .DeleteAtResourceGroupWithHttpMessagesAsync(resourceGroupName, name, resourcesCleanupAction, resourceGroupsCleanupAction, managementGroupsCleanupAction, unmanageActionResourcesWithoutDeleteSupport: resourcesWithoutDeleteSupport, bypassStackOutOfSyncError)
+                .DeleteAtResourceGroupWithHttpMessagesAsync(resourceGroupName, name, resourcesCleanupAction, resourceGroupsCleanupAction, managementGroupsCleanupAction, unmanageActionResourcesWithoutDeleteSupport: resourcesWithoutDeleteSupport, bypassStackOutOfSyncError: bypassStackOutOfSyncError ? (bool?)true : null)
```
dantedallag: Does this need to be true or null and not false?

src/Resources/ResourceManager/SdkClient/DeploymentStacksSdkClient.cs
```diff
@@ -846,14 +847,14 @@ public DeploymentStack CreateDeploymentStackModel(
...
-            var actionOnUnmanage = new ActionOnUnmanage
+            var actionOnUnmanage = new Microsoft.Azure.Management.Resources.DeploymentStacks.Models.ActionOnUnmanage
             {
                 Resources = resourcesCleanupAction,
                 ResourceGroups = resourceGroupsCleanupAction,
                 ManagementGroups = managementGroupsCleanupAction,
                 ResourcesWithoutDeleteSupport = resourcesWithoutDeleteSupport
             };
-            var denySettings = new DenySettings
+            var denySettings = new Microsoft.Azure.Management.Resources.DeploymentStacks.Models.DenySettings
```
dantedallag: Could we remove the explicit path here.

src/Resources/ResourceManager/SdkClient/NewResourceManagerSdkClient.cs
```diff
@@ -424,14 +424,14 @@ private Deployment CreateBasicDeployment(PSDeploymentCmdletParameters parameters
 
             if (!string.IsNullOrEmpty(parameters.TemplateSpecId))
             {
-                deployment.Properties.TemplateLink = new TemplateLink
+                deployment.Properties.TemplateLink = new Microsoft.Azure.Management.Resources.Models.TemplateLink
```
dantedallag: Remove explicit path here too.

src/Resources/ResourceManager/SdkClient/DeploymentStacksWhatIfSdkClient.cs
```diff
@@ -0,0 +1,960 @@
...
+        public IDeploymentStacksClient DeploymentStacksClient { get; set; }
+        public Action<string> VerboseLogger { get; set; }
+        public Action<string> ErrorLogger { get; set; }
+        public Action<string> WarningLogger { get; set; }
+        private IAzureContext azureContext;
+
+        public DeploymentStacksWhatIfSdkClient(IDeploymentStacksClient deploymentStacksClient)
+        {
+            this.DeploymentStacksClient = deploymentStacksClient;
+        }
```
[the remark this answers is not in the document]
dantedallag: We can do this for extra safety.

src/Resources/ResourceManager/SdkClient/DeploymentStacksWhatIfSdkClient.cs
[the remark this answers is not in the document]
dantedallag: I think this is a reference to the AI comment at the top of the file?

src/Resources/ResourceManager/SdkClient/DeploymentStacksWhatIfSdkClient.cs
```diff
@@ -0,0 +1,963 @@
...
+                Location = location,
+                Properties = properties
+            };
+
+            return whatIfResult;
+        }
+
+        #region What-If Resource CRUD Operations
+
+        public PSDeploymentStackWhatIfResult GetResourceGroupDeploymentStackWhatIfResult(string resourceGroupName, string stackName, bool throwIfNotExists = true)
```
dantedallag: Should this be stackWhatIfName or whatIfResultName?

src/Resources/ResourceManager/SdkClient/DeploymentStacksWhatIfSdkClient.cs
```diff
@@ -0,0 +1,963 @@
...
+                            return null;
+                        throw new PSArgumentException($"WhatIf result '{stackName}' in Resource Group '{resourceGroupName}' not found.");
+                    }
+                    throw new PSArgumentException(dex.Body?.Error?.Message ?? dex.Message);
+                }
+                throw;
+            }
+        }
+
+        public PSDeploymentStackWhatIfResult GetResourceGroupDeploymentStackWhatIfResultWithPropertyChanges(string resourceGroupName, string stackName)
```
dantedallag: The `WithPropertyChange` methods should also handle throwIfNotExists as the Get does.

src/Resources/ResourceManager/SdkClient/DeploymentStacksWhatIfSdkClient.cs
```diff
@@ -0,0 +1,963 @@
...
+            }
+            catch (Exception ex)
+            {
+                if (ex is ErrorResponseException dex)
+                    throw new PSArgumentException(dex.Body?.Error?.Message ?? dex.Message);
+                throw;
+            }
+        }
+
+        public void DeleteSubscriptionDeploymentStackWhatIfResult(string stackName)
```
dantedallag: Most instances of "stackName" in this file should probably refer to the what-if result name for clarity.

src/Resources/ResourceManager/Implementation/CmdletBase/DeploymentStackWhatIfCreateCmdlet.cs
```diff
@@ -0,0 +1,119 @@
...
+                }
+                else
+                {
+                    existing = DeploymentStacksWhatIfSdkClient.GetSubscriptionDeploymentStackWhatIfResult(
+                        parameters.StackName, throwIfNotExists: false);
+                }
+            }
+            catch
+            {
+                // Any other error during existence check is non-fatal.
```
dantedallag, line 67: Is this what we do in stacks? I think that there is an argument for making an existence check failure a fatal error.

src/Resources/ResourceManager/Implementation/CmdletBase/DeploymentStackWhatIfCreateCmdlet.cs
```diff
@@ -0,0 +1,119 @@
...
+            }
+
+            Action executeAction = () =>
+            {
+                PSDeploymentStackWhatIfResult whatIfResult = this.ExecuteWhatIf(parameters);
+                // The ToString() method on PSDeploymentStackWhatIfResult calls the formatter
+                this.WriteObject(whatIfResult);
+            };
+
+            if (existing != null)
```
dantedallag: This check should be different for the New and Set cmdlets. The New cmdlet should issue a warning if the stack what-if exists and the Set cmdlet should issue a warning if the stack does not exist.

src/Resources/ResourceManager/Implementation/CmdletBase/DeploymentStackWhatIfCreateCmdlet.cs
```diff
@@ -0,0 +1,119 @@
...
+            }
+            else
+            {
+                executeAction();
+            }
+        }
+
+        protected PSDeploymentStackWhatIfResult ExecuteWhatIf(PSDeploymentStackWhatIfParameters parameters)
+        {
+            const string statusMessage = "Getting the latest status of all resources...";
```
dantedallag, line 104: I may need to get some context on the output logic here.

src/Resources/ResourceManager/Implementation/DeploymentStacksWhatIf/RemoveAzSubscriptionDeploymentStackWhatIfResult.cs
```diff
@@ -0,0 +1,97 @@
...
+                        DeploymentStacksWhatIfSdkClient.DeleteSubscriptionDeploymentStackWhatIfResult(Name);
+                        if (PassThru.IsPresent)
+                        {
+                            WriteObject(true);
+                        }
+                    });
+            }
+            catch (Exception ex)
+            {
+                WriteExceptionError(ex);
```
dantedallag: Wondering if we should follow the catch logic in stacks cmdlets for similar formatting.

src/Resources/Resources.Test/ScenarioTests/DeploymentStackTests.cs
```diff
@@ -303,5 +303,158 @@ public void TestSaveAndRemoveManagementGroupDeploymentStackWithPipeOperator()
         {
             TestRunner.RunTestScript("Test-SaveAndRemoveManagementGroupDeploymentStackWithPipeOperator");
         }
+
+        // ---- WhatIf Resource Group Scope ----
```
dantedallag: Can we move all the what-if scenario tests to a separate file?

src/Resources/Resources.Test/ScenarioTests/DeploymentStackTests.ps1
```diff
@@ -1904,4 +1904,735 @@ function Test-SaveAndRemoveManagementGroupDeploymentStackWithPipeOperator
...
+	$rglocation = "West Central US"
+
+	try
+	{
+		New-AzResourceGroup -Name $rgname -Location $rglocation
+
+		# Test - Success with template file
+		$result = New-AzResourceGroupDeploymentStackWhatIfResult -Name $rname -ResourceGroupName $rgname -TemplateFile StacksRGTemplate.json -TemplateParameterFile StacksRGTemplateParams.json -StackResourceId (Get-ResourceGroupDeploymentStackId $rgname $rname) -RetentionInterval "P1D" -ActionOnUnmanage DetachAll -DenySettingsMode None -Force
+		Assert-NotNull $result
+		Assert-DeploymentStackWhatIfResultCompletedOrUnavailable $result
```
dantedallag: We should be asserting it completed. I don't think we want to assert on it not being available.

dantedallag: We should also be asserting that it completes sucessfully.

dantedallag (commented):
Some general review comments:

1. We need to adjust testing to ensure things are completing as anticipated. 
2. We need to integrate changes that have been checked-in since we started working on this, like the additional action on unmanage value.

src/Resources/ResourceManager/SdkClient/DeploymentStacksSdkClient.cs
anamikapan11: Updated this to pass true only when -BypassStackOutOfSyncError is specified and null otherwise, so the generated SDK omits the nullable query/body field when the switch is not used.

src/Resources/ResourceManager/Implementation/CmdletBase/DeploymentStackWhatIfCreateCmdlet.cs
anamikapan11: The output flow here follows the existing deployment what-if pattern: we write a transient status message while the operation runs, clear it before returning, and then emit the typed PSDeploymentStackWhatIfResult object via WriteObject. The formatted display is handled by the result formatter/ToString() path rather than writing formatted text directly from the cmdlet.

.github/copilot-instructions.md
```diff
@@ -268,4 +268,8 @@ cd generated/ModuleName/ModuleName.Autorest
 ./test-module.ps1                  # If available
 ```
 
-Remember: This is a large, complex repository with extensive build times. Plan accordingly and never cancel long-running operations.
\ No newline at end of file
+Remember: This is a large, complex repository with extensive build times. Plan accordingly and never cancel long-running operations.
```
VeryEarly: please exclude these changes

VeryEarly (approved):

PR #29842, opened 2026-07-14
Disable WAM broker on Linux and macOS; keep WAM on Windows only

xuming-ms:
## Description

WAM (Web Account Manager) relies on the operating system broker, which is only reliably available on Windows. On non-Windows platforms this previously caused interactive login to fail with a broker/`msalruntime` error, because:
- The broker is not supported on Linux, and
- The native broker runtime (`msalruntime.dylib`) is not shipped with the module for macOS.

This change gates the broker to Windows only:
- `EnableLoginByWam` now defaults to enabled on Windows and disabled on non-Windows (Linux/WSL, macOS).
- `AzConfigReader.IsWamEnabled` requires `IsWindowsPlatform()`, so on Linux and macOS the broker path is never selected and interactive login uses the browser directly — even if the user explicitly enables the config.

Windows WAM behavior is unchanged.

## Mandatory Checklist

- Please choose the target release of Azure PowerShell.

* **SHOULD** update `ChangeLog.md` file(s) appropriately — done (`src/Accounts/Accounts/ChangeLog.md`).
* No cmdlet API change, so no markdown help regeneration required.

### DIFF
--- a/src/Accounts/Accounts/ChangeLog.md
+++ b/src/Accounts/Accounts/ChangeLog.md
@@ -19,6 +19,9 @@
 -->
 
 ## Upcoming Release
+* Disabled login by WAM (Web Account Manager) broker on non-Windows platforms.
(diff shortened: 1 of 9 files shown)


YangAn-microsoft:
Please resolve conflicts

